PR #5, opened 2025-04-13
Find flows 🔨 

brovatten:
Hey, thanks for reviewing this.

Uses the readme, and a prompt for entrypoints that are either python code, or CLI. I stole the basemodel and configparser from you :) feel free to change anything, and I will review it later 

- I tried to do it general, way too complex for now. I scratched it all and did this instead. 
- Thinking about whether we want to here already look for dependecies. If people input any repo into our demo, perhaps we should simply install all of the dependencies using their requirements.txt or .toml onto our server. This would be simplest.

Subset of output when you run `entrypoints.py`:

```----------------------------------------
Entrypoint Type: code
from markitdown import MarkItDown

md = MarkItDown(docintel_endpoint="<document_intelligence_endpoint>")
result = md.convert("test.pdf")
print(result.text_content)
----------------------------------------
Entrypoint Type: code
from markitdown import MarkItDown
from openai import OpenAI

client = OpenAI()
md = MarkItDown(llm_client=client, llm_model="gpt-4o")
result = md.convert("example.jpg")
print(result.text_content)
----------------------------------------
```

### DIFF
--- a/entrypoints.py
+++ b/entrypoints.py
@@ -10,7 +10,7 @@ from langchain_google_genai import ChatGoogleGenerativeAI
 
 def init_llm():
     load_dotenv()
-    api_key = os.getenv("API_KEY")
+    api_key = os.getenv("GOOGLE_API_KEY")
     return ChatGoogleGenerativeAI(
         model="gemini-2.0-flash-001",
         temperature=0,
--- a/main.py
+++ b/main.py
@@ -10,7 +10,7 @@ def main():
 # result = md.convert("./resources/test.xlsx")
 # """
 
-    entrypoints = EntrypointScanner(root="markitdown").run()
+    entrypoints = EntrypointScanner(root="./repos/markitdown").run()
     for entrypoint in entrypoints.root:
 
         stat_analyzer = Analyzer(module_name='markitdown', code=entrypoint.python_code)


entrypoints.py
```diff
@@ -0,0 +1,121 @@
...
+from dotenv import load_dotenv
+from pydantic import BaseModel, RootModel, Field
+from langchain.output_parsers import PydanticOutputParser
+from langchain_core.prompts import ChatPromptTemplate
+from langchain_google_genai import ChatGoogleGenerativeAI
+
+
+def init_llm():
+    load_dotenv()
+    api_key = os.getenv("API_KEY")
```
ivanmilevtues: Let's use GOOGLE_API_KEY
As now in the env we have multiple api keys.

main.py
```diff
@@ -1,31 +1,36 @@
...
-    mermaid_str += '```'
-    with open("markitdown_diagram.md", "w") as f:
-        f.write(mermaid_str)
+#     code = """from markitdown import MarkItDown
+
+# md = MarkItDown(enable_plugins=False) # Set to True to enable plugins
+# result = md.convert("./resources/test.xlsx")
+# """
+
+    entrypoints = EntrypointScanner(root="markitdown").run()
```
ivanmilevtues: Let's keep analyzed projects in the /repo folder.

entrypoints.py
brovatten: Good catch, didn't check this after merging

PR #6, opened 2025-04-13
Added hardcoded web demo server.

ivanmilevtues:
This I don't think we should merge. Just take a look and we can rediscuss tomorrow :-)

### DIFF
--- a/web.py
+++ b/web.py
@@ -7,7 +7,7 @@ app = FastAPI()
 @app.post("/generate_repository")
 async def analyze_repo():
     mermaid_str = analyze()
-    with open("/home/ivan/StartUp/CodeBoarding/repos/markitdown/README.md", "r") as f:
+    with open("/home/ivan/StartUp/CodeBoarding/repos/markitdown/README_original.md", "r") as f:
         existing_content = f.read()
     with open("/home/ivan/StartUp/CodeBoarding/repos/markitdown/README.md", "w") as f:
         f.write("# High Level Diagram\n")


web.py
```diff
@@ -0,0 +1,22 @@
...
+        existing_content = f.read()
+    with open("/home/ivan/StartUp/CodeBoarding/repos/markitdown/README.md", "w") as f:
+        f.write("# High Level Diagram\n")
+        f.write(mermaid_str)
+        f.write("\n")
+        f.write(existing_content)
+
+    # Now git add, commit and push
+    import subprocess
+    subprocess.run(["git", "-C", "/home/ivan/StartUp/CodeBoarding/repos/markitdown", "add", "/home/ivan/StartUp/CodeBoarding/repos/markitdown/README.md"])
```
brovatten, line 20: This could be a github action, but as long as it works :) I think this should push to a pretermined branch (not main imo).

web.py
```diff
@@ -0,0 +1,22 @@
+from fastapi import FastAPI
+
+from main import analyze
+
+app = FastAPI()
+
+@app.post("/generate_repository")
+async def analyze_repo():
+    mermaid_str = analyze()
+    with open("/home/ivan/StartUp/CodeBoarding/repos/markitdown/README.md", "r") as f:
```
brovatten: Before the merge - since we are close to adding more repos - adding flexibility to add more repos here is useful. Creating a repo class. And using the repo select from the user to determine the name. Then finding the README.md of the project if there is one.

web.py
brovatten: Maybe check if it exists before committing it. Could easily cloud the repo with commits.

PR #13, opened 2025-04-27
Detail agent

ivanmilevtues:
Agent which drills down on specific component and gives information for it.
- Generates a flow diagram
-  Links the main high-level diagram with the child diagram
- Fixes mermaid templates

### DIFF
--- a/.github/workflows/main_diagramgenerator.yml
+++ b/.github/workflows/main_diagramgenerator.yml
@@ -1,0 +1,83 @@
+# Docs for the Azure Web Apps Deploy action: https://github.com/azure/functions-action
+# More GitHub Actions for Azure: https://github.com/Azure/actions
+# More info on Python, GitHub Actions, and Azure Functions: https://aka.ms/python-webapps-actions
+
+name: Build and deploy Python project to Azure Function App - DiagramGenerator
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+env:
+  AZURE_FUNCTIONAPP_PACKAGE_PATH: '.' # set this to the path to your web app project, defaults to the repository root
+  PYTHON_VERSION: '3.12' # set this to the python version to use (supports 3.6, 3.7, 3.8)
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read #This is required for actions/checkout
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Python version
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Create and start virtual environment
+        run: |
+          python -m venv venv
+          source venv/bin/activate
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      # Optional: Add step to run tests here
+
+      - name: Zip artifact for deployment
+        run: zip release.zip ./* -r
+
+      - name: Upload artifact for deployment job
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-app
+          path: |
+            release.zip
+            !venv/
+
+  deploy:
+    runs-on: ubuntu-latest
+    needs: build
+    permissions:
+      id-token: write #This is required for requesting the JWT
+      contents: read #This is required for actions/checkout
+
+    steps:
+      - name: Download artifact from build job
+        uses: actions/download-artifact@v4
+        with:
+          name: python-app
+
+      - name: Unzip artifact for deployment
+        run: unzip release.zip     
+        
+      - name: Login to Azure
+        uses: azure/login@v2
+        with:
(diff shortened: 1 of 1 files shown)


agents/agent.py
```diff
@@ -0,0 +1,85 @@
...
+    sub_graph: str = Field(description="The sub control flow graph in DOT format.")
+
+
+class AnalysisInsights(BaseModel):
+    abstract_components: List[Component] = Field(
+        description="List of the abstract components identified in the project.")
+
+    def llm_str(self):
+        if not self.abstract_components:
+            return "No abstract components found."
```
brovatten, line 41: Great basecase and inheritance use!

agents/agent.py
```diff
@@ -0,0 +1,85 @@
...
+    def _invoke(self, prompt):
+        """Unified agent invocation method."""
+        response = self.agent.invoke(
+            {"messages": [self.system_message, HumanMessage(content=prompt)]}
+        )
+        agent_response = response["messages"][-1]
+        assert isinstance(agent_response, AIMessage), f"Expected AIMessage, but got {type(agent_response)}"
+        if type(agent_response.content) == str:
+            return agent_response.content
+        if type(agent_response.content) == list:
```
brovatten, line 84: When is it a list here?

agents/prompts.py
```diff
@@ -0,0 +1,154 @@
...
+You a software architecture expert. Now you have observed the CFG and got insights from the structure of the project.
+
+Here is the analysis result of the components, and abstractions as well as their related source code files:
+{insight_so_far}
+
+Now as final step in order to validate and enhance their relationship you can make use of their **source code** via the provided `read_source_code` tool.
+
+**Your Tasks:**
+1. Use the read_source_code tool to read the source code of the modules and components you need further details about.
+2. Refine or expand the earlier high-level classes/components, in the end you have to have **LESS THAN 10 COMPONENTS** (best if they are 5), based on new details from the source code.
```
brovatten: (best if they are 5) I would probably remove. Improving explanations of what you want (already quite clear) is better for generalizing. But for this use-case I can be wrong here

brovatten: Or maybe we should try it for some project we believe has 10 components, and see if it degrades the abstract diagrams or not (by e.g. showing 5 ones)

main.py
```diff
@@ -1,32 +1,133 @@
...
+
+def upload_onboarding_materials(project_name):
+    onboarding_repo = "/home/ivan/StartUp/GeneratedOnBoardings/"
+    repo = Repo(onboarding_repo)
+    origin = repo.remote(name='origin')
+    origin.pull()
+
+    onboarding_repo_location = onboarding_repo + project_name
+    if os.path.exists(onboarding_repo_location):
+        # remove the directory
```
brovatten: Maybe allow caching as an option, so that for the deployment we can flag it to True

static_analyzer/pylint_analyze/structure_graph_builder.py
```diff
@@ -0,0 +1,68 @@
...
+        for path in tqdm(paths, desc="Building structure graphs for packages", unit="package"):
+            self.run_pyreverse(path)
+
+
+def collect_paths(root: Path) -> list[Path]:
+    collected: list[Path] = []
+
+    def _walk(dir_: Path):
+        # Skip tests for now.
+        if "tests" in dir_.parts:
```
brovatten: add "test" or "testing"

static_analyzer/pylint_graph_transform.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def bfs_scan_directory(self):
+        queue = deque()
+        queue.append(self.repo)
+
+        while queue:
+            current_dir = queue.popleft()
+            try:
+                entries = os.listdir(current_dir)
+            except PermissionError:
```
brovatten, line 28: We should probably print this - could easily lead to fault searching

static_analyzer/pylint_graph_transform.py
```diff
@@ -0,0 +1,90 @@
...
+                continue
+
+            # Check if this directory has an __init__.py file
+            if '__init__.py' in entries and "test" not in current_dir and "example" not in current_dir:
```
brovatten, line 33: Shouldn't Pylint already have filtered away the tests and examples in def collect_paths. I would add example to the list of ignored collect_paths. And if we find one here, we should raise an Exception

main.py
```diff
@@ -1,32 +1,133 @@
...
+    for f in os.listdir(directory):
+        if f.endswith('.dot'):
+            os.remove(os.path.join(directory, f))
+        if f.endswith('.md') and f != "README.md":
+            os.remove(os.path.join(directory, f))
+    print("Files clean up done.")
+
+
+def upload_onboarding_materials(project_name):
+    onboarding_repo = "/home/ivan/StartUp/GeneratedOnBoardings/"
```
brovatten: Make this one work in prod

agents/agent.py
brovatten: When it returns too long of a response and need to split it up.

PR #26, opened 2025-06-18
Polling

brovatten:
Polling that allows you to see how your job is doing. The you can check up for a user or a job_id (return by the post whne process starts). The checkup for the job_id is not working - but for the user is. 

Have a look and see what you think.

I also integrated it into Framer and it seems to be working.

### DIFF
--- a/local_app.py
+++ b/local_app.py
@@ -1,21 +1,20 @@
+import dotenv
+dotenv.load_dotenv()
+
 import asyncio
 import logging
 import os
 import uuid
 from datetime import datetime
 from pathlib import Path
-from typing import Optional
 from urllib.parse import urlparse
-
-import duckdb
-from filelock import FileLock
-import dotenv
 from fastapi import FastAPI, HTTPException, Query, BackgroundTasks
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from starlette.concurrency import run_in_threadpool
 
 from diagram_generator import DiagramGenerator
+from duckdb_crud import fetch_job, init_db, insert_job, update_job
 from utils import (
     CFGGenerationError,
     RepoDontExistError,
@@ -24,14 +23,18 @@ from utils import (
     remove_temp_repo_folder,
 )
 
-dotenv.load_dotenv()
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+
+class JobStatus:
+    PENDING = "pending"
+    RUNNING = "running"
+    SUCCESS = "success"
+    FAILED = "failed"
+
 # Environment variables
-DB_PATH = os.getenv("JOB_DB", "jobs.duckdb")
-LOCK_PATH = DB_PATH + ".lock"
 REPO_ROOT = os.getenv("REPO_ROOT")
 if not REPO_ROOT:
     logger.error("REPO_ROOT environment variable not set")
@@ -57,97 +60,8 @@ app.add_middleware(
 MAX_CONCURRENT_JOBS = 5
 job_semaphore = asyncio.Semaphore(MAX_CONCURRENT_JOBS)
 
-class JobStatus:
-    PENDING = "pending"
-    RUNNING = "running"
-    SUCCESS = "success"
-    FAILED = "failed"
-
-# -- DuckDB Connection Helper --
-def _connect():
-    return duckdb.connect(DB_PATH)
-
-# Initialize DB on startup
-def init_db():
-    # ensure directory exists
-    dir_path = os.path.dirname(DB_PATH)
-    if dir_path and not os.path.exists(dir_path):
-        os.makedirs(dir_path, exist_ok=True)
-    # wipe existing DB and lock files
-    if os.path.exists(DB_PATH):
-        try:
-            os.remove(DB_PATH)
-            os.remove(LOCK_PATH)
-        except OSError:
-            pass
-    # create fresh table
-    with FileLock(LOCK_PATH):
-        conn = _connect()
-        conn.execute(
-            """
-            CREATE TABLE IF NOT EXISTS jobs (
-              id TEXT PRIMARY KEY,
-              repo_url TEXT,
-              status TEXT,
-              result TEXT,
-              error TEXT,
-              created_at TIMESTAMP,
-              started_at TIMESTAMP,
-              finished_at TIMESTAMP
-            )
-            """
-        )
-        conn.close()
-
 app.add_event_handler("startup", init_db)
 
-# -- CRUD operations --
-def insert_job(job: dict):
-    with FileLock(LOCK_PATH):
-        conn = _connect()
-        conn.execute(
-            "INSERT INTO jobs VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
-            [
-                job["id"], job["repo_url"], job["status"], job["result"],
-                job["error"], job["created_at"], job["started_at"], job["finished_at"]
-            ],
-        )
-        conn.close()
-
-
-def update_job(job_id: str, **fields):
-    cols, vals = zip(*fields.items())
-    set_clause = ", ".join(f"{c} = ?" for c in cols)
-    with FileLock(LOCK_PATH):
-        conn = _connect()
-        conn.execute(
-            f"UPDATE jobs SET {set_clause} WHERE id = ?",
-            list(vals) + [job_id],
-        )
-        conn.close()
-
-
-def fetch_job(job_id: str) -> Optional[dict]:
-    conn = _connect()
-    res = conn.execute(
-        "SELECT id, repo_url, status, result, error, created_at, started_at, finished_at"
-        " FROM jobs WHERE id = ?",
-        [job_id]
-    ).fetchall()
-    conn.close()
-    if not res:
-        return None
-    id_, repo_url, status, result, error, created_at, started_at, finished_at = res[0]
-    return {
-        "id": id_,
-        "repo_url": repo_url,
-        "status": status,
-        "result": result,
-        "error": error,
-        "created_at": created_at.isoformat() if created_at else None,
-        "started_at": started_at.isoformat() if started_at else None,
-        "finished_at": finished_at.isoformat() if finished_at else None,
-    }
 
 # -- Utility Functions --
 def extract_repo_name(repo_url: str) -> str:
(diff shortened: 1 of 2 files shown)


local_app.py
```diff
@@ -37,152 +35,116 @@
...
-
-def generate_documents(repo_path, temp_repo_folder, repo_name):
-    generator = DiagramGenerator(repo_location=repo_path, temp_folder=temp_repo_folder, repo_name=repo_name,
-                                 output_dir=temp_repo_folder)
-    analysis_files = generator.generate_analysis()
-    return analysis_files
-
-
-@app.get(
-    "/github_action",
```
ivanmilevtues, line 113: Github action is removed, while we still need it for the github action

local_app.py
```diff
@@ -37,152 +35,116 @@
...
     finally:
-        # cleanup temp folder for this run
-        remove_temp_repo_folder(temp_repo_folder)
+        job["finished_at"] = datetime.utcnow().isoformat()
+
+@app.post("/job", response_class=JSONResponse, summary="Create a new onboarding job", responses={
+    200: {"description": "Job created", "content": {"application/json": {}}},
+    400: {"description": "Missing user_id or url"},
+})
+async def create_job(user_id: str = Query(..., description="User ID"), url: str = Query(..., description="GitHub repo URL"), background_tasks: BackgroundTasks = None):
```
ivanmilevtues: create_job and run_job ar enot informative. Can you name them with create_markdown or demo_markdown generation. etc.

local_app.py
```diff
@@ -37,152 +35,116 @@
...
+    404: {"description": "Job not found"},
+})
+async def get_job(job_id: str):
+    job = jobs.get(job_id)
+    if not job:
+        raise HTTPException(404, detail="Job not found")
+    return {
+        "id": job["id"],
+        "user_id": job["user_id"],
+        "url": job["url"],
```
ivanmilevtues: os this url, the repo_url. maybe better to name it repo_url. That I don't mind.

local_app.py
```diff
@@ -37,152 +35,116 @@
...
+    200: {"description": "Job status/result"},
+    404: {"description": "Job not found"},
+})
+async def get_job(job_id: str):
+    job = jobs.get(job_id)
+    if not job:
+        raise HTTPException(404, detail="Job not found")
+    return {
+        "id": job["id"],
+        "user_id": job["user_id"],
```
ivanmilevtues: This user_id seems kind of un needed atm. Looked at the front end and it is always random or am I missing something?

local_app.py
```diff
@@ -37,152 +35,116 @@
...
+        "status": JobStatus.PENDING,
+        "result": None,
+        "error": None,
+        "created_at": datetime.utcnow().isoformat(),
+        "started_at": None,
+        "finished_at": None,
+    }
+    return job
+
+async def run_job(job_id: str):
```
ivanmilevtues: run_diagram_generation_job, or something more informative maybe

local_app.py
ivanmilevtues: same goes for the url "/job" etc.

local_app.py
```diff
@@ -37,152 +35,109 @@
...
-
-    except Exception as e:
-        logger.exception("Unexpected error processing repo %s", url)
-        raise HTTPException(500, detail="Internal server error")
-
+        async with job_semaphore:
+            temp_repo_folder = create_temp_repo_folder()
+            try:
+                repo_name = await run_in_threadpool(
+                    generate_docs_remote,
```
ivanmilevtues: This is calling the wrong method. I thas to call generate_coduments.

local_app.py
```diff
@@ -37,152 +35,109 @@
...
-
-        if not docs_content:
-            logger.warning("No documentation files generated for: %s", url)
-            raise HTTPException(404, detail="No documentation files were generated")
-
-        logger.info("Successfully generated %d doc files for %s", len(docs_content), url)
-        resp = JSONResponse(content={
-            "files": docs_content
-        })
-        return resp
```
ivanmilevtues, line 172: The new method is not returning the correct stuff.

local_app.py
```diff
@@ -37,152 +35,109 @@
...
     finally:
-        # cleanup temp folder for this run
-        remove_temp_repo_folder(temp_repo_folder)
+        job["finished_at"] = datetime.utcnow().isoformat()
+
+@app.post("/generation", response_class=JSONResponse, summary="Create a new onboarding job", responses={
+    200: {"description": "Job created", "content": {"application/json": {}}},
+    400: {"description": "Missing repo_url"},
+})
+async def start_generation_job(repo_url: str = Query(..., description="GitHub repo URL"), background_tasks: BackgroundTasks = None):
```
ivanmilevtues: This looks good :)
Thanks for updating.

local_app.py
```diff
@@ -37,152 +42,151 @@
...
-                analysis = AnalysisInsights.model_validate_json(f.read())
-                logging.info(f"Generated analysis file: {file}")
-                markdown_response = generate_markdown_content(analysis, repo_name, link_files=("analysis.json" in file),
-                                                              repo_url=url,
-                                                              reference_link=f"{url}/blob/main/.codeboarding/")
-                fname = Path(file).name.split(".json")[0]
-                fname = "on_boarding" if fname.endswith("analysis") else fname
-                docs_content[f"{fname}.md"] = markdown_response.strip()
+async def github_action(job_id: str):
+    return await generate_onboarding(job_id)
```
ivanmilevtues: This still doesn't retrun the correct thing. It has to be JSON Response which is mapping of the files with their contents.

```Python
        for file in analysis_files:
            with open(file, 'r') as f:
                analysis = AnalysisInsights.model_validate_json(f.read())
                logging.info(f"Generated analysis file: {file}")
                markdown_response = generate_markdown_content(analysis, repo_name, link_files=("analysis.json" in file),
                                                              repo_url=url,
                                                              reference_link=f"{url}/blob/main/.codeboarding/")
                fname = Path(file).name.split(".json")[0]
                fname = "on_boarding" if fname.endswith("analysis") else fname
                docs_content[f"{fname}.md"] = markdown_response.strip()

        if not docs_content:
            logger.warning("No documentation files generated for: %s", url)
            raise HTTPException(404, detail="No documentation files were generated")

        logger.info("Successfully generated %d doc files for %s", len(docs_content), url)
        resp = JSONResponse(content={
            "files": docs_content
        })
        return resp
```

ivanmilevtues: The generate_onboarding method doesn't give the proper format that is expected from the github action

local_app.py
```diff
@@ -37,152 +42,151 @@
...
+        "error": job["error"],
+        "created_at": job["created_at"],
+        "started_at": job["started_at"],
+        "finished_at": job["finished_at"],
+    }
 
-    finally:
-        # cleanup temp folder for this run
-        remove_temp_repo_folder(temp_repo_folder)
+def generate_documents(repo_path, temp_repo_folder, repo_name):
```
ivanmilevtues: This is fine but if you are editing the file anyway. Probably better to put it toghether with the other non-rest methods so above the start_job or move the other non end-point funcs below here

local_app.py
```diff
@@ -37,152 +42,151 @@
...
-    Example:
-        GET /myroute?url=https://github.com/your/repo
-    """
-    logger.info("Received request to generate docs for %s", url)
-
-    # Setup a dedicated temp folder for this run
-    temp_repo_folder = create_temp_repo_folder()
+MAX_CONCURRENT_JOBS = 5
+job_semaphore = asyncio.Semaphore(MAX_CONCURRENT_JOBS)
+jobs: Dict[str, dict] = {}
```
ivanmilevtues: I was doing polling for the Github action and this solution with global variables doesn't work:
```Why does it work with one worker but “lose” jobs when you start uvicorn with --workers 4?
Because every worker is a totally independent process.
 

A new process is forked four times.
Your module is imported four times – therefore job_store = JobStore() is created four times.
When a request is routed to worker-1 the job is stored in worker-1’s private dictionary.
When you later poll /github_action/jobs/<id> the request may be handled by worker-2, -3 or -4. In those processes the dictionary never contained that job, so you get “Job not found”.
Even the background task that is doing the real work lives only inside the worker that accepted the POST request; the other workers know nothing about it.
In-memory objects are never shared between Gunicorn / Uvicorn workers, only within the same
process and its threads.
That is the whole problem – nothing is wrong with your FastAPI code apart from the assumption that a “global” variable stays global when you scale out.

```

ivanmilevtues: Basically global are not global when ran in uvicorn as it just forks X times the main process.

local_app.py
brovatten: Fixed on his own branch

local_app.py
brovatten: ok i put endpoint at top

local_app.py
```diff
@@ -1,34 +1,50 @@
...
 import os
+import uuid
+from datetime import datetime
 from pathlib import Path
+from typing import Optional
+from urllib.parse import urlparse
 
-from dotenv import load_dotenv
-from fastapi import FastAPI, HTTPException, Query
+import duckdb
```
ivanmilevtues: add depenedency to requirements.txt

local_app.py
```diff
@@ -37,152 +53,212 @@
...
+        "error": None,
+        "created_at": now,
+        "started_at": None,
+        "finished_at": None,
+    }
+
+async def generate_onboarding(job_id: str):
+    update_job(job_id, status=JobStatus.RUNNING, started_at=datetime.utcnow())
+    try:
+        async with job_semaphore:
```
ivanmilevtues, line 103: I don't we need the job_semaphore.

local_app.py
```diff
@@ -1,34 +1,50 @@
...
+    remove_temp_repo_folder,
+)
+
+dotenv.load_dotenv()
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# Environment variables
+DB_PATH = os.getenv("JOB_DB", "jobs.duckdb")
```
ivanmilevtues: Add in the readme that we need the DB_PATH in the .env

local_app.py
```diff
@@ -37,152 +53,212 @@
...
-
-        result_url = (
-            f"https://github.com/CodeBoarding/GeneratedOnBoardings"
-            f"/blob/main/{repo_name}/on_boarding.md"
+        conn.close()
+
+app.add_event_handler("startup", init_db)
+
+# -- CRUD operations --
+def insert_job(job: dict):
```
ivanmilevtues: maybe the setup of the db has to go into its own file

PR #28, opened 2025-07-04
Vscode integration

ivanmilevtues:
- fixes on the endpoint
- fixes for the vscode integration.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,10 +32,6 @@ on:
     branches:
       - master
       - main
-  pull_request:
-    branches:
-      - master
-      - main
 
 jobs:
   update-docs:


README.md
```diff
@@ -21,6 +21,7 @@ GITHUB_TOKEN= # Github token for accessing private repositories
 PROJECT_ROOT= # The source project root => Has to end with /CodeBoarding
 DIAGRAM_DEPTH_LEVEL= # max level of depth for the generations
 JOB_DB=.duckdb/duckdb # The database for the jobs
+GOOGLE_API_KEY=# # Google API key for the Google Cloud Platform
```
brovatten, line 24: Good!

agents/tools/read_cfg.py
```diff
@@ -39,11 +39,16 @@ def component_cfg(self, component: Component):
         component_cfg = {}
         for k, v in self.cfg.items():
             for ref in component.referenced_source_code:
-                if ref.qualified_name == k or ref.qualified_name in k or k in ref.qualified_name:
+                qual_name = ref.qualified_name
+                if "/" in qual_name:
+                    qual_name = qual_name.replace("/", ".")
```
brovatten, line 44: How come?

demo.py
```diff
@@ -84,17 +86,18 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
...
         # if "python" not in langs.lower():
         #     continue
-        try:
-            generate_docs_remote(repo, temp_repo_folder, local_dev=True)
-            # companies.add(company)
-        except Exception as e:
-            logging.error(f"Failed to generate docs for {repo}: {e}")
-        finally:
-            remove_temp_repo_folder(temp_repo_folder)
+        # try:
```
brovatten: these changes should prolly not be pushed

agents/tools/read_cfg.py
ivanmilevtues: Models tend to hallucinate from time to time and put the qualified name as the file path. This usually happens on the highest level of abstractions because at that stage it is whole files...
This just makes the tool more robust and I change it with "." as this is how we process those :)

demo.py
ivanmilevtues: Yes I guess. They shouldn't make a difference but I will revert. This is the manual run of the demo.py so we change it all the time for specific generations.

PR #29, opened 2025-07-21
Depth for vscode

ivanmilevtues:
- depth exploration for VSCode
- better error handling for model throttle
- better validation (via static analysis or just deterministic validation :D)

### DIFF
--- a/agents/tools/read_docs.py
+++ b/agents/tools/read_docs.py
@@ -39,19 +39,11 @@ class ReadDocsTool(BaseTool):
         """
         Walk the directory and collect all markdown files.
         """
-        for path in root_project_dir.rglob('*.md'):
-            if "tests" not in path.parts and "test" not in path.name.lower():
-                self.cached_files.append(path)
-            self.cached_files.append(path)
-        for path in root_project_dir.rglob('*.rst'):
-            if "tests" not in path.parts and "test" not in path.name.lower():
-                self.cached_files.append(path)
-        for path in root_project_dir.rglob('*.txt'):
-            if "tests" not in path.parts and "test" not in path.name.lower():
-                self.cached_files.append(path)
-        for path in root_project_dir.rglob('*.html'):
-            if "tests" not in path.parts and "test" not in path.name.lower():
-                self.cached_files.append(path)
+        for pattern in ['*.md', '*.rst', '*.txt', '*.html']:
+            for path in root_project_dir.rglob(pattern):
+                # Exclude test files and directories
+                if "tests" not in path.parts and "test" not in path.name.lower():
+                    self.cached_files.append(path)
         self.cached_files.sort(key=lambda x: len(x.parts))
 
     def _run(self, file_path: Optional[str] = None, line_number: Optional[int] = 0) -> str:
--- a/agents/prompts.py
+++ b/agents/prompts.py
@@ -89,12 +89,14 @@ Tasks:
 2. Identify critical interaction pathways from provided data
 3. Produce final components (max 8, optimally 5) following {project_type} patterns
 4. Ensure max 2 relationships between any component pair
-5. Provide single paragraph project explanation
+5. Provide single paragraph which explains the architecture and main flow
 
 CONSTRAINTS:
 - Use only provided analysis data - no additional tool calls
 - Focus on highest level architectural components
-- Exclude utility/logging components from final analysis"""
+- Exclude utility/logging components from final analysis
+- Give a brief overview of the architecture and main flow in a single paragraph called description
+- Describe the overview of the architecture and the main flow in a single paragraph, to be used as intro description to the project"""
 
 FEEDBACK_MESSAGE = """You are a software architect receiving expert feedback on your analysis.
 
--- a/static_analyzer/pylint_graph_transform.py
+++ b/static_analyzer/pylint_graph_transform.py
@@ -38,7 +38,7 @@ class DotGraphTransformer:
                 if entry.endswith('.py'):
                     has_python = True
                     break
-            if has_python and "test" not in current_dir and "example" not in current_dir:
+            if has_python and "test" not in str(current_dir) and "example" not in str(current_dir):
                 package_name = Path(current_dir).name
                 self.packages.append(package_name)
                 continue


agents/agent.py
```diff
@@ -87,14 +100,54 @@ def fix_source_code_reference_lines(self, analysis: AnalysisInsights):
...
+
+        # Check for direct directory path references
+        if dir_path_lower in lines:
+            return True
+
+        # Check for import statements that might reference the package
+        for part in parts:
+            part_lower = part.lower()
+            # Look for import patterns
+            if f"import {part_lower}" in lines or f"from {part_lower}" in lines:
```
brovatten, line 147: smart!

agents/agent.py
```diff
@@ -87,14 +100,54 @@ def fix_source_code_reference_lines(self, analysis: AnalysisInsights):
...
+                return True
+            # Look for path-like references
+            if f"/{part_lower}/" in lines or f".{part_lower}." in lines:
```
brovatten, line 150: you think directory references are relevant? example pls :D

agents/tools/get_method_invocations.py
```diff
@@ -12,12 +12,11 @@ class MethodInvocationsInput(BaseModel):
...
     name: str = "getMethodInvocationsTool"
     description: str = (
-        "Retrieves the **complete control flow graph (CFG)** for the entire project. "
-        "This tool provides a graphical representation (in DOT format) of "
-        "all function and method calls across the codebase, illustrating "
-        "the execution flow and inter-function relationships. "
-        "It's useful for understanding the overall architecture and call sequences "
-        "within the project. This tool requires no input arguments."
+        "Retrieves complete project control flow graph (CFG) in DOT format. "
+        "**FOUNDATIONAL TOOL** - Use once at the start of analysis to understand overall architecture. "
```
brovatten, line 16: Just curious, did you observe this to be the case? With "Once at the start..."

agents/tools/read_docs.py
```diff
@@ -36,19 +40,29 @@ def walk_dir(self, root_project_dir):
         Walk the directory and collect all markdown files.
         """
         for path in root_project_dir.rglob('*.md'):
+            if "tests" not in path.parts and "test" not in path.name.lower():
+                self.cached_files.append(path)
             self.cached_files.append(path)
+        for path in root_project_dir.rglob('*.rst'):
```
brovatten: how come not nested for loop? u expect we change the patterns inside these for loops? subjective but still :P

agents/tools/read_file.py
```diff
@@ -16,9 +16,10 @@ class ReadFileInput(BaseModel):
 class ReadFileTool(BaseTool):
     name: str = "readFile"
     description: str = (
-        "Reads a specified Python file, returning 200 lines of content centered around the requested line number. "
-        "This tool is useful for examining source code in detail, by focusing on a specific section of a file. "
-        "It only works with Python (.py) files."
+        "Reads specific Python file content around a target line number. "
+        "**PRECISION USE** - Only when specific implementation details are needed that CFG cannot provide. "
+        "Returns 300 lines centered on the requested line. "
```
brovatten, line 21: How did u come up with the number 300? just curious, did u test others?

agents/tools/read_structure.py
```diff
@@ -12,16 +12,16 @@
...
     name: str = "getClassHierarchy"
     description: str = (
-        "Retrieves the internal class structure and hierarchy for a given root package within a project. "
-        "This tool is useful for understanding how classes are organized, inherit from each other, "
-        "and relate within a specific software package or module. "
-        "The output is a detailed representation (e.g., a DOT graph format string) illustrating "
-        "these class relationships and their hierarchy. "
-        "It helps to visualize the internal architecture of a package."
+        "Retrieves class hierarchy and structure for a specific root package. "
+        "**STRATEGIC USE ONLY** - Use once per analysis phase when component relationships are unclear from CFG. "
```
brovatten, line 16: Nice!

outreach_utils/pr_util.py
```diff
@@ -169,9 +170,9 @@ def main():
     forked_ssh_url = fork_repository(original_owner, repo_name, github_token, organization)
 
     # # Define paths
-    forks_dir = Path("/home/ivan/StartUp/forks")
+    forks_dir = Path("/Users/imilev/StartUp/forks")
     repo_path = forks_dir / repo_name
-    source_dir = Path(f"/home/ivan/StartUp/GeneratedOnBoardings/{repo_name}")
+    source_dir = Path(f"/Users/imilev/StartUp/GeneratedOnBoardings/{repo_name}")
```
brovatten, line 175: Finally mac paths 😂

agents/agent.py
ivanmilevtues: So on the highest level of abstraction it makes more sense to reference a full package rather then listing all the classes.

I.e. imagine you have a restendpoints in the .rest package.
Here I think it makes more sense to put as a reference the rest folder then listing all the files in it. Then when you click the component and expand you can actually see all the files and how they interact with each other, but on this level even if we are referencing them, they are not on the diagram or explanations so makes no sense.

agents/tools/get_method_invocations.py
ivanmilevtues: Some models like i.e. `claude` like using tools too much. So Claude could get stuck into using tools again and again that is why I added these statements.

agents/tools/read_docs.py
ivanmilevtues: I don't understand this can you give example how you would nest?
.rglob(pattern) just lists the files with certain pattern. That is why we have multiple of those, you cannot combine all patterns at once ;/

agents/tools/read_file.py
ivanmilevtues: So I noticed that most files are about 300 lines and `claude` would usually read the first 300 lines in 3 steps (around 0, 100, 200) and to reduce tool calling and steps I bumped it up to 300.

agents/tools/get_method_invocations.py
brovatten: Okay cool, didn't know that :)

agents/tools/read_docs.py
brovatten: Sorry, my comment is unclear. I meant doing 4 separate loops here as opposed to this:
```
for file_extension in ['*.md', ...] :  
        for path in root_project_dir.rglob(file_extension):
              if "tests" not in path.parts and "test" not in path.name.lower():
                  self.cached_files.append(path)
```
Just curious, not super relevant

agents/tools/read_file.py
brovatten: I see, nice observation!

PR #40, opened 2025-08-05
Fix broken demo

Zonotora:
demo.py does not work if the environment variable "REPO_ROOT" is not set. 
- Check if "REPO_ROOT" is set, or default to "repositories"
- Check if one of the API keys is set, otherwise fail fast.

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -12,9 +12,33 @@ from agents.agent_responses import AnalysisInsights
 from diagram_analysis import DiagramGenerator
 from logging_config import setup_logging
 from output_generators.markdown import generate_markdown_file
-from repo_utils import store_token, clone_repository, upload_onboarding_materials, get_branch
-from utils import create_temp_repo_folder, caching_enabled, remove_temp_repo_folder
+from repo_utils import (
+    clone_repository,
+    get_branch,
+    store_token,
+    upload_onboarding_materials,
+)
+from utils import caching_enabled, create_temp_repo_folder, remove_temp_repo_folder
 
+
+def validate_env_vars():
+    api_provider_keys = ["OPENAI_API_KEY", "ANTHROPIC_API_KEY", "GOOGLE_API_KEY", "AWS_BEARER_TOKEN_BEDROCK"]
+    api_env_keys = [(key, os.getenv(key)) for key in api_provider_keys if os.getenv(key) is not None]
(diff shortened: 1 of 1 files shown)


demo.py
```diff
@@ -12,9 +12,29 @@
...
+    exit(1)
+elif len(API_ENV_KEYS) > 1:
+    logging.error(f"Detected multiple API keys set ({API_ENV_KEYS}), set ONE of the following: {API_PROVIDER_KEYS}")
+    exit(2)
+
+REPO_ROOT = os.getenv("REPO_ROOT")
+if not REPO_ROOT:
+    logging.error("REPO_ROOT environment variable not set, setting REPO_ROOT environment variable to 'repositories'")
+    REPO_ROOT = "repositories"
+    os.environ["REPO_ROOT"] = REPO_ROOT
```
ivanmilevtues: All of this validation has to go after we call `load_dotenv` at line 144, as at this point the .env is not loaded.

I suppose it is better to have it in a function then, validate_env_vars or something else.

demo.py
```diff
@@ -27,7 +47,7 @@ def onboarding_materials_exist(project_name: str, source_dir: str):
 
 def generate_docs(repo_name: str, temp_repo_folder: Path, repo_url: str = None):
     # Create directories if they don't exist
-    repos_dir = Path(os.getenv("REPO_ROOT"))
+    repos_dir = Path(REPO_ROOT)
```
ivanmilevtues: If you move the validation to a separate func then we'd have to revert this back to os.genev

demo.py
```diff
@@ -129,11 +149,11 @@ def copy_files(temp_folder: Path, output_dir: Path):
...
-            
+
             # Copy markdown files to output directory if specified
             if args.output_dir:
                 copy_files(temp_repo_folder, args.output_dir)
-                
+
         except Exception as e:
             logging.error(f"Failed to generate docs for {repo}: {e}")
         # finally:
```
ivanmilevtues, line 164: Not a comment for you, but we have to uncomment this :D

ivanmilevtues (changes requested):
Thanks for the contribution. Slight changes requested :)

ivanmilevtues (approved):
Amazing thanks

Issue #72, opened 2025-10-01
Cleaner way for UV usage


seppzer0:
Hi! I noticed that there is a more proper way to use UV to install and manage dependencies.

## Current Solution

Right now the dependency management is done via `uv pip install -r requirements.txt` in [setup.py](https://github.com/CodeBoarding/CodeBoarding/blob/f066a7d1fad66101799a65202994c68772ff795d/setup.py#L35C1-L52C20):

```py
def install_requirements():
    """Install Python requirements using uv pip."""
    print("Step: Requirements installation started")
    
    requirements_file = 'requirements.txt'

    if not os.path.exists(requirements_file):
        print("Step: Requirements installation finished: failure - No requirements file found")
        sys.exit(1)

    try:
        subprocess.run([
            'uv', 'pip', 'install', '-r', requirements_file
        ], check=True, capture_output=True, text=True)
        print("Step: Requirements installation finished: success")
    except subprocess.CalledProcessError as e:
        print(f"Step: Requirements installation finished: failure - {e}")
        sys.exit(1)
```

However, this is not the only way.
With a properly configured project, we can use `uv sync` functionality to manage the .venv and install dependencies simultaneously.

## Proposed Changes

Add UV configuration to prepare the entire venv and remove standalone dependency download from setup.py .

ivanmilevtues:
That sounds like a good solution, however we still need our install scripts as it gets some additional dependencies for the statci analysis to be done.

I don't mind the uv sync configuration which you are suggesting. So you can go ahead and do that for us! Thank you @seppzer0.

You can join our discord for more info and collaboration :)

PR #75, opened 2025-10-02
build: update dependency management

seppzer0:
## Related Items

Resolves https://github.com/CodeBoarding/CodeBoarding/issues/72

## Changes

- Added `pyproject.toml` as a central configuration for everything Python-related
- Added `.python-version` to automatically manage Python version via uv
- Removed redundant `requirements.txt`, moved dependency info into `pyproject.toml`
- Removed redundant logic in `setup.py`
- Added `uv.lock` for frozen dependency versions
- (!) Bumped `numpy` from 1.26.4 to 2.1.0 as it was causing dependency conflict with `langchain-community==0.3.27`
- (!) Set minimum Python version to 3.12 as it is required for `langchain-aws==0.2.27`
- Updated `README.md`

image
image


.python_version
```diff
@@ -0,0 +1 @@
+3.13
```
ivanmilevtues, line 1: Can we use python 3.10 please.

pyproject.toml
```diff
@@ -0,0 +1,244 @@
...
+    "wcwidth==0.2.13",
+    "websockets==15.0.1",
+    "werkzeug==3.1.3",
+    "xxhash==3.5.0",
+    "yarl==1.20.1",
+    "zstandard==0.23.0"
+]
+
+[project.urls]
+Repository = "https://github.com/CodeBerg/CodeBerg"
```
ivanmilevtues: Url is wrong: https://github.com/CodeBoarding/CodeBoarding

pyproject.toml
```diff
@@ -0,0 +1,244 @@
...
+Documentation = "https://www.codeboarding.org"
```
ivanmilevtues: The link here should be: https://github.com/CodeBoarding/CodeBoarding/blob/main/.codeboarding/overview.md

README.md
```diff
@@ -56,9 +56,10 @@ graph LR
 Setup the environment:
 
 ```bash
-uv venv --python 3.11
+python3 -m pip install uv==0.8.22
```
ivanmilevtues: I prefer this to be with just python, and do we need to specify the exact uv version, probably some peopel already have uv installed :)

README.md
```diff
@@ -56,9 +56,10 @@ graph LR
...
+uv sync --frozen
 source .venv/bin/activate
-python setup.py
+python3 setup.py
```
ivanmilevtues: python setup.py

pyproject.toml
```diff
@@ -0,0 +1,244 @@
+[project]
+name = "codeboarding"
+version = "0.2.0"
+description = "🪄 Interactive Diagrams for Code"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
```
ivanmilevtues, line 7: requires-python=">=3.10"

(optional)Can you somehow clean-up the dependencies, I know that this is what the requirements.txt says and probably we should've cleaned them ourselves, but if we can have only the main deps, without the secondary that would be awesome!

ivanmilevtues (changes requested):
Very good job overall, few minor comments, if you can address these we should be ready to go. I will try the install on Win/Lin and Mac myself and if it all works it will be ready for merge!

pyproject.toml
seppzer0: Yes, that's an honest mistake. Will fix.

README.md
seppzer0: > do we need to specify the exact uv version, probably some peopel already have uv installed :)

I would say yes, specifying UV version would be best.
Main reason is that UV is still in development, and there can be breaking API changes.

E.g., validation of pyproject.toml : 0.8.x has extra enforced checks that 0.6.x doesn't. This can break project's configuration process overall.

It's a general practice to have as many of the project dependencies specified to a certain version.

> I prefer this to be with just python

`python3` comes from when `python` still could point to Python 2.7. This is only an extra safety belt.
I do recognize that now it can be somewhat redundant.
If you think it still makes sense to change to `python`, please let me know.

pyproject.toml
seppzer0: > requires-python=">=3.10"

As I mentioned in the description, there would be a problem as one of the dependencies requires Python `>3.11`.

> optional)Can you somehow clean-up the dependencies

I was thinking about that but decided not to do so because of being unsure of the project prefers it's dependencies :)
It might take quite some time, but I'll look into it.

README.md
seppzer0: Decided I don't have additional comments here. Done.

pyproject.toml
seppzer0: Done

pyproject.toml
seppzer0: Done

pyproject.toml
seppzer0: Reviewed the dependencies, left only the direct ones.
For reference, I attached a requirements.txt file for which all dependencies (direct+transitive) were specified.
[requirements-bak.txt](https://github.com/user-attachments/files/22663746/requirements-bak.txt)

README.md
ivanmilevtues: Hmm I think the default for `python` is now python3 for newer version of ubuntu and generally newer OSs

pyproject.toml
ivanmilevtues: Awesome, let's do 3.11. I agree with you, also checked that 3.10 is not LTS, for some reason I thought it is an LTS. We can keep it with 3.11 :)

pyproject.toml
```diff
@@ -0,0 +1,77 @@
+[project]
+name = "codeboarding"
+version = "0.2.0"
+description = "🪄 Interactive Diagrams for Code"
+readme = "README.md"
+requires-python = ">=3.12"
```
ivanmilevtues, line 6: 3.12 is not the same as the one we have in .python_version :)

pyproject.toml
seppzer0: Again, 3.12 is the minimum allowed version :)
That is evident by the second screenshot in PR description.

pyproject.toml
seppzer0: That is intentional.
I thought that it would make sense for the project to move to 3.13, but indicate that technically the lowest supported version is 3.12 .

In case some of the users have reason to use 3.12 .

But I suppose this is open for discussion. Let me know what you think.

ivanmilevtues:
I agree with this co-pilot comment:
Installing uv via pip when the project is designed to use uv for dependency management creates a circular dependency. Consider using the official uv installation method (curl/powershell script) or document that users should install uv first through their system package manager.

As i.e. on my Mac I don't have a global level python installed I always use uv and a venv.

Apart from that I have tested it so, I suggest you just skip this step in the readme, link the uv install link. Update with main and it is ready for merge!

pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+    "py2puml==0.10.0",
+    "pydot==3.0.4",
+    "pyinstaller==6.13.0",
+    "pylint==3.3.6",
+    "pyright==1.1.403",
+    "pytest==8.3.5",
+    "python-call-graph==2.1.3",
+    "pyyaml==6.0.3",
+    "regex==2024.11.6",
+    "rich==12.6.0", # pdm
```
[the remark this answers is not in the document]
seppzer0: Rudimentary comment when cleaning up dependencies. Fixed.

ivanmilevtues:
It is all good, I am merging. Well done! Are you part of our discord server, i will give you the HacktoberFest role:
https://discord.gg/T5zHTJYFuy

Issue #74, opened 2025-10-02
TypeScript config resolution is not dynamic


ivanmilevtues:
We support TypeScript via our LSP integration. However at the moment we are looking for environment setup only at the root-directory, and in the case of a mono-repo with multi-project setup (i.e. django + react, or something else) we won't be able to scan the typescript.

## What has to be done?
- We need to add a scanner for the typescript config files,
- This scanner has to be able to use the concrete client for the concrete sub-directories (not use one config for all directories in the case of a multi-typescript project setup)
- You can try to run it with the [humanlayer](https://github.com/humanlayer/humanlayer) project as it doesn't contain typescript at its main layer.

Anshumancanrock:
@ivanmilevtues Please assign this to me. I already started working on this 🙏

PR #79, opened 2025-10-04
feat: Implement dynamic TypeScript config resolution for monorepos

SinghCod3r:
# TypeScript Config Resolution for Monorepos

## Summary
This PR implements dynamic TypeScript config resolution for monorepo structures, allowing proper scanning of TypeScript files in multi-project setups.

## Problem
Previously, the TypeScript LSP integration only looked for environment setup at the root directory, preventing proper scanning of TypeScript files in monorepos with multiple projects.

### 1. TypeScript Config Scanner (`typescript_scanner.py`)
- Recursively finds `tsconfig.json` and `jsconfig.json` files throughout the repository
- Skips `node_modules` directories automatically
- Validates that projects contain actual TypeScript/JavaScript files
- Handles UTF-8 BOM in JSON files using `utf-8-sig` encoding

### 2. Dynamic Client Creation (`__init__.py`)
- Modified `create_clients` function to detect multiple TypeScript projects
- Creates individual `TypeScriptClient` instances for each valid project directory
- Maintains backward compatibility with single-project repositories

### 3. Project-Specific Configuration (`typescript_client.py`)
- Each TypeScript client handles config files relative to its `project_path`
- Proper workspace folder management for multi-project setups
- Individual project initialization and validation

## Key Features
- **Recursive scanning**: Finds TypeScript projects anywhere in the repository structure
- **Project validation**: Ensures detected projects contain actual TypeScript/JavaScript files
- **Multiple project support**: Creates separate LSP clients for each TypeScript project
- **Node_modules exclusion**: Automatically skips dependency directories
- **Robust error handling**: Gracefully handles malformed config files

## Testing
- Verified with test monorepo structure containing multiple TypeScript projects
- Each project correctly gets its own dedicated TypeScript client
- UTF-8 BOM files are handled properly
- Backward compatibility maintained for single-project repositories

## Files Changed
- `static_analyzer/typescript_scanner.py` (new)
- `static_analyzer/__init__.py` (modified)
- `static_analyzer/lsp_client/typescript_client.py` (enhanced)

## Dependencies
- Added `pathspec` package for efficient file pattern matching

This implementation resolves the original issue where only root-level TypeScript configurations were considered, enabling proper TypeScript analysis in complex monorepo environments.

### DIFF
--- a/PR_TYPE_SCRIPT_CONFIG_RESOLUTION.md
+++ b/PR_TYPE_SCRIPT_CONFIG_RESOLUTION.md
@@ -1,0 +1,48 @@
+# TypeScript Config Resolution for Monorepos
+
+## Summary
+This PR implements dynamic TypeScript config resolution for monorepo structures, allowing proper scanning of TypeScript files in multi-project setups.
+
+## Problem
+Previously, the TypeScript LSP integration only looked for environment setup at the root directory, preventing proper scanning of TypeScript files in monorepos with multiple projects.
+
+## Solution
+
+### 1. TypeScript Config Scanner (`typescript_scanner.py`)
+- Recursively finds `tsconfig.json` and `jsconfig.json` files throughout the repository
+- Skips `node_modules` directories automatically
+- Validates that projects contain actual TypeScript/JavaScript files
+- Handles UTF-8 BOM in JSON files using `utf-8-sig` encoding
+
+### 2. Dynamic Client Creation (`__init__.py`)
+- Modified `create_clients` function to detect multiple TypeScript projects
+- Creates individual `TypeScriptClient` instances for each valid project directory
+- Maintains backward compatibility with single-project repositories
+
+### 3. Project-Specific Configuration (`typescript_client.py`)
+- Each TypeScript client handles config files relative to its `project_path`
+- Proper workspace folder management for multi-project setups
+- Individual project initialization and validation
+
+## Key Features
+- **Recursive scanning**: Finds TypeScript projects anywhere in the repository structure
+- **Project validation**: Ensures detected projects contain actual TypeScript/JavaScript files
+- **Multiple project support**: Creates separate LSP clients for each TypeScript project
+- **Node_modules exclusion**: Automatically skips dependency directories
+- **Robust error handling**: Gracefully handles malformed config files
+
+## Testing
+- Verified with test monorepo structure containing multiple TypeScript projects
+- Each project correctly gets its own dedicated TypeScript client
+- UTF-8 BOM files are handled properly
+- Backward compatibility maintained for single-project repositories
+
+## Files Changed
+- `static_analyzer/typescript_scanner.py` (new)
+- `static_analyzer/__init__.py` (modified)
+- `static_analyzer/lsp_client/typescript_client.py` (enhanced)
+
+## Dependencies
+- Added `pathspec` package for efficient file pattern matching
+
+This implementation resolves the original issue where only root-level TypeScript configurations were considered, enabling proper TypeScript analysis in complex monorepo environments.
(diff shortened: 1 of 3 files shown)


PR_TYPE_SCRIPT_CONFIG_RESOLUTION.md
```diff
@@ -0,0 +1,48 @@
+# TypeScript Config Resolution for Monorepos
```
ivanmilevtues, line 1: Remove this file, and just add it as the description of the PR

static_analyzer/__init__.py
```diff
@@ -5,21 +5,44 @@
...
+    
+    # Create clients for each TypeScript project found
+    for project_dir in ts_projects:
+        try:
+            # Create a special TypeScript client for this specific project
+            client = TypeScriptClient(
+                language=ProgrammingLanguage(language='TypeScript', size=0, percentage=0, suffixes=['.ts', '.tsx', '.js', '.jsx'], server_commands=None),
+                project_path=project_dir
+            )
+            clients.append(client)
```
ivanmilevtues, line 28: I don't really like this pre-patched creation, as now it is confusing, we create something here than we skip it typescript.

I would suggest to unify it like:
```python
for pl in programming_language:
        if not pl.is_supported_lang():
            continue
          if pl.language in ['TypeScript']:
               ts_scanner = TypeScript....
               ts_projects = ts_scanner.get...
                # now here you add your new logic which appends the client to the clients.
                # and now it is unified we don't need to skip stuff and so on
            else:
                clients.append(LSPClient(language=pl, project_path=repository_path))
```
This way the scanner will only run for typescript projects and the logic of the initailization is all at the same place.

static_analyzer/typescript_scanner.py
```diff
@@ -0,0 +1,95 @@
...
+            for config_path in config_paths:
+                project_dir = config_path.parent
+                
+                # Skip node_modules directories
+                if 'node_modules' in str(project_dir):
+                    continue
+                
+                # Parse the config file to validate it's a valid TypeScript project
+                config_data = self._parse_config_file(config_path)
+                if config_data:
```
ivanmilevtues, line 41: you can filter the valid projects already here:
```python
if config_data and self._validate_project
```

static_analyzer/typescript_scanner.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        Get all directories that contain valid TypeScript projects.
+        
+        Returns:
+            List[Path]: List of project directory paths
+        """
+        projects = self.scan()
+        return [project['project_dir'] for project in projects if project['is_valid']]
+
+    def get_configurations(self) -> List[Dict]:
```
ivanmilevtues, line 88: I don't think you need this method, we can keep it at the scan method just. To me it is clear what it gives back.

static_analyzer/typescript_scanner.py
```diff
@@ -0,0 +1,95 @@
...
+            # Filter out node_modules files
+            ts_files = [f for f in ts_files if 'node_modules' not in str(f)]
+            
+            if ts_files:
+                return True
+        
+        logger.debug(f"No TypeScript/JavaScript files found in project: {project_dir}")
+        return False
+
+    def get_project_directories(self) -> List[Path]:
```
ivanmilevtues, line 78: Make the `scan` method return directly the directories and use that directly in the __init__.py

ivanmilevtues (changes requested):
Awesome work, just small refinements on style and we should be ready to go!

ivanmilevtues:
@SinghCod3r Can you also comment in #74 so I can assign you to the task, I cannot if you haven't comment or at least I don't know how to :)

PR #80, opened 2025-10-04
feat: add TypeScript multi-project support for mono-repos

### LINKED ISSUE #74: TypeScript config resolution is not dynamic

Anshumancanrock:
## What does this PR do?

This PR adds support for TypeScript multi-project configurations in mono-repo setups. Previously, CodeBoarding only looked for TypeScript configuration files (`tsconfig.json`/`jsconfig.json`) at the repository root, which prevented proper analysis of mono-repos where TypeScript projects exist in subdirectories.

- Fixes #74 

**Changes:**
- Added `TypeScriptConfigScanner` to recursively discover all TypeScript projects in the repository
- Modified `create_clients()` to create separate LSP client instances for each discovered TypeScript project
- Each TypeScript project now gets its own properly scoped LSP server with the correct configuration

**Impact:**
-  Supports mono-repos with multiple TypeScript projects 
-  Each project uses its own `tsconfig.json` with proper scoping
-  More efficient analysis (each LSP server loads only its project)
-  Backward compatible (falls back to root if no configs found)

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -5,6 +5,7 @@ from typing import List
 from static_analyzer.lsp_client.client import LSPClient
 from static_analyzer.lsp_client.typescript_client import TypeScriptClient
 from static_analyzer.programming_language import ProgrammingLanguage
+from static_analyzer.typescript_config_scanner import TypeScriptConfigScanner
 
 logger = logging.getLogger(__name__)
 
@@ -17,7 +18,19 @@ def create_clients(programming_languages: List[ProgrammingLanguage], repository_
             continue
         try:
             if pl.language in ['TypeScript']:
-                clients.append(TypeScriptClient(language=pl, project_path=repository_path))
+                # For TypeScript, scan for multiple project configurations (mono-repo support)
+                config_scanner = TypeScriptConfigScanner(repository_path)
+                typescript_projects = config_scanner.find_typescript_projects()
+                
+                if typescript_projects:
+                    # Create a separate client for each TypeScript project found
+                    for project_path in typescript_projects:
+                        logger.info(f"Creating TypeScript client for project at: {project_path.relative_to(repository_path)}")
+                        clients.append(TypeScriptClient(language=pl, project_path=project_path))
+                else:
+                    # Fallback: No config files found, use repository root
+                    logger.info("No TypeScript config files found, using repository root")
+                    clients.append(TypeScriptClient(language=pl, project_path=repository_path))
             else:
                 clients.append(LSPClient(language=pl, project_path=repository_path))
         except RuntimeError as e:
--- a/static_analyzer/typescript_config_scanner.py
+++ b/static_analyzer/typescript_config_scanner.py
@@ -1,0 +1,47 @@
+import logging
+from pathlib import Path
+from typing import List
+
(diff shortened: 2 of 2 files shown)


Anshumancanrock:
@ivanmilevtues Please review. Thanks !

static_analyzer/__init__.py
```diff
@@ -17,7 +18,19 @@ def create_clients(programming_languages: List[ProgrammingLanguage], repository_
...
+                
+                if typescript_projects:
+                    # Create a separate client for each TypeScript project found
+                    for project_path in typescript_projects:
+                        logger.info(f"Creating TypeScript client for project at: {project_path.relative_to(repository_path)}")
+                        clients.append(TypeScriptClient(language=pl, project_path=project_path))
+                else:
+                    # Fallback: No config files found, use repository root
+                    logger.info("No TypeScript config files found, using repository root")
+                    clients.append(TypeScriptClient(language=pl, project_path=repository_path))
```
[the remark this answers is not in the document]
Anshumancanrock: I double-checked the indentation and the code is actually correct. The else block at line 30 is properly paired with the "if typescript_projects:" at line 25 and the fallback code (lines 31-33) is correctly indented inside it.

The logic works as intended, when the scanner finds no TypeScript configs (empty list), the else block executes and falls back to the repository root. I've tested this scenario and it works perfectly.

No changes needed here!

static_analyzer/typescript_config_scanner.py
```diff
@@ -0,0 +1,47 @@
...
+            # Find all config files recursively
+            for config_path in self.repo_location.rglob(config_file):
+                if config_path.is_file():
+                    project_dir = config_path.parent
+                    
+                    # Avoid duplicates (e.g., if both tsconfig.json and jsconfig.json exist)
+                    if project_dir not in seen_dirs:
+                        seen_dirs.add(project_dir)
+                        project_roots.append(project_dir)
+                        logger.info(f"Found TypeScript project at: {project_dir.relative_to(self.repo_location)}")
```
ivanmilevtues: You can remove this logger here, I think it is redundant with the final logging message.

When that is done. I think we are ready to merge :)

ivanmilevtues (changes requested):
Almost done, just a nit-pick when you resolve it I will merge it in :) 

Well done!

static_analyzer/typescript_config_scanner.py
Anshumancanrock: Done !! @ivanmilevtues

ivanmilevtues:
@Anshumancanrock sorry I meant just the logger.info, you seen logic is good, please put it back :)

                    if project_dir not in seen_dirs:
                        seen_dirs.add(project_dir)
                        project_roots.append(project_dir)

Issue #81, opened 2025-10-09
refactor: Data control and logic implementation update


seppzer0:
## Overview

There are several locations in Python codebase can be improved in terms of tidiness and performance. Given that project uses Python 3.10+, there are some implementation features that can make the code more maintanable.

Additionally, some of the datatypes don't match to what is declared in type hints. This can be verified with pyright.

The app's logic is not expected to be changed.

## Suggestion

There can be general refactoring done for the Python codebase. There are several objective points that can be taken into account.

`#Hacktoberfest2025`

PR #82, opened 2025-10-09
refactor: update trivial implementations and datatypes

seppzer0:
## Related Items

Closes https://github.com/CodeBoarding/CodeBoarding/issues/81

## Changes

- Omit unused variables
- Remove f-strings where required
- Replace `if-elif-else` with `match-case`
- Replace dynamic arrays with fixed size
- Update deprecated `utcnow()`

### DIFF
--- a/agents/tools/read_docs.py
+++ b/agents/tools/read_docs.py
@@ -142,5 +142,4 @@ class ReadDocsTool(BaseTool):
         for i in range(len(full) - len(sub) + 1):
             if full[i:i + len(sub)] == sub:
                 return True
-        
         return False


setup.py
```diff
@@ -145,15 +145,16 @@ def download_binary_from_gdrive():
     }
 
     system = platform.system()
```
seppzer0, line 147: Forgot to remove this, as match-case structure makes the variable redundant.
Will fix with enough comments collected.

local_app.py
```diff
@@ -212,8 +214,9 @@ async def get_job(job_id: str):
     if job["status"] == JobStatus.COMPLETED:
         if job.get("result"):
             # Check if result is a JSON string containing files
+            # NOTE: JSON is imported here to avoid unbound usage in exception
+            import json
             try:
```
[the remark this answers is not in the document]
seppzer0: I agree with this, but did not do so because was not sure if non-module level import here is intended. I moved it one level higher with an appropriate comment.

@ivanmilevtues , if you also agree with this, I'll add an update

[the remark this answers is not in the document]
ivanmilevtues: You can move it on top, it will work fine. We are already importing on a global level and json is a third party dependency which we use anyway on global level in other files.

github_action.py
```diff
@@ -101,20 +101,49 @@ def generate_analysis(repo_url: str, source_branch: str, target_branch: str, ext
...
-    elif extension == ".html":
-        generate_html(analysis_files, repo_name, repo_url, target_branch, temp_repo_folder, output_dir)
-    elif extension == ".mdx":
-        generate_mdx(analysis_files, repo_name, repo_url, target_branch, temp_repo_folder, output_dir)
-    elif extension == ".rst":
-        generate_rst(analysis_files, repo_name, repo_url, target_branch, temp_repo_folder, output_dir)
-    else:
-        raise ValueError(f"Unsupported extension: {extension}")
+    match extension:
+        case ".md":
```
ivanmilevtues, line 110: This kind of formatting seems too much. No need to have every funtion call on 10 lines :).

I usually use 120 chars of length for the lines not 80.

seppzer0: I would say it's subjective and is a matter of preference :)
Updated the formatting to match the project's style.

setup.py
seppzer0: Nevermind, it's used in the last `case` for error output.
Makes sense to have a re-usable variable

agents/prompts/prompt_factory.py
```diff
@@ -36,21 +36,22 @@ def __init__(self, llm_type: LLMType = LLMType.GEMINI_FLASH, prompt_type: Prompt
     
     def _create_prompt_factory(self) -> AbstractPromptFactory:
         """Create the appropriate prompt factory based on LLM type and prompt type."""
-        if self.llm_type == LLMType.GEMINI_FLASH:
-            if self.prompt_type == PromptType.BIDIRECTIONAL:
+        match self.llm_type:
+            case LLMType.GEMINI_FLASH:
+                if self.prompt_type == PromptType.BIDIRECTIONAL:
+                    return GeminiFlashBidirectionalPromptFactory()
+                else:
```
ivanmilevtues: No need to have this else statement, as the if is with a return.
Just an aesthetic nitpick.

Same for the folloing if/elses

agents/prompts/prompt_factory.py
```diff
@@ -36,21 +36,22 @@ def __init__(self, llm_type: LLMType = LLMType.GEMINI_FLASH, prompt_type: Prompt
...
+                    return GeminiFlashUnidirectionalPromptFactory()
+            case LLMType.CLAUDE: #Adding Claude support
+                if self.prompt_type == PromptType.BIDIRECTIONAL:
+                    from .claude_prompts_bidirectional import ClaudeBidirectionalPromptFactory
```
ivanmilevtues: Move all the imports on top, no need for them to be local.

github_action.py
ivanmilevtues: It is up-to personal preference I agree, thanks for updating.

local_app.py
```diff
@@ -102,8 +105,8 @@ async def generate_onboarding(job_id: str):
                     raise ValueError("REPO_ROOT environment variable not set")
 
                 # run generation
-                repo_name = extract_repo_name(job["repo_url"])
-                generated_repo_name = await run_in_threadpool(
+                _ = extract_repo_name(job["repo_url"])
```
ivanmilevtues: remove the assign whatsoever instead of calling the assing.
These methods are not loading lazily or something like that so you can just call them:
extract_repo_name(..)
run_int_threadpool(..)

You can also remove extrat_repo_name, it doesn't do anything here, I double checked :)

ivanmilevtues (changes requested):
It is almost ready to go, just a few nitpicks and we are ready to ship this!

Thanks for spending the time on this @seppzer0!

local_app.py
seppzer0: Done. Left only `await run_int_threadpool(..)` .

agents/prompts/prompt_factory.py
seppzer0: Done

agents/prompts/prompt_factory.py
seppzer0: > No need to have this else statement, as the if is with a return.

I agree. It is a better style choice.

ivanmilevtues:
Nice, we are done here. Going to main :)

PR #91, opened 2025-10-20
Test evals

brovatten:
Please take a quick look here - the `end_to_end_eval` is quite slow & still running. I will squash this on approval :) 

Reverted blacksmith, it could not overcome the bandwidth issues GHA had with API calls. Rather than the SECURITY.md file we discussed, I went for separate reports instead, much easier to maintain imo. 

Keeping the static analysis as part of the CI/CD and leaving it up to us to generate when evals are needed.

TY for review 🙏

### DIFF
--- a/evals/reports/static-analysis-report.md
+++ b/evals/reports/static-analysis-report.md
@@ -1,16 +1,16 @@
 # Static Analysis Performance Evaluation
 
-**Generated:** 2025-11-04T12:32:05.883790
+**Generated:** 2025-11-04T12:52:32.533065
 
-**Evaluation Time:** 2.24 seconds
+**Evaluation Time:** 2.83 seconds
 
 ### Summary
 
 | Project | Language | Status | Time (s) | Files | Errors |
 |---------|----------|--------|----------|-------|--------|
-| markitdown | Python | ❌ Failed | 0.54 | 0 | 0 |
-| tsoa | TypeScript | ❌ Failed | 1.15 | 0 | 0 |
-| cobra | Go | ❌ Failed | 0.54 | 0 | 0 |
+| markitdown | Python | ❌ Failed | 0.68 | 0 | 0 |
+| tsoa | TypeScript | ❌ Failed | 1.38 | 0 | 0 |
+| cobra | Go | ❌ Failed | 0.77 | 0 | 0 |
 ## System Specifications
 
 **Operating System:** Linux (Linux-6.11.0-1018-azure-x86_64-with-glibc2.39)


agents/agent.py
```diff
@@ -108,7 +110,7 @@ def _initialize_llm(self):
         elif self.ollama_base_url:
             logging.info("Using Ollama LLM")
             return ChatOllama(
-                model="qwen3:30b",
+                model="qwen3:30b", #qwen3:30b
```
ivanmilevtues: left-out comment

agents/monitoring.py
```diff
@@ -0,0 +1,72 @@
...
+            tool_name = self._tool_names[run_id]
+        self.tool_errors[tool_name] += 1
+        # Clean up any in-flight timing
+        if run_id and run_id in self._tool_start_times:
+            self._tool_start_times.pop(run_id, None)
+            self._tool_names.pop(run_id, None)
+
+
+
+### Build this basecallbackhandler into the agent callbacks if the run is an evaluation run
```
ivanmilevtues: this comment can be removed.

diagram_analysis/diagram_generator.py
```diff
@@ -175,4 +186,94 @@ def generate_analysis(self):
...
         logger.info(f"Analysis complete. Generated {len(files)} analysis files")
         print("Generated analysis files: %s", [os.path.abspath(file) for file in files])
+        
+        # Collect and save monitoring results if enabled
+        self._collect_and_save_monitoring_results()
+        
         return files
+
+    def generate_static_analysis(self):
+        import time
```
ivanmilevtues: No need to have local imports.

evals/reports/static-analysis-report.md
```diff
@@ -0,0 +1,13 @@
+# Static Analysis Performance Evaluation
```
ivanmilevtues, line 1: very cool format

evals/end_to_end_eval.py
```diff
@@ -0,0 +1,367 @@
...
+from typing import Dict, Any, List
+
+# Add the parent directory to the Python path
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+from dotenv import load_dotenv
+from evals.report_generator import generate_header, generate_e2e_section, write_report
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
```
ivanmilevtues, line 26: I saw that we are using print statements in the code, I don't mind either prints or logs, but no need to have both at the same time, so I propose to pick logs and just keep it consistent across the codebase.

evals/end_to_end_eval.py
```diff
@@ -0,0 +1,367 @@
...
+            "project": project_name,
+            "url": repo_url,
+            "expected_language": project_info.get("expected_language"),
+            "total_time_seconds": elapsed_time,
+            "error": str(e),
+            "success": False
+        }
+
+
+def run_end_to_end_eval(projects=None):
```
ivanmilevtues, line 223: I would say either remove the projects parameter, as it is never used (we never pass it) or add argparse and add the name, url, expected_language as parameters that we can use with the script from the console.

Tbh I'd probably like the argparse addition as I personally like a few projects so I'd test with all of them (my taste changes with time XDD). If curios I am thinking about fastAPI, django and flask :)

evals/static_analysis_eval.py
```diff
@@ -0,0 +1,262 @@
+#!/usr/bin/env python3
```
ivanmilevtues, line 1: I think adding the machine specs would be a nice addition:
CPU core count, CPU speed

These should be easily accessible from os. Unsure for speed, but count definetely.

setup.py
```diff
@@ -259,7 +258,7 @@ def init_dot_env_file():
 # ============================================================================
 
 # LLM Provider Configuration (uncomment and configure one)
-OLLAMA_BASE_URL=http://localhost:11434
+# OLLAMA_BASE_URL=http://localhost:11434
```
ivanmilevtues: Why did you comment this line?
I have it as the only uncommented line to prevent people running CodeBoarding with an API key which is set in their environment by default and use money on something they didn't want to :)

agents/monitoring.py
```diff
@@ -0,0 +1,72 @@
+import sys, json, time
+from pathlib import Path
+from collections import defaultdict
+from typing import Any, Dict, Optional
+
+from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.outputs import LLMResult
+
+# ----- Callback that counts tools + tokens -----
+class UsageAndToolAggregator(BaseCallbackHandler):
```
ivanmilevtues: Very cool never used this, I have just used the LLMs before: https://python.langchain.com/docs/how_to/chat_token_usage_tracking/

ivanmilevtues (changes requested):
Well, done! A lot of code has been created!

I read it carefully (haven't run), my main comment is around the design of monitoring/timing methods, imo we should do it with a decorator to keep logic cleaner and avoid propagating parameters everywhere. Add a global flag which activates the decorators (if not FLAT -> don't do anything).

Down to discuss this further and see what will work best!

ivanmilevtues:
Also update there is merge conflict, shouldn't be anything big :)

setup.py
brovatten: When the GH action builds with the setup, and fetches the `AWS_BEDROCK_KEY` GH secret, it complains that two LLM options are set. But since we are not using calls any more, I will bring it back!

OK, smart!

agents/monitoring.py
brovatten: These `CallbackHandler` was my nemisis during my whole thesis 😂 This metadata is nice when it works, but for example, the `total_tokens` I tried to fetch always gave 0

evals/end_to_end_eval.py
brovatten: Yup I will remove this!

I think keeping the same projects are better for the long-term purpose of evals. Switching them up - I'd say we either add them to the eval list, or test them with the `demo.py`

edit: this is supposed to be the purpose of projects, like a few different configs we save in the repo and load in to test ourselves; or a suite of python tests, go-tests, etc., but it's redundant atm :)

evals/static_analysis_eval.py
brovatten: Yup good idea!

brovatten: I will also add the git user to this

evals/end_to_end_eval.py
ivanmilevtues: Yea the demo.py will do now that you said it :D

evals/end_to_end_eval.py
```diff
@@ -0,0 +1,357 @@
...
+                "name": "cobra",
+                "url": "https://github.com/spf13/cobra",
+                "expected_language": "Go"
+            }
+        ]
+    
+    logger.info("Starting end-to-end pipeline evaluation")
+    logger.info(f"Testing {len(projects)} projects: {[p['name'] for p in projects]}")
+    
+    output_base_dir = Path("evals")
```
ivanmilevtues: Hardcoded relative path, might bite us in the ass if for some reason we start running the script from the wrong path (i.e. test suite run/docker run/etc.)

evals/end_to_end_eval.py
```diff
@@ -0,0 +1,357 @@
...
+from evals.report_generator import generate_header, generate_e2e_section, write_report
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def read_monitoring_results(project_name: str) -> Dict[str, Any]:
+    """Read monitoring results from the generated JSON file."""
+    monitoring_file = Path(f"evals/monitoring_results/{project_name}_monitoring.json")
```
ivanmilevtues: I am really scared from relative paths, as we had quite some probs with those and you can see that we have them usually as configuration or full paths :).

The problem is that we'd forget and we will have the problem it works when I run it, but some process runs it from the wrong dir and it is cooked (i.e. test suite run, docker run etc.)

diagram_analysis/diagram_generator.py
```diff
@@ -15,17 +19,21 @@
...
 
 class DiagramGenerator:
-    def __init__(self, repo_location, temp_folder, repo_name, output_dir, depth_level: int):
+    def __init__(self, repo_location, temp_folder, repo_name, output_dir, depth_level: int, enable_monitoring: bool = False):
         self.repo_location = repo_location
         self.temp_folder = temp_folder
         self.repo_name = repo_name
         self.output_dir = output_dir
+        self.enable_monitoring = enable_monitoring
+        self.monitoring_output_dir = Path("evals/monitoring_results")
```
ivanmilevtues: I am scare of relative paths, as if you run the script from the wrong dir we are cooked :D

ivanmilevtues:
👀 @[brovatten](https://github.com/brovatten) brovatten [deployed](https://github.com/CodeBoarding/CodeBoarding/actions/runs/18807683176/job/53664273098) to [github-pages](https://codeboarding.github.io/CodeBoarding/) [15 hours ago](https://github.com/CodeBoarding/CodeBoarding/pull/91#event-20504359703) — with  [GitHub Pages](https://github.com/apps/github-pages)

What is this, when I clicked it gave 404 but looked very interesting :D

agents/agent.py
```diff
@@ -35,6 +36,7 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, syste
         self._setup_env_vars()
         self.llm = self._initialize_llm()
         self.extractor_llm = self._initialize_llm()
+        self._monitoring_callback = None
```
ivanmilevtues:
```suggestion
        self._monitoring_callback = None  # Used by monitor decorator
```

diagram_analysis/diagram_generator.py
```diff
@@ -175,4 +182,44 @@ def generate_analysis(self):
 
         logger.info(f"Analysis complete. Generated {len(files)} analysis files")
         print("Generated analysis files: %s", [os.path.abspath(file) for file in files])
+        
         return files
+
+    def generate_static_analysis(self):
```
ivanmilevtues:
```suggestion
    def generate_static_analysis(self):
```

evals/run_all_evals.py
```diff
@@ -0,0 +1,123 @@
...
+        "name": "cobra",
+        "url": "https://github.com/spf13/cobra",
+        "expected_language": "Go"
+    }
+]
+
+
+def generate_security_md(static_results: Dict[str, Any], e2e_results: Dict[str, Any], output_path: Path) -> None:
+    """Deprecated: Unified SECURITY.md is no longer generated."""
+    logger.info("Skipping unified SECURITY.md generation (deprecated).")
```
ivanmilevtues:
```suggestion
```

static_analysis_config.yml
```diff
@@ -2,7 +2,7 @@ lsp_servers:
   python:
     name: Pyright Language Server
     command:
-    - ./static_analyzer/servers/py-lsp
+    - /Users/brovatten/Documents/github_folder/CodeBoarding/static_analyzer/servers/py-lsp
```
ivanmilevtues:
```suggestion
    - ./static_analyzer/servers/py-lsp
```

static_analysis_config.yml
```diff
@@ -13,7 +13,7 @@ lsp_servers:
   typescript:
     name: TypeScript Language Server
     command:
-    - ./static_analyzer/servers/node_modules/.bin/typescript-language-server
+    - /Users/brovatten/Documents/github_folder/CodeBoarding/static_analyzer/servers/node_modules/.bin/typescript-language-server
```
ivanmilevtues:
```suggestion
    - ./static_analyzer/servers/node_modules/.bin/typescript-language-server
```

static_analysis_config.yml
```diff
@@ -28,7 +28,7 @@ lsp_servers:
   go:
     name: Go Language Server (gopls)
     command:
-    - ./static_analyzer/servers/gopls
+    - /Users/brovatten/Documents/github_folder/CodeBoarding/static_analyzer/servers/gopls
```
ivanmilevtues:
```suggestion
    - ./static_analyzer/servers/gopls
```

static_analysis_config.yml
```diff
@@ -39,7 +39,7 @@ tools:
   tokei:
     name: tokei
     command:
-    - ./static_analyzer/servers/tokei
+    - /Users/brovatten/Documents/github_folder/CodeBoarding/static_analyzer/servers/tokei
```
ivanmilevtues:
```suggestion
    - ./static_analyzer/servers/tokei
```

ivanmilevtues (approved):
Just accept my suggestions and this can get in :)

PR #97, opened 2025-10-25
Package CodeBoarding

brovatten:
I finally got around to packaging the codeboarding module and making it runnable from the CLI. 

Most of the changes are moving files and adjusting the import paths. It is battle tested!


codeboarding/agents/details_agent.py
```diff
@@ -4,14 +4,14 @@
 
 from langchain_core.prompts import PromptTemplate
 
-from agents.agent import CodeBoardingAgent
-from agents.agent_responses import AnalysisInsights, CFGAnalysisInsights, ValidationInsights, Component, \
+from codeboarding.agents.agent import CodeBoardingAgent
```
ivanmilevtues, line 7: Not a bad comment, but this looks weird, I think this can be a good refactor to have:
```Python
from codeboarding.agents.agent import CodeBoardingAgent
```
Generally speaking I like that import pattern with the __init__.py

But this will be for another issue :)

codeboarding/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .agents.agent import CodeBoardingAgent
+from .diagram_analysis.diagram_generator import DiagramGenerator
+
+__all__ = [
+    "CodeBoardingAgent",
+    "DiagramGenerator",
+]
```
ivanmilevtues, line 7: Didn't know the __all__ syntax:
__all__ is a special Python variable used in modules.

It controls what gets exported when someone does:
```python
from package_name import *
```

In this case, only CodeBoardingAgent and DiagramGenerator will be accessible when the package is imported with *.

README.md
```diff
@@ -59,7 +59,7 @@ First, make you you have uv installed. Check official installation guide [Instal
 Setup the environment:
 
 ```bash
-uv sync --frozen
+uv sync
```
ivanmilevtues, line 62: the --frozen flag is good to have, if we don't have it then the lock file will get overriden and then we will get changes in git and eventually someone will commit it and there will be comments to revert :D

From GPT:
When you pass --frozen to uv sync, you’re telling it:
Do not rewrite or update the lockfile;
Only install (sync) what’s already locked, and fail if things don’t align.

static_analysis_config.yml
```diff
@@ -2,7 +2,7 @@ lsp_servers:
   python:
     name: Pyright Language Server
     command:
-    - ./static_analyzer/servers/py-lsp
+    - /Users/brovatten/Documents/github_folder/cleanup/static_analyzer/servers/py-lsp
```
ivanmilevtues: Do not expose your personal paths :)
That's why I had it with relative paths, the setup.py script actually adds proper paths

ivanmilevtues (changes requested):
Awesome job!
Just few very small comments and this is ready to be merged 🥳

codeboarding/agents/prompts/gpt_prompts_unidirectional.py
```diff
@@ -1,5 +1,6 @@
 """Prompt factory implementation for GPT-4 models in unidirectional analysis mode.
 """
+from .abstract_prompt_factory import AbstractPromptFactory
```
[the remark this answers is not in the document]
ivanmilevtues: It is needed...

README.md
brovatten: We need to skip the --frozen to install codeboarding as a package itself. If you try the setup, you will see the uv.lock file doesn't change :)

static_analysis_config.yml
brovatten: Thanks, this was slop on my side!

codeboarding/agents/details_agent.py
brovatten: Yup agreed!

PR #101, opened 2025-11-07
Clustering is failing analysis

ivanmilevtues:
Our analysis used to fail in the case where a node from a cluster is refering to one which is outside of the clusters (returns None).

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -138,11 +138,12 @@ class CodeBoardingAgent(ReferenceResolverMixin):
                 callback_list = callbacks or []
                 if self._monitoring_callback:
                     callback_list.append(self._monitoring_callback)
-                
-                response = self.agent.invoke(
-                    {"messages": [self.system_message, HumanMessage(content=prompt)]},
-                    config={"callbacks": callback_list} if callback_list else None
-                )
+
+                if callback_list:
+                    response = self.agent.invoke({"messages": [self.system_message, HumanMessage(content=prompt)]},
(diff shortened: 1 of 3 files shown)


static_analyzer/graph.py
```diff
@@ -121,8 +121,10 @@ def __cluster_str(communities, graph_x, top_nodes) -> str:
             src_cluster = node_to_cluster.get(src)
             dst_cluster = node_to_cluster.get(dst)
 
-            # Only process edges where both nodes are in top clusters
-            if src in top_nodes and dst in top_nodes and src_cluster != dst_cluster:
+            # Skip edges where either node doesn't belong to any cluster
+            if src_cluster is None or dst_cluster is None:
```
Svilen-Stefanov, line 125: Thank you for addressing this! 

My main question here was why the src / dst are not found in the `node_to_cluster` if they are part of `graph_x`.
@ivanmilevtues and I chatted and realized that's the case because `communities` only considers the top communities.
So what we do here is iterate over all edges in the entire graph even though we only want to find the edges that connect the clusters. Can we create a ticket to fix this and add some tests in the future?

Since we're modifying the `__cluster_str` function, can we also add types for all attributes for clarity?

Apart from that, the changes make sense and lgtm.

ivanmilevtues: I agree with the comments, we should start adding types and addressing tech depth, step by step.

I have added the types and also renamed the parameters to be easier to understand.

brovatten:
LGTM!

PR #102, opened 2025-11-08
Add gitattributes file

Svilen-Stefanov:


.gitattributes
```diff
@@ -0,0 +1,13 @@
+# Handle text files automatically
```
ivanmilevtues: I did a quick research, and I think we need:
```
# Auto-detect text and ensure LF by default
* text=auto eol=lf

# Force Windows scripts to use CRLF
*.bat eol=crlf
*.ps1 eol=crlf
```

I don't like the exhaustive list as i.e. we now have .md which isn't in the list, as well as .tml.

* text -> detects all text formats w/o binaries so we can set it directly there and just have a small patch for the windows specific fomrat bat and ps1 :)

Svilen-Stefanov: I agree. I'll update it in a second.

PR #105, opened 2025-11-18
Fix black formatting

Svilen-Stefanov:
Since I suggested to split the formatting from the rest of the changes in [this PR](https://github.com/CodeBoarding/CodeBoarding/pull/103/files), I went ahead and just opened a new PR for the formatting changes only.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.12
+  python: python3
 
 repos:
   - repo: https://github.com/psf/black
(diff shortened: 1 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.12
```
brovatten, line 2: How come this changed?

brovatten (approved):
LGTM

.pre-commit-config.yaml
Svilen-Stefanov: I made the change because I thought we might want to support other python3 versions - e.g I'm using python3.13.

PR #107, opened 2025-11-19
Fix demo.py for Windows

Svilen-Stefanov:
The main changes @brovatten and I did here:
- replace the `py-lsp` executable with `pyright` that is now downloaded using npm
- fixed `npm` usage for Windows
- fixed the download from gdrive function, which was downloading HTML files due to a warning

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,12 +40,12 @@ def check_npm():
 
     if npm_path:
         try:
-            result = subprocess.run(["npm", "--version"], capture_output=True, text=True, check=True)
+            result = subprocess.run([npm_path, "--version"], capture_output=True, text=True, check=True)
             print(f"Step: npm check finished: success (version {result.stdout.strip()})")
             return True
-        except Exception:
+        except Exception as e:
             print(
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -219,14 +239,19 @@ def update_static_analysis_config():
...
 
-    # Update Python LSP server path
-    py_lsp_path = servers_dir / "py-lsp"
-    if py_lsp_path.exists():
-        config["lsp_servers"]["python"]["command"][0] = str(py_lsp_path)
+    # Update Python LSP server path (using pyright from node_modules)
+    pyright_path = servers_dir / "node_modules" / ".bin" / "pyright-langserver"
+    # On Windows, use .cmd files; on Unix, use the shell scripts
+    if platform.system() == "Windows":
+        pyright_path = servers_dir / "node_modules" / ".bin" / "pyright-langserver.cmd"
```
ivanmilevtues, line 246: Nice catch, just saw it in their repo that they explictly say this. Usually this is not the case :)
Also unsure what .cmd is, usually these scripts are done with .bat files in windows.

ivanmilevtues (approved):

PR #108, opened 2025-11-24
Remove pathlib from dependencies

Svilen-Stefanov:
Remove pathlib from the dependencies package is an obsolete backport of a standard library package.


ivanmilevtues (approved):
In the latest versions of python it is part of the default lib. Good catch!

PR #109, opened 2025-11-30
Added Intelephense as php LSP

brovatten:
It installs as a node module now to handle php ⭐


vscode_constants.py
```diff
@@ -22,6 +22,11 @@ def update_command_paths(bin_dir):
                 if platform.system().lower() == "windows":
                     # Use node to run the .mjs file on Windows
                     value["command"].insert(0, "node")
+            elif key == "php":
+                # Handle intelephense separately as it might be in node_modules/.bin
+                intelephense_path = os.path.join(bin_dir, "node_modules", ".bin", "intelephense")
```
ivanmilevtues: this is bad, I will fix it.

vscode_constants.py
```diff
@@ -63,6 +68,13 @@ def update_config(bin_dir=None):
             "file_extensions": [".go"],
             "install_commands": "go install golang.org/x/tools/gopls@latest",
         },
+        "php": {
+            "name": "Intelephense",
+            "command": ["intelephense", "--stdio"],
+            "languages": ["php"],
+            "file_extensions": [".php"],
+            "install_commands": "npm install intelephense",
```
ivanmilevtues, line 93: well done

vscode_constants.py
```diff
@@ -4,33 +4,52 @@
...
         for key, value in section.items():
-            if key == "typescript":
-                # Scan the bin dir to fine the cli.mjs path
-                value["command"][0] = find_cli_js(bin_dir) or value["command"][0]
-                if platform.system().lower() == "windows":
-                    # Use node to run the .mjs file on Windows
-                    value["command"].insert(0, "node")
+            if key == 'typescript':
+                # Scan the bin dir to find the cli.mjs path
+                value['command'][0] = find_runnable(bin_dir, 'cli.mjs', 'typescript-language-server') or \
```
ivanmilevtues: second or is for the local-dev setup. Because whenever you run the VSCode without vsce package it will ship things in the node_modules instead of producing a .vscx file which is later "extracted" and has different structure.

setup.py
```diff
@@ -256,6 +263,14 @@ def update_static_analysis_config():
         config["lsp_servers"]["typescript"]["command"][0] = str(ts_lsp_path)
         updates += 1
```
ivanmilevtues: Feels like these deserve their own function now, the updates started small and simple but now it is copy paste 4 times :D

setup.py
```diff
@@ -84,6 +84,7 @@ def install_node_servers():
         # Verify the installation
         ts_lsp_path = Path("./node_modules/.bin/typescript-language-server")
```
Svilen-Stefanov, line 85: Can we refactor these to something like:
```
    lsp_servers: List[Tuple[str, Path]] = [
        ("TypeScript Language Server", Path("./node_modules/.bin/typescript-language-server")),
        ("Pyright Language Server", Path("./node_modules/.bin/pyright-langserver")),
        ("Intelephense Language Server", Path("./node_modules/.bin/intelephense")),
    ]
    success_status = True

    for name, path in lsp_servers:
        if path.exists():
            print(f"Step: {name} installation finished: success")
        else:
            print(f"Step: {name} installation finished: warning - Binary not found at {path}")
            # Only set to False if it's currently True (or just set it directly)
            success_status = False 

    return success_status
```

vscode_constants.py
```diff
@@ -4,33 +4,52 @@
 
 def get_bin_path(bin_dir):
     system = platform.system().lower()
-    subdirs = {"windows": "win", "darwin": "macos", "linux": "linux"}
+    subdirs = {
```
Svilen-Stefanov: Something strange is happening with the linting here, probably this is causing the black failures.

vscode_constants.py
```diff
@@ -4,33 +4,52 @@
...
-            f"Unsupported platform: {system}. The extension currently supports Windows, macOS, and Linux."
-        )
-    return os.path.join(bin_dir, "bin", subdirs[system])
+            f"Unsupported platform: {system}. The extension currently supports Windows, macOS, and Linux.")
+    return os.path.join(bin_dir, 'bin', subdirs[system])
 
 
 def update_command_paths(bin_dir):
     bin_path = get_bin_path(bin_dir)
+    is_windows = platform.system().lower() == 'windows'
```
Svilen-Stefanov: nice!

setup.py
Svilen-Stefanov: They are too similar for separate functions, I suggested something similar below (which could technically be extracted as a function).

vscode_constants.py
brovatten: Can you elaborate on the requested change? :D don't get "second or is for the local-dev setup" atm

setup.py
brovatten: Thanks for the suggestion, agreed!

Svilen-Stefanov: I think this hasn't been addressed yet, right?

setup.py
```diff
@@ -239,22 +246,24 @@ def update_static_analysis_config():
...
-    # Update Python LSP server path (using pyright from node_modules)
-    pyright_path = servers_dir / "node_modules" / ".bin" / "pyright-langserver"
-    # On Windows, use .cmd files; on Unix, use the shell scripts
-    if platform.system() == "Windows":
-        pyright_path = servers_dir / "node_modules" / ".bin" / "pyright-langserver.cmd"
-    if pyright_path.exists():
-        config["lsp_servers"]["python"]["command"][0] = str(pyright_path)
-        updates += 1
+    # Update Node.js based LSP servers
+    lsp_servers = {
```
Svilen-Stefanov, line 250: Thank you for the changes here, there will be a merge conflict with [this PR](https://github.com/CodeBoarding/CodeBoarding/pull/111/files), which also modifies this in a different way. 
The other PR introduces slightly more generic changes (also slightly more complex), we can discuss which one we want.

PR #110, opened 2025-11-30
Add initial version of mypy type checker

Svilen-Stefanov:
I'm currently ignoring a lot of files to incrementally add them and make sure we don't merge things that change the behavior. I'll open a separate PR to introduce the changes to the other files and remove them from the exclude list of mypy.

I also added this to the pre-commit git action so that it runs on every commit.

### DIFF
--- a/demo.py
+++ b/demo.py
@@ -96,7 +96,7 @@ def generate_docs(repo_name: str, temp_repo_folder: Path, repo_url: str | None =
             )
 
 
-def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False) -> str:
+def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False):
     """
     Clone a git repo to target_dir/<repo-name>.
     Returns the Path to the cloned repository.
@@ -114,7 +114,8 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
         store_token()
     if caching_enabled() and onboarding_materials_exist(repo_name, ROOT_RESULT):
         logger.info(f"Cache hit for '{repo_name}', skipping documentation generation.")
-        return repo_name
+        return
+
     repo_name = clone_repository(repo_url, Path(REPO_ROOT))
     generate_docs(repo_name, temp_repo_folder, repo_url)
     if os.path.exists(ROOT_RESULT):
@@ -123,7 +124,6 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
         logger.warning(
             f"ROOT_RESULT directory '{ROOT_RESULT}' does not exist. Skipping upload of onboarding materials."
         )
-    return repo_name
 
 
 def copy_files(temp_folder: Path, output_dir: Path):


evals/end_to_end_eval.py
```diff
@@ -29,7 +29,7 @@
 load_dotenv()
 
 # Get project root from environment variable
-PROJECT_ROOT = Path(os.getenv("PROJECT_ROOT"))
+PROJECT_ROOT = Path(os.getenv("PROJECT_ROOT") or ".")
```
Svilen-Stefanov: If `PROJECT_ROOT` is undefined, we need a default. I don't think this is the cleanest way to do it but it's one way.

demo.py
```diff
@@ -57,9 +58,9 @@ def onboarding_materials_exist(project_name: str, source_dir: str):
         return False
 
 
-def generate_docs(repo_name: str, temp_repo_folder: Path, repo_url: str = None):
+def generate_docs(repo_name: str, temp_repo_folder: Path, repo_url: Optional[str] = None):
     # Create directories if they don't exist
-    repos_dir = Path(os.getenv("REPO_ROOT"))
+    repos_dir = Path(os.getenv("REPO_ROOT") or ".")
```
brovatten: I would use os.getenv("REPO_ROOT", ".")

brovatten: Also, I'd argue if the REPO_ROOT is not set something is wrong, since it's part of the `setup.py`. Are you sure we shouldn't just throw an error?

demo.py
```diff
@@ -101,12 +102,12 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
     repo_name = get_repo_name(repo_url)
     if not local_dev:
         store_token()
-    if caching_enabled() and onboarding_materials_exist(repo_name, ROOT_RESULT):
+    if ROOT_RESULT and caching_enabled() and onboarding_materials_exist(repo_name, ROOT_RESULT):
```
brovatten: This to me can lead to confusion... if caching is enabled in the .env then an unset `ROOT_RESULT`can lead to no caching? But we also set it before? 

More than two if-statements usually smells in my opinion.

I can't provide any good suggestions, maybe @ivanmilevtues can pitch in.

demo.py
```diff
@@ -101,12 +102,12 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
...
         logger.info(f"Cache hit for '{repo_name}', skipping documentation generation.")
-        return
-    repo_name = clone_repository(repo_url, Path(os.getenv("REPO_ROOT")))
+        return repo_name
```
brovatten: Good catch with consistencies, although I would instead remove the other `return repo_name`. We are not using the return for anything, better to define it when we need it. Expected return value is unclear

output_generators/html.py
```diff
@@ -68,7 +68,8 @@ def generate_html(
     cytoscape_data = generate_cytoscape_data(insights, linked_files, project, demo)
     cytoscape_json = json.dumps(cytoscape_data, indent=2)
 
-    root_dir = str(Path(os.getenv("REPO_ROOT")) / project)
+    repo_root = os.getenv("REPO_ROOT")
+    root_dir = (repo_root + "/" + project) if repo_root else project
```
brovatten: Doesn't this break windows? Not sure.

output_generators/markdown.py
```diff
@@ -57,7 +57,8 @@ def generate_markdown(insights: AnalysisInsights, project: str = "", repo_ref=""
 
     detail_lines = ["\n## Details\n", f"{insights.description}\n"]
 
-    root_dir = str(Path(os.getenv("REPO_ROOT")) / project)
+    repo_root = os.getenv("REPO_ROOT")
+    root_dir = (repo_root + "/" + project) if repo_root else project
```
brovatten: Same question here as above. I would assume we'd use the os.sep, even if this means making it a str after.

vscode_runnable.py
```diff
@@ -57,8 +57,8 @@ def partial_updates(component_to_update_name: str, analysis_to_update_name: str)
     generator.pre_analysis()
 
     # Load the analysis for which we want to extend the component
-    analysis = analysis_folder / f"{analysis_to_update_name}.json"
-    with open(analysis, "r") as file:
+    analysis_file = analysis_folder / f"{analysis_to_update_name}.json"
+    with open(analysis_file, "r") as file:
```
brovatten, line 61: Good!

vscode_constants.py
```diff
@@ -16,16 +17,18 @@ def update_command_paths(bin_dir):
...
-                # Scan the bin dir to fine the cli.mjs path
-                value["command"][0] = find_cli_js(bin_dir) or value["command"][0]
-                if platform.system().lower() == "windows":
-                    # Use node to run the .mjs file on Windows
-                    value["command"].insert(0, "node")
-            elif "command" in value:
-                cmd = value["command"]
-                if isinstance(cmd, list) and cmd:
-                    value["command"][0] = os.path.join(bin_path, cmd[0])
+            if isinstance(value, Dict) and "command" in value:
```
brovatten: Have to leave it here, @ivanmilevtues can u pitch in?

output_generators/html.py
ivanmilevtues: It will break windows, we need the os.sep thing as it is not a Path object good catch

output_generators/mdx.py
```diff
@@ -39,7 +39,7 @@ def generated_mermaid_str(
     return "\n".join(lines)
 
 
-def generate_frontmatter(file_name: str, component_name: str = None) -> str:
+def generate_frontmatter(file_name: str, component_name: Optional[str] = None) -> str:
```
ivanmilevtues: As your comment from the other commit, we want to use `str|None` here instead of optional?

It could be that I misunderstood, I don't have strong preference

output_generators/mdx.py
```diff
@@ -84,7 +84,8 @@ def generate_mdx(
 
     detail_lines = ["\n### Details\n", f"{insights.description}\n"]
 
-    root_dir = str(Path(os.getenv("REPO_ROOT")) / project)
+    repo_root = os.getenv("REPO_ROOT")
+    root_dir = (repo_root + "/" + project) if repo_root else project
```
ivanmilevtues: Windows sep thing

output_generators/sphinx.py
```diff
@@ -91,7 +91,8 @@ def generate_rst(
     lines.append("")
 
     # Add component details
-    root_dir = str(Path(os.getenv("REPO_ROOT")) / project)
+    repo_root = os.getenv("REPO_ROOT")
+    root_dir = (repo_root + "/" + project) if repo_root else project
```
ivanmilevtues: Windows sep

demo.py
ivanmilevtues: I agree probably this should be an error.
In that case better use: value = os.environ["MY_ENV_VAR"]  # raises KeyError if not set

local_app.py
```diff
@@ -169,7 +170,7 @@ async def generate_onboarding(job_id: str):
 # -- API Endpoints --
 @app.post("/generation", response_class=JSONResponse, summary="Create a new onboarding job")
 async def start_generation_job(
-    repo_url: str = Query(..., description="GitHub repo URL"), background_tasks: BackgroundTasks = None
+    repo_url: str = Query(..., description="GitHub repo URL"), background_tasks: Optional[BackgroundTasks] = None
```
ivanmilevtues: unsure what we do with the optional, as I said I don't have anything against it. Just mentioning for consistency.

vscode_constants.py
```diff
@@ -16,16 +17,18 @@ def update_command_paths(bin_dir):
...
+                cmd: List[str] = value["command"]  # type: ignore[assignment]
```
ivanmilevtues: list[str]

demo.py
ivanmilevtues: I understand it but cansee the confusion. I think the previous thing was fine as:
if ROOT_RESULT is unset and caching is enabled, the onboarding_materials_exist would be false.

This whole thing is quite confusing as it is patch work after patch work. We should probably reconsider the CLI (demo.py).

As now it does 3 things. Generates an analysis to an output dir with an arg + uploads to our own repo GeneratedOnBoardings + can produce .md/.html.

Half of those things are done via sys envs, the other half with args.

I would leave this for now and create a story to refactor this class and name it `codeboarding_cli.py`.

Migrate most things to flags and probably remove the uploading to somewhere else.

demo.py
```diff
@@ -101,12 +102,12 @@ def generate_docs_remote(repo_url: str, temp_repo_folder: Path, local_dev=False)
...
+    repo_name = clone_repository(repo_url, Path(os.getenv("REPO_ROOT") or "repos"))
     generate_docs(repo_name, temp_repo_folder, repo_url)
-    if os.path.exists(ROOT_RESULT):
+    if ROOT_RESULT and os.path.exists(ROOT_RESULT):
```
ivanmilevtues: path.exists(none) is false

vscode_constants.py
ivanmilevtues: not much to add, it is a type safe approach :D
we should decide if we do List/list Dict/dict and so on.

demo.py
Svilen-Stefanov: I agree we should refactor these. The way that I'll resolve this is similar to how I've resolved `Path(os.getenv("REPO_ROOT") or ".")` - throwing an exception when the env is not present, which makes a couple of the other changes redundant.

ivanmilevtues (approved):
Looks good, a big step in the direction of being typesafe <3

PR #111, opened 2025-12-01
Javascript LSP selection fix

ivanmilevtues:
We have fixed the Javascript LSP selection in this change.
Also enhanced the ProgrammingLanguage builder by allowing selection based on the file suffix.

### DIFF
--- a/static_analyzer/programming_language.py
+++ b/static_analyzer/programming_language.py
@@ -33,7 +33,7 @@ class ProgrammingLanguage:
         # id for the language, used in LSP server
         return self.language.lower().replace(" ", "_")
 
-    def get_server_parameters(self) -> List[str]:
+    def get_server_parameters(self) -> list[str]:
         if not self.server_commands:
             raise ValueError(
                 f"No server commands defined for {self.language}. "


static_analyzer/programming_language.py
```diff
@@ -39,5 +42,132 @@ def is_supported_lang(self) -> bool:
...
+        """
+        if self.lsp_key != other.lsp_key:
+            raise ValueError(f"Cannot merge languages with different lsp_keys: {self.lsp_key} vs {other.lsp_key}")
+        
+        # Combine suffixes
+        merged_suffixes = list(set(self.suffixes) | set(other.suffixes))
+        
+        # Use the lsp_key as the canonical language name
+        return ProgrammingLanguage(
+            language=self.lsp_key,
```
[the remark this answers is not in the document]
ivanmilevtues: I kept it like this and changed __init__.py

static_analyzer/programming_language.py
```diff
@@ -39,5 +42,132 @@ def is_supported_lang(self) -> bool:
...
+        Args:
+            lsp_configs: Dictionary of LSP server configurations from config file
+        """
+        self.lsp_configs = lsp_configs
+        # Build reverse index: extension -> lsp_config_key
+        self._extension_to_lsp: Dict[str, str] = {}
+        for lsp_key, config in lsp_configs.items():
+            for ext in config.get("file_extensions", []):
+                # Normalize extension (ensure it starts with '.')
+                normalized_ext = ext if ext.startswith('.') else f'.{ext}'
```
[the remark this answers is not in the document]
ivanmilevtues: Should be fine as long as some LSP actually processses it. Also should not happen generally.

setup.py
```diff
@@ -255,9 +255,22 @@ def update_static_analysis_config():
     if ts_lsp_path.exists():
         config["lsp_servers"]["typescript"]["command"][0] = str(ts_lsp_path)
         updates += 1
+        # Also update JavaScript LSP server (uses same typescript-language-server)
+        config["lsp_servers"]["javascript"]["command"][0] = str(ts_lsp_path)
+        updates += 1
+
+    # Update Go Language Server path
+    gopls_path = servers_dir / "gopls"
+    if platform.system() == "Windows":
```
brovatten: How come the go paths updated for this PR?

static_analyzer/scanner.py
```diff
@@ -36,7 +37,10 @@ def scan(self) -> List[ProgrammingLanguage]:
             logger.warning("No total code count found in Tokei output")
             return []
 
-        programming_languages = []
+        # Use dict to accumulate and merge by lsp_key
+        merged_languages: Dict[str, ProgrammingLanguage] = {}
```
brovatten: This is never used

static_analyzer/programming_language.py
```diff
@@ -39,5 +47,101 @@ def is_supported_lang(self) -> bool:
...
-        return f"ProgrammingLanguage(language={self.language}, size={self.size}, percentage={self.percentage}, suffixes={self.suffixes})"
+        return f"ProgrammingLanguage(language={self.language}, lsp_key={self.lsp_key}, size={self.size}, percentage={self.percentage:.2f}%, suffixes={self.suffixes})"
+
+
+class ProgrammingLanguageBuilder:
+    """Builder to create ProgrammingLanguage instances from tokei output with greedy LSP matching."""
+
+    def __init__(self, lsp_configs: dict):
+        """
+        Initialize builder with LSP server configurations.
```
brovatten: I would skip the docs and use CodeBoarding ;) 

Or use 2 line descriptions like we've done before

static_analyzer/programming_language.py
```diff
@@ -1,18 +1,26 @@
...
+        server_commands: List[str] = None,
+        lsp_key: Optional[str] = None,
     ):
         self.language = language
         self.size = size
         self.percentage = percentage
         self.suffixes = suffixes
         self.server_commands = server_commands
+        # lsp_key is used for grouping related languages (e.g., JS, TSX, JSX -> typescript)
+        self.lsp_key = lsp_key or language.lower()
```
brovatten: Use a more descriptive name. What they have in common is the LSP that's used, so maybe reference that?

static_analyzer/programming_language.py
```diff
@@ -39,5 +47,101 @@ def is_supported_lang(self) -> bool:
...
+            percentage: Percentage of total codebase
+            file_suffixes: Set of file suffixes extracted from tokei reports
+
+        Returns:
+            ProgrammingLanguage instance with LSP config if available
+        """
+        lsp_key = self._find_lsp_config_key(tokei_language, file_suffixes)
+
+        server_commands = None
+        config_suffixes: Set[str] = set()
```
brovatten, line 103: I would just define the set normally for consistency

brovatten: Actually nvm, this is nice, good thinking!

setup.py
Svilen-Stefanov: I suppose you're fixing other things as well, not just the javascrip LSP selection. I also wonder if we can refactor this with something like:
```
# The Plan: (Binary Name, Is_Node_App, List of Config Targets)
# "True" means it lives in node_modules/.bin and needs .cmd on Windows
# "False" means it lives in the root and needs .exe on Windows
server_definitions = [
    ("pyright-langserver",         True,  [("lsp_servers", "python")]),
    ("typescript-language-server", True,  [("lsp_servers", "typescript"), 
                                           ("lsp_servers", "javascript")]),
    ("gopls",                      False, [("lsp_servers", "go")]),
    ("tokei",                      False, [("tools",       "tokei")]),
]

for binary, is_node, targets in server_definitions:
    # 1. Determine the extension and folder based on the type
    ext = (".cmd" if is_node else ".exe") if is_win else ""
    folder = (servers_dir / "node_modules" / ".bin") if is_node else servers_dir
    
    # 2. Build the full path once
    full_path = folder / (binary + ext)

    # 3. Apply to all relevant targets in config
    if full_path.exists():
        for section, key in targets:
            config[section][key]["command"][0] = str(full_path)
            updates += 1
```

The `server_definitions ` still seem a bit primitive.

static_analysis_config.yml
```diff
@@ -2,46 +2,65 @@ lsp_servers:
   python:
     name: Pyright Language Server
     command:
-    - ./static_analyzer/servers/node_modules/.bin/pyright-langserver
-    - --stdio
+      - ./static_analyzer/servers/py-lsp
```
Svilen-Stefanov: Hmm, this seems wrong, no? We just introduced pyright for Python with npm, what am I missing?

static_analysis_config.yml
```diff
@@ -2,46 +2,65 @@ lsp_servers:
...
+      - .py
+      - .pyi
+    install_commands: pip install pyright
   typescript:
     name: TypeScript Language Server
     command:
-    - ./static_analyzer/servers/node_modules/.bin/typescript-language-server
-    - --stdio
-    - --log-level=2
+      - ./static_analyzer/servers/node_modules/.bin/typescript-language-server
```
Svilen-Stefanov, line 16: I hope the linter fixes these things because most of these changes seem like noise.

setup.py
```diff
@@ -255,9 +255,22 @@ def update_static_analysis_config():
...
+        gopls_path = servers_dir / "gopls.exe"
+    if gopls_path.exists():
+        config["lsp_servers"]["go"]["command"][0] = str(gopls_path)
+        updates += 1
 
     # Update tokei tool path
     tokei_path = servers_dir / "tokei"
+    if platform.system() == "Windows":
+        tokei_path = servers_dir / "tokei.exe"
```
Svilen-Stefanov: If we make the above change, we should update the [PHP PR](https://github.com/CodeBoarding/CodeBoarding/pull/109/files) accordingly.

static_analyzer/programming_language.py
```diff
@@ -1,18 +1,26 @@
...
 class ProgrammingLanguage:
     def __init__(
-        self, language: str, size: int, percentage: float, suffixes: List[str], server_commands: List[str] = None
+        self,
+        language: str,
+        size: int,
+        percentage: float,
+        suffixes: List[str],
+        server_commands: List[str] = None,
+        lsp_key: Optional[str] = None,
```
Svilen-Stefanov: I was just checking what's the preffered way to type this wrt my open mypy PR. Apparently since 3.9+ and `PEP 585`, the preferred options here are `list[str]` and `str | None`, sorry if I introduced confusion in my PR. 
One of the reasons for it being that you no longer need an extra import.

Thank you for introducing the types in the first place!

static_analyzer/programming_language.py
```diff
@@ -39,5 +47,101 @@ def is_supported_lang(self) -> bool:
...
+
     def __str__(self):
-        return f"ProgrammingLanguage(language={self.language}, size={self.size}, percentage={self.percentage}, suffixes={self.suffixes})"
+        return f"ProgrammingLanguage(language={self.language}, lsp_key={self.lsp_key}, size={self.size}, percentage={self.percentage:.2f}%, suffixes={self.suffixes})"
+
+
+class ProgrammingLanguageBuilder:
+    """Builder to create ProgrammingLanguage instances from tokei output with greedy LSP matching."""
+
+    def __init__(self, lsp_configs: dict):
```
Svilen-Stefanov, line 62: Can this type be more specific?

static_analyzer/programming_language.py
```diff
@@ -39,5 +47,101 @@ def is_supported_lang(self) -> bool:
...
+
+        if lsp_key and lsp_key in self.lsp_configs:
+            config = self.lsp_configs[lsp_key]
+            server_commands = config.get("command")
+            config_suffixes = set(config.get("file_extensions", []))
+
+        # Merge suffixes from tokei and config
+        all_suffixes = file_suffixes | config_suffixes
+
+        return ProgrammingLanguage(
```
Svilen-Stefanov, line 113: I'm probably missing something here but why do we need a Builder object? Can't we improve the existing `ProgrammingLanguage` class?

static_analyzer/programming_language.py
ivanmilevtues: I don't get the `str|None ` in this context?
I assume it is instead of optional so I will just do that.

static_analyzer/programming_language.py
ivanmilevtues: lsp_server_key it is

static_analyzer/programming_language.py
ivanmilevtues: sadly not really as this is a dict read from yml file ;/

static_analyzer/programming_language.py
ivanmilevtues: yes removed few comments, I think some of them are needed thou. I.e. the find_lsp_server_key

static_analyzer/programming_language.py
ivanmilevtues: Because the constructor will get quite heavy.
Builder pattern is usually for constructing complex objects I agree with that, however in this case with time and already there are multiple steps to decide which PL do we need to have.

1. if tokei found JavaScript that is an easymatch the normal PL constructor works just fine
2. if tokei found JSX as PL, now we have to match it in the suffixes of files from the configuration so we have to do these iteartions from the other method. I think PL should be a minimal DAO class and not have logic in it. Hense I created this Builder.

Open to converse around this, but I feel like DAO objects should be slim

setup.py
ivanmilevtues: yes it is an error I spotted and I fixed it

static_analysis_config.yml
ivanmilevtues: I ran the linter so that is why we have them :D, hopefully for the last time

static_analyzer/programming_language.py
Svilen-Stefanov: In such cases, what I usually do for JSON files is use Pydantic [BaseModels](https://docs.pydantic.dev/latest/api/base_model/) to define the general structure of the file and do basic validation instead of using dict (for which you always need to do validation after the fact). 
It doesn't support YAML out of the box but it's not super complicated to configure iirc. We could also consider modifying the yaml file to a JSON file.

Svilen-Stefanov: We can create this as a tec depth task we investigate later, we don't have to do this in this PR.

static_analyzer/__init__.py
```diff
@@ -19,7 +19,7 @@ def create_clients(programming_languages: List[ProgrammingLanguage], repository_
             logger.warning(f"Unsupported programming language: {pl.language}. Skipping.")
             continue
         try:
-            if pl.language in ["TypeScript"]:
+            if pl.language.lower() in ["typescript"]:
```
Svilen-Stefanov, line 22: Can the language here actually be an `Enum` we compare against? In the long run, I think we should try to avoid string parsings as much as possible.

static_analysis_config.yml
Svilen-Stefanov: What's strange though is that the linter is part of our build pipeline and it seems to have passed previously so shouldn't there be another reason for this as well or was this somehow ignored by black?

static_analyzer/programming_language.py
```diff
@@ -34,10 +42,82 @@ def get_server_parameters(self) -> List[str]:
...
+        return f"ProgrammingLanguage(language={self.language}, lsp_server_key={self.lsp_server_key}, size={self.size}, percentage={self.percentage:.2f}%, suffixes={self.suffixes})"
+
+
+class ProgrammingLanguageBuilder:
+    """Builder to create ProgrammingLanguage instances from tokei output with greedy LSP matching."""
+
+    def __init__(self, lsp_configs: dict):
+        self.lsp_configs = lsp_configs
+        # Build reverse index: extension -> lsp_config_key
+        self._extension_to_lsp: Dict[str, str] = {}
```
Svilen-Stefanov: nit:
```suggestion
        self._extension_to_lsp: dict[str, str] = {}
```

I see it in a couple other places in the builder class. Thank you for the types and again sorry for the original confusion with the types (the main idea of the latest PEP is to reduce the dependencies on typing).

brovatten (approved):

static_analyzer/__init__.py
ivanmilevtues: Doesn't that mean that we'd have to update it in config files + in the enum? So having it at two places.
I would like this to be yaml driven so we don't have to update the codebase at many places for adding new config.
This is a corner case, specific handling for TS. Probably will have to have such thing for other cases as well, don't get me wrong.

ivanmilevtues: But I will add it in the pydantic story so we can come up with a solution that fits all of these cases.

ivanmilevtues: Added as story.

PR #112, opened 2025-12-01
Add Cerebras support, and increase recursion limits

brovatten:
Any opinions on the recursion limit? It would stall with the default recursion value...

Could be weaker models that more easily gets stuck/needs more tool calls? Should follow up on this at some point.

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -154,7 +154,7 @@ class CodeBoardingAgent(ReferenceResolverMixin):
                 if callback_list:
                     response = self.agent.invoke(
                         {"messages": [self.system_message, HumanMessage(content=prompt)]},
-                        config={"callbacks": callback_list, "recursion_limit": 100000},
+                        config={"callbacks": callback_list, "recursion_limit": 40},
                     )
                 else:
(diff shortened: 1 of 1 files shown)


agents/agent.py
```diff
@@ -142,10 +154,13 @@ def _invoke(self, prompt, callbacks: list = None) -> str:
                 if callback_list:
                     response = self.agent.invoke(
                         {"messages": [self.system_message, HumanMessage(content=prompt)]},
-                        config={"callbacks": callback_list},
+                        config={"callbacks": callback_list, "recursion_limit": 100000},
```
ivanmilevtues: This recursion limit of 1M doesn't seem too safe :D.

Probably make it lower so that somones token doesn't explode. Usually it is around 35 :D, but as gpt-oss is a smaller model I understand why it needs to be bigger.

brovatten: Thanks for the suggestion, let's set it to 40 to trade-off :)

I also see 25 is the default https://github.com/langchain-ai/langgraphjs/issues/1524

PR #113, opened 2025-12-02
Fix all other mypy issues and enable mypy for the entire repo

Svilen-Stefanov:
I'm trying to fix all other mypy files here. The diff is quite big and there are some functional changes - it seems there were outdated bits and pieces that are potentially never executed or still wrong, which should be fixed in this PR. 

Please take a close look at the changes that don't just add types.

### DIFF
--- a/agents/tools/read_docs.py
+++ b/agents/tools/read_docs.py
@@ -153,9 +153,9 @@ class ReadDocsTool(BaseTool):
         # exclude the analysis_dir from the comparison
         if self.repo_dir is None:
             return False
-        sub_parts: tuple[str, ...] = sub.parts
-        full_parts: tuple[str, ...] = full.parts
-        repo_dir_parts: tuple[str, ...] = self.repo_dir.parts
+        sub_parts = sub.parts
+        full_parts = full.parts
+        repo_dir_parts = self.repo_dir.parts
         full_parts = full_parts[len(repo_dir_parts) :]
         for i in range(len(full_parts) - len(sub_parts) + 1):
             if full_parts[i : i + len(sub_parts)] == sub_parts:
--- a/agents/tools/read_file_structure.py
+++ b/agents/tools/read_file_structure.py
@@ -116,9 +116,9 @@ class FileStructureTool(BaseTool):
         # exclude the analysis_dir from the comparison
         if self.repo_dir is None:
             return False
-        sub_parts: tuple[str, ...] = sub.parts
-        full_parts: tuple[str, ...] = full.parts
-        analysis_parts: tuple[str, ...] = self.repo_dir.parts
+        sub_parts = sub.parts
+        full_parts = full.parts
+        analysis_parts = self.repo_dir.parts
         full_parts = full_parts[len(analysis_parts) :]
         for i in range(len(full_parts) - len(sub_parts) + 1):
             if full_parts[i : i + len(sub_parts)] == sub_parts:
@@ -149,25 +149,24 @@ def get_tree_string(
         return tree_lines
 
     try:
-        entries = sorted(os.listdir(str(startpath)))
+        entries = sorted([p for p in startpath.iterdir()])
     except (PermissionError, FileNotFoundError):
         # Handle permission errors or non-existent directories
         return [indent + "└── [Error reading directory]"]
 
-    for i, entry in enumerate(entries):
+    for i, entry_path in enumerate(entries):
         # Check if we've exceeded the maximum number of lines
         if len(tree_lines) >= max_lines:
             tree_lines.append(indent + "└── [Output truncated due to size limits]")
             return tree_lines
 
-        path = str(startpath / entry)
         connector = "└── " if i == len(entries) - 1 else "├── "
-        tree_lines.append(indent + connector + entry)
+        tree_lines.append(indent + connector + entry_path.name)
 
-        if os.path.isdir(path):
+        if entry_path.is_dir():
             extension = "    " if i == len(entries) - 1 else "│   "
             subtree = get_tree_string(
-                Path(path), indent + extension, max_depth, current_depth + 1, max_lines - len(tree_lines)
+                entry_path, indent + extension, max_depth, current_depth + 1, max_lines - len(tree_lines)
             )
             tree_lines.extend(subtree)
 
--- a/agents/tools/read_file.py
+++ b/agents/tools/read_file.py
@@ -114,9 +114,9 @@ class ReadFileTool(BaseTool):
         # exclude the analysis_dir from the comparison
         if self.repo_dir is None:
             return False
-        sub_parts: tuple[str, ...] = sub.parts
-        full_parts: tuple[str, ...] = full.parts
-        repo_dir_parts: tuple[str, ...] = self.repo_dir.parts
+        sub_parts = sub.parts
+        full_parts = full.parts
+        repo_dir_parts = self.repo_dir.parts
         full_parts = full_parts[len(repo_dir_parts) :]
         for i in range(len(full_parts) - len(sub_parts) + 1):
             if full_parts[i : i + len(sub_parts)] == sub_parts:


agents/agent.py
```diff
@@ -87,21 +87,21 @@ def _initialize_llm(self):
             return ChatOpenAI(
                 model="gpt-4o",
                 temperature=0,
-                max_tokens=None,
+                max_tokens=None,  # type: ignore[call-arg]
```
Svilen-Stefanov, line 90: This seems redundant, before removing it, I wanted to check with you if there is a reason for adding it?

agents/agent.py
```diff
@@ -87,21 +87,21 @@ def _initialize_llm(self):
...
                 timeout=None,
                 max_retries=0,
-                api_key=self.openai_api_key,
+                api_key=self.openai_api_key,  # type: ignore[arg-type]
```
Svilen-Stefanov, line 93: These things are currently ignored becuase they were throwing an error that seems legit:
```
agents\agent.py:93: error: Argument "api_key" to "ChatOpenAI" has incompatible type "str"; expected "SecretStr | None"  [arg-type]
```
Seems the amount of functional changes is quite big, I thought we can follow up with a PR that removes these and tests if the agents still behave the way we expect them to.

tests/test_file_structure_tool.py
```diff
@@ -7,7 +7,7 @@
 class TestFileStructureTool(unittest.TestCase):
     def setUp(self):
         # Set up any necessary state or mocks before each test
-        self.tool = FileStructureTool(analysis_dir=Path("./repos/django"))
+        self.tool = FileStructureTool(repo_dir=Path("./repos/django"))
```
Svilen-Stefanov, line 10: Are any of these tests run / executed at any time? It seems to me like the APIs changed but the tests never did. 
We should eventually add all tests to the pipeline

static_analyzer/lsp_client/client.py
```diff
@@ -1013,7 +1019,7 @@ def _extract_package_from_import(module_name: str) -> str:
 
         # Get the top-level package
         parts = module_name.split(".")
-        return parts[0] if parts else None
+        return parts[0] if parts else ""
```
Svilen-Stefanov, line 1021: Does this make sense, should we return `None` here?

tests/test_file_structure_tool.py
brovatten: Good catch, they are not

static_analyzer/lsp_client/client.py
brovatten: It is only used to compare with a package name later (string comparison)? I personally would expect None if no superclasses are found and not ""

agents/agent.py
brovatten: If you remove it and run the mypy checks, is this added again? I can't really see why either, perhaps @ivanmilevtues knows?

agents/agent.py
brovatten: It's a str that when printed hashes its output (to not leak api keys). We actually load the env variables in `demo.py` so we should stick to those and not pass them expclitly here. Good catch!

I think it still accepts the string though...

agents/agent_responses.py
```diff
@@ -20,11 +20,12 @@ def extractor_str(cls):
...
             if get_origin(ftype) is list:
                 # get the type of the list:
-                ftype = ftype.__args__[0]
+                if ftype is not None and hasattr(ftype, "__args__"):
+                    ftype = ftype.__args__[0]
                 result_str += f"{fname} which is a list ("
             if ftype is Optional:
                 result_str += f"{fname} ({fvalue.description}), "
-            elif issubclass(ftype, LLMBaseModel):
+            elif ftype is not None and isinstance(ftype, type) and issubclass(ftype, LLMBaseModel):
```
brovatten, line 28: Is this really needed? it was so clean before 😢

agents/agent_responses.py
```diff
@@ -191,7 +192,7 @@ class UpdateAnalysis(LLMBaseModel):
     feedback: str = Field(description="Feedback provided on the analysis.")
 
     def llm_str(self):
-        return f"**Updated Analysis:**\n{self.analysis.llm_str()}\n\n**Feedback:**\n{self.feedback}"
+        return f"**Feedback:**\n{self.feedback}"
```
brovatten, line 195: Hmmmm... @ivanmilevtues ?

agents/tools/read_docs.py
```diff
@@ -140,11 +151,13 @@ def _run(self, file_path: Optional[str] = None, line_number: Optional[int] = 0)
...
         # exclude the analysis_dir from the comparison
-        sub = sub.parts
-        full = full.parts
-        repo_dir = self.repo_dir.parts
-        full = full[len(repo_dir) :]
-        for i in range(len(full) - len(sub) + 1):
-            if full[i : i + len(sub)] == sub:
+        if self.repo_dir is None:
+            return False
+        sub_parts: tuple[str, ...] = sub.parts
```
brovatten: I am a bit against doing typing for declaration of local variables to avoid clutter, do u think we need them here?

agents/tools/read_file_structure.py
```diff
@@ -69,28 +69,33 @@ def _run(self, dir: Optional[str] = None) -> str:
...
-            if self.is_subsequence(dir, d):
-                logger.info(f"[File Structure Tool] Found directory {d}")
-                searching_dir = d
-                break
+        if not dir:
+            return "Error: No directory specified."
+
+        dir_path = Path(dir)
+        searching_dir: Path | None = None
+        if self.cached_dirs:
```
brovatten, line 77: Nice, this also makes it clear there may not be cached dirs

agents/tools/read_file_structure.py
```diff
@@ -151,14 +160,14 @@ def get_tree_string(startpath, indent="", max_depth=float("inf"), current_depth=
             tree_lines.append(indent + "└── [Output truncated due to size limits]")
             return tree_lines
 
-        path = os.path.join(startpath, entry)
+        path = str(startpath / entry)
```
brovatten: Avoid string conversions when possible. If we use Pathlib then stick to it. I would do Path.is_dir() below instead of str -> os.path.isdir

agents/tools/read_structure.py
```diff
@@ -36,6 +36,10 @@ def _run(self, qualified_class_name: str) -> str:
         """
         Run the tool with the given input.
         """
+        if self.static_analysis is None:
+            logger.error("[CodeStructureTool] static_analysis is not set")
+            return "Error: Static analysis is not set."
```
brovatten, line 41: I am not sure, but shouldn't this thrown an error? @ivanmilevtues

evals/static_analysis_eval.py
```diff
@@ -30,7 +30,10 @@
 
 # Get project root from environment variable
 load_dotenv()
-PROJECT_ROOT = Path(os.getenv("PROJECT_ROOT"))
+project_root_env = os.getenv("PROJECT_ROOT")
+if not project_root_env:
+    raise ValueError("PROJECT_ROOT environment variable is not set")
```
brovatten, line 35: Nice!

evals/static_analysis_eval.py
```diff
@@ -57,7 +63,7 @@ def analyze_project_static_analysis(project_info: Dict[str, str]) -> Dict[str, A
         )
 
         # Just run static analysis, not full generation
-        static_analysis = generator.generate_static_analysis()
+        static_analysis = generator.generate_analysis()
```
brovatten, line 66: Why remove the static from analysis?

repo_utils/__init__.py
```diff
@@ -17,12 +17,12 @@
     GIT_AVAILABLE = True
 except ImportError:
     GIT_AVAILABLE = False
-    Repo = None
-    Git = None
-    GitCommandError = None
+    Repo = None  # type: ignore[misc, assignment]
```
brovatten, line 20: What does these do?

agents/agent.py
Svilen-Stefanov: So when I look at the definition of the actual models, I see:
```
openai_api_key: Optional[SecretStr] = Field(
        alias="api_key", default_factory=secret_from_env("OPENAI_API_KEY", default=None)
    )
```
This mean the right type here would technically be `SecretStr` instead of a regular string. Also the alias is `api_key` but I guess mypy would prefer to use the actual variable name `openai_api_key`.

agents/agent.py
Svilen-Stefanov: Oh, actually this is an oversight on my end, you don't need the `# type: ignore[call-arg]` here. I think I added it only to mention it in the PR. 
My question here was if this is needed in the first place, because the default seems to be None, so it is technically not needed.

agents/agent_responses.py
Svilen-Stefanov: It is because if `ftype` is None or is not a type, it will raise an exception - that's what mypy complains about here.

agents/agent_responses.py
Svilen-Stefanov: Thank you for noticing this, I wanted to point out this one!!! Where is this coming from? I couldn't find what `self.analysis` refers to?

agents/tools/read_docs.py
Svilen-Stefanov: Yup, the type declaration here is not mandatory, it can be deducted directly from the path. I can remove these.

agents/tools/read_file_structure.py
Svilen-Stefanov: Good point. I originally didn't want to make functional changes but using pathlib actually makes the code here much nicer. I'll fix this

agents/tools/read_structure.py
Svilen-Stefanov: Probably, yes. Iirc, if the static analysis doesn't work, we still try to make the LLM figure things out but this is definitely not an expected behavior - hence we should throw an exception.

evals/static_analysis_eval.py
Svilen-Stefanov: Ah, thank you for pointing this out as well!  
That's because there's no such method :O !!! This variable is also never used here. @ivanmilevtues what the idea behind this?

repo_utils/__init__.py
Svilen-Stefanov: It ignores the following 2 errors:
```
repo_utils\__init__.py:20: error: Cannot assign to a type  [misc]
repo_utils\__init__.py:20: error: Incompatible types in assignment (expression has type "None", variable has type "type[Repo]")  [assignment]
```

static_analyzer/lsp_client/client.py
Svilen-Stefanov: The only reason I kept it is because of the way the lines before were handled:
```
if not module_name or module_name.startswith("."):
            return ""
```
I can make this return `str | None` but I guess that we'd like to return `None` for both of these cases.

agents/agent.py
brovatten: Yup, but like I mentioned I think env variables are the standard, we should not be passing it

agents/agent_responses.py
brovatten: Gemini gave me this:

```
# Mypy is happy here because isinstance checks if it's a class first
if isinstance(ftype, type) and issubclass(ftype, LLMBaseModel):
```

repo_utils/__init__.py
brovatten: Smart, thanks!

agents/tools/read_structure.py
ivanmilevtues: Svilen is right if there is no static analysis we still go through with the agents just doing stuff, this way the extension can be used on anything even though inaccurate :)

agents/agent.py
ivanmilevtues: so it will pick it up automatically from the env vars is what you are saying right?

agents/agent_responses.py
```diff
@@ -41,14 +42,14 @@ class SourceCodeReference(LLMBaseModel):
         description="Qualified name of the source code, e.g., `langchain.tools.tool` or `langchain_core.output_parsers.JsonOutputParser` or `langchain_core.output_parsers.JsonOutputParser:parse`."
     )
 
-    reference_file: Optional[str] = Field(
+    reference_file: str | None = Field(
```
ivanmilevtues, line 45: I am unsure if this is okay to stay like this.
As these Pydantic classes are picked up for the LLM to send context. I am unsure if not having Optional is fine with it. As it does some prompt generaiton based on the defined types and str or None, is different than optional string in terms of prompting. Unsure though just something to keep in mind.

agents/agent_responses.py
ivanmilevtues: That is my bad :D, I have no idea how it didn't break tbh :D

agents/details_agent.py
```diff
@@ -62,12 +62,12 @@ def __init__(
             ),
         }
 
-        self.context = {}
+        self.context: dict[str, AnalysisInsights] = {}
```
ivanmilevtues, line 65: This is not correct.

Line 70 breaks it already. The step_cfg just takes a subpart of the cfg which is relevant and gives it back as string, so for that it'd be dict[str, str]

diagram_analysis/diagram_generator.py
```diff
@@ -27,19 +26,24 @@ def __init__(self, repo_location, temp_folder, repo_name, output_dir, depth_leve
         self.temp_folder = temp_folder
         self.repo_name = repo_name
         self.output_dir = output_dir
-        self.details_agent = None
-        self.abstraction_agent = None
-        self.planner_agent = None
-        self.validator_agent = None
-        self.diff_analyzer_agent = None
-        self.meta_agent = None
+        self.details_agent: DetailsAgent | None = None
```
ivanmilevtues, line 29: Now this is ugly in my opinion :D. as it is only to define them as properties and have them lazily instnatiated :(

If there is no other way I get it, but it is so much boiler plate :D

evals/static_analysis_eval.py
ivanmilevtues: oooh no I am dying, @brovatten didn't you merge the change where we fixed this. So I reworked the way we do the diagram generation, there is a StaticAnalysis class, the generation has just a generate_analysis method.
It got messed up at some point around websummit.

What is the state of that thing?

Svilen-Stefanov: It seems like this was last changed ~4 weeks ago so that would make sense, ref to the commit:
https://github.com/CodeBoarding/CodeBoarding/commit/051ef6b5e8ae1371a12b69e25cd30999b44b0c6e

agents/agent_responses.py
Svilen-Stefanov: They are equivalent so we can revert this. It's just a nit that I wanted to be consistent across the codebase.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Yeah, BaseModel + builders could potentially solve all of these problems. I think you already created a ticket to fix these later.

agents/details_agent.py
Svilen-Stefanov: Is it? If this was `dict[str, str]`, how would that work:
```
cfg_insight = self.context.get("cfg_insight")
cfg_insight_str = cfg_insight.llm_str() if cfg_insight else "No CFG insight available."
```
Here are the exact errors from mypy if you try to change the type to `dict[str, str]`:
```
agents\details_agent.py:96: error: "str" has no attribute "llm_str"  [attr-defined]                                                                                                                           
agents\details_agent.py:115: error: "str" has no attribute "llm_str"  [attr-defined]                                                                                                                          
Found 2 errors in 1 file (checked 74 source files)
```
So the typing seems correct to me.

PR #114, opened 2025-12-03
Analysis doesn't fail on bad git state

ivanmilevtues:
We used to fail whenever the git state was bad, when instantiating a Repo.
Now it is passively handled.

### DIFF
--- a/repo_utils/__init__.py
+++ b/repo_utils/__init__.py
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 
 # Handle the case where git is not installed on the system
 try:
-    from git import Repo, Git, GitCommandError, InvalidGitRepositoryError, GitError
+    from git import Repo, Git, GitCommandError, GitError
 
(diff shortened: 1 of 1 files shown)


repo_utils/__init__.py
```diff
@@ -38,7 +38,14 @@ def wrapper(*args, **kwargs):
                     return default
                 logger.error(f"Git module required for {func.__name__} but not installed")
                 raise ImportError("GitPython is not installed. Install it with: pip install gitpython")
-            return func(*args, **kwargs)
+            try:
+                return func(*args, **kwargs)
+            except GitError as e:
```
Svilen-Stefanov, line 43: Thank you for adding this! Makes sense. 
I don't quite like the global state in this file and the typing in the function but let's leave them for later.

Svilen-Stefanov: Maybe one comment - the name `require_git_import` might be misleading now given that it is not "required".

PR #115, opened 2025-12-03
Reworked evaluations

brovatten:
This PR:
- Splits the monitoring into different classes inside a `monitoring` folder.
- Removes custom logic from the eval files for monitoring, and puts it inside the main code, with the flag `ENABLE_MONITORING` as an env variable.
- Streams the results to json as opposed to having it stateful.

I would really appreciate if you guys could read through and understand how we save the results and produce them.

I am considering if we should keep the report generator as is, OR, use jupyter notebook file for evals to be more agile. In either case we save the json run files. The hassle now is having to run the whole eval end to end. If the plotting fails, we don't have to have to rerun the whole pipeline. Of course this could be solved, but requires more robust pipeline that I'm not sure is worth atm.

Don't let this PR die with me, review and let's revive the evals 🎉 

PS: have to update the `uv.lock` (waiting to not break anything), and not sure if `seaborn` is needed, too bloated for my taste

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -170,7 +170,7 @@ class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
                     {"messages": [self.system_message, HumanMessage(content=prompt)]},
                     config={"callbacks": callback_list, "recursion_limit": 40},
                 )
-                
+
                 agent_response = response["messages"][-1]
                 assert isinstance(agent_response, AIMessage), f"Expected AIMessage, but got {type(agent_response)}"
                 if isinstance(agent_response.content, str):


agents/abstraction_agent.py
```diff
@@ -2,7 +2,7 @@
 import os.path
 from pathlib import Path
 
-from langchain.prompts import PromptTemplate
+from langchain_core.prompts import PromptTemplate
```
ivanmilevtues, line 5: Are you sure this is properly setup?

There is no change in dependency versions but this is changed 🤔

agents/abstraction_agent.py
```diff
@@ -61,7 +61,7 @@ def __init__(
             "feedback": PromptTemplate(template=get_feedback_message(), input_variables=["analysis", "feedback"]),
         }
 
-    @monitoring
+    @trace_step("step_cfg")
```
ivanmilevtues: Pretty sure you can get the name in the decorator:
```python
# Source - https://stackoverflow.com/a
# Posted by Zach Estela, modified by community. See post 'Timeline' for change history
# Retrieved 2025-12-05, License - CC BY-SA 4.0

import time

def timeit(func):
    def timed(*args, **kwargs):
        ts = time.time()
        result = func(*args, **kwargs)
        te = time.time()
        print('Function', func.__name__, 'time:', round((te -ts)*1000,1), 'ms')
        print()
        return result
    return timed

@timeit
def math_harder():
    [x**(x%17)^x%17 for x in range(1,5555)]
math_harder()

@timeit
def sleeper_agent():
    time.sleep(1)
sleeper_agent()
```

output:
```# Source - https://stackoverflow.com/a
# Posted by Zach Estela, modified by community. See post 'Timeline' for change history
# Retrieved 2025-12-05, License - CC BY-SA 4.0

Function math_harder time: 8.4 ms
Function sleeper_agent time: 1003.7 ms
```

agents/agent.py
```diff
@@ -28,22 +29,24 @@
...
+
 
 class CodeBoardingAgent(ReferenceResolverMixin):
     def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str):
         super().__init__(repo_dir, static_analysis)
         self._setup_env_vars()
         self.llm = self._initialize_llm()
         self.extractor_llm = self._initialize_llm()
-        self._monitoring_callback = None  # Used by monitor decorator
+        # self._monitoring_callback = None  # Used by monitor decorator
```
ivanmilevtues: forgotten comment.

diagram_analysis/diagram_generator.py
```diff
@@ -83,27 +90,62 @@ def process_component(self, component):
...
 
+        # --- Capture Static Analysis Stats ---
+        static_stats = {"repo_name": self.repo_name, "languages": {}}
+
+        for language in static_analysis.get_languages():
+            files = static_analysis.get_source_files(language)
+            file_count = len(files)
+            total_loc = 0
+
+            # Calculate LOC (simplistic approach)
```
ivanmilevtues: You can use `tokei` to get that information, or directly extract it and get it from `static_analyzer.scanner.ProjectScanner`, tokei gives the LoC, they are unused atm

evals/definitions/scalability.py
```diff
@@ -0,0 +1,168 @@
...
+from evals.base import BaseEval
+from evals.utils import generate_header
+
+
+class ScalabilityEval(BaseEval):
+    def extract_metrics(self, project: Dict, run_data: Dict) -> Dict[str, Any]:
+        code_stats = run_data.get("code_stats", {})
+        llm_usage = run_data.get("llm_usage", {})
+
+        # Calculate total LOC
```
ivanmilevtues, line 28: You can use tokei or the `static_analyzer.scanner.ProjectScanner` class.

diagram_analysis/diagram_generator.py
```diff
@@ -129,78 +199,92 @@ def generate_analysis(self):
...
+                current_level_components = next_level_components
+
+            logger.info(f"Analysis complete. Generated {len(files)} analysis files")
+            print("Generated analysis files: %s", [os.path.abspath(file) for file in files])
+
+            return files
+        except Exception as e:
+            error_message = str(e)
+            raise
+        finally:
```
ivanmilevtues: This method has grown way too much, we should split it 🤪

pyproject.toml
```diff
@@ -32,14 +32,14 @@ dependencies = [
     "langchain==0.3.27",
     "langchain-anthropic==0.3.18",
     "langchain-aws==0.2.27",
-    "langchain-cerebras>=0.8.1",
```
ivanmilevtues, line 35: This should proabably stay right?

agents/abstraction_agent.py
brovatten: We can yes, I just think this is more flexible. Also for backwards compatability. Agreed? Otherwise I'll change it

agents/abstraction_agent.py
brovatten: The import from `langchain` is deprecated, `langchain_core` is the correct one

pyproject.toml
brovatten: Yup, thanks!

evals/definitions/scalability.py
brovatten: Great point!

diagram_analysis/diagram_generator.py
brovatten: Given that it's only 10 lines more than before, I think it's okay for now. Would rather keep that in a separate PR

diagram_analysis/diagram_generator.py
brovatten: Great point, this is done!

agents/abstraction_agent.py
```diff
@@ -21,7 +21,7 @@
     get_feedback_message,
     get_classification_message,
 )
-from agents.monitoring import monitoring
+from monitoring import trace_step
```
Svilen-Stefanov: Nice, I like that we're introducing a separate module for that.

agents/abstraction_agent.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         self.project_name = project_name
         self.meta_context = meta_context
 
-        self.context: dict[str, AnalysisInsights | list[AnalysisInsights]] = {
+        self.context: dict[str, AnalysisInsights | CFGAnalysisInsights | list[AnalysisInsights]] = {
```
Svilen-Stefanov: This attribute starts to look overloaded. I think BaseModels can help here as well because it seems we only need that because different string attributes match to different values. In a BaseModel, we can say exactly what attributes we expect and what types they should be, which can significantly improve things.

agents/agent.py
```diff
@@ -147,8 +152,9 @@ def _invoke(self, prompt, callbacks: list | None = None) -> str:
         for _ in range(max_retries):
             try:
                 callback_list = callbacks or []
-                if self._monitoring_callback:
-                    callback_list.append(self._monitoring_callback)
+                # Always append monitoring callback - logging config controls output
+                callback_list.append(MONITORING_CALLBACK)
+                callback_list.append(self.agent_monitoring_callback)
```
Svilen-Stefanov, line 167: Do we need both of these? I see:
```
MONITORING_CALLBACK = MonitoringCallback()
...
self.agent_monitoring_callback = MonitoringCallback(stats_container=self.agent_stats)
```

Why do we need `MONITORING_CALLBACK`? Can't we use a single monitoring callback instead of two?

agents/agent.py
```diff
@@ -193,8 +199,8 @@ def _parse_response(self, prompt, response, return_type, max_retries=5):
         if response is None or response.strip() == "":
             logger.error(f"Empty response for prompt: {prompt}")
         try:
-            config = {"callbacks": [self._monitoring_callback]} if self._monitoring_callback else None
-            result = extractor.invoke(return_type.extractor_str() + response, config=config)
+            config: Dict[str, Any] = {"callbacks": [MONITORING_CALLBACK, self.agent_monitoring_callback]}
```
Svilen-Stefanov:
```suggestion
            config: dict[str, Any] = {"callbacks": [MONITORING_CALLBACK, self.agent_monitoring_callback]}
```

demo.py
```diff
@@ -174,13 +175,16 @@ def copy_files(temp_folder: Path, output_dir: Path):
     setup_logging()
     logger.info("Starting up…")
 
-    for repo in tqdm(args.repositories, desc="Generating docs for repos"):
-        temp_repo_folder = create_temp_repo_folder()
-        try:
-            generate_docs_remote(repo, temp_repo_folder, local_dev=True)
+    with monitor_execution(run_id="demo_run", enabled=is_monitoring_enabled()) as mon:
```
Svilen-Stefanov: Can we improve the naming here, e.g:
```suggestion
    with monitor_execution(run_id="demo_run", enabled=is_monitoring_enabled()) as monitoring_context:
```

diagram_analysis/diagram_generator.py
```diff
@@ -34,6 +40,8 @@ def __init__(self, repo_location, temp_folder, repo_name, output_dir, depth_leve
         self.meta_agent: MetaAgent | None = None
         self.meta_context = None
         self.depth_level = depth_level
+        self._monitoring_agents: dict[str, CodeBoardingAgent | None] = {}
```
Svilen-Stefanov: Hmm, why are we adding `None` values to the dictionary. Wouldn't it make more sense if they were not in the dictionary if they were `None`? 
I'd like this to be:
```suggestion
        self._monitoring_agents: dict[str, CodeBoardingAgent] = {}
```
we'd need to change the callers to handle this accordingly.

diagram_analysis/diagram_generator.py
```diff
@@ -83,27 +91,53 @@ def process_component(self, component):
...
         )
+        self._monitoring_agents["MetaAgent"] = self.meta_agent
         meta_context = self.meta_agent.analyze_project_metadata()
         self.details_agent = DetailsAgent(
             repo_dir=self.repo_location,
             project_name=self.repo_name,
             static_analysis=static_analysis,
             meta_context=meta_context,
         )
+        self._monitoring_agents["AbstractionAgent"] = None  # placeholder for type; assigned below
```
Svilen-Stefanov: I don't understand, why can't we just assign this later (the value is modified 3 lines below again)?

diagram_analysis/diagram_generator.py
```diff
@@ -113,13 +147,41 @@ def pre_analysis(self):
                 ).model_dump_json(indent=2)
             )
 
+        enable_monitoring = os.getenv("ENABLE_MONITORING", "").lower() in ("true", "1", "yes", "on")
```
Svilen-Stefanov: What is the desired bahavior when the env variable is not there? Does it make sense to get a default of empty string?

diagram_analysis/diagram_generator.py
Svilen-Stefanov: I think we can refactor this here already, it's kind of difficult to see which 10 lines changed :D

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class BaseEval(ABC):
+    def __init__(self, name: str, output_dir: Path):
+        self.name = name
+        self.output_dir = output_dir
+        self.results: List[Dict[str, Any]] = []
```
Svilen-Stefanov:
```suggestion
        self.results: list[dict[str, Any]] = []
```
and remove the imports from `typing`. Reference: https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict
Applies to the rest of the PR as well.

evals/cli.py
Svilen-Stefanov, line 1: I imagine you're thinking we will run these manually to test if things work as expected.
Instead of creating a separate way to run the code, should we write a test that executes the things we want and evaluate the things we expect? What's the benefit of creating it this way?
Edit: we might want to keep this for the snapshot generation but I still see a benefit of having actual tests we can run to validate things behave as expected (within our expected ranges).

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+
+            self.results.append(metrics)
+
+            if metrics["success"]:
+                logger.info(f"✅ {project['name']} completed in {metrics['duration_seconds']:.1f}s")
+            else:
+                logger.error(f"❌ {project['name']} failed: {metrics.get('error', 'Unknown')[:100]}")
+
+        # 4. Generate & Save Report
+        report_content = self.generate_report(self.results)
```
Svilen-Stefanov, line 237: Iiuc, the current approach assumes we run the CLI, which then generates a bunch of files that someone needs to manually look at, is that correct? 
The way that I imagine it is that we generate a report once and then on each change we make, we run a test that generates something based on our baselines, we compare that with our expectation and fail the test if it doesn't match what we expect.
We can of course regenerate the files if we want / need to because of change of expectations but I'm not sure we want to just generate files for manual review.

evals/definitions/end_to_end.py
```diff
@@ -0,0 +1,118 @@
+from typing import Dict, Any, List
+from evals.base import BaseEval
+from evals.utils import generate_header
+
+
+class EndToEndEval(BaseEval):
```
Svilen-Stefanov, line 39: Can we document exactly what these evals are doing? What does end to end mean here or scalability?
By looking at the generated MD files, it seems like nothing got generated, how did you test these?

diagram_analysis/diagram_generator.py
Svilen-Stefanov: I just saw we also have:
```
def is_monitoring_enabled():
    return os.getenv("ENABLE_MONITORING", "false").lower() in ("1", "true", "yes")
```
Do we need both? Can we reuse `is_monitoring_enabled`?

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "markdown==3.8",
     "markdown-it-py==3.0.0",
     "markitdown==0.1.1",
+    "matplotlib>=3.10.7",
```
Svilen-Stefanov: Can we fix the dependencies (`==`) and not use `>=`?

evals/definitions/end_to_end.py
brovatten: Thanks, I am adding the explanations now.

The MD files are not generated, I will include this for the final iteration (takes quite some time). We run the with the CLI

evals/base.py
brovatten: Check my description at the top - my concern is that we overengineer this.

IMO we should have a consistent benchmark. We should run this manually at every major change to the pipeline. But running it too often is hard to setup with the API calls, and it is kind of expensive, and should not be run on most PRs.

I also think the report generation will change so much that a Jupyter notebook is easier

evals/cli.py
brovatten: It's not a health test, it's a performance benchmark.

I thought of having one to run it all, like we had before, but like I write down below we often times don't want to run these at all. Sometimes we do, but we are often interested in either the static analyisis OR the LLM performance

demo.py
```diff
@@ -174,13 +175,16 @@ def copy_files(temp_folder: Path, output_dir: Path):
...
-            remove_temp_repo_folder(temp_repo_folder)
+            temp_repo_folder = create_temp_repo_folder()
+            try:
+                generate_docs_remote(repo, temp_repo_folder, local_dev=True)
+
+                # Copy markdown files to output directory if specified
+                if args.output_dir:
+                    copy_files(temp_repo_folder, args.output_dir)
+            finally:
+                remove_temp_repo_folder(temp_repo_folder)
```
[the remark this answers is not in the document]
ivanmilevtues: @brovatten this sounds like a legit comment

agents/abstraction_agent.py
ivanmilevtues: This has to be LLMBaseModel, as we discussed in #119

agents/abstraction_agent.py
ivanmilevtues: What do you mean by backwards compatibility? I am unsure what backwards compatibility you mean, old analysis? old python? unsure :D 

To keep it flexible I will add it as optional parameter - if not set take the function name. (I just think it is easier to maintain with less hardcoded strings).

agents/agent.py
```diff
@@ -193,8 +199,8 @@ def _parse_response(self, prompt, response, return_type, max_retries=5):
         if response is None or response.strip() == "":
             logger.error(f"Empty response for prompt: {prompt}")
         try:
-            config = {"callbacks": [self._monitoring_callback]} if self._monitoring_callback else None
-            result = extractor.invoke(return_type.extractor_str() + response, config=config)
+            config: dict[str, Any] = {"callbacks": [MONITORING_CALLBACK, self.agent_monitoring_callback]}
```
ivanmilevtues: nitpick, I don't like this type of inline typing when the type is on the right

But if we want to do it, we should also have it on line 234.

Personally, I would NOT have it.

monitoring/writers.py
```diff
@@ -0,0 +1,135 @@
...
+class StreamingStatsWriter:
+    """
+    Handles periodic writing of monitoring stats to a JSON file.
+    Also tracks run timing and saves run_metadata.json on stop.
+    """
+
+    def __init__(
+        self,
+        monitoring_dir: Path,
+        agents_dict: dict,
```
ivanmilevtues: Add typing to the agents_dict [str, CodeBoardingAgent]
This way we don't have to do attribute checks i.e. line 75

agents/agent.py
```diff
@@ -236,22 +242,5 @@ def _try_parse(self, message_content, parser):
         raise ValueError(f"Couldn't parse {message_content}")
 
     def get_monitoring_results(self) -> dict:
```
ivanmilevtues: Much cleaner than before, if we want to be superbly clean I would make it a Mixin like the ReferenceResolverMixin.

MonitoringMixin and it will have the instantiations of the agent_stats in it + this method.

agents/abstraction_agent.py
ivanmilevtues, line 5: Good catch, unsure what the difference is, but we should use _core.

agents/agent.py
```diff
@@ -28,22 +30,23 @@
...
+    ReadDocsTool,
 )
 from agents.tools.external_deps import ExternalDepsTool
-from agents.tools.read_docs import ReadDocsTool
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.reference_resolve_mixin import ReferenceResolverMixin
 
 logger = logging.getLogger(__name__)
 
+MONITORING_CALLBACK = MonitoringCallback()
```
ivanmilevtues, line 41: To add to what Copilot things about this :D. If you decide to go with the mixin approach I think it should be a class parameter there, instead of hanging here as a global variable.

agents/details_agent.py
```diff
@@ -22,7 +22,7 @@
     get_feedback_message,
     get_classification_message,
 )
-from agents.monitoring import monitoring
+from monitoring import trace_step
```
ivanmilevtues: Nice I like this way of importing stuff rather than having module.file_name import file_name

diagram_analysis/diagram_generator.py
```diff
@@ -83,27 +91,53 @@ def process_component(self, component):
     def pre_analysis(self):
         static_analysis = StaticAnalyzer(self.repo_location).analyze()
 
+        # --- Capture Static Analysis Stats ---
+        static_stats = {"repo_name": self.repo_name, "languages": {}}
```
ivanmilevtues: These lines should go to someplace not here, either a method here (which I don't like too much) or have a Monitoring class or something which does all the setup that is done in this pre_analysis:
- the stats
- the dictionary building with the agents.

diagram_analysis/diagram_generator.py
ivanmilevtues: I don't like this being in the DiagramGenerator, it doesn't have much to do with it. 

Continuing on my previous comment I think this should go out into a new class which is a Monitor or something. Which does all of this logic in its methods and we just call it here once.

diagram_analysis/diagram_generator.py
```diff
@@ -129,78 +191,92 @@ def generate_analysis(self):
...
+                logger.info(f"Completed level {level}. Found {len(next_level_components)} components for next level")
+                current_level_components = next_level_components
+
+            logger.info(f"Analysis complete. Generated {len(files)} analysis files")
+            print("Generated analysis files: %s", [os.path.abspath(file) for file in files])
+
+            return files
+        except Exception as e:
+            error_message = str(e)
+            raise
```
ivanmilevtues: raise e, so that we have the exception in the trace.

diagram_analysis/diagram_generator.py
```diff
@@ -129,78 +191,92 @@ def generate_analysis(self):
...
+        finally:
+            # Stop monitoring (saves run_metadata.json with timing)
+            if self.stats_writer:
```
ivanmilevtues: Add this to the monitoring class and have monitor.start/monitor.end

To be completely honest I would override the with and make the whole thing look something like:

with monitor:
   ...existing logic

and then we don't have this try except here which looks sus and we don't have additional logic around start/stop timers.

You can check it here quickly:
https://stackoverflow.com/questions/1984325/explaining-pythons-enter-and-exit

monitoring/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    from monitoring import RunStats, LeanMonitoringCallback, StreamingStatsWriter
+    from monitoring import monitor_execution, trace_step, current_step
+"""
+
+from monitoring.stats import RunStats, stats
+from monitoring.callbacks import MonitoringCallback
+from monitoring.writers import StreamingStatsWriter, save_static_stats
+from monitoring.context import monitor_execution, trace_step, current_step
+
+__all__ = [
```
ivanmilevtues: I don't like this style of exports. Because with them PyCharm and other IDEs cannot resolve the linking. I.e. wherever you've imported RunStats on my Ctrl + click it cannot open and resolve it properly.


Do you have something against the way we do it in `agents.tool.__init__`?
If not can we stick to that style pls :)

monitoring/callbacks.py
```diff
@@ -0,0 +1,161 @@
+"""
```
ivanmilevtues: Would rather have this comment together with the class comment, not a big fan of file comments. Feels like we are writing C code :D

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+                text=True,
+                timeout=1800,  # 30 minutes
+                env=env,
+                cwd=self.project_root,
+            )
+            duration = time.time() - start_time
+            success = result.returncode == 0
+            stderr = result.stderr[-500:] if result.stderr else ""
+
+            return {
```
ivanmilevtues: You can create a PyDantic class/Dataclass for this return type so we don't have to do the -> dict[str -> any]. This way we will be more typesafe and I think it is easer to read the code Return EvalResult i.e.

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+            return {}
+
+        data["metadata"] = read_json("run_metadata.json")
+        data["code_stats"] = read_json("code_stats.json")
+        data["llm_usage"] = read_json("llm_usage.json")
+
+        return data
+
+    def run_pipeline(self, project: dict[str, str], extra_args: list[str] | None = None) -> dict[str, Any]:
+        """Run the CodeBoarding pipeline for a project."""
```
ivanmilevtues: No need of this comment imo.

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+        # Fallback to finding it relative to this file if not set
+        if not project_root_env:
+            # Assuming evals/base.py is 2 levels deep from root
+            root = Path(__file__).parent.parent
+            os.environ["PROJECT_ROOT"] = str(root)
+            return root
+        return Path(project_root_env)
+
+    def get_latest_run_dir(self, project_name: str) -> Path | None:
+        """Find the most recent monitoring run directory for a project."""
```
ivanmilevtues, line 44: no need of this comment imo.

evals/base.py
```diff
@@ -0,0 +1,221 @@
...
+        # Add system specs
+        system_specs = generate_system_specs()
+        full_report = f"{report_content}\n\n{system_specs}"
+
+        report_path = self.output_dir / f"{self.name}-report.md"
+        self._write_report(report_path, full_report)
+        logger.info(f"Report generated: {report_path}")
+
+        # Save raw JSON results too
+        json_path = self.project_root / "evals/artifacts/monitoring_results/reports" / f"{self.name}_eval.json"
```
ivanmilevtues, line 248: I am not sure if the middle part with the "/" won't break on windows and be interpreted as a whole string and not path.

ivanmilevtues: Same for other places in this file i.e. line 87

evals/cli.py
```diff
@@ -0,0 +1,58 @@
...
+from evals.definitions.static_analysis import StaticAnalysisEval
+from evals.definitions.end_to_end import EndToEndEval
+from evals.definitions.scalability import ScalabilityEval
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Run evaluations for CodeBoarding")
```
ivanmilevtues, line 22: I would move the parser creation in a different method to reduce length of the main (L23-L26).

```python 
cli_args = get_args()
```
or something like that.

To keep the flow clean and easy on the eye.

evals/cli.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Set up REPO_ROOT if not set, similar to original main()
+    if not os.getenv("REPO_ROOT"):
+        os.environ["REPO_ROOT"] = "repos"
+
+    evals_to_run: list[tuple[Any, list[dict[str, str]], list[str]]] = []
+
+    if args.type in ["static", "all"]:
+        evals_to_run.append(
+            (StaticAnalysisEval("static-analysis", args.output_dir), PROJECTS_STATIC_ANALYSIS, ["--depth-level", "1"])
```
ivanmilevtues: It seems like each of the evals has its own config. I might suggest to do a quick for loop here with a class create method.

So we have something like:
```python
for eval in BaseEval.__subclasses__():
    if eval.matches(args.type):
        evals_to_run.append(eval.create())

```

This way all the config for the evals stays in the object/class itself. So when we create a new eval it will be picked up automatically.

evals/utils.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
ivanmilevtues: I think there is no need for this type of comment

evals/utils.py
```diff
@@ -0,0 +1,58 @@
...
+from __future__ import annotations
+
+import os
+import platform
+import subprocess
+from datetime import datetime, timezone
+
+
+def generate_header(title: str, timestamp: str | None = None, extra_lines: list[str] | None = None) -> str:
+    """Generate a standard Markdown header with timestamp."""
```
ivanmilevtues: These func comments are not really useful imo.
If you think it is unclear probably rename the function, here I think it is clear so would just remove the comment if you don't think otherwise.

monitoring/stats.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self):
+        self._lock = threading.Lock()
+        self.total_tokens = 0
+        self.input_tokens = 0
+        self.output_tokens = 0
+        self.tool_counts = defaultdict(int)
+        self.tool_errors = defaultdict(int)
+        self.tool_latency_ms = defaultdict(list)
+
+    def to_dict(self):
```
ivanmilevtues, line 27: You can make the object a pydantic object.
They have the dict option by default.

As you have some custom stuff going on: https://docs.pydantic.dev/1.10/usage/exporting_models/#custom-json-deserialisation

I think it is better to stick to pydantic across the repo and reduce custom logic as much as possible looks clearner :).

monitoring/stats.py
```diff
@@ -0,0 +1,42 @@
...
+                    "avg_latency_ms": {
+                        tool: sum(latencies) / len(latencies) if latencies else 0
+                        for tool, latencies in self.tool_latency_ms.items()
+                    },
+                },
+            }
+
+
+# Global stats instance - Lean & Simple for CLI tools
+stats = RunStats()
```
ivanmilevtues: I don't like this global state here, makes it incredibly hard to trackdown. Should be in the CLI script imo, or somewhere else no idea.

Also should be STATS right as it is global :)

monitoring/callbacks.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Update State
+        with self.stats._lock:
+            self.stats.total_tokens += usage.get("total_tokens", 0)
+            self.stats.input_tokens += usage.get("prompt_tokens", 0)
+            self.stats.output_tokens += usage.get("completion_tokens", 0)
+
+        # Log Event
+        logger.info(
+            json.dumps(
```
ivanmilevtues: Is this the best format, wouldn't be easier to read it as a normal string?

Event: LLM Usage, Step: Name, Model: gpt-4o, Usage: usage

Instead of having it as json? Don't mind either way just asking.

monitoring/callbacks.py
```diff
@@ -0,0 +1,161 @@
...
+            self._tool_start_times[run_id] = now
+            self._tool_names[run_id] = tool_name
+
+    def on_tool_end(self, output: Any, **kwargs: Any) -> None:
+        run_id_any = kwargs.get("run_id")
+        run_id: str | None = str(run_id_any) if run_id_any else None
+        if run_id and run_id in self._tool_start_times:
+            start = self._tool_start_times.pop(run_id)
+            tool_name = self._tool_names.pop(run_id, "unknown_tool")
+            latency = int((time.time() - start) * 1000)
```
ivanmilevtues, line 80: Why * 1000 -> to move it into seconds? (I assume)

monitoring/context.py
```diff
@@ -0,0 +1,136 @@
+"""
```
ivanmilevtues: I think this comment is not giving more than the file name + module (these are informative)

monitoring/context.py
```diff
@@ -0,0 +1,134 @@
...
+    # Allow the user to manually log steps via the yielded context
+    class MonitorContext:
+        def step(self, name):
+            trace_logger.info(json.dumps({"event": "phase_change", "step": name, "timestamp": time.time()}))
+            # Also update the ContextVar for other components
+            self._token = current_step.set(name)
+
+        def end_step(self):
+            # Reset if needed, though trace_step decorator usually handles this
+            pass
```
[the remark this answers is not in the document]
ivanmilevtues: Check on this comment :D, I don't know enough to give an opinion. But it seems suspicious to override end with nothing

monitoring/context.py
```diff
@@ -0,0 +1,136 @@
...
+        except Exception as e:
+            logger.error(f"Failed to save monitoring summary: {e}")
+
+        # Cleanup handler
+        trace_logger.removeHandler(trace_handler)
+        trace_handler.close()
+        logger.info(f"✨ Execution traces saved to {trace_file}")
+
+
+def trace_step(step_name: str):
```
ivanmilevtues: nice

demo.py
ivanmilevtues: Nice, that is what I was also thinking about one of the other monitoring stuff, I made a comment on it so you'd not miss it :D

evals/definitions/end_to_end.py
```diff
@@ -0,0 +1,123 @@
...
+            total_tokens += token_usage.get("total_tokens", 0)
+            input_tokens += token_usage.get("input_tokens", 0)
+            output_tokens += token_usage.get("output_tokens", 0)
+
+            for tool, count in agent_data.get("tool_usage", {}).get("counts", {}).items():
+                tool_counts[tool] = tool_counts.get(tool, 0) + count
+            for tool, count in agent_data.get("tool_usage", {}).get("errors", {}).items():
+                tool_errors[tool] = tool_errors.get(tool, 0) + count
+
+        return {
```
ivanmilevtues: I would make this a data class/pydantic object instead of raw JSON/dict.

evals/definitions/scalability.py
```diff
@@ -0,0 +1,173 @@
...
+            # Store detailed usage for charts
+            agent_token_usage[agent_name] = {
+                "input": token_usage.get("input_tokens", 0),
+                "output": token_usage.get("output_tokens", 0),
+            }
+
+            tool_usage = agent_data.get("tool_usage", {}).get("counts", {})
+            agent_tool_usage[agent_name] = tool_usage
+
+        return {
```
ivanmilevtues: I would make this a data class/pydantic object instead of raw JSON/dict.

evals/definitions/static_analysis.py
```diff
@@ -0,0 +1,72 @@
...
+        languages_summary = {}
+
+        for lang, stats in code_stats.get("languages", {}).items():
+            file_count = stats.get("file_count", 0)
+            loc = stats.get("lines_of_code", 0)
+            total_files += file_count
+            total_loc += loc
+            languages_summary[lang] = {"files": file_count, "loc": loc}
+
+        return {
```
ivanmilevtues: I would make this a data class/pydantic object instead of raw JSON/dict.

agents/agent.py
brovatten: Agree, in this instance typing is quite useless

monitoring/writers.py
brovatten: Great point!!

evals/utils.py
brovatten: Agreed! comments removed

monitoring/callbacks.py
brovatten: Yes, it's milli by default

evals/cli.py
brovatten: Good idea! added

evals/base.py
brovatten: It will not, this equiv to `os.path.join` for `pathlib`

diagram_analysis/diagram_generator.py
brovatten: Thanks I overlooked this, will add a print to here also

monitoring/stats.py
brovatten: My gut feeling was this is not the best for the stats, here's what Opus said:
`RunStats` is a stateful container that accumulates metrics with thread safety. Pydantic models are for validating and serializing data shapes, not managing mutable runtime state.

evals/definitions/end_to_end.py
brovatten: Great point, thx for elaborating IRL

monitoring/callbacks.py
brovatten: I see what you mean, this would clutter the logs quite a bit indeed. Changed it to string. 

Could be like it is, but if so for `logger.debug` imo

agents/abstraction_agent.py
brovatten: Thanks for the reference, I agree with you both!

monitoring/context.py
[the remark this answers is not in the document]
brovatten: We now reset in the beginning instead. Will let the `end_step` be placeholder for now

agents/agent.py
brovatten: Yes you do, one is per-agent and one is the global state

diagram_analysis/diagram_generator.py
brovatten: Yup you're right, it was before when we initialized the values as None.

agents/agent.py
brovatten: Makes sense, good idea!

evals/cli.py
brovatten: I think this is easier to maintain as it is now. Since we don't have plans to include many more evals, I will leave it as is

diagram_analysis/diagram_generator.py
brovatten: This is really clean! thanks for the suggestion

brovatten:
Alright guys, please have another look.

FYI the ´config.py´ will be updated last. I am running the E2E and Scalability evals overnight :)

agents/abstraction_agent.py
```diff
@@ -39,10 +41,7 @@ def __init__(
 
         self.project_name = project_name
         self.meta_context = meta_context
-
-        self.context: dict[str, AnalysisInsights | list[AnalysisInsights]] = {
-            "structure_insight": []
-        }  # Store evolving insights here
+        self.context: dict[str, LLMBaseModel] = {"structure_insight": []}
```
Svilen-Stefanov: This might be the reason why you can't automatically merge, I removed the last part because @ivanmilevtues mentioned we don't need it.
```suggestion
        self.context: dict[str, LLMBaseModel] = {}
```

agents/abstraction_agent.py
```diff
@@ -97,9 +96,11 @@ def step_source(self):
                 insight_str += "\n".join([f"- {insight.llm_str()}" for insight in analysis_insight]) + "\n\n"
             elif isinstance(analysis_insight, AnalysisInsights):
                 insight_str += analysis_insight.llm_str() + "\n\n"
+            elif isinstance(analysis_insight, CFGAnalysisInsights):
```
Svilen-Stefanov: Actually, nvm, there are more things, you should rebase on top of main and fix the rest of the conflicts here. This should remain as in main atm.

agents/agent.py
```diff
@@ -3,6 +3,7 @@
 import os
 import time
 from pathlib import Path
+from typing import Any, Dict
```
Svilen-Stefanov: Let's not add `Dict` anymore as discussed in other PRs, and use `dict` instead.

demo.py
```diff
@@ -14,7 +14,8 @@
 from logging_config import setup_logging
```
Svilen-Stefanov: Oh, no, this is not exciting. This will introduce a painful merge conflict after the unification of the two entry points and the main question here for me is why I don't see similar changes for `vscode_runnable.py`. 
In which order should we merge these PRs?

agents/agent.py
```diff
@@ -28,22 +31,24 @@
...
 from static_analyzer.reference_resolve_mixin import ReferenceResolverMixin
 
 logger = logging.getLogger(__name__)
 
+MONITORING_CALLBACK = MonitoringCallback()
 
-class CodeBoardingAgent(ReferenceResolverMixin):
+
+class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
     def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str):
```
ivanmilevtues, line 45: Now that is clean thanks

monitoring/callbacks.py
```diff
@@ -0,0 +1,160 @@
+import json
```
ivanmilevtues, line 1: unused json import

agents/agent.py
```diff
@@ -147,8 +152,9 @@ def _invoke(self, prompt, callbacks: list | None = None) -> str:
         for _ in range(max_retries):
             try:
                 callback_list = callbacks or []
-                if self._monitoring_callback:
-                    callback_list.append(self._monitoring_callback)
+                # Always append monitoring callback - logging config controls output
+                callback_list.append(MONITORING_CALLBACK)
+                callback_list.append(self.agent_monitoring_callback)
 
                 if callback_list:
```
ivanmilevtues: It seems that this if is no longer needed, the list will always be non-empty.

You can remove the if/else and just directly send the callback list same as L202.

I think you can also directly have self.callback_list so that you modify it from a single place.

ivanmilevtues: Same comment applies for L202, L234

diagram_analysis/diagram_generator.py
```diff
@@ -81,29 +103,56 @@ def process_component(self, component):
...
+        loc_by_language = {pl.language: pl.size for pl in scanner.scan()}
+
+        for language in static_analysis.get_languages():
+            files = static_analysis.get_source_files(language)
+
+            static_stats["languages"][language] = {
+                "file_count": len(files),
+                "lines_of_code": loc_by_language.get(language, 0),
+            }
+        # ------------------------------------------
```
ivanmilevtues: not a fan of this cursor line comment :d

diagram_analysis/diagram_generator.py
```diff
@@ -129,78 +206,87 @@ def generate_analysis(self):
...
-            logger.info(f"Completed level {level}. Found {len(next_level_components)} components for next level")
-            current_level_components = next_level_components
-
-        logger.info(f"Analysis complete. Generated {len(files)} analysis files")
-        print("Generated analysis files: %s", [os.path.abspath(file) for file in files])
-
-        return files
+        # Start monitoring (tracks start time)
+        monitor = self.stats_writer if self.stats_writer else nullcontext()
+        with monitor:
```
ivanmilevtues, line 212: Call me biased, but if this isn't beauty I don't know what is.
Noice

evals/reports/end-to-end-report.md
```diff
@@ -1,78 +1,20 @@
 # End-to-End Pipeline Evaluation
```
ivanmilevtues, line 1: Just a question, the way we'd track these over time is with the githistory right?
Or are they appending one below the other?

This is a question for all 3 repots

evals/base.py
```diff
@@ -0,0 +1,270 @@
...
+        if not project_root_env:
+            # Assuming evals/base.py is 2 levels deep from root
+            root = Path(__file__).parent.parent
+            os.environ["PROJECT_ROOT"] = str(root)
+            return root
+        return Path(project_root_env)
+
+    def get_latest_run_dir(self, project_name: str) -> Path | None:
+        """Find the most recent monitoring run directory for a project."""
+        runs_dir = self.project_root / "evals/artifacts/monitoring_results/runs"
```
ivanmilevtues, line 45: This string path is hardcoded and will probably break on windows

evals/config.py
```diff
@@ -0,0 +1,38 @@
...
+        env_vars={"DIAGRAM_DEPTH_LEVEL": "1"},
+    ),
+    ProjectSpec(
+        name="bertopic-depth-1",
+        url="https://github.com/MaartenGr/BERTopic",
+        expected_language="Python",
+        env_vars={"DIAGRAM_DEPTH_LEVEL": "1"},
+    ),
+    # ProjectSpec(
+    #     name="markitdown-depth-2",
```
ivanmilevtues, line 33: Left out comment

evals/config.py
```diff
@@ -0,0 +1,38 @@
+from evals.schemas import ProjectSpec
+
+PROJECTS_STATIC_ANALYSIS = [
+    ProjectSpec(name="markitdown", url="https://github.com/microsoft/markitdown", expected_language="Python"),
+    ProjectSpec(name="bertopic", url="https://github.com/MaartenGr/BERTopic", expected_language="Python"),
+    ProjectSpec(name="browser-use", url="https://github.com/browser-use/browser-use", expected_language="Python"),
+    ProjectSpec(name="fastapi", url="https://github.com/fastapi/fastapi", expected_language="Python"),
+    ProjectSpec(name="django", url="https://github.com/django/django", expected_language="Python"),
+    # ProjectSpec(name="tsoa", url="https://github.com/lukeautry/tsoa", expected_language="TypeScript"),
+    # ProjectSpec(name="cobra", url="https://github.com/spf13/cobra", expected_language="Go"),
```
ivanmilevtues, line 10: Are this WIP (or used as notes). Atm seem like forgotten comments

monitoring/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+Usage:
+    from monitoring import RunStats, MonitoringCallback, StreamingStatsWriter
+    from monitoring import monitor_execution, trace, current_step
+"""
+
+from .stats import RunStats
+from .callbacks import MonitoringCallback
+from .writers import StreamingStatsWriter, save_static_stats
+from .context import monitor_execution, trace, current_step
```
ivanmilevtues, line 12: Thanks for switching the imports to this style.
It really seems to work much better with IDE/LSP tooling for reference resolution!

monitoring/stats.py
ivanmilevtues: mm okay, I don't see what thread safety has to do with Pyrdantic objects (why are they not safe). My comment was mostly around eye candy in a way, maybe I spent too much time with typed langs (Java/Kotlin/C#).

This seems to Javascripty for my taste :D, but I am fine with it.

demo.py
ivanmilevtues: Now that is rough :D

demo.py
```diff
@@ -175,12 +191,24 @@ def copy_files(temp_folder: Path, output_dir: Path):
...
-
-            # Copy markdown files to output directory if specified
-            if args.output_dir:
-                copy_files(temp_repo_folder, args.output_dir)
-        finally:
-            remove_temp_repo_folder(temp_repo_folder)
+        repo_name = get_repo_name(repo)
+        run_id = args.project_name if args.project_name else f"demo_run_{repo_name}"
+
+        with monitor_execution(run_id=run_id, enabled=monitoring_enabled()) as mon:
```
ivanmilevtues: Now that is some Gordon Ramsay stuff here

demo.py
```diff
@@ -166,6 +180,8 @@ def copy_files(temp_folder: Path, output_dir: Path):
     )
     parser.add_argument("repositories", nargs="+", help="One or more Git repository URLs to generate documentation for")
     parser.add_argument("--output-dir", type=Path, help="Directory to copy generated markdown files to")
+    parser.add_argument("--static-only", action="store_true", help="Run only static analysis without agents")
+    parser.add_argument("--project-name", help="Custom project name for monitoring runs")
```
ivanmilevtues: hmm this project-name will be problematic for the vscode runnable.

evals/reports/end-to-end-report.md
brovatten: Git history

evals/base.py
brovatten: It's not, pathlib resolves this

evals/config.py
brovatten: Like I mentioned in my comment b4 the review, the config is work in progress :D

demo.py
brovatten: It's optional, only to help out when we run monitoring for different depths. It's since we save the stuff inside the main pipeline and not the eval itself.

Should we rename all saved files to `repo_name_DEPTH=X` instead? This would solve it

main.py
```diff
@@ -318,12 +326,13 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
     """
     Adds all command-line arguments and groups to the ArgumentParser.
     """
-    # Repository specification (mutually exclusive groups)
-    repo_group = parser.add_mutually_exclusive_group(required=True)
```
brovatten, line 322: I liked this, but I tried running it and got:
```
Traceback (most recent call last):
  File "/Users/brovatten/Downloads/temp_code/CodeBoarding/main.py", line 464, in <module>
    main()
    ~~~~^^
  File "/Users/brovatten/Downloads/temp_code/CodeBoarding/main.py", line 389, in main
    define_cli_arguments(parser)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/brovatten/Downloads/temp_code/CodeBoarding/main.py", line 331, in define_cli_arguments
    repo_group.add_argument(
    ~~~~~~~~~~~~~~~~~~~~~~~^
        "repositories", nargs="+", help="One or more Git repository URLs to generate documentation for"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/argparse.py", line 1489, in add_argument
    return self._add_action(action)
           ~~~~~~~~~~~~~~~~^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/argparse.py", line 1724, in _add_action
    raise ValueError(msg)
ValueError: mutually exclusive arguments must be optional
```

agents/agent.py
```diff
@@ -204,8 +210,8 @@ def _parse_response(self, prompt, response, return_type, max_retries=5):
         if response is None or response.strip() == "":
             logger.error(f"Empty response for prompt: {prompt}")
         try:
-            config = {"callbacks": [self._monitoring_callback]} if self._monitoring_callback else None
-            result = extractor.invoke(return_type.extractor_str() + response, config=config)
+            config = {"callbacks": [MONITORING_CALLBACK, self.agent_monitoring_callback]}
+            result = extractor.invoke(return_type.extractor_str() + response, config=config)  # type: ignore[arg-type]
```
Svilen-Stefanov, line 208: I don't quite like that we ignore the type here. What's the reason for that?

diagram_analysis/diagram_generator.py
```diff
@@ -14,18 +19,33 @@
...
+from utils import monitoring_enabled
 
 logger = logging.getLogger(__name__)
 
 
 class DiagramGenerator:
-    def __init__(self, repo_location, temp_folder, repo_name, output_dir, depth_level: int):
+    def __init__(
+        self,
+        repo_location,
```
Svilen-Stefanov: nit: can we add types here?

agents/abstraction_agent.py
```diff
@@ -1,14 +1,16 @@
 import logging
 import os.path
+from typing import Any
```
Svilen-Stefanov: Unused import, remove.

agents/agent.py
Svilen-Stefanov: These also seem unused and can be removed.

agents/agent.py
Svilen-Stefanov: Yup, we know there's a list so let's remove this check.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Same, let's get rid of it.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Yup, I agree.

main.py
Svilen-Stefanov: Thank you! My bad.

monitoring/context.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Tracks the current high-level operation (e.g., "static_analysis", "code_generation")
+current_step: ContextVar[str] = ContextVar("current_step", default="startup")
+
+logger = logging.getLogger("monitoring")
+
+
+@contextlib.contextmanager
+def monitor_execution(
+    run_id: str | None = None, output_dir: str = "evals/artifacts/monitoring_results", enabled: bool = True
```
Svilen-Stefanov, line 20: Can this not be a string but a path?

monitoring/context.py
```diff
@@ -0,0 +1,149 @@
...
+                pass
+
+        yield DummyContext()
+        return
+
+    # Default run_id if none provided
+    if not run_id:
+        from uuid import uuid4
+
+        run_id = str(uuid4())[:8]
```
Svilen-Stefanov, line 46: What's the reason we take the first 8 chars here? Can't we just use the uuid4?

PR #116, opened 2025-12-03
Fix main again

brovatten:
Saying "again" as I am on a roll!

Fixed by downgrading the `langchain-openai` package. 

Pydantic validation also threw an error, that I fixed in the `external_deps.py`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "isort==6.0.1",
     "jiter==0.10.0",
     "jsonschema==4.25.0",
-    "langchain>=0.3.0",
+    "langchain==0.3.27",
     "langchain-anthropic==0.3.18",
     "langchain-aws==0.2.27",
     "langchain-cerebras>=0.8.1",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -29,13 +29,14 @@ dependencies = [
     "isort==6.0.1",
     "jiter==0.10.0",
     "jsonschema==4.25.0",
+    "langchain>=0.3.0",
```
Svilen-Stefanov: Let's fix the version to something specific. Apart from that, lgtm.

agents/tools/external_deps.py
```diff
@@ -52,8 +52,7 @@ class ExternalDepsTool(BaseTool):
     ]
 
     def __init__(self, repo_dir: Path):
-        super().__init__()
-        self.repo_dir = repo_dir
+        super().__init__(repo_dir=repo_dir)
```
ivanmilevtues, line 55: Now this seems more like the proper fix, unlike the other PR :)

PR #117, opened 2025-12-04
Don't set ROOT_RESULT automatically

brovatten:
See comments below. Both of these are crashing `main`


agents/tools/external_deps.py
```diff
@@ -51,10 +51,6 @@ class ExternalDepsTool(BaseTool):
         "tsconfig.json",  # TypeScript compiler configuration (not dependencies, but relevant)
     ]
 
-    def __init__(self, repo_dir: Path):
-        super().__init__()
-        self.repo_dir = repo_dir
```
Svilen-Stefanov: The fix here seems different than the one here: https://github.com/CodeBoarding/CodeBoarding/pull/116/files
Is this the main reason for the failure?

ivanmilevtues: The tool needs the repo_dir i.e. line 68

This should not resolve this issue I think, or am I missing something?

agents/abstraction_agent.py
```diff
@@ -95,7 +95,7 @@ def step_source(self):
             insight_str += f"## {insight_type.capitalize()} Insight\n"
             if isinstance(analysis_insight, list):
                 insight_str += "\n".join([f"- {insight.llm_str()}" for insight in analysis_insight]) + "\n\n"
-            elif isinstance(analysis_insight, AnalysisInsights):
+            elif isinstance(analysis_insight, (AnalysisInsights, CFGAnalysisInsights)):
```
ivanmilevtues, line 98: This is addressed by #119. So when you rebase just make sure we use LLMBaseModel

ivanmilevtues (approved):
Looks good, just Rebase after #119

PR #118, opened 2025-12-04
Allow the user to pick their own model via environment variable

Svilen-Stefanov:


agents/agent.py
```diff
@@ -80,12 +80,16 @@ def _setup_env_vars(self):
...
         self.ollama_base_url = os.getenv("OLLAMA_BASE_URL")
 
+        # Model selection via environment variable
+        self.codeboarding_model = os.getenv("CODEBOARDING_MODEL")
+
     def _initialize_llm(self):
         """Initialize LLM based on available API keys with priority order."""
         if self.openai_api_key:
-            logger.info("Using OpenAI LLM")
+            model = self.codeboarding_model if self.codeboarding_model else "gpt-4o"
```
ivanmilevtues, line 89: I would put gemini-2.5-flash as default as this is the one that we've been using the most and probably is the most prompt optimised.

PR #119, opened 2025-12-06
Fix abstact_agent context type causing failures

Svilen-Stefanov:
Revert some of the mypy confusion, which causes the following error atm:
```
TypeError: Expected analysis_insight to be either list or AnalysisInsights, but got CFGAnalysisInsights for insight_type 'cfg_insight'
```

### DIFF
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -41,7 +41,7 @@ class AbstractionAgent(CodeBoardingAgent):
         self.project_name = project_name
         self.meta_context = meta_context
 
-        self.context: dict[str, LLMBaseModel] = dict()
+        self.context: dict[str, LLMBaseModel] = {}
 
         self.prompts = {
             "cfg": PromptTemplate(
@@ -92,13 +92,7 @@ class AbstractionAgent(CodeBoardingAgent):
         insight_str = ""
         for insight_type, analysis_insight in self.context.items():
             insight_str += f"## {insight_type.capitalize()} Insight\n"
-            if isinstance(analysis_insight, LLMBaseModel):
-                insight_str += analysis_insight.llm_str() + "\n\n"
-            else:
-                raise TypeError(
-                    f"Expected analysis_insight to be either list or LLMBaseModel, "
-                    f"but got {type(analysis_insight).__name__} for insight_type '{insight_type}'"
-                )
+            insight_str += analysis_insight.llm_str() + "\n\n"
 
         meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
         project_type = self.meta_context.project_type if self.meta_context else "unknown"


agents/abstraction_agent.py
```diff
@@ -40,7 +41,7 @@ def __init__(
         self.project_name = project_name
         self.meta_context = meta_context
 
-        self.context: dict[str, AnalysisInsights | list[AnalysisInsights]] = {
+        self.context: dict[str, LLMBaseModel | list[LLMBaseModel]] = {
```
ivanmilevtues: This is just a dict[LLMBaseModel]

I checked if we ever use list, and we don't.

This "structure_insight" is also unsused so it can be removed.

```python

**Review by ivanmilevtues — CHANGES_REQUESTED** [review]
Good catch, a small request to make the typing tighter.

Also we should update accordingly in all agents (dict of string -> LLMBaseModel):
agents/details_agent.py line 65 # Also please check line 70 which has an ignore at the moment. 

We can also add return types to the validator_agent.py

I quickly skimmed them all that is what I think has to be change, but take a look as well :)

File: agents/abstraction_agent.py (hunk 1 above)
Comment (Svilen-Stefanov) [reply agents/abstraction_agent.py]: I was wondering if we use the list but this confused me a bit:
```
for insight_type, analysis_insight in self.context.items():
            insight_str += f"## {insight_type.capitalize()} Insight\n"
            if isinstance(analysis_insight, list):
            elif isinstance(analysis_insight, LLMBaseModel):
                insight_str += analysis_insight.llm_str() + "\n\n"
            else:
                raise TypeError(
                    f"Expected analysis_insight to be either list or LLMBaseModel, "
                    f"but got {type(analysis_insight).__name__} for insight_type '{insight_type}'"
                )
```
Are you saying we should overall remove the list there too? From what I saw, we assign different things in the context:
```
self.context["cfg_insight"] = parsed_response
...
self.context["source"] = parsed_response
```
both of which weren't typed, which made me hesitant to change this. If you're sure we can't get a list here, I can remove it.

Comment (ivanmilevtues) [reply agents/abstraction_agent.py]:
```python
    parsed_response = self._parse_invoke(prompt, CFGAnalysisInsights)
    self._parse_invoke(prompt, AnalysisInsights)
```

the _parse_invoke always return the type which is passed as second parameter.

So I am sure that this there is no list.

Svilen-Stefanov: Done

agents/abstraction_agent.py
```diff
@@ -95,11 +96,11 @@ def step_source(self):
...
                 insight_str += "\n".join([f"- {insight.llm_str()}" for insight in analysis_insight]) + "\n\n"
-            elif isinstance(analysis_insight, AnalysisInsights):
+            elif isinstance(analysis_insight, LLMBaseModel):
                 insight_str += analysis_insight.llm_str() + "\n\n"
             else:
                 raise TypeError(
-                    f"Expected analysis_insight to be either list or AnalysisInsights, "
+                    f"Expected analysis_insight to be either list or LLMBaseModel, "
                     f"but got {type(analysis_insight).__name__} for insight_type '{insight_type}'"
                 )
```
ivanmilevtues: You are right there is no need to have this isinstance check now, with the new typing. it can be just .llm_str(). It can also be a list comprehension if you like that better.

agents/abstraction_agent.py
```diff
@@ -40,9 +41,7 @@ def __init__(
         self.project_name = project_name
         self.meta_context = meta_context
 
-        self.context: dict[str, AnalysisInsights | list[AnalysisInsights]] = {
-            "structure_insight": []
-        }  # Store evolving insights here
+        self.context: dict[str, LLMBaseModel] = dict()
```
ivanmilevtues: cool any difference of = {} and = dict().

Don't care which one we use, just wondering :D

agents/abstraction_agent.py
```diff
@@ -93,13 +92,11 @@ def step_source(self):
...
             insight_str += f"## {insight_type.capitalize()} Insight\n"
-            if isinstance(analysis_insight, list):
-                insight_str += "\n".join([f"- {insight.llm_str()}" for insight in analysis_insight]) + "\n\n"
-            elif isinstance(analysis_insight, AnalysisInsights):
+            if isinstance(analysis_insight, LLMBaseModel):
                 insight_str += analysis_insight.llm_str() + "\n\n"
             else:
                 raise TypeError(
-                    f"Expected analysis_insight to be either list or AnalysisInsights, "
+                    f"Expected analysis_insight to be either list or LLMBaseModel, "
```
ivanmilevtues: message is misleading, no list is expected.

agents/abstraction_agent.py
```diff
@@ -93,13 +92,11 @@ def step_source(self):
         insight_str = ""
         for insight_type, analysis_insight in self.context.items():
             insight_str += f"## {insight_type.capitalize()} Insight\n"
-            if isinstance(analysis_insight, list):
-                insight_str += "\n".join([f"- {insight.llm_str()}" for insight in analysis_insight]) + "\n\n"
-            elif isinstance(analysis_insight, AnalysisInsights):
+            if isinstance(analysis_insight, LLMBaseModel):
```
ivanmilevtues: Do we even need this if/else now, as the self.context is typed. Shouldn't MyPy automatically raise an exception if something is wrong?

PR #121, opened 2025-12-11
Unify entry points for remote and local analysis

Svilen-Stefanov:
I think it would be great if we already merge the two existing entry points into a single CLI, which is what I've done in this PR. 
To keep the changes small, I've kept all behavior from both files and didn't modify what we add to the .env file during setup, we can refactor this later if we want.

### DIFF
--- a/main.py
+++ b/main.py
@@ -314,30 +314,10 @@ def validate_arguments(args, parser, is_local: bool):
         parser.error("--partial-component is required when using --partial-analysis")
 
 
-def main():
-    """Main entry point for the unified CodeBoarding CLI."""
-    parser = argparse.ArgumentParser(
-        description="Generate onboarding documentation for Git repositories (local or remote)",
-        formatter_class=argparse.RawDescriptionHelpFormatter,
-        epilog="""
-Examples:
-  # Remote repositories
-  python main.py https://github.com/user/repo1
-  python main.py https://github.com/user/repo1 --output-dir ./docs
-  python main.py https://github.com/user/repo1 https://github.com/user/repo2 --output-dir ./output
-
-  # Local repository
-  python main.py --local /path/to/repo --project-name MyProject --output-dir ./analysis
-
-  # Partial update
-  python main.py --local /path/to/repo --project-name MyProject --output-dir ./analysis \\
-                 --partial-component ComponentName --partial-analysis analysis_name
-
-  # Use custom binary location
-  python main.py --local /path/to/repo --project-name MyProject --binary-location /path/to/binaries
-        """,
-    )
-
+def define_cli_arguments(parser: argparse.ArgumentParser):
+    """
+    Adds all command-line arguments and groups to the ArgumentParser.
+    """
     # Repository specification (mutually exclusive groups)
     repo_group = parser.add_mutually_exclusive_group(required=True)
     repo_group.add_argument(
@@ -373,6 +353,32 @@ Examples:
         "--no-cache-check", action="store_true", help="Skip checking if materials already exist (remote repos only)"
     )
     parser.add_argument("--project-root", type=Path, help="Project root directory (default: current directory)")
+
+
+def main():
+    """Main entry point for the unified CodeBoarding CLI."""
+    parser = argparse.ArgumentParser(
+        description="Generate onboarding documentation for Git repositories (local or remote)",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  # Remote repositories
+  python main.py https://github.com/user/repo1
+  python main.py https://github.com/user/repo1 --output-dir ./docs
+  python main.py https://github.com/user/repo1 https://github.com/user/repo2 --output-dir ./output
+
+  # Local repository
+  python main.py --local /path/to/repo --project-name MyProject --output-dir ./analysis
+
+  # Partial update
+  python main.py --local /path/to/repo --project-name MyProject --output-dir ./analysis \\
+                 --partial-component ComponentName --partial-analysis analysis_name
+
+  # Use custom binary location
+  python main.py --local /path/to/repo --project-name MyProject --binary-location /path/to/binaries
+        """,
+    )
+    define_cli_arguments(parser)
 
     args = parser.parse_args()
 


main.py
```diff
@@ -0,0 +1,520 @@
+"""
+Unified entry point for CodeBoarding documentation generation.
+Supports both local and remote repository analysis.
+"""
```
ivanmilevtues: I don't like these kind of comments ngl :D.

It is so AI generated :D.

main.py
```diff
@@ -0,0 +1,520 @@
...
+        "CEREBRAS_API_KEY",
+    ]
+    api_env_keys = [(key, os.getenv(key)) for key in api_provider_keys if os.getenv(key) is not None]
+
+    if len(api_env_keys) == 0:
+        logger.error(f"API key not set, set one of the following: {api_provider_keys}")
+        exit(1)
+    elif len(api_env_keys) > 1:
+        logger.error(f"Detected multiple API keys set ({api_env_keys}), set ONE of the following: {api_provider_keys}")
+        exit(2)
```
ivanmilevtues, line 40: No need to change it just a quick rant.
 
This validation I know is old, but honestly it sucks :D.

It works only if you have .env file, otherwise it fails all the time esp if you are someone like me who has multple keys, I think a flag is much better here.

main.py
```diff
@@ -0,0 +1,520 @@
...
+    Generate analysis files for a repository.
+
+    Args:
+        repo_name: Name of the repository
+        repo_path: Path to the repository
+        output_dir: Directory for output files
+        depth_level: Depth level for diagram generation
+
+    Returns:
+        List of paths to generated analysis files
```
ivanmilevtues: This comment doesn't give me much, if you want you can rename the method.

Otherwise I think you should delete it and safe a little bit of space and tokens :D

ivanmilevtues: This same comment applies for all methods, Claude loves those comments doesn't it.

ivanmilevtues (approved):
minor comments

README.md
```diff
@@ -116,8 +116,58 @@ LANGCHAIN_API_KEY=                # Optional: LangChain API key
...
+- `--project-name <name>` - Name of the project (required when using `--local`)
+
+**Partial Update Options (local repos only):**
+- `--partial-component <name>` - Specific component to update in existing analysis
+- `--partial-analysis <name>` - Analysis file name to update (both component and analysis must be specified together)
+
+**Advanced Options:**
+- `--depth-level <int>` - Depth level for diagram generation (default: 1)
+- `--prompt-type <type>` - Prompt type: `bidirectional` or `unidirectional` (default: bidirectional for remote, unidirectional for local)
+- `--binary-location <path>` - Custom path to language server binaries (for VSCode integration)
```
ivanmilevtues: remove (for VSCode integration), I think it is not relevant for the open-source repo.

main.py
```diff
@@ -0,0 +1,520 @@
...
+            depth_level=args.depth_level,
+            component_name=args.partial_component,
+            analysis_name=args.partial_analysis,
+        )
+        logger.info(f"Documentation generated successfully in {args.output_dir or './analysis'}")
+    else:
+        # Remote repositories processing
+        if args.repositories:
+            # Store GitHub token if needed
+            if not args.no_upload:
```
ivanmilevtues: I would rather have the arg as upload, and change the default value.

main.py
```diff
@@ -0,0 +1,520 @@
...
+
+    if args.partial_analysis and not args.partial_component:
+        parser = argparse.ArgumentParser()
+        parser.error("--partial-component is required when using --partial-analysis")
+
+    # Determine prompt type: bidirectional for remote (github repo link), unidirectional for local
+    if args.prompt_type:
+        prompt_type = PromptType.BIDIRECTIONAL if args.prompt_type == "bidirectional" else PromptType.UNIDIRECTIONAL
+    else:
+        prompt_type = PromptType.UNIDIRECTIONAL if is_local else PromptType.BIDIRECTIONAL
```
ivanmilevtues, line 402: This has to be unified, we can make a comment, the reasoning is so baaad :D.

We couldn't visualize two-way graphs for a bit and it was easier to force the diagrams to be unidrectional than to fix it 🤭

main.py
ivanmilevtues: On a second look it seems that not all referenced arguments are actually defined.

I.e. no_upload is not registered with argparse?

ivanmilevtues (changes requested):
I haven't finished the review.
Would have to finish it tmrw with more energy in me.

Awesome step, towards easier to maintain repo. The arguments seemed to be hallucinated :D

main.py
Svilen-Stefanov: I agree. The reason why I didn't want to do it here is because there's still some more work that would need to be done - we need to get rid of the .env file overall, aka remove its generation, which I didn't want to do here already to reduce the scope of the PR.

main.py
Svilen-Stefanov: It is defined here:
```
 "--no-upload", action="store_true", help="Skip uploading onboarding materials (remote repos only)"
```
I agree with the upload suggestion

main.py
```diff
@@ -0,0 +1,444 @@
+import argparse
```
brovatten, line 1: For next review, it would be easiest if you change demo.py->main.py as a last step to see the changes more easily

main.py
```diff
@@ -0,0 +1,444 @@
...
+    repo_group.add_argument("--local", type=Path, help="Path to a local repository")
+
+    # Output configuration
+    parser.add_argument("--output-dir", type=Path, help="Directory to output generated files to")
+
+    # Local repository specific options
+    parser.add_argument("--project-name", type=str, help="Name of the project (required for local repositories)")
+
+    # Partial update options
+    parser.add_argument("--partial-component", type=str, help="Component to update (for partial updates only)")
```
brovatten, line 335: Lots of arguments here... As in #115 , should `parse_args()` maybe be a function here?

README.md
```diff
@@ -116,8 +116,58 @@ LANGCHAIN_API_KEY=                # Optional: LangChain API key
...
+- `--project-root <path>` - Project root directory (default: current directory)
+- `--upload` - Upload onboarding materials to GeneratedOnBoardings repo (remote repos only)
+- `--no-cache-check` - Skip checking if documentation already exists (remote repos only)
+
+#### Examples
+
 ```bash
-python demo.py <github_repo_url> --output-dir <output_path>
+# Analyze a single remote repository
+python main.py https://github.com/pytorch/pytorch --output-dir ./pytorch-docs
```
brovatten, line 157: Why give a specific one here but do `repo1` and `repo2` below? Looks strange

README.md
```diff
@@ -116,8 +116,58 @@ LANGCHAIN_API_KEY=                # Optional: LangChain API key
...
+
+# Analyze a local repository
+python main.py --local ./my-project --project-name MyProject --output-dir ./analysis
+
+# Partial update of a specific component
+python main.py --local ./my-project --project-name MyProject --output-dir ./analysis \
+               --partial-component "API Service" --partial-analysis on_boarding
+
+# Use custom depth level and disable caching
+python main.py https://github.com/user/repo --depth-level 2 --no-cache-check --output-dir ./deep-analysis
```
brovatten, line 170: Maybe show depth-level as a set [1,2,3] or something, setting this to a specific value is not easy to gauge

main.py
brovatten: Let's keep it for now, when generating I sometimes have use for this. But ur right, it's crap!

main.py
```diff
@@ -0,0 +1,444 @@
...
+    process_remote_repository(
+        repo_url=repo_url,
+        output_dir=temp_repo_folder,
+        depth_level=int(os.getenv("DIAGRAM_DEPTH_LEVEL", "1")),
+        upload=not local_dev,  # Only upload if not in local dev mode
+        cache_check=True,
+    )
+
+
+def process_remote_repository(
```
brovatten, line 190: Great naming here and overall!

main.py
```diff
@@ -0,0 +1,444 @@
...
+def main():
+    """Main entry point for the unified CodeBoarding CLI."""
+    parser = argparse.ArgumentParser(
+        description="Generate onboarding documentation for Git repositories (local or remote)",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  # Remote repositories
+  python main.py https://github.com/user/repo1
+  python main.py https://github.com/user/repo1 --output-dir ./docs
```
brovatten: I would keep this for the readme, and keep lean on the docs

brovatten: AKA trim the examples here

Svilen-Stefanov: The reason why this is added here is because people can just do `python main.py -h` and see how they can use it, which is super useful when using CLIs overall.

README.md
Svilen-Stefanov: I think it's great to give a realistic example - e.g ppl might be confused if they should pass the clone link, the http one or something else, hence the example.

main.py
Svilen-Stefanov: Ah, good point. I guess I can still rename the `main.py` to `demo.py` to see the changes but they are quite significant so I'm not sure how much easier it's actually going to be.

main.py
Svilen-Stefanov: Sure, we can do that.

ivanmilevtues (approved):
Well done, nothing to add :)

PR #123, opened 2025-12-12
Remove default values for env variables and do not load them for local repo analysis

Svilen-Stefanov:


main.py
```diff
@@ -41,35 +41,6 @@ def validate_env_vars():
         exit(2)
 
 
-def setup_environment(project_root: Path | None = None):
-    if project_root is None:
-        project_root = Path.cwd()
-
-    # Set defaults for environment variables that can be deduced
-    if not os.getenv("REPO_ROOT"):
```
Svilen-Stefanov, line 49: This will probably fail down the line if these are not set in the .evn file. We could eventually remove them from the file and set them as constants instead.

ivanmilevtues: setup.py should cover this.

main.py
```diff
@@ -401,9 +367,10 @@ def main():
     logger.info("Starting CodeBoarding documentation generation...")
 
     # Load environment from .env file if it exists
-    load_dotenv()
-
-    setup_environment(args.project_root)
+    # Validate environment variables (only for remote repos due to .env file - should be modified soon)
+    if not is_local:
```
ivanmilevtues: We can go with this, but I still don't like it. because with this you cannot use the CLI main.py for local project.

Because the difference is if one is using it via VSCode, we have to pickup the vscode setup (vscode_constants and if not we have to pick-up the dotenv).

But this probably has to be resolved by removing all the .env vars or something a bit more major.

Svilen-Stefanov: I've added a new flag for this so now these should be independent.

static_analyzer/lsp_client/client.py
```diff
@@ -201,7 +201,7 @@ def _get_document_symbols(self, file_uri: str) -> list:
         """Fetches all document symbols (functions, classes, etc.) for a file."""
         params = {"textDocument": {"uri": file_uri}}
         req_id = self._send_request("textDocument/documentSymbol", params)
-        logger.info(f"Requesting document symbols for {file_uri} with ID {req_id}")
+        logger.debug(f"Requesting document symbols for {file_uri} with ID {req_id}")
```
Svilen-Stefanov, line 204: While I was making the change, I thought I'll change the log level here because there are a lot of messages of this type, which I'm not sure we actually want to log as info.

brovatten: Super, nice!

main.py
```diff
@@ -401,9 +374,10 @@ def main():
     logger.info("Starting CodeBoarding documentation generation...")
 
     # Load environment from .env file if it exists
-    load_dotenv()
-
-    setup_environment(args.project_root)
+    # Validate environment variables (only for remote repos due to .env file - should be modified soon)
+    if args.load_env_variables:
```
brovatten, line 378: Having this args variable seems a bit chaotic without further explanation

main.py
```diff
@@ -345,6 +311,13 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
     parser.add_argument("--partial-analysis", type=str, help="Analysis file to update (for partial updates only)")
 
     # Advanced options
+    parser.add_argument(
```
brovatten, line 314: didnt we make an parse_args() for all of these?

Svilen-Stefanov: Yeah, this is the method, it's just called `define_cli_arguments` :D

main.py
Svilen-Stefanov: What do you mean, it's used as all other env variables (e.g local / binary_location are used the same way a couple lines further down)?

PR #124, opened 2025-12-13
Save monitoring into "runs" and stream the token usage

brovatten:
Since not only evals, but also ordinary `main.py`, is to be saved we should do so in a new folder: `runs`.

Also, to stream the outputs of the token usage to be captured by the extension.

Edit: Now we log the model choice in the TokenUsage (to stream to the extension).
image

Here's an example of how the runs are structured:
image

### DIFF
--- a/monitoring/context.py
+++ b/monitoring/context.py
@@ -109,10 +109,6 @@ def monitor_execution(
 
         logger.info(f"✨ Run results saved to {out_path}")
 
-        # Cleanup app.log handler
-        root_logger.removeHandler(app_log_handler)
-        app_log_handler.close()
-
         # Reset context var
         current_stats.reset(stats_token)
 
--- a/utils.py
+++ b/utils.py
@@ -43,6 +43,7 @@ def get_project_root() -> Path:
 
 
 def monitoring_enabled():
+    print("Monitoring enabled:", os.getenv("ENABLE_MONITORING", "false"))
     return os.getenv("ENABLE_MONITORING", "false").lower() in ("1", "true", "yes")
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,3 +100,9 @@ dev-dependencies = [
     "mypy==1.19.0",
     "pytest-cov==7.0.0",
 ]
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+norecursedirs = ["repos", "build", "dist", "temp", "node_modules"]
+addopts = "--ignore=repos"
+pythonpath = "."


main.py
```diff
@@ -16,6 +16,7 @@
 from repo_utils import clone_repository, get_branch, get_repo_name, store_token, upload_onboarding_materials
```
ivanmilevtues, line 16: Please update to latest main (be careful :d), says that there are conflicts.

monitoring/paths.py
```diff
@@ -0,0 +1,34 @@
+import os
+from datetime import datetime
+from pathlib import Path
+
+
+def get_project_root() -> Path:
```
Svilen-Stefanov: Don't we have such helper elsewhere - we probably should and it shouldn't be under `monitoring`.

monitoring/paths.py
```diff
@@ -0,0 +1,34 @@
...
+    # Fallback to current working directory
+    return Path.cwd().resolve()
+
+
+def get_monitoring_base_dir() -> Path:
+    return get_project_root() / "runs"
+
+
+def get_monitoring_runs_dir() -> Path:
+    return get_monitoring_base_dir()
```
Svilen-Stefanov: We can just call `get_monitoring_base_dir()` :D

monitoring/mixin.py
```diff
@@ -13,4 +13,6 @@ def __init__(self):
 
     def get_monitoring_results(self) -> dict:
         """Return monitoring statistics."""
-        return self.agent_stats.to_dict()
+        results = self.agent_stats.to_dict()
+        results["model_name"] = self.agent_monitoring_callback.model_name
```
Svilen-Stefanov: This looks suspicious to me. It seems like this should be part of `agent_stats`. We use ` self.agent_stats.to_dict()` to avoid such manual dict manipulations and mostly work with typed objects. 
So this code smells to me.

main.py
```diff
@@ -50,20 +51,28 @@ def setup_environment(project_root: Path | None = None):
         repo_root = project_root / "repos"
         os.environ["REPO_ROOT"] = str(repo_root)
         logger.info(f"REPO_ROOT not set, using default: {repo_root}")
+    else:
+        logger.info(f"Using REPO_ROOT from environment: {os.getenv('REPO_ROOT')}")
```
Svilen-Stefanov: Do we need these? What's the value of having that info? I think we overall want to get rid of these env vars so logging them here seems suspicious.

diagram_analysis/diagram_generator.py
```diff
@@ -41,13 +42,16 @@ def __init__(
         depth_level: int,
         static_only: bool = False,
         project_name: str | None = None,
+        run_id: str | None = None,
     ):
         self.repo_location = repo_location
         self.temp_folder = temp_folder
         self.repo_name = repo_name
         self.output_dir = output_dir
+        self.depth_level = depth_level
```
Svilen-Stefanov, line 52: This variable is defined twice now - look at line 62. 
Also based on this init method, this object screams to me `BaseModel` to avoid all of that manual patching of init attributes. Can you convert it to a BaseModel?

monitoring/paths.py
```diff
@@ -0,0 +1,34 @@
...
+def get_project_root() -> Path:
+    project_root_env = os.getenv("PROJECT_ROOT")
+    if project_root_env:
+        return Path(project_root_env).resolve()
+    # Fallback to current working directory
+    return Path.cwd().resolve()
+
+
+def get_monitoring_base_dir() -> Path:
+    return get_project_root() / "runs"
```
Svilen-Stefanov, line 9: I would definitely avoid creating more folders in their repo. I think all things we create should be under the .codeboarding dir. So I suggest something like:
.codeboarding/logs:
1. run_id:
    - app.log
    - monitoring.log

If we do that and the run_id is effectively a timestamp, that would simplify things for the extension as well.

evals/base.py
```diff
@@ -42,18 +42,26 @@ def _get_project_root(self) -> Path:
 
     def get_latest_run_dir(self, project_name: str) -> Path | None:
         """Find the most recent monitoring run directory for a project."""
-        runs_dir = self.project_root / "evals/artifacts/monitoring_results/runs"
+        # Use the standard runs directory defined in monitoring/paths.py
+        runs_dir = self.project_root / "runs"
```
Svilen-Stefanov: Here and in other places we should try to use the helpers that were created.

evals/base.py
```diff
@@ -122,7 +130,7 @@ def run_pipeline(
         try:
             result = subprocess.run(
                 cmd,
-                capture_output=True,
+                capture_output=False,
```
Svilen-Stefanov: Why is that?

monitoring/writers.py
```diff
@@ -29,13 +29,15 @@ def __init__(
...
         start_time: float | None = None,
+        static_stats: dict | None = None,
     ):
         self.monitoring_dir = Path(monitoring_dir)
         self.llm_usage_file = self.monitoring_dir / "llm_usage.json"
         self.agents_dict = agents_dict
         self.repo_name = repo_name
         self.output_dir = output_dir
         self.interval = interval
+        self.static_stats = static_stats
```
Svilen-Stefanov: Same here, this class also screams `BaseModel`.

evals/base.py
brovatten: Sorry, I was watching the terminal logs. Will fix this

evals/base.py
brovatten: Yup, this function should also be moved to the paths.py in the monitoring, it is not specific to the evals

monitoring/paths.py
brovatten: I would generally agree, but nesting folders for the sake of root folders will create another mess. Now the projectname of the runs makes it super easy to identify which run is which. e.g. experiments can be named "markitdown-try-fewer-workers-experiments"

I thought a bit about how to structure them, and having a `/runs` folders is very standard. I would not deviate from this. Including the app.log for each run is a great idea, this would atleast remove the `logs` folder

monitoring/mixin.py
brovatten: I purposefully choose not to include it there as that is runtime statistics, not static information

brovatten: but gemini is not on my side 😂 i will implement it like you said!

diagram_analysis/diagram_generator.py
brovatten: Great thank you!! I like this sugggestion

brovatten:
Now uses basemodels and saves the logs into separate runs. 

For the PROJECT_ROOT and handling of env variables better, let's do that in a separate PR and assume it's set for now.

diagram_analysis/diagram_generator.py
```diff
@@ -164,23 +161,21 @@ def pre_analysis(self):
...
+                run_id = generate_run_id(name=run_name)
+
+            monitoring_dir = get_monitoring_run_dir(run_id, create=True)
+            logger.info(f"Monitoring enabled. Writing stats to {monitoring_dir}")
 
             # Save code_stats.json
             code_stats_file = monitoring_dir / "code_stats.json"
             with open(code_stats_file, "w") as f:
                 json.dump(static_stats, f, indent=2)
+            logger.info(f"Written code_stats.json to {code_stats_file}")
```
ivanmilevtues: logger.debug I would say for this.

diagram_analysis/diagram_generator.py
```diff
@@ -31,33 +33,28 @@
...
-        self.abstraction_agent: AbstractionAgent | None = None
-        self.planner_agent: PlannerAgent | None = None
-        self.validator_agent: ValidatorAgent | None = None
-        self.diff_analyzer_agent: DiffAnalyzingAgent | None = None
-        self.meta_agent: MetaAgent | None = None
-        self.meta_context = None
-        self.depth_level = depth_level
-        self._monitoring_agents: dict[str, MonitoringMixin] = {}
-        self.stats_writer: StreamingStatsWriter | None = None
+class DiagramGenerator(BaseModel):
```
ivanmilevtues: I don't like this being BaseModel as it implys it is a Data Object, which it is not.

And I don't see an advantage for it being a BaseModel

monitoring/writers.py
```diff
@@ -10,38 +10,37 @@
...
 
+from pydantic import BaseModel, ConfigDict, PrivateAttr, Field
+
 from monitoring.mixin import MonitoringMixin
 
 logger = logging.getLogger("monitoring")
 
 
-class StreamingStatsWriter:
+class StreamingStatsWriter(BaseModel):
```
ivanmilevtues: Same comment as the DiagramGenerator.

ivanmilevtues (changes requested):
Nice overall, I just don't like the BaseModel for a non-dataclasses.

main.py
```diff
@@ -389,6 +409,8 @@ def main():
     if args.binary_location:
         update_config(args.binary_location)
 
+    should_monitor = args.enable_monitoring or monitoring_enabled()
```
brovatten, line 412: accept monitoring as args now instead. before it was `load_dotenv` in the utils, but this does not make sense with our new `load_envvars` arg

monitoring/callbacks.py
```diff
@@ -49,13 +52,13 @@ def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
...
-            self.stats.input_tokens += usage.get("prompt_tokens", 0)
-            self.stats.output_tokens += usage.get("completion_tokens", 0)
+            self.stats.input_tokens += usage.get("input_tokens", 0)
+            self.stats.output_tokens += usage.get("output_tokens", 0)
 
         # Log Event
         if self.log_results:
             model = self.model_name or "unknown"
-            logger.info(f"LLM Usage: step={step_name} model={model} usage={usage}")
+            logger.info(f"Token Usage: step={step_name} model={model} usage={json.dumps(usage)}")
```
brovatten, line 61: Makes more sense this way, to use this as the logger for each token usage, and use the context.py as a central monitor, to aggregate results and then write it. It does this every X second.

monitoring/context.py
```diff
@@ -87,18 +107,24 @@ def end_step(self):
...
-            logger.info(f"✨ Monitoring summary saved to {summary_file}")
+            logger.debug(f"✨ Monitoring summary saved to {summary_file}")
         except Exception as e:
             logger.error(f"Failed to save monitoring summary: {e}")
 
         # Cleanup handler
         trace_logger.removeHandler(trace_handler)
         trace_handler.close()
-        logger.info(f"✨ Execution traces saved to {trace_file}")
+        logger.debug(f"✨ Execution traces saved to {trace_file}")
```
brovatten, line 108: Here u go Ivan :) just below this we clarify where the run is saved

monitoring/paths.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return run_dir
+
+
+def generate_run_id(name: str) -> str:
+    timestamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    return f"{name}/{timestamp}"
+
+
+def get_latest_run_dir(project_name: str) -> Path | None:
```
brovatten, line 27: This made more sense here than in the base eval

utils.py
```diff
@@ -34,8 +34,15 @@ def caching_enabled():
     return os.getenv("CACHING_DOCUMENTATION", "false").lower() in ("1", "true", "yes")
 
 
+def get_project_root() -> Path:
```
brovatten, line 37: this should be here and not in the base eval

monitoring/context.py
ivanmilevtues: thanks

PR #125, opened 2025-12-14
Logging fix

brovatten:
We now log the model in the token monitoring, and we don't duplicate the log. This was caused by both the callbacks logging it, and not only the global callback one.

See the outputs here:
image


ivanmilevtues (approved):
looks good.

PR #126, opened 2025-12-15
Use smaller model for parsing

brovatten:
When we format the output, we are using the same model as the analysis. Instead we could set a smaller (and cheaper) default model to parse this.

On a small scale I tested this : `markitdown` with `depth=1`  and the parsing went from 59.3s to 24.5s on a run of ~360s. This would scale nicely for larger prompts.


agents/agent.py
```diff
@@ -87,6 +87,7 @@ def _setup_env_vars(self):
 
         # Model selection via environment variable
         self.codeboarding_model = os.getenv("CODEBOARDING_MODEL")
+        self.parsing_model = os.getenv("PARSING_MODEL")
```
ivanmilevtues: My only comment here is to handle this somehow with default value. I.e. if none (as the vscode extension will keep it at None) we should do something idk what :D.

I would like to know if we are using our API key or the vscode extension user uses a custom one and if a custom one is used we just use whichever model they've used.

We can add later down the line an advance settings/config where people can set such things. (we can also add it now).

agents/agent.py
```diff
@@ -163,6 +164,76 @@ def _initialize_llm(self):
         MONITORING_CALLBACK.model_name = model_name
         return model
 
+    def _initialize_parsing_llm(self):
```
ivanmilevtues: I suppose what svilen said can be done, have a single method which has a mapping for model families:
Smart LLM, small LLM and the method get a flag or something like that (maybe not as with my other comment it becomes more complicated.)

agents/agent.py
brovatten: 1. Totally agree, I will default it to None.
2. For the second, I think the extension should handle this with it's settings like we discussed? Since that's where the input from the user (the key) is registered. Only the extension needs to be aware of this.
3. If I understood, this is what I added

agents/agent.py
```diff
@@ -47,8 +42,8 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, syste
         ReferenceResolverMixin.__init__(self, repo_dir, static_analysis)
         MonitoringMixin.__init__(self)
         self._setup_env_vars()
-        self.llm = self._initialize_llm()
-        self.extractor_llm = self._initialize_llm()
+        self.llm = self._initialize_llms(is_small_model=False)
```
Svilen-Stefanov: The refactors look better.

However, this still looks very strange - for a single agent we initialize two different LLMs. 
My thinking here is to create another level of the class hierarchy here - one for a small agent and one for a large one. Then each of the specific agents will inherit either from the small or from the large LLM class but not for `CodeBoardingAgent` directly. 

Currently, this class breaks the single responsibility principle, which is a code smell.

brovatten:
I agree. To add on what you mentioned, I let two of our agents inherit from the "small model", although I refrained from this variable naming because I think this would lead to more confusion. Instead welcome `parsing_model` and `agent_model`.

Great idea to split them into two different classes. And the config now looks much cleaner!

I tried to run it with AWS and Google and checks out!

agents/agent.py
```diff
@@ -73,95 +69,50 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, syste
...
-        # Model selection via environment variable
-        self.codeboarding_model = os.getenv("CODEBOARDING_MODEL")
-
-    def _initialize_llm(self):
+    @classmethod
+    def get_parsing_llm(cls) -> BaseChatModel:
+        """Shared access to the small model for parsing tasks."""
+        if cls._parsing_llm is None:
+
+            parsing_model = os.getenv("PARSING_MODEL", None) or os.getenv("CODEBOARDING_MODEL")
```
ivanmilevtues, line 77: smart I love these almost ternary operators, they give me a sense of nostalgia for C/C++/Kotlin <3

agents/agent.py
```diff
@@ -73,95 +69,50 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, syste
...
+                kwargs["api_key"] = api_key
+
+            model = config.chat_class(**kwargs)  # type: ignore[call-arg, arg-type]
+
+            # Update global monitoring callback
+            MONITORING_CALLBACK.model_name = model_name
+            return model, model_name
+
+        raise ValueError(
+            "No valid API key found. Please set one of: OPENAI_API_KEY, ANTHROPIC_API_KEY, "
```
ivanmilevtues: This doesn't support open ai compatible self-hosted/azure hosted solutions as they also require the openai url + ollama/lm studio

agents/llm_config.py
```diff
@@ -0,0 +1,119 @@
...
+
+# Define supported providers in priority order
+LLM_PROVIDERS = {
+    "openai": LLMConfig(
+        chat_class=ChatOpenAI,
+        api_key_env="OPENAI_API_KEY",
+        agent_model="gpt-4o",
+        parsing_model="gpt-4o-mini",
+        extra_args={
+            "base_url": lambda: os.getenv("OPENAI_BASE_URL"),
```
ivanmilevtues, line 64: This is what I meant in the previous comment i.e. with LM studio you provide OPENAI base url without the openai api key, it's not a big deal tbh :D

agents/llm_config.py
```diff
@@ -0,0 +1,119 @@
...
+        extra_args={
+            "max_tokens": 8192,
+            "timeout": None,
+            "max_retries": 0,
+        },
+    ),
+    "google": LLMConfig(
+        chat_class=ChatGoogleGenerativeAI,
+        api_key_env="GOOGLE_API_KEY",
+        agent_model="gemini-2.5-flash",
```
ivanmilevtues, line 84: @Svilen-Stefanov can you confirm, I think we are passing the model name now from the extension and then using it here, not as hardcoded string. Can you talk with alex to see how to resolve this as it was partly hardcoded/partly user settable.

tests/test_read_docs.py
```diff
@@ -10,33 +11,39 @@ def setUp(self):
...
     def test_readme_not_found(self):
         # Test when README.md doesn't exist (using a non-existent repo path)
         tool_no_readme = ReadDocsTool(repo_dir=Path("/tmp/nonexistent"))
         content = tool_no_readme._run()
         self.assertIsInstance(content, str)
-        self.assertIn("No markdown documentation files found in this repository.", content)
+        # Fixed: Updated expectation to match actual output
+        self.assertIn("No documentation files found in this repository.", content)
 
+    @unittest.skipIf(not os.path.exists("./repos/yfinance"), "yfinance repo not found")
```
ivanmilevtues: This is sus, @Svilen-Stefanov we should've added those repos to our repo :D, otherwise the rests are just always skipped from what I see here.

agents/agent.py
```diff
@@ -252,3 +205,19 @@ def _try_parse(self, message_content, parser):
                 except:
                     pass
         raise ValueError(f"Couldn't parse {message_content}")
+
+
+class LargeModelAgent(CodeBoardingAgent):
```
ivanmilevtues, line 215: This is very nice, I think it makes the usage super clear. Nice.

agents/meta_agent.py
```diff
@@ -13,7 +13,7 @@
 logger = logging.getLogger(__name__)
 
 
-class MetaAgent(CodeBoardingAgent):
+class MetaAgent(SmallModelAgent):
```
ivanmilevtues: Why do you think the MetaAgent needs less power than the others?

agents/planner_agent.py
```diff
@@ -3,14 +3,14 @@
...
-from agents.agent import CodeBoardingAgent
+from agents.agent import SmallModelAgent
 from agents.agent_responses import AnalysisInsights, ExpandComponent, Component
 from agents.prompts import get_expansion_prompt, get_planner_system_message
 from monitoring import trace
 from static_analyzer.analysis_result import StaticAnalysisResults
 
 
-class PlannerAgent(CodeBoardingAgent):
+class PlannerAgent(SmallModelAgent):
```
ivanmilevtues: Same question as for the MetaAgent.
I think for this change now I'd like to see the markitdown/codeboarding e2e results to see if quality isn't degraded.

ivanmilevtues (changes requested):
Great job, my comments are babies, not super code related.

My main concern is the quality, can we get some e2e benchmark action here. This touches agents so that is a perfect use case to harvest the fruits of your labour :D

agents/llm_config.py
brovatten: Good point, I added it now!

agents/llm_config.py
brovatten: We do still handle this, check `model_override` in the `_static_initialize_llm`

agents/planner_agent.py
brovatten: Let's do this for a separate PR, where we decide which to take where. I started running it, but it will take some time to analyze for AWS and Gemini. The planner agents also depends on depth level (1 and 2)

brovatten: Hence I will set them to LargeModelAgent for now

PR #127, opened 2025-12-19
Chunk logs

brovatten:
This separates the logs into separate runs, titled by the timestamp, and using a symlink to `_latest.log` to easily see you ongoing/latest run.

Here's a screenshot where I put `--output-dir `

image

I tested by running it with and without `output_dir` specified.

### DIFF
--- a/logging_config.py
+++ b/logging_config.py
@@ -8,10 +8,10 @@ from pathlib import Path
 
 def setup_logging(
     default_level: str = "INFO",
-    log_filename: str = "codeboarding.log",
     max_bytes: int = 10 * 1024 * 1024,
     backup_count: int = 5,
     log_dir: Path | None = None,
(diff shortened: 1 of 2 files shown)


logging_config.py
```diff
@@ -1,36 +1,47 @@
...
+import os
+import shutil
+from datetime import datetime
 from pathlib import Path
 
 
 def setup_logging(
     default_level: str = "INFO",
-    log_filename: str = "app.log",
+    log_filename: str = "codeboarding.log",
```
ivanmilevtues: That seems better, you should reflect this change in the extension as well

ivanmilevtues: But this isn't used as it is just the timestamped, maybe remove the default value whatsoever.

brovatten: Good catch!

PR #128, opened 2025-12-19
Unify exclusions and simplify agent constructors

brovatten:
This resolves the exclusion issues we had with node modules, by creating a class `BaseRepoTool` that the tools inherit.

It also puts the tools into a tool kit that says simplifies the constructors for the codeboarding agent and the others.

I tested it by running the pipeline.

### DIFF
--- a/static_analyzer/lsp_client/client.py
+++ b/static_analyzer/lsp_client/client.py
@@ -1098,10 +1098,6 @@ class LSPClient:
 
         return filtered
 
-    def reload_ignore_manager(self):
-        """Reload the ignore manager to pick up changes to .gitignore."""
-        self.ignore_manager.reload()
-
     def get_exclude_dirs(self) -> pathspec.PathSpec:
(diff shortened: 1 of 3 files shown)


static_analyzer/lsp_client/client.py
```diff
@@ -1084,33 +1087,21 @@ def _get_all_symbols_recursive(self, symbols: list) -> list:
...
+        # First use the centralized manager
+        filtered = [f for f in src_files if not self.ignore_manager.should_ignore(f)]
+
+        # Then apply the additional spec if provided (backward compatibility for tests)
+        if spec is not None:
+            filtered = [f for f in filtered if not spec.match_file(str(f))]
+
+        return filtered
+
+    def reload_ignore_manager(self):
```
ivanmilevtues: This method seems useless as it is used only in tests, you can directly do lspclient.ignore_manager.reload() in the test if you need it

PR #130, opened 2025-12-25
Adding a more sophisticated clustering.

ivanmilevtues:
In this change I added a more sophisticated way of doing the clustering.

The issue before was that we were using just greedy_modularity. Which for small codebases and not only would result in a single huge cluster and that is it, which is very uninformative for the analysis.

Now we have a multi-step approach:
1. We try a vanilla algorithm: for algorithm in ['louvain', 'leiden', 'greedy_modularity']:
2. We try to group things togheter in classes as it might be the case that the codebase is way too sparse:         
        # Strategy 2: Try structural clustering only if connectivity fails
        for level in ['method', 'class']:  # Skip file/package levels as they create too large clusters
3. We try to combine both approaches, to do a balanced greed modularity, so that we group some of the non-grouped nodes toghether by class/package etc. on top of the already ran algorithm (greedy_modularity_communities) + we try to break a "monolith" cluster

Apart from that I bumped model usage to be 3.0 flash with 2.5-flash from 2.5/2.0

Fixed a binary arg to be a flag rather than a value argument.

### DIFF
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -55,7 +55,7 @@ class Node:
     METHOD_TYPE = 6
 
     def __init__(
-        self, fully_qualified_name: str, node_type: str, file_path: str, line_start: int, line_end: int
+        self, fully_qualified_name: str, node_type: int, file_path: str, line_start: int, line_end: int
     ) -> None:
         self.fully_qualified_name = fully_qualified_name
         self.file_path = file_path
--- a/tests/static_analyzer/test_analysis_result.py
+++ b/tests/static_analyzer/test_analysis_result.py
@@ -45,8 +45,8 @@ class TestStaticAnalysisResults(unittest.TestCase):
     def test_add_and_get_cfg(self):
         # Test CFG storage and retrieval
         cfg = CallGraph()
-        node1 = Node("test.func1", "function", "test.py", 1, 5)
-        node2 = Node("test.func2", "function", "test.py", 6, 10)
+        node1 = Node("test.func1", 12, "test.py", 1, 5)
+        node2 = Node("test.func2", 12, "test.py", 6, 10)
         cfg.add_node(node1)
         cfg.add_node(node2)
         cfg.add_edge("test.func1", "test.func2")
@@ -79,8 +79,8 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_add_and_get_references(self):
         # Test source code references with case-insensitive lookup
-        node1 = Node("MyClass.method", "method", "test.py", 1, 5)
-        node2 = Node("utils.helper", "function", "utils.py", 10, 15)
+        node1 = Node("MyClass.method", 6, "test.py", 1, 5)
+        node2 = Node("utils.helper", 12, "utils.py", 10, 15)
         self.results.add_references("python", [node1, node2])
 
         # Test case-insensitive retrieval
@@ -92,7 +92,7 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_get_reference_not_found(self):
         # Test error when reference not found
-        node = Node("MyClass.method", "method", "test.py", 1, 5)
+        node = Node("MyClass.method", 6, "test.py", 1, 5)
         self.results.add_references("python", [node])
 
         with self.assertRaises(ValueError) as context:
@@ -101,7 +101,7 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_get_reference_file_path_error(self):
         # Test file path detection
-        node = Node("mymodule.file.Class", "class", "mymodule/file.py", 1, 5)
+        node = Node("mymodule.file.Class", 5, "mymodule/file.py", 1, 5)
         self.results.add_references("python", [node])
 
         with self.assertRaises(FileExistsError) as context:
@@ -110,7 +110,7 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_get_loose_reference(self):
         # Test loose reference matching
-        node = Node("mypackage.module.MyClass.method", "method", "test.py", 1, 5)
+        node = Node("mypackage.module.MyClass.method", 6, "test.py", 1, 5)
         self.results.add_references("python", [node])
 
         # Should match by suffix
@@ -121,7 +121,7 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_get_loose_reference_unique_substring(self):
         # Test loose reference with unique substring
-        node = Node("mypackage.unique_function", "function", "test.py", 1, 5)
+        node = Node("mypackage.unique_function", 12, "test.py", 1, 5)
         self.results.add_references("python", [node])
 
         message, retrieved = self.results.get_loose_reference("python", "unique")
@@ -131,7 +131,7 @@ class TestStaticAnalysisResults(unittest.TestCase):
 
     def test_get_loose_reference_not_found(self):
         # Test when no loose match found
-        node = Node("mypackage.module.Class", "class", "test.py", 1, 5)
+        node = Node("mypackage.module.Class", 5, "test.py", 1, 5)
         self.results.add_references("python", [node])
 
         message, retrieved = self.results.get_loose_reference("python", "nonexistent")
--- a/tests/static_analyzer/test_graph.py
+++ b/tests/static_analyzer/test_graph.py
@@ -60,7 +60,7 @@ class TestNode(unittest.TestCase):
         caller = Node("module.caller", 12, "/file.py", 1, 10)
 
         with self.assertRaises(ValueError) as context:
-            caller.added_method_called_by_me("invalid_string")
+            caller.added_method_called_by_me("invalid_string")  # type: ignore[arg-type]
 
         self.assertIn("Expected a Node instance", str(context.exception))
 
@@ -336,7 +336,7 @@ class TestCallGraph(unittest.TestCase):
         # Test LLM string for large graph (exceeds size limit)
         graph = CallGraph()
 
-        # Create many method nodes (type 6)
+        # Create many method nodes (type "6")
         for i in range(50):
             node = Node(f"class{i % 5}.ClassA.method{i}", 6, "/file.py", i * 10, i * 10 + 5)
             graph.add_node(node)


agents/agent.py
```diff
@@ -246,12 +248,4 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, syste
         agent_model = os.getenv("AGENT_MODEL")
         llm, model_name = self._static_initialize_llm(model_override=agent_model, is_parsing=False)
         super().__init__(repo_dir, static_analysis, system_message, llm)
-        self.agent_monitoring_callback.model_name = model_name
-
-
-class SmallModelAgent(CodeBoardingAgent):
```
ivanmilevtues, line 252: This was unused so it is removed now :)

brovatten: Fair enough! :D

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
             logger.warning(summary)
             return summary
 
-        communities = list(nx.community.greedy_modularity_communities(cfg_graph_x))
+        communities, strategy_used = self._adaptive_clustering(cfg_graph_x, target_clusters=20, min_cluster_size=2)
```
brovatten: How did u pick the params?

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+        if not communities:
+            return False
+
+        valid_clusters = [c for c in communities if len(c) >= min_cluster_size]
+
+        if len(valid_clusters) == 0:
+            return False
+
+        cluster_count = len(valid_clusters)
+        if cluster_count < max(2, target_clusters // 6):
```
brovatten: // 6? :D add comment here please

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+            return False
+
+        if cluster_count > target_clusters * 2:
```
brovatten: a small comment on assumption/thinking here would be cool

main.py
```diff
@@ -333,9 +333,8 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
     # Advanced options
     parser.add_argument(
         "--load-env-variables",
-        action="store",
+        action="store_true",
         default=False,
```
Svilen-Stefanov: Shouldn't we remove this, having `store_true` and default `False` seems a bit confusing.

static_analyzer/graph.py
```diff
@@ -2,75 +2,74 @@
...
         return hash(self.fully_qualified_name)
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return f"Node({self.fully_qualified_name}, {self.file_path}, {self.line_start}-{self.line_end})"
 
 
 class Edge:
-    def __init__(self, src_node: Node, dst_node: Node):
+    def __init__(self, src_node: Node, dst_node: Node) -> None:
```
Svilen-Stefanov, line 81: Thanks for adding types everywhere.

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+            if abstract_src != abstract_dst:
+                edge_weights[(abstract_src, abstract_dst)] += 1
+
+        for (src, dst), weight in edge_weights.items():
+            abstracted_graph.add_edge(src, dst, weight=weight)
+
+        return abstracted_graph
+
+    def _get_abstract_node_name(self, node_name: str, level: str) -> str:
+        parts = node_name.split('.')
```
brovatten: This may break for other languages, not sure if C++ for example would use other than ".". Use delimiter maybe?

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+                return [set(comm) for comm in sub_communities if len(comm) >= 2]
+        except:
+            pass
+
+        components = list(nx.connected_components(subgraph.to_undirected()))
+        if len(components) > 1:
+            return [set(comp) for comp in components]
+
+        cluster_list = list(large_cluster)
+        mid = len(cluster_list) // 2
```
brovatten, line 412: Is there a smarter way we can split this? 😅  it's quite hard to pick up when this split is wrong (does not follow any structure, or breaks it)

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+        logger.info(f"Used clustering strategy: {strategy_used}, found {len(communities)} clusters")
         top_nodes = set().union(*communities) if communities else set()
 
-        cluster_str = self.__cluster_str(top_communities, cfg_graph_x)
+        cluster_str = self.__cluster_str(communities, cfg_graph_x)
         non_cluster_str = self.__non_cluster_str(cfg_graph_x, top_nodes)
-
         return cluster_str + non_cluster_str
 
+    def _adaptive_clustering(self, graph: nx.DiGraph, target_clusters: int = 20, min_cluster_size: int = 2) -> tuple[
```
Svilen-Stefanov: Do you want to add some info about the default numbers here? I know they are not coming from proper research but even the experiments you did might be worth mentioning.

main.py
```diff
@@ -333,9 +333,8 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
     # Advanced options
     parser.add_argument(
         "--load-env-variables",
-        action="store",
+        action="store_true",
```
brovatten, line 336: Nice!!

static_analyzer/graph.py
```diff
@@ -207,16 +428,14 @@ def llm_str(self, size_limit=2_500_000, skip_nodes=[]):
         for _, node in self.nodes.items():
             if node in skip_nodes:
                 continue
-            if node.type == 6 and node.methods_called_by_me:  # type 6 = method
-                # Extract class name from fully qualified name
+            if node.type == 6 and node.methods_called_by_me:
```
brovatten: Can this be a constant? Also, these magic numbers are a bit scary

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
         non_cluster_str = self.__non_cluster_str(cfg_graph_x, top_nodes)
-
         return cluster_str + non_cluster_str
 
+    def _adaptive_clustering(self, graph: nx.DiGraph, target_clusters: int = 20, min_cluster_size: int = 2) -> tuple[
+        list[set[str]], str]:
+        total_nodes = graph.number_of_nodes()
+        logger.info(f"Starting adaptive clustering for {total_nodes} nodes, target: {target_clusters} clusters")
+
+        for algorithm in ['louvain', 'leiden', 'greedy_modularity']:
```
brovatten: Does it make sense to prioritze the ordering based on the repo size, or other qualities?

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+                        remaining.pop(i)
+                        merged_any = True
+                        break
+
+            if len(current_cluster) >= min_cluster_size:
+                merged_clusters.append(current_cluster)
+
+        return merged_clusters
+
+    def _is_good_clustering(self, communities: list[set[str]], target_clusters: int, min_cluster_size: int,
```
Svilen-Stefanov: Seems like this method deserves a comment about what we believe is a good clustering.

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+        if level == 'class' and len(parts) > 1:
+            return '.'.join(parts[:-1])
+        elif level == 'file' and len(parts) > 2:
+            return '.'.join(parts[:-2])
+        elif level == 'package' and len(parts) > 3:
+            return parts[0]
+        else:
+            return node_name
+
+    def _map_abstract_to_original(self, abstract_communities: list[set[str]], original_graph: nx.DiGraph, level: str) -> \
```
brovatten: To make this O(N), create a look table instead

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+            try:
+                communities = self._cluster_at_level(graph, level, target_clusters, min_cluster_size)
+                if self._is_good_clustering(communities, target_clusters, min_cluster_size, total_nodes):
+                    return communities, f"structural_{level}"
+            except Exception as e:
+                logger.debug(f"Structural level {level} failed: {e}")
+                continue
+
+        try:
+            initial_communities = list(nx.community.greedy_modularity_communities(graph))
```
Svilen-Stefanov, line 218: Aren't we doing this in the first loop as well or is the difference with `greedy_modularity` about the number of target clusters? 
The order in which we do the clustering seems very important as we pick the first that is considered good, should we add a comment about it in this method as well?

static_analyzer/graph.py
brovatten: I think greedy is poor for big codebases for example. I also wonder how these algorithms scale to LoC

static_analyzer/graph.py
```diff
@@ -91,37 +90,279 @@ def to_cluster_string(self) -> str:
...
+            return balanced_communities, "balanced_greedy_modularity"
+        except Exception as e:
+            logger.warning(f"All clustering strategies failed: {e}")
+            components = list(nx.connected_components(graph.to_undirected()))
+            return components[:target_clusters], "connected_components"
+
+    def _cluster_at_level(self, graph: nx.DiGraph, level: str, target_clusters: int, min_cluster_size: int) -> list[
+        set[str]]:
+        if level == 'method':
+            return self._cluster_with_algorithm(graph, 'louvain', target_clusters)
```
Svilen-Stefanov: Hmm, this seems a bit redundant - won't we try exactly this in the other for loop:
```
for algorithm in ['louvain', 'leiden', 'greedy_modularity']:
            try:
                communities = self._cluster_with_algorithm(graph, algorithm, target_clusters)
                if self._is_good_clustering(communities, target_clusters, min_cluster_size, total_nodes):
                    return communities, f"connectivity_{algorithm}"
```
Reaching here seems like the above didn't work.

brovatten (changes requested):
I would love to see E2E and Static analysis eval results for the different clustering algorithms, to see the results on the codeboarding repo and the time to run it.

The louvain and leiden uses shuffling, I think we should pass the seed for reproducibility (or does it pick up our np seed)?:
image


Also, black formatter.

I realised while reading this is a really important topic :D Ping me once this is done, and I will review again.

One random advice from gemini (not sure if relevant or not):
Advice: The _edge_set: set[tuple[str, str]] in CallGraph is redundant once the graph is built. You should del self._edge_set after the graph construction is complete to free up memory before the expensive clustering begins.

main.py
ivanmilevtues: store_true means that it is a flag and not an arg, so if --load-env-variables is there it is true, if you skip the arg it is false.

The other was way more confusing to me as if you have it as "store" and default True, any value you put there is treated as True i.e.

--load-env-variables asdasdas and --load-env-variables "false" is true for both

Svilen-Stefanov: Yeah, I meant if we want to remove `default=False` - I like the change, I just find the default after the change a bit misleading.

static_analyzer/graph.py
ivanmilevtues: I added a big comment that both me and Kiro liked :D + I added a Note comment for your comment here to explain that we add the balance_clusters on top of the initial_communities.

PR #132, opened 2025-12-27
Enhancing depth analysis

ivanmilevtues:
Enhancing depth analysis with proper cluster associations.

Before we just used the source code for depth analysis. Now each component has its related clusters, for further analysis for the details agent.

To reduce LLM complexity the steps are simplified in their output so that we don't ask the LLM to do too much in one step.

### DIFF
--- a/tests/agents/prompts/test_prompts.py
+++ b/tests/agents/prompts/test_prompts.py
@@ -22,10 +22,8 @@ class TestAbstractPromptFactory(unittest.TestCase):
         # Test that all expected abstract methods are defined
         expected_methods = [
             "get_system_message",
-            "get_cfg_message",
-            "get_source_message",
-            "get_classification_message",
-            "get_conclusive_analysis_message",
+            "get_cluster_analysis_message",
+            "get_final_analysis_message",
             "get_feedback_message",
             "get_system_details_message",
             "get_subcfg_details_message",
@@ -242,10 +240,8 @@ class TestConvenienceFunctions(unittest.TestCase):
 
         convenience_functions = [
             "get_system_message",
-            "get_cfg_message",
-            "get_source_message",
-            "get_classification_message",
-            "get_conclusive_analysis_message",
+            "get_cluster_analysis_message",
+            "get_final_analysis_message",
             "get_feedback_message",
             "get_system_details_message",
             "get_subcfg_details_message",


agents/abstraction_agent.py
```diff
@@ -45,13 +49,23 @@ def __init__(
...
+            # NEW: Cluster-aware prompts
+            "analyze_clusters": PromptTemplate(
+                template=get_cluster_analysis_message(),
+                input_variables=["project_name", "cfg_clusters", "meta_context", "project_type"]
+            ),
+            "final_analysis": PromptTemplate(
+                template=get_final_analysis_message(),
+                input_variables=["project_name", "cluster_analysis", "meta_context", "project_type"],
+            ),
+            # OLD: Keep for backward compatibility (will be removed later)
```
ivanmilevtues: remove now.

agents/abstraction_agent.py
```diff
@@ -193,9 +399,12 @@ def classify_files(self, analysis: AnalysisInsights):
...
             comp.assigned_files = files
 
+        logger.info(f"[AbstractionAgent] Classified {len(all_files)} files into {len(analysis.components)} components")
+
     def run(self):
-        self.step_cfg()
-        self.step_source()
-        analysis = self.generate_analysis()
+        """New simplified two-step flow"""
+        self.analyze_clusters()  # Step 1: Understand clusters
```
ivanmilevtues: remove step_cfg from the class no longer used, same for step_source.

agents/details_agent.py
```diff
@@ -65,10 +66,71 @@ def __init__(
...
+                cfg_lines.append(f"\n## {lang.capitalize()} - Component CFG\n")
+                cfg_lines.append(filtered_clusters)
+                cfg_lines.append("\n")
+
+        result = "".join(cfg_lines)
+        if not result.strip():
+            logger.warning(f"[DetailsAgent] No CFG found for component {component.name}, cluster IDs: {cluster_ids}")
+            return "No relevant CFG clusters found for this component."
+
+        return result
```
ivanmilevtues: All these are the same as the abstraction agent.

evals/reports/end-to-end-report.md
```diff
@@ -1,24 +1,68 @@
...
+**Generated:** 2025-12-27T00:05:32.012040+00:00
 
 ### Summary
 
 | Project | Language | Status | Time (s) | Total Tokens | Tool Calls |
 |---------|----------|--------|----------|--------------|------------|
-| webscrap | Python | ❌ Failed | 1.05 | 0 | 0 |
+| markitdown | Python | ✅ Success | 356.2 | 808,081 | 69 |
+| codeboarding | Python | ✅ Success | 315.7 | 497,793 | 46 |
+| django | Python | ❌ Failed | 1800.0 | 0 | 0 |
```
ivanmilevtues: 30 mins, it is quite slow :(

evals/cli.py
```diff
@@ -33,6 +35,9 @@ def parse_args():
 
 
 def main():
+    # Load environment variables from .env file
+    load_dotenv()
```
ivanmilevtues, line 38: without this you cannot run the evals, it used to be the case that main would loaddotenv, now we have the flag which will trigger such load.

static_analyzer/graph.py
ivanmilevtues, line 317: I have to rebase on main ...

agents/abstraction_agent.py
```diff
@@ -146,42 +117,51 @@ def apply_feedback(self, analysis: AnalysisInsights, feedback: ValidationInsight
...
-            file_block = [str(f) for f in all_files[i : i + 100]]
-            prompt = self.prompts["classification"].format(
-                project_name=self.project_name, components=component_str, files="\n".join(file_block)
-            )
-            classification = self._parse_invoke(prompt, ComponentFiles)
-            files.extend(classification.file_paths)
-        for file in files:
-            component_match: Component | None = next(
-                (c for c in analysis.components if c.name == file.component_name), None
+        # Build file-to-cluster mapping for deterministic classification
```
ivanmilevtues: remove comment

agents/abstraction_agent.py
```diff
@@ -146,42 +117,51 @@ def apply_feedback(self, analysis: AnalysisInsights, feedback: ValidationInsight
...
                 name="Unclassified",
                 description="Component for all unclassified files and utility functions (Utility functions/External Libraries/Dependencies)",
-                referenced_source_code=[],
+                key_entities=[],
+                source_cluster_ids=[],
             )
         )
-        component_str = "\n".join([component.llm_str() for component in analysis.components])
 
+        # Reset assigned_files for all components
```
ivanmilevtues: remove comment

agents/details_agent.py
```diff
@@ -65,10 +66,46 @@ def __init__(
...
+
     def step_subcfg(self, component: Component):
```
ivanmilevtues, line 95: is this used, i think not?

evals/cli.py
```diff
@@ -33,6 +35,9 @@ def parse_args():
 
 
 def main():
+    # Load environment variables from .env file
```
ivanmilevtues: remove

static_analyzer/graph.py
```diff
@@ -55,7 +55,7 @@ class Node:
     METHOD_TYPE = 6
 
     def __init__(
-        self, fully_qualified_name: str, node_type: int, file_path: str, line_start: int, line_end: int
+        self, fully_qualified_name: str, node_type: str, file_path: str, line_start: int, line_end: int
```
ivanmilevtues: this should be int

agents/details_agent.py
ivanmilevtues: it is used

.codeboarding/Agent_Orchestration_Prompting.json
```diff
@@ -0,0 +1,145 @@
...
+        },
+        {
+          "qualified_name": "diagram_analysis.diagram_generator.DiagramGenerator.process_component",
+          "reference_file": "diagram_analysis/diagram_generator.py",
+          "reference_start_line": 69,
+          "reference_end_line": 112
+        }
+      ],
+      "assigned_files": [
+        "../../diagram_analysis/diagram_generator.py"
```
Svilen-Stefanov: Will these be re-generated upon merge? 
I'd be interested to see if the paths will be indeed fixed after the latest changes I've merged (we shouldn't see such ../../ prefixes anymore).

agents/agent_responses.py
```diff
@@ -85,33 +80,60 @@ def llm_str(self):
...
+
+
+class Component(LLMBaseModel):
     name: str = Field(description="Name of the component")
     description: str = Field(description="A short description of the component.")
-    referenced_source_code: list[SourceCodeReference] = Field(
-        description="A list of source code names of referenced methods and classes to the component. THIS CANNOT BE EMPTY."
+
+    # LLM picks these: The MOST IMPORTANT/critical methods and classes
+    key_entities: list[SourceCodeReference] = Field(
```
Svilen-Stefanov, line 123: Isn't this a breaking change? My thinking is that this is part of the component files so people that already have generated analysis and use `referenced_source_code` will no longer be able to read their diagrams, right?
At least I couldn't see a backwards compatibility thing here.

You can add an alias for backwards compatibility or if this is intentional, we should properly document it in the release notes.

Wdyt?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,208 @@
...
+        for line in lines:
+            # Check if this is a cluster header: "Cluster N (...)"
+            if line.startswith("Cluster "):
+                parts = line.split()
+                if len(parts) >= 2:
+                    try:
+                        cluster_num = int(parts[1])
+                        include_current = cluster_num in cluster_ids
+                    except ValueError:
+                        include_current = False
```
Svilen-Stefanov: Should we add logging here?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,208 @@
...
+            if component.name == "Unclassified":
+                continue
+
+            # Check 1: Does file contain any key entities?
+            for key_entity in component.key_entities:
+                if key_entity.reference_file:
+                    # Normalize paths for comparison
+                    ref_file_norm = os.path.normpath(key_entity.reference_file)
+                    file_path_norm = os.path.normpath(file_path)
+                    if file_path_norm.endswith(ref_file_norm) or ref_file_norm in file_path_norm:
```
Svilen-Stefanov: Can't this result in false positives - e.g utils.py and test_utils.py would be a match here, right?
Maybe something like this would be better:
```
Path(file_path_norm).resolve() == Path(ref_file_norm).resolve()
```

agents/agent_responses.py
```diff
@@ -85,33 +80,60 @@ def llm_str(self):
...
+
+    # Deterministic from static analysis: ALL files belonging to this component
     assigned_files: list[str] = Field(
-        description="A list of source code names of files assigned to the component.",
+        description="All source files assigned to this component (populated deterministically).",
         default_factory=list,
         exclude=True,
     )
 
+    source_cluster_ids: list[int] = Field(
```
Svilen-Stefanov, line 134: Do we validate somewhere that these IDs actually exist in the CFG if this is populated by the LLMs?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,208 @@
...
+            cfg = self.static_analysis.get_cfg(lang)
+            nx_graph = cfg.to_networkx()
+
+            if nx_graph.number_of_nodes() == 0:
+                continue
+
+            # Get cluster mapping
+            communities, _ = cfg._adaptive_clustering(
+                nx_graph,
+                target_clusters=20,  # Use same default as to_cluster_string()
```
Svilen-Stefanov: Should we define a constant and use it in both places? It would also be good to add a reason (even if it is primitive) so that we know how we picked those.

evals/reports/end-to-end-report.md
Svilen-Stefanov: I don't see previous reports, was this ever succeeding? Do we have a baseline if the changes actually improve or make things worse in terms of performance?

agents/prompts/abstract_prompt_factory.py
```diff
@@ -14,6 +14,14 @@ class AbstractPromptFactory(ABC):
     def get_system_message(self) -> str:
         pass
 
+    @abstractmethod
+    def get_cluster_analysis_message(self) -> str:
```
Svilen-Stefanov, line 18: Something I don't quite in this factory pattern is that all subclasses technically have the exact same implementation but refer to different constants. 
For most models, we will probably use the same or very similar prompt, which seems like a maintenance nightmare - adding a new prompt requires updating:
1. The abstract class
2. All 6 concrete factory classes
3. The convenience functions in `prompt_factory.py`
4. The `__all__` list in `__init__.py`

Seems a bit overengineered to me - do we expect the prompts to be very different for all or most of the models? If not, should we define defaults and overwrite occasionally? 

This is probably a problem for another PR but just wanted to mention it here.

Svilen-Stefanov (changes requested):
I couldn't finalize the review but I'm leaving some comments that you can start working on if you have time. Thank you for working on this, it will make everything much better!

agents/abstraction_agent.py
```diff
@@ -45,99 +42,61 @@ def __init__(
...
+                template=get_final_analysis_message(),
+                input_variables=["project_name", "cluster_analysis", "meta_context", "project_type"],
             ),
             "feedback": PromptTemplate(template=get_feedback_message(), input_variables=["analysis", "feedback"]),
         }
 
     @trace
-    def step_cfg(self):
-        logger.info(f"[AbstractionAgent] Analyzing CFG for project: {self.project_name}")
+    def analyze_clusters(self) -> ClusterAnalysis:
```
brovatten, line 53: Much better naming! gj

agents/agent_responses.py
```diff
@@ -85,33 +80,60 @@ def llm_str(self):
...
+    )
+
     def llm_str(self):
         n = f"**Component:** `{self.name}`"
         d = f"   - *Description*: {self.description}"
         qn = ""
-        if self.referenced_source_code:
-            qn += "   - *Related Classes/Methods*: "
-            qn += ", ".join(f"`{q.llm_str()}`" for q in self.referenced_source_code)
+        if self.key_entities:
```
brovatten, line 143: Key components may be more consistent naming (if i understand right)

brovatten: or "core components"

agents/agent_responses.py
```diff
@@ -235,8 +243,6 @@ def llm_str(self):
 
 
 class ComponentFiles(LLMBaseModel):
-    """Files classified to components."""
```
brovatten, line 238: Good that you cleaned up here

agents/cluster_methods_mixin.py
brovatten: Agreed!

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,208 @@
...
+                        include_current = False
+
+            if include_current:
+                result_lines.append(line)
+
+        return "\n".join(result_lines)
+
+    def _build_file_cluster_mapping(self) -> dict[str, set[int]]:
+        """
+        Build mapping of file_path -> set of cluster_ids that have nodes in that file.
```
brovatten: "set of cluster_ids that have nodes in that file." 🤔 this description feels hard to grasp. maybe it's just me?

agents/cluster_methods_mixin.py
brovatten: Great point. Even if it's just "we just used default. can be tuned later"

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,208 @@
...
+        This is purely deterministic - no LLM needed.
+        """
+        file_to_clusters: dict[str, set[int]] = defaultdict(set)
+
+        for lang in self.static_analysis.get_languages():
+            cfg = self.static_analysis.get_cfg(lang)
+            nx_graph = cfg.to_networkx()
+
+            if nx_graph.number_of_nodes() == 0:
```
brovatten: Should we log this?

agents/agent_responses.py
ivanmilevtues: It is a breaking change. You can check the PR in the vscode, there it is handled.

I renamed it as it matters to the LLM. Namings are important as they give bias. The way it is handled there is, both keys are accepted when loading. And I think we should put it as depricated if you want to.

I am unsure if many people have this and even if it is incompatible it is good for them to rerun the analysis as we introduced lot's of improvements lately. I would rather focus on new things then backward compatibility at this moment.

agents/agent_responses.py
ivanmilevtues: We don't, that seems like a good idea, my only concern is that there is no cluster ids assigned :D. Which is a valid use case if a component is small enough it might be outside of the clusters.

agents/cluster_methods_mixin.py
ivanmilevtues: We should!

agents/prompts/abstract_prompt_factory.py
ivanmilevtues: Yes, the prompts will be very different sadly. As Anthropic i.e. loves tool calling and there we should explicitly note to use tools less often.

Further in the future we'd have to have also custom prompts for the tools and custom tools for different models. That is industry standard, we are still far from that point, but it is good to have in mind

agents/cluster_methods_mixin.py
ivanmilevtues: I will add

.codeboarding/Agent_Orchestration_Prompting.json
ivanmilevtues: I will update them, should be resolved. This is an older version.

agents/prompts/claude_prompts_unidirectional.py
```diff
@@ -24,6 +24,65 @@
...
+The CFG has been pre-clustered into groups of related methods/functions. Each cluster represents methods that call each other frequently.
+
+CFG Clusters:
+{cfg_clusters}
+
+Instructions:
+1. For each cluster shown above, provide:
+   - The cluster ID (exactly as shown, e.g., "1", "2", "3")
+   - A short descriptive name (2-4 words)
+   - One sentence explaining what this cluster does
```
brovatten, line 41: Really would be interesting to see a cluster description with and without this constraint of "One sentence". to see if it would be valuable for the subsequent analysis

agents/cluster_methods_mixin.py
brovatten: I think `Path.parts` would be even better

agents/abstraction_agent.py
```diff
@@ -146,42 +105,41 @@ def apply_feedback(self, analysis: AnalysisInsights, feedback: ValidationInsight
...
-            files.extend(classification.file_paths)
-        for file in files:
-            component_match: Component | None = next(
-                (c for c in analysis.components if c.name == file.component_name), None
-            )
-            if component_match is None:
-                logger.warning(f"[AbstractionAgent] File {file.component_name} not found in analysis")
-                continue
-            component_match.assigned_files.append(file.file_path)
+        file_to_clusters = self._build_file_cluster_mapping()
```
brovatten: Is this called multiple times? if so caching may be nice as it's deterministic

evals/reports/end-to-end-report.md
ivanmilevtues: I will rerun with vercel hopefully it passes now :)

agents/agent_responses.py
Svilen-Stefanov: Makes sense. Should we maybe only make sure there's a clear error we show if that's the case?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                logger.debug(f"[ClusterMethodsMixin] Skipping {lang} CFG - no nodes found")
+                continue
+
+            communities, _ = cfg._adaptive_clustering(
```
Svilen-Stefanov: I have a question here that might be a potential problem:

Afaik, Louvain and Leiden algorithms are generally considered non-deterministic so we could generate different clusters each time the clustering is performed. Iiuc, we run this before and after the LLM steps so the clusters in the communities here might differ from the previous ones - did you think about that?

Example:
- analyze_clusters() calls to_cluster_string() → produces cluster IDs 1, 2, 3...
- LLM assigns component.source_cluster_ids = [1, 3]
- classify_files() re-runs _adaptive_clustering() → different cluster IDs due to non-determinism
- Can files get assigned to the wrong components? I think it is possible

agents/abstraction_agent.py
```diff
@@ -45,99 +42,61 @@ def __init__(
         self.context: dict[str, LLMBaseModel] = {}
 
         self.prompts = {
-            "cfg": PromptTemplate(
-                template=get_cfg_message(), input_variables=["project_name", "cfg_str", "meta_context", "project_type"]
-            ),
-            "source": PromptTemplate(
-                template=get_source_message(), input_variables=["insight_so_far", "meta_context", "project_type"]
+            "analyze_clusters": PromptTemplate(
+                template=get_cluster_analysis_message(),
```
Svilen-Stefanov, line 42: I see we are sing new prompt messages but you've also kept the old ones, which seem no longer needed. 
Do we want to remove `get_cfg_message`, `get_source_message` and `get_classification_message`?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                    # Get file path for this node
+                    if node_name in nx_graph.nodes:
+                        node_data = nx_graph.nodes[node_name]
+                        file_path = node_data.get("file_path")
+                        if file_path:
+                            file_to_clusters[file_path].add(cluster_id)
+
+        return file_to_clusters
+
+    def _match_file_to_components(
```
Svilen-Stefanov: I'm not sure how to handle this just yet but I think we should have some sort of validation that there aren't too many components a file is assigned to (especially if multiple files are assigned to many of the same components) as this can indicate poor clustering. 
At this point, can we maybe just add metrics that track this?

agents/details_agent.py
```diff
@@ -63,15 +64,49 @@ def __init__(
...
+            logger.warning(f"[DetailsAgent] Component {component.name} has no source_cluster_ids, using fallback")
+            # Fallback to old method if no cluster IDs
+            return self.read_cfg_tool.component_cfg(component)  # type: ignore[return-value]
+
+        cluster_ids = set(component.source_cluster_ids)
+        cfg_lines = []
+
+        for lang in self.static_analysis.get_languages():
+            cfg = self.static_analysis.get_cfg(lang)
+            cluster_str = cfg.to_cluster_string()
```
Svilen-Stefanov: I see we have `_build_cluster_string`, which does something similar. Can it be reused here - why are we filtering the clusters here but not there?

Also I'm not sure how often this is called but it could be benefitial to cache this if called multiple times as this is re-running clustering iiuc.

evals/reports/end-to-end-report.md
```diff
@@ -1,24 +1,66 @@
...
+**Generated:** 2026-01-01T23:16:52.961160+00:00
 
 ### Summary
 
 | Project | Language | Status | Time (s) | Total Tokens | Tool Calls |
 |---------|----------|--------|----------|--------------|------------|
-| webscrap | Python | ❌ Failed | 1.05 | 0 | 0 |
+| markitdown | Python | ✅ Success | 450.0 | 1,661,125 | 97 |
+| codeboarding | Python | ✅ Success | 381.5 | 1,536,205 | 66 |
+| django | Python | ❌ Failed | 53454.4 | 0 | 0 |
```
Svilen-Stefanov: Crazy, what happened here - do I understand this correctly or was this running almost 15h?

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                    except ValueError:
+                        logger.warning(f"[ClusterMethodsMixin] Failed to parse cluster ID from line: {line}")
+                        include_current = False
+
+            if include_current:
+                result_lines.append(line)
+
+        return "\n".join(result_lines)
+
+    def _build_file_cluster_mapping(self) -> dict[str, set[int]]:
```
Svilen-Stefanov: Hmm, I see potential for confusion here as we have two somewhat distinct clustering paths / approaches. 
In `graph.py` we have `to_cluster_string`, which sorts communities based on their size:
```
top_communities = sorted(valid_communities, key=len, reverse=True)
```

Here, we don't sort them so I'm wondering, can this result in a problem? Not sure where / what kind of a test would help here, but I'm thinking that there might be a problem if:
```
communities = [{x,y}, {a,b,c,d,e}]:
```
As they will be handled differently by the above two and we seem to use both.

agents/agent_responses.py
ivanmilevtues: There is no error, we will load the old thing in VSCode, it accepts both

evals/reports/static-analysis-report.md
```diff
@@ -1,23 +1,23 @@
 # Static Analysis Performance Evaluation
 
-**Generated:** 2025-12-11T23:21:14.580915+00:00
+**Generated:** 2026-01-15T23:13:06.223708+00:00
```
ivanmilevtues: We should improve this, I've spoken with @brovatten on making it easier to read

agents/cluster_methods_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                        ref_path = Path(key_entity.reference_file)
+                        file_path_obj = Path(file_path)
+                        # Handle both absolute and relative paths
+                        if ref_path.is_absolute() and file_path_obj.is_absolute():
+                            if file_path_obj.resolve() == ref_path.resolve():
+                                matched_components.append(component)
+                                break
+                        else:
+                            # For relative paths, check if file_path ends with ref_file
+                            if file_path_obj.resolve().match(f"*/{ref_path}") or file_path_obj.name == ref_path.name:
```
Svilen-Stefanov: I think the second part of this could still result in false positives - example `src/services/utils.py` and `utils.py` (this file is just under the repo path), won't this be considered the same?

Do you want to add some meaningful tests for some of the utility functions here to make sure we know the behavior exactly?
For the paths this should be quite simple and reliable. It might be useful for other methods as well but that's up to you - I'd probably love to see some tests that could have caught the issues we're discussing here.

agents/abstraction_agent.py
```diff
@@ -193,9 +151,11 @@ def classify_files(self, analysis: AnalysisInsights):
...
             comp.assigned_files = files
 
+        logger.info(f"[AbstractionAgent] Classified {len(all_files)} files into {len(analysis.components)} components")
+
     def run(self):
-        self.step_cfg()
-        self.step_source()
-        analysis = self.generate_analysis()
+        self.analyze_clusters()  # Step 1: Understand clusters
+        analysis = self.generate_analysis()  # Step 2: Create final components
```
Svilen-Stefanov: Iiuc, the LLM here determines the components and which clusters their assigned to. LLM could hallucinate cluster ID 99 when only clusters 1-3 exist, potentially breaking file assignment. So I'm thinking something like:
```
cluster_analysis = self.analyze_clusters()  # Step 1: Understand clusters
analysis = self.generate_analysis()  # Step 2: Create final components
    
# Add step here: Validate cluster ID references
self._validate_cluster_ids(analysis, cluster_analysis)
```

I also find it somewhat suspicious that `analyze_clusters()` has a return type, which we don't use. I see that's because of `self.context["cluster_analysis"]` - I see that the context is also only used for that so I'm wondering if this is a mistake and whether we should only use the return type instead or just direcly store the `cluster_analysis` in a var.

agents/agent_responses.py
Svilen-Stefanov: Iiuc, you are referring to file to cluster mapping, I'm referring to the LLM halucinating that fileA belongs to cluster 99 even though we have 3 clusters only. Refer to my comment here: https://github.com/CodeBoarding/CodeBoarding/pull/132#discussion_r2700840877

nit: why `source_cluster_ids`, how about `assigned_cluster_ids` or just `cluster_ids`?

agents/cluster_methods_mixin.py
Svilen-Stefanov: Maybe both `_build_file_cluster_mapping` and `_match_file_to_components` should be part of the static analyser and not part of this class.

Svilen-Stefanov (approved):
The screenshots look great, I think this significantly improves our current flow. 
My main concern is about the use of the clustering and the separation of concerns between the `cluster_methods_mixin.py` and the static analyser, which is related to this comment: https://github.com/CodeBoarding/CodeBoarding/pull/132#discussion_r2700879450

The mixin should potentially only convert requests and responses in the expected format and call the static analyser to do the rest. We might also make these methods standalone and not part of a mixin to have a single place that handles the parsing of the clustering. Not sure what's the best approach would be here.

.codeboarding/analysis.json
```diff
@@ -1,268 +1,372 @@
...
-        "agents/prompts/gemini_flash_prompts_unidirectional.py",
-        "agents/prompts/claude_prompts_unidirectional.py"
+        "agents/details_agent.py",
+        "agents/validator_agent.py",
+        "agents/agent.py",
+        "agents/diff_analyzer.py",
+        "agents/tools/toolkit.py"
+      ],
+      "source_cluster_ids": [
+        5,
```
Svilen-Stefanov: Something I saw just now - none of the child components seem to have source_cluster_ids . Now that we're actually adding this, we need to think about cluser ids in relationship with child components - e.g do we want to have hierarchical clusters - e.g for a child on the third level to have:
```
source_cluster_ids: [
  1.3.5 # belongs to the first cluster of the overview, third cluster of the parent and 5th cluster of the leve
]
```

And if we don't, do we want to have these cluster IDs as part of the analysis or will they lead to confusion (if parent and child refer to the same cluster but the clusters actually mean 2 completely different things).

brovatten (approved):

PR #134, opened 2025-12-30
Ignore more file patterns and fail fast

Svilen-Stefanov:
The goal of this PR is to fix some parts of the static analysis, in particular:
- fail faster for files we cannot analyze (reduce the timeout for `_analyze_single_file` from 6min to 1min) - I kept the 6min timeout for initialization as it may involve full workspace indexing
- ignore derivative files (e.g `.min.js` or `.bundle.js`)

As I also got some black formatter issues, I noticed the pre-commit hook is not installed in `setup.py` and added that as well to make sure we reduce the number of build failures.

### DIFF
--- a/repo_utils/ignore.py
+++ b/repo_utils/ignore.py
@@ -30,12 +30,12 @@ class RepoIgnoreManager:
 
     # Build artifacts and minified files that should be ignored
     DEFAULT_IGNORED_FILE_PATTERNS = [
-        "*.bundle.js",      # Webpack/bundler output
+        "*.bundle.js",  # Webpack/bundler output
         "*.bundle.js.map",  # Source maps for bundles
-        "*.min.js",         # Minified JavaScript
-        "*.min.css",        # Minified CSS
-        "*.chunk.js",       # Code-split chunks
-        "*.chunk.js.map",   # Source maps for chunks
+        "*.min.js",  # Minified JavaScript
+        "*.min.css",  # Minified CSS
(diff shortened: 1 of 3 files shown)


repo_utils/ignore.py
```diff
@@ -28,6 +28,16 @@ class RepoIgnoreManager:
         "tests",
     }
 
+    # Build artifacts and minified files that should be ignored
+    DEFAULT_IGNORED_FILE_PATTERNS = [
```
ivanmilevtues, line 32: Nice, I wonder isnt' there a way to get the list of .gitignores from github: 
https://github.com/github/gitignore <- something like this.

static_analyzer/lsp_client/client.py
```diff
@@ -193,7 +193,8 @@ def _initialize(self):
             "workspace": {"applyEdit": True, "workspaceEdit": {"documentChanges": True}},
         }
         init_id = self._send_request("initialize", params)
-        response = self._wait_for_response(init_id)
+        # Use longer timeout for initialization as it may involve full workspace indexing
```
ivanmilevtues, line 196: This is better than always waiting 6 mins!

setup.py
```diff
@@ -380,6 +380,45 @@ def init_dot_env_file():
         print(f"Step: .env file creation finished: failure - {e}")
 
 
+def install_pre_commit_hooks():
```
ivanmilevtues, line 383: Nice

PR #137, opened 2026-01-06
Add java support

Svilen-Stefanov:
This PR adds Java static analysis support using Eclipse JDT Language Server (jdtls), which seems to be the industry-standard LSP implementation for Java used by VSCode and other major editors, here's some more info:
- LSP Server Choice: Eclipse jdtls is the only viable production-ready Java LSP server (alternatives are deprecated or immature)
- Runtime Requirement: Java 21+ is required to run jdtls (hard requirement), though projects can target Java 1.8+
- Build System Support: Automatic detection and configuration for Maven, Gradle, and Eclipse projects with multi-module support
- Distribution Model: JDTLS is platform-independent Java bytecode distributed as a single tar.gz archive containing all platform configs (config_linux/, config_mac/, config_win/)

Architecture: 
• JavaConfigScanner: Detects Maven/Gradle projects and multi-module structures via pom.xml and settings.gradle parsing 
• JavaClient: LSP client extending existing LSPClient base class with JDTLS-specific initialization, project import tracking, and workspace management 
• java_utils.py: Cross-platform utilities for Java version detection, JDK discovery, and JDTLS command construction via Eclipse Equinox launcher 
• Integration follows existing patterns from TypeScript multi-project support in create_clients factory

Some other things to note about the implementation: 
• Java installation auto-detection across platforms (macOS, Linux, Windows) with fallback to JAVA_HOME 
• Build system detection following priority: Gradle → Maven → Eclipse → fallback 
• Project import tracking with timeout handling (default 5 minutes) 
• Heap size calculation based on project size (1G-8G range)

### DIFF
--- a/static_analyzer/java_config_scanner.py
+++ b/static_analyzer/java_config_scanner.py
@@ -1,9 +1,3 @@
-"""
-Java project configuration scanner.
-
-Detects Maven, Gradle, and multi-module Java projects.
-"""
-
 from pathlib import Path
 import xml.etree.ElementTree as ET
 import logging
@@ -14,7 +8,6 @@ logger = logging.getLogger(__name__)
 
 
 class JavaProjectConfig:
-    """Configuration for a Java project."""
 
     def __init__(
         self,
@@ -38,7 +31,6 @@ class JavaProjectConfig:
 
 
 class JavaConfigScanner:
-    """Scanner for Java project configurations."""
 
     def __init__(self, repo_path: Path, ignore_manager: RepoIgnoreManager | None = None):
         self.repo_path = repo_path
--- a/static_analyzer/lsp_client/java_client.py
+++ b/static_analyzer/lsp_client/java_client.py
@@ -9,10 +9,15 @@ import tempfile
 import time
 from pathlib import Path
 
-from .client import LSPClient
-from ..java_config_scanner import JavaProjectConfig
-from ..java_utils import create_jdtls_command, get_java_version, find_java_21_or_later, detect_java_installations
-from ..programming_language import ProgrammingLanguage, JavaConfig
+from static_analyzer.lsp_client.client import LSPClient
+from static_analyzer.java_config_scanner import JavaProjectConfig
+from static_analyzer.java_utils import (
+    create_jdtls_command,
+    get_java_version,
+    find_java_21_or_later,
+    detect_java_installations,
+)
+from static_analyzer.programming_language import ProgrammingLanguage, JavaConfig
 from repo_utils.ignore import RepoIgnoreManager
 
 logger = logging.getLogger(__name__)


static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+        Extract package name from Java file.
+
+        Overrides the base implementation to use Java package declarations.
+        """
+        try:
+            # Try to read package declaration from file
+            content = file_path.read_text(encoding="utf-8")
+            lines = content.split("\n")
+
+            for line in lines[:20]:  # Check first 20 lines
```
brovatten: Increase this limit. There may be a quite long licensing text for big project at the top

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+
+        params = {
+            "processId": os.getpid(),
+            "rootUri": self.project_path.as_uri(),
+            "capabilities": self._get_capabilities(),
+            "initializationOptions": self._get_initialization_options(),
+        }
+
+        init_id = self._send_request("initialize", params)
+        response = self._wait_for_response(init_id)
```
brovatten: Should we set the timeout explciitly here like for the other calls?

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+        # Initialize base LSPClient
+        super().__init__(project_path, language, ignore_manager)
+
+        # Track import status
+        self.import_complete = False
+        self.import_errors: List[str] = []
+        self.java_home: Optional[Path] = None  # Will be detected in start()
+        self.workspace_indexed = False  # Track if workspace symbols are available
+
+    def start(self):
```
brovatten, line 74: Are we cleaning this up now if start produces an error? I would try except finally to remove the dir if fail

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+            if location.exists() and (location / "plugins").exists():
+                logger.info(f"Found JDTLS at {location}")
+                return location
+
+        return None
+
+    def _calculate_heap_size(self) -> str:
+        """Calculate appropriate heap size for project."""
+        # Count Java files as rough size indicator
+        java_files = list(self.project_config.root.rglob("*.java"))
```
brovatten: Should this exclude the ones ignored by the `self.ignore_manager.should_ignore()`?

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
...
+from pathlib import Path
+from typing import Optional, List
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def get_java_version(java_cmd: str = "java") -> int:
+    """
+    Get major version of Java executable.
```
brovatten, line 14: We aren't using this type of docstring no?

brovatten: although it does feel nice here

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Returns:
+        List of paths to JDK home directories
+    """
+    candidates = []
+
+    # Check JAVA_HOME
+    import os
+
+    if java_home := os.getenv("JAVA_HOME"):
```
brovatten, line 56: Rare walrus operator :D

brovatten: If not found, I would tell the user to set this env varialbe (or warn java may not be instsalled)

static_analyzer/java_config_scanner.py
```diff
@@ -0,0 +1,217 @@
...
+                return JavaProjectConfig(gradle_dir, "gradle", False)
+
+        except Exception as e:
+            logger.warning(f"Failed to parse {settings_file}: {e}")
+            return JavaProjectConfig(gradle_dir, "gradle", False)
+
+    def _has_java_files(self, directory: Path) -> bool:
+        """Check if directory contains any .java files."""
+        try:
+            next(directory.rglob("*.java"))
```
brovatten, line 176: Maybe use the ignore files here, and make that a helper fucntion?

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
...
+    Get major version of Java executable.
+
+    Args:
+        java_cmd: Path to java executable or "java" for system default
+
+    Returns:
+        Major version number (e.g., 21) or 0 if not found
+    """
+    try:
+        result = subprocess.run([java_cmd, "-version"], capture_output=True, text=True, timeout=5)
```
brovatten, line 23: If you have many SDKs this could take a long time perhaps? Should we cache this maybe? To avoid repeatdly running it

static_analyzer/java_config_scanner.py
```diff
@@ -0,0 +1,217 @@
...
+
+        try:
+            content = settings_file.read_text()
+
+            # Look for include statements
+            # Examples: include 'app', 'lib'
+            #           include 'services:api', 'services:impl'
+            import re
+
+            includes = re.findall(r"include\s+['\"]([^'\"]+)['\"]", content)
```
brovatten: Not sure, what do you think about gemini here?

```
Impact: This fails for common Gradle patterns like:
include("app") (Kotlin DSL)
include 'module1', 'module2' (multiple modules on one line)
Dynamic includes (common in large Gradle builds)
Fix: Consider a more robust parser or, better yet, rely on JDTLS's own ability to discover projects once given the root.
```

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+        Args:
+            timeout: Maximum time to wait in seconds (default: 5 minutes)
+
+        Raises:
+            TimeoutError: If import doesn't complete within timeout
+        """
+        logger.info("Waiting for Java project import to complete...")
+        start = time.time()
+        last_log = start
+
```
brovatten, line 276: If JDTLS fails to import the project (e.g., due to a broken pom.xml), proceeding with the analysis will result in empty or highly inaccurate data (missing references, broken call graphs). Perhaps mark as failed if core  imports fail

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+    def _get_package_name(self, file_path: Path) -> str:
+        """
+        Extract package name from Java file.
+
+        Overrides the base implementation to use Java package declarations.
+        """
+        try:
+            # Try to read package declaration from file
+            content = file_path.read_text(encoding="utf-8")
```
brovatten: It can sometimes use ISO-8859-1, use `errors="replace"`

static_analyzer/lsp_client/client.py
```diff
@@ -175,6 +177,23 @@ def _wait_for_response(self, message_id: int, timeout: int = 60):
...
+        Process a notification from the LSP server.
+
+        Subclasses can override handle_notification() to process specific notifications.
+        This method extracts the method and params and calls the subclass handler if available.
+        """
+        method = notification.get("method", "")
+        params = notification.get("params", {})
+
+        # Call subclass handler if it exists
+        if hasattr(self, "handle_notification"):
```
ivanmilevtues: hmm, I really don't like this :D. It feels almost as we should have an LSPClient as abstract class and have the handle_notification as @abstractmethod and then have a GenericLSPClient which is kinda of empty.

I am saying it because I guarantee you that I will always miss to implement this in subclasses. And this already raised a question for me why is it that only the java_client needs it? Did we forget to implement it in the others? idk, but if it is an abstract method we have to implement it and add a comment: skipped not needed for TypeScript.

static_analyzer/lsp_client/client.py
```diff
@@ -417,11 +436,17 @@ def build_static_analysis(self) -> dict:
                 for future in as_completed(future_to_file):
                     file_path = future_to_file[future]
                     try:
-                        result = future.result()
+                        # Add timeout to prevent infinite blocking if worker thread hangs
+                        # 5 minutes should be more than enough for any single file
+                        result = future.result(timeout=300)
```
ivanmilevtues, line 476: Nice

static_analyzer/lsp_client/java_client.py
ivanmilevtues: legit comment.

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+    def start(self):
+        """Start the JDTLS server with proper command construction."""
+        # Create workspace directory
+        self.workspace_dir = Path(tempfile.mkdtemp(prefix="jdtls-workspace-"))
+        self.temp_workspace = True
+
+        # Find Java 21+
+        self.java_home = find_java_21_or_later()
+        if self.java_home is None:
+            raise RuntimeError("Java 21+ required to run JDTLS. Please install JDK 21 or later.")
```
ivanmilevtues: Make sure that this gets to the user somehow, I think atm in VSCode we don't have such mechanism, but with integrations for legit enterprise technologies I think it will be inevitable to have complex setups where we need to communicate errors with the users.

ivanmilevtues: This can be added as a new linear task, and you can resolve the comment :)

static_analyzer/lsp_client/java_client.py
ivanmilevtues: Hmm unsure, as the eclipse will scan the full project it won't follow the gitignore.

ivanmilevtues: I would rather include all JVM related files as maven/gradle will build with them -> meaning that proly the LSP will detect as well so also add *.kt, *.groovy

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+        elif file_count < 5000:
+            return "6G"
+        else:
+            return "8G"
+
+    def _initialize(self):
+        """Performs the LSP initialization handshake with JDTLS-specific options."""
+        logger.info(f"Initializing JDTLS for {self.language_id}...")
+
+        import os
```
ivanmilevtues: can be a global import.

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
...
+                self.jdtls_root = Path(jdtls_root_str)
+            else:
+                self.jdtls_root = None
+
+        # Initialize base LSPClient
+        super().__init__(project_path, language, ignore_manager)
+
+        # Track import status
+        self.import_complete = False
+        self.import_errors: List[str] = []
```
ivanmilevtues: Reprompt ClaudeCode to use list[str] and also instead of Optional[Path] to have the None | Path. As we discussed earlier, now this is something that can be nice to be added to our claude.md (Alex and me can reference it in our tooling). Further this would be nice to add to codeboarding.md if you see what I mean :D

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,489 @@
+"""
+Java LSP client using Eclipse JDT Language Server.
+"""
+
+import time
+import tempfile
+import shutil
+from pathlib import Path
+from typing import Dict, List, Optional
```
ivanmilevtues: Reprompt so we don't have these, we discuseed to use the dict/list/None |

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
+"""
```
ivanmilevtues: Remove this file comment :D, it gives me Linux kernel vibe and I know that Claude loves it which makes me not like it XDD

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+import subprocess
+import re
+import platform
+from pathlib import Path
+from typing import Optional, List
+import logging
+
+logger = logging.getLogger(__name__)
```
ivanmilevtues, line 9: I saw Alex's comments around this file. I tend to agree with some of them. I have just one major comment here.

In my experience of working with Java:
- Java setups are painful and complicated
- Every company has their own way of using it - custom choice of JDK/JRE env, custom choice of maven version (custom maven i.e. atlassian have their own).

This util is an amazing attempt of automating this, and I think it has place here and is nice for people who are just playing with the extension at home.

This said I would open a new Linear task, which will be to enable a comprehensive setup of those system variables JAVA_HOME, JRE location, Maven location, custom config files and so on for the companies which we work with.

This will be for a later stage and will proly come up through our iterative enhancements with our early adopter customers which are using Java.

static_analyzer/programming_language.py
```diff
@@ -13,6 +13,7 @@ def __init__(
         suffixes: list[str],
         server_commands: list[str] | None = None,
         lsp_server_key: str | None = None,
+        config_extra: dict | None = None,
```
ivanmilevtues: this seems a bit like it can be **config_extra, instead of having it as dict | None = None.

Very minor comment I have been in love with *args, **kwargs since I got to code in python and using the opportunity to make it 1 bit nicer :D.

I think it will also make it nicer when calling explicitly with the arguments:
PL(lang=sss,size=2,...,jdtls_root='some value')

It is not that deep, so if you don't like it you can keep it with a param.
In that case just make it:
config_extra: dict = {}

We never end up using the None :)

.gitignore
```diff
@@ -187,5 +187,14 @@ runs/
 static_analyzer/servers/
 static_analyzer/servers/*
 
+# JDTLS workspace data (temporary workspaces created during analysis)
+.jdtls-workspace/
```
ivanmilevtues, line 191: It is nice to have it here, still I think Alex's suggestion of cleaning up is better :)

setup.py
```diff
@@ -156,6 +156,56 @@ def download_file_from_gdrive(file_id, destination):
     return response.status_code == 200
 
 
+def download_jdtls():
```
ivanmilevtues, line 159: I will test if this works on Linux in a minute and update the review accordingly if it doesn't

static_analyzer/lsp_client/java_client.py
Svilen-Stefanov: Sounds good. I'll increase it to 100 and if we see failures, we can increase it further.

static_analyzer/lsp_client/java_client.py
Svilen-Stefanov: I also think we should include all JVM related files - I've improved the comment to make this more explicit.

static_analyzer/java_utils.py
Svilen-Stefanov: We already have that, it's just done further up the call stack:
```
self.java_home = find_java_21_or_later()
if self.java_home is None:
      raise RuntimeError("Java 21+ required to run JDTLS. Please install JDK 21 or later.")
```
Pointer: https://github.com/CodeBoarding/CodeBoarding/pull/137/files#r2685880534

static_analyzer/java_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Returns:
+        Command as list of strings
+
+    Raises:
+        RuntimeError: If Java 21+ not found or JDTLS components missing
+    """
+    # Find Java 21+
+    if java_home is None:
+        java_home = find_java_21_or_later()
```
Svilen-Stefanov, line 198: pointer

static_analyzer/java_utils.py
Svilen-Stefanov: Hmm, I'm not sure this is needed - the timeout here is set to 5sec, which I'm not sure is worth caching. 
Should we cache it if we see an example where this is indeed slow?

static_analyzer/lsp_client/java_client.py
Svilen-Stefanov: I think this is already visible to the user both in the extension and in the CLI. Did you see an issue with the propagation?

static_analyzer/lsp_client/java_client.py
Svilen-Stefanov: Hmm, that's something we should definitely explore further but I generally agree. 
The reason I did it like this is because the current Java analysis behavior is consistent with what we do for all other languages but you're probably right that we should fix it for all of them. 

@ivanmilevtues what do you think, should we just continue and try to do something or do we want to fail early and not provide a potentially highly inaccurate diagram? I'd suggest to create a ticket to fix this separately for all languages, wdyt?

.gitignore
Svilen-Stefanov: We'd definitely delete this when we encounter an exception but when the analysis succeeds, the dir will still be there. We could delete it but I was thinking we would need this if we want to do an iterative analysis to update the project diagram.

static_analyzer/programming_language.py
Svilen-Stefanov: Good point. I generally very much don't like *args, **kwargs but I think it's gonna be much cleaner if this is not a dict but a class with a proper structure. For now, I defined a BaseModel class `LanguageConfig`, which we can extend for language specific configs.

static_analyzer/lsp_client/client.py
```diff
@@ -175,6 +182,53 @@ def _wait_for_response(self, message_id: int, timeout: int = 60):
...
+        method = notification.get("method", "")
+        params = notification.get("params", {})
+
+        # Call subclass handler
+        try:
+            self.handle_notification(method, params)
+        except Exception as e:
+            logger.debug(f"Error in notification handler for {method}: {e}")
+
+    def handle_notification(self, method: str, params: dict):
```
ivanmilevtues, line 204: The @abstractmethod on top of the method is missing.

static_analyzer/lsp_client/java_client.py
```diff
@@ -0,0 +1,513 @@
...
+"""
+
+import logging
+import os
+import shutil
+import tempfile
+import time
+from pathlib import Path
+
+from .client import LSPClient
```
ivanmilevtues: I have to be honest, I don't particularly like this style of relative imports as it might cause problems when using refactor features in the IDEs, also tends to mess with type resolution from time to time.

What is your opinion on the topic, I am more of an absolute imports enthusiast :D

static_analyzer/java_config_scanner.py
```diff
@@ -0,0 +1,209 @@
+"""
+Java project configuration scanner.
+
+Detects Maven, Gradle, and multi-module Java projects.
+"""
```
ivanmilevtues: This comment doesn't seem too useful :D

static_analyzer/java_config_scanner.py
```diff
@@ -0,0 +1,209 @@
...
+import xml.etree.ElementTree as ET
+import logging
+
+from repo_utils.ignore import RepoIgnoreManager
+
+logger = logging.getLogger(__name__)
+
+
+class JavaProjectConfig:
+    """Configuration for a Java project."""
```
ivanmilevtues: This comment also :D, ClaudeCode likes to document

static_analyzer/java_config_scanner.py
```diff
@@ -0,0 +1,209 @@
...
+        return (
+            f"JavaProjectConfig(root={self.root}, "
+            f"build_system={self.build_system}, "
+            f"is_multi_module={self.is_multi_module}, "
+            f"modules={len(self.modules)})"
+        )
+
+
+class JavaConfigScanner:
+    """Scanner for Java project configurations."""
```
ivanmilevtues: this as well ;d

static_analyzer/java_utils.py
brovatten: We call it many times, see e.g. https://github.com/CodeBoarding/CodeBoarding/blob/ac4787bd4aa96adb9b99595cfc602cfaf856cbee/static_analyzer/java_utils.py#L104
so I would try it with and without to time it

static_analyzer/java_utils.py
Svilen-Stefanov: I like the comment as it explains the input and output more clearly so I'd keep it.

static_analyzer/java_config_scanner.py
Svilen-Stefanov: What do you mean - we use this to check if no build system found but Java files exist to create a basic config?

static_analyzer/java_utils.py
Svilen-Stefanov: All calls of this function actually won't be cached because the function is called with different arguments, hence there will be zero cache hits and we will only cache things unnecessarily.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: We can't make this abstract at this point, because we use this class as a default client when we don't recongnise the language:
```
clients.append(LSPClient(language=pl, project_path=repository_path, ignore_manager=ignore_manager))
```

static_analyzer/java_utils.py
brovatten: thanks for clarifying

static_analyzer/lsp_client/client.py
ivanmilevtues: Aah okay. That's what I meant by adding a new client - GenericClient, which jsut implements and does a pass on that method :D

setup.py
ivanmilevtues: It works

static_analyzer/lsp_client/java_client.py
ivanmilevtues: Ultimately I think we should be accurate and working. But I think if I download a product run it and get nothing I will be more disapointed than if I get something bad :D.

Imo we should keep it running pure LLM BUT add a NOTICE in the UI - I will open a Linear task.

[DISCLAIMER: Static analysis failed for your project - the diagram might have innacuracies!]

I think we generally need to have a way to communicated with the UI :D, it might be just a field in the analysis.json or something else idk.

PR #138, opened 2026-01-07
Improve validator agent paths

Svilen-Stefanov:
In all analysis executions I'm seeing consistent failures with the path resolution, the failures are of the type:
```
Component Web Interface has App.tsx as an incorrect reference: 'QName:`App.tsx` FileRef: `ecs_fargate_app/frontend/src/App.tsx`'. App.tsx is INCORRECT reference, there is no such module or function/class/method in the project. Please reconsider by using `getSourceCode` tool to find the correct reference or validate with `readFile` tool if it is a file reference.
Component Web Interface has useAnalyzer.ts as an incorrect reference: 'QName:`useAnalyzer.ts` FileRef: `ecs_fargate_app/frontend/src/hooks/useAnalyzer.ts`'. useAnalyzer.ts is INCORRECT reference, there is no such module or function/class/method in the project. Please reconsider by using `getSourceCode` tool to find the correct reference or validate with `readFile` tool if it is a file reference.
Component API & Security Gateway has auth.module.ts as an incorrect reference: 'QName:`auth.module.ts` FileRef: `ecs_fargate_app/backend/src/modules/auth/auth.module.ts`'. auth.module.ts is INCORRECT reference, there is no such module or function/class/method in the project. Please reconsider by using `getSourceCode` tool to find the correct reference or validate with `readFile` tool if it is a file reference.
```

This leads to a significant slowdown of the LLM agents and waste of tokens as we need to apply the feedback in a second round. This PR fixes these issues by converting the paths to absolute before comparison and after LLM path resolution.

### DIFF
--- a/static_analyzer/reference_resolve_mixin.py
+++ b/static_analyzer/reference_resolve_mixin.py
@@ -145,15 +145,24 @@ class ReferenceResolverMixin(abc.ABC):
                     reference.reference_file = abs_path
                 else:
                     # File might be just a filename - search for it recursively
-                    from pathlib import Path
-
                     matches = list(Path(self.repo_dir).rglob(os.path.basename(reference.reference_file)))
-                    if matches:
-                        # Use the first match (convert to absolute path)
+                    if len(matches) == 1:
+                        # Unambiguous case: exactly one match found
                         reference.reference_file = str(matches[0])
                         logger.info(
-                            f"[Reference Resolution] Found file '{os.path.basename(reference.reference_file)}' at {reference.reference_file}"
+                            f"[Reference Resolution] Found unique file '{os.path.basename(reference.reference_file)}' at {reference.reference_file}"
                         )
+                    else:
+                        # Ambiguous case: multiple files with the same name
+                        match_paths = [str(m) for m in matches]
+                        if len(matches) > 1:
+                            logger.error(
+                                f"[Reference Resolution] Ambiguous file name '{os.path.basename(reference.reference_file)}' "
+                                f"for reference '{qname}'. Found {len(matches)} matches: {match_paths}. "
+                                f"Cannot determine the correct file with certainty."
+                            )
+                        # Clear the reference to signal resolution failure
+                        reference.reference_file = None
 
             if reference.reference_file is None:
                 logger.error(
--- a/tests/static_analyzer/test_reference_resolve_mixin.py
+++ b/tests/static_analyzer/test_reference_resolve_mixin.py
@@ -270,8 +270,8 @@ class TestReferenceResolverMixin(unittest.TestCase):
 
         self.resolver._try_llm_resolution(reference, "nonexistent.Class", ["nonexistent.py"])
 
-        # Should still set the path (even if it doesn't exist)
-        self.assertEqual(reference.reference_file, "nonexistent.py")
+        # The reference should be None to signal resolution failure
+        self.assertIsNone(reference.reference_file)
 
     def test_relative_paths_conversion(self):
         """Test conversion of absolute paths to relative paths"""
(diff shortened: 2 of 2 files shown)


static_analyzer/reference_resolve_mixin.py
```diff
@@ -137,6 +137,24 @@ def _try_llm_resolution(self, reference, qname, assigned_files):
...
 
+            # Normalize the path if it's not absolute and doesn't exist
+            if reference.reference_file and not os.path.isabs(reference.reference_file):
+                # Try as relative path from repo root
+                abs_path = os.path.join(self.repo_dir, reference.reference_file)
+                if os.path.exists(abs_path):
+                    reference.reference_file = abs_path
+                else:
+                    # File might be just a filename - search for it recursively
+                    from pathlib import Path
```
brovatten: How come local import?

static_analyzer/reference_resolve_mixin.py
```diff
@@ -137,6 +137,24 @@ def _try_llm_resolution(self, reference, qname, assigned_files):
...
+
+                    matches = list(Path(self.repo_dir).rglob(os.path.basename(reference.reference_file)))
```
brovatten, line 148: Can this give us multiple matches? `utils.py` for example. This may lead to problems when we pick the first match later on

brovatten (commented):
Really nice you caught this slowdown!

agents/validator_agent.py
```diff
@@ -65,12 +65,21 @@ def validate_references(self, analysis: AnalysisInsights):
                         f"Retry finding the proper source code reference via `getSourceCode` tool or if it is a file reference validate with `readFile`."
                     )
                     continue
+
+                # Normalize reference file to absolute path for comparison
+                # (references may be relative after _relative_paths() conversion)
+                ref_file_abs = (
```
ivanmilevtues, line 71: smart.

static_analyzer/reference_resolve_mixin.py
ivanmilevtues: Agree, just move it to global import, we use pathlib at a lot of locations, so no need to have it locally imported.

static_analyzer/reference_resolve_mixin.py
ivanmilevtues: hmm, that is a good point. Maybe it should be a feedback message from the validator agent:
The reference: "file reference" is ambigiouys there are multiple potential matches "list of matches".

As I am unsure if we can determine which of the files is the actual reference, we can do by common directory with other files in the component, but I don't think this is reliable enough as we might have multiple reference to the same file.

tests/agents/test_validator_agent.py
```diff
@@ -322,6 +323,110 @@ def test_validate_references_file_exists(self, mock_exists, mock_static_init):
         # Should be valid because file exists
         self.assertTrue(result.is_valid)
 
+    @patch("os.path.exists")
+    @patch("agents.agent.CodeBoardingAgent._static_initialize_llm")
+    def test_validate_references_relative_path_comparison(self, mock_static_init, mock_exists):
```
ivanmilevtues, line 328: These tests are nice.

tests/static_analyzer/test_reference_resolve_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+        component = Component(
+            name="TestComponent",
+            description="Test",
+            referenced_source_code=[reference],
+            assigned_files=["test.py"],
+        )
+
+        analysis = AnalysisInsights(description="Test", components=[component], components_relations=[])
+
+        result = self.resolver.fix_source_code_reference_lines(analysis)
```
ivanmilevtues, line 71: Maybe we shouldn't return anything from `fix_source_code_reference_lines` as this result confused me, in the next test we test if it is true :D.

I checked we don't have return types in the mixin, if you don't minding some that would be nice progress to our typing journey

tests/static_analyzer/test_reference_resolve_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+        # Mock get_reference to return a node
+        mock_node = MagicMock()
+        mock_node.file_path = str(self.repo_dir / "test.py")
+        mock_node.line_start = 0
+        mock_node.line_end = 2
+        self.mock_static_analysis.get_reference.return_value = mock_node
+
+        result = self.resolver._try_exact_match(reference, "test.TestClass", "python")
+
+        self.assertTrue(result)
```
ivanmilevtues, line 91: This is confusing as it is True, becuase it is not None :D

ivanmilevtues (approved):
Nicely done, I reallly like this change. Few minor comments, which you can adress and insta merge 🚢

tests/static_analyzer/test_reference_resolve_mixin.py
Svilen-Stefanov: I think here, I'd just fix the below test to keep the blast radius smaller. I'll look into the above and we can fix it in a separate PR.

tests/static_analyzer/test_reference_resolve_mixin.py
Svilen-Stefanov: I'm not sure I understand, `_try_exact_match` returns a bool and in this case it finds an exact match because we define `mock_node.file_path = str(self.repo_dir / "test.py")`.

PR #140, opened 2026-01-14
Fix Vercel API

brovatten:
https://github.com/vercel/ai/issues/10344

The issue is google is using a new signature method. In my mind, OpenAI (that is our python wrapper for vercel) is only using a REST API to send to Vercel.

We should try to test and upgrade `langchain-openai`. Problem there is it breaks our other dependencies.

For now let's downgrade to `gemini-2.5-flash`


Svilen-Stefanov (approved):
LGTM, thank you for the quick fix!

PR #142, opened 2026-01-14
Start using instructor for structure output

ivanmilevtues:
Remove our dependency to trustcall as it is problematic with vercel's API.

Instead of that we are now using instructor.

This comes with updates in the llm config, so that we minimize instantiation logic in the agent. It was getting bloated so now it is back to being leaner.

All tests that were agent related are updated.


agents/agent.py
```diff
@@ -305,34 +256,22 @@ def _parse_response(self, prompt, response, return_type, max_retries=5, attempt=
...
 
 class LargeModelAgent(CodeBoardingAgent):
     def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str):
         agent_model = os.getenv("AGENT_MODEL")
-        llm, model_name = self._static_initialize_llm(model_override=agent_model, is_parsing=False)
-        super().__init__(repo_dir, static_analysis, system_message, llm)
+        llm, model_name, _ = create_llm_from_env(model_override=agent_model, is_parsing=False)
+        instructor_client, instructor_model_name = create_instructor_client_from_env()
+
+        MONITORING_CALLBACK.model_name = model_name
```
ivanmilevtues, line 271: This used to be in the try_parse method.
Unsure why it was there, it seems to be better placed at init or no? @brovatten

agents/llm_config.py
```diff
@@ -51,6 +66,33 @@ def get_resolved_extra_args(self) -> Dict[str, Any]:
...
+
+                openai_client = OpenAI(api_key=api_key, base_url=base_url)
+                return instructor.from_openai(openai_client, mode=instructor.Mode.MD_JSON)
+            return instructor.from_provider(f"openai/{model_name}", api_key=api_key)
+
+        if self.instructor_provider == InstructorProvider.GOOGLE:
+            return instructor.from_provider(
+                f"google/{model_name}",
+                api_key=api_key,
+                mode=instructor.Mode.MD_JSON,
```
ivanmilevtues: This line is critical for Gemini :D

pyproject.toml
```diff
@@ -56,11 +56,11 @@ dependencies = [
     "python-call-graph==2.1.3",
     "pyyaml==6.0.3",
     "regex==2024.11.6",
-    "rich==12.6.0",
+    "rich>=13.7.0",
```
ivanmilevtues: we needed a newer version cause of instructor

agents/agent.py
```diff
@@ -34,12 +29,22 @@
...
-
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str, llm: BaseChatModel):
+    def __init__(
+        self,
+        repo_dir: Path,
+        static_analysis: StaticAnalysisResults,
+        system_message: str,
+        llm: BaseChatModel,
+        model_name: str,
+        instructor_client,
```
Svilen-Stefanov: can we add a type here?

agents/agent.py
```diff
@@ -266,34 +221,30 @@ def _parse_invoke(self, prompt, type):
         assert isinstance(response, str), f"Expected a string as response type got {response}"
         return self._parse_response(prompt, response, type)
 
-    def _parse_response(self, prompt, response, return_type, max_retries=5, attempt=0):
+    def _parse_response(self, prompt, response, return_type: Type[LLMBaseModel], max_retries=5, attempt=0):
```
Svilen-Stefanov: Thank you for adding more typing here, can we further improve this - the other attributes / return type?

agents/agent_responses.py
```diff
@@ -152,6 +195,39 @@ class CFGAnalysisInsights(LLMBaseModel):
     components: list[CFGComponent] = Field(description="List of components identified in the CFG.")
     components_relations: list[Relation] = Field(description="List of relations among the components in the CFG.")
 
+    @field_validator("components", mode="before")
+    @classmethod
+    def parse_stringified_components(cls, v: Any) -> Any:
```
Svilen-Stefanov: How / where is this and `parse_stringified_relations` used?

agents/agent_responses.py
```diff
@@ -1,8 +1,24 @@
...
+import json
 from abc import abstractmethod
-from typing import get_origin, Optional
+from typing import get_origin, get_args, Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
+
+
+def _parse_stringified_json(v: Any) -> Any:
```
Svilen-Stefanov: Can we improve the name of the var here? How about the typing - would it help to create a structured `BaseModel` object - is this even possible supporting that many models?

pyproject.toml
Svilen-Stefanov: Should we specify the exact version we use to prevent ppl having different versions, leading to potential conflicts?

pyproject.toml
```diff
@@ -56,11 +56,11 @@ dependencies = [
...
     "seaborn==0.13.2",
     "sqlparse==0.5.3",
     "tomli==2.2.1",
-    "trustcall==0.0.39",
```
Svilen-Stefanov, line 63: Glad to see this!

agents/llm_config.py
```diff
@@ -43,14 +58,40 @@ def is_active(self) -> bool:
...
                 resolved[k] = value
         return resolved
 
+    def create_instructor_client(self, model_name: str):
+        api_key = self.get_api_key()
+
+        if self.instructor_provider == InstructorProvider.OPENAI:
+            base_url = self.get_resolved_extra_args().get("base_url")
+            if base_url:
+                from openai import OpenAI
```
Svilen-Stefanov: Should we move this to the top?

agents/llm_config.py
```diff
@@ -71,7 +113,8 @@ def get_resolved_extra_args(self) -> Dict[str, Any]:
         chat_class=ChatOpenAI,
         api_key_env="VERCEL_API_KEY",
         agent_model="gemini-2.5-flash",
-        parsing_model="gpt-4o-mini",  # Use OpenAI model for parsing to avoid trustcall compatibility issues with Gemini
+        parsing_model="gemini-2.5-flash",
```
Svilen-Stefanov, line 115: Nice.

agents/agent.py
brovatten: I think either all args should be typed or none

agents/agent.py
```diff
@@ -266,34 +221,30 @@ def _parse_invoke(self, prompt, type):
...
-            parser = PydanticOutputParser(pydantic_object=return_type)
-            return self._try_parse(response, parser)
-        except IndexError as e:
-            # try to parse with the json parser if possible
-            logger.warning(f"IndexError while parsing response (attempt {attempt + 1}/{max_retries}): {e}")
+            extraction_prompt = return_type.extractor_str() + response
+            create_kwargs = {
+                "response_model": return_type,
+                "messages": [{"role": "user", "content": extraction_prompt}],
+                "max_retries": 2,
```
brovatten, line 242: How long time does each extraction take? I would maybe log the time it takes. 
With this I would consider also putting the max retries higher.

agents/agent_responses.py
ivanmilevtues: It is used by Instructor API when it parses the LLM Outputs

agents/llm_config.py
```diff
@@ -43,14 +58,40 @@ def is_active(self) -> bool:
...
         for k, v in self.extra_args.items():
             value = v() if callable(v) else v
             if value is not None:
                 resolved[k] = value
         return resolved
 
+    def create_instructor_client(self, model_name: str):
+        api_key = self.get_api_key()
+
+        if self.instructor_provider == InstructorProvider.OPENAI:
```
brovatten, line 73: Did you try to run this for Anthropic and Cerebras?

brovatten (commented):
I wonder if you checked the implications of time and cost of the new parser. `Trustcall` is nice that it patches by making the llm using operations like `add`, `replace`, etc. whereas `Instructor` regenerates the schema with an LLM. This could be costly/time consuming if we are structuring a long output. 

I would try this/think twice before merging as it trade-offs my responsibility area: speed :D adding vercel compatability may in the end slow it down.

agents/llm_config.py
ivanmilevtues: Cerebas I haven't Anthropic works. I think Cerebras is too nieche to properly support atm.

Anthropic, Google and OpenAI we should though!

agents/agent.py
ivanmilevtues: ============================================================
Model: gemini-2.5-flash
============================================================

📦 Instructor:
  Instructor run 1: 5.500s
  Instructor run 2: 10.355s
  Instructor run 3: 7.668s
  Avg: 7.841s | Min: 5.500s | Max: 10.355s

🤝 Trustcall:
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1768473653.683386 5971869 alts_credentials.cc:93] ALTS creds ignored. Not running on GCP and untrusted ALTS is not enabled.
Key '$defs' is not supported in schema, ignoring
  Trustcall run 1: 10.315s
Key '$defs' is not supported in schema, ignoring
  Trustcall run 2: 9.644s
Key '$defs' is not supported in schema, ignoring
  Trustcall run 3: 8.689s
  Avg: 9.549s | Min: 8.689s | Max: 10.315s

📊 Comparison:
  Instructor is 1.708s faster (17.9% improvement)

✅ Result Validation:
  Instructor: 5 components, 5 relations
  Trustcall: 5 components, 5 relations

============================================================
Model: gemini-2.0-flash-lite
============================================================

📦 Instructor:
  Instructor run 1: 5.207s
  Instructor run 2: 5.228s
  Instructor run 3: 5.397s
  Avg: 5.277s | Min: 5.207s | Max: 5.397s

🤝 Trustcall:
E0000 00:00:1768473698.216736 5971869 alts_credentials.cc:93] ALTS creds ignored. Not running on GCP and untrusted ALTS is not enabled.
Key '$defs' is not supported in schema, ignoring
  Trustcall run 1: 5.312s
Key '$defs' is not supported in schema, ignoring
  Trustcall run 2: 5.532s
Key '$defs' is not supported in schema, ignoring
  Trustcall run 3: 5.156s
  Avg: 5.334s | Min: 5.156s | Max: 5.532s

📊 Comparison:
  Instructor is 0.056s faster (1.1% improvement)

✅ Result Validation:
  Instructor: 5 components, 5 relations
  Trustcall: 5 components, 5 relations

============================================================
Benchmark complete!

Turns out it is faster. I tried with a realistic scenario where I had the prompt for an AnalysisResultInsight (So a long analysis - real one)

PR #143, opened 2026-01-15
Fix diff filtering to consider other languages as well

Svilen-Stefanov:
We shouldn't be checking only for Python files when checking what has changed in the analysis. 
This PR fixes this and first checks what are the supported extensions based on the static analysis and falls back to some default extensions otherwise.


agents/constants.py
```diff
@@ -0,0 +1,15 @@
+"""Constants for agents, including language support configuration."""
+
+# Fallback extensions for supported languages if static analysis is unavailable
```
brovatten: Can you give an example when this is unavailable?

agents/constants.py
```diff
@@ -0,0 +1,15 @@
...
+# IMPORTANT: Update this constant when adding new language support
```
brovatten: This is good for the LLM to know!

agents/diff_analyzer.py
```diff
@@ -42,6 +53,33 @@ def get_analysis(self) -> AnalysisInsights:
         with open(analysis_file_path, "r") as analysis_file:
             return AnalysisInsights.model_validate_json(analysis_file.read())
 
+    def _get_supported_extensions(self, static_analysis: StaticAnalysisResults) -> set[str]:
+        """
+        Extract supported file extensions from static analysis results.
```
brovatten: Is this docstring really necessary? Same for the other ones

agents/diff_analyzer.py
```diff
@@ -42,6 +53,33 @@ def get_analysis(self) -> AnalysisInsights:
...
+        Args:
+            static_analysis: Static analysis results containing detected languages
+
+        Returns:
+            Set of supported file extensions (e.g., {'.py', '.ts', '.go'})
+        """
+        extensions = set()
+        for lang in static_analysis.programming_languages:
+            if lang.is_supported_lang():  # Only include languages with LSP support
+                extensions.update(lang.suffixes)
```
brovatten: should we include an else here to Warn? Since you're gonna start looking at the logs, I would include logs for weird scenarios. Like here, if the static analysis contains a language we don't support, isn't that strange?

agents/diff_analyzer.py
```diff
@@ -42,6 +53,33 @@ def get_analysis(self) -> AnalysisInsights:
...
+        return extensions
+
+    def _get_fallback_extensions(self) -> set[str]:
+        """
+        Get fallback extensions when static analysis is unavailable or incomplete.
+
+        Returns:
+            Set of default supported file extensions
+        """
+        return SUPPORTED_EXTENSIONS_FALLBACK.copy()
```
brovatten: Good copy!

static_analyzer/analysis_result.py
```diff
@@ -1,9 +1,11 @@
 from static_analyzer.graph import Node, CallGraph
+from static_analyzer.programming_language import ProgrammingLanguage
 
 
 class StaticAnalysisResults:
     def __init__(self):
         self.results = {}
+        self.programming_languages: list[ProgrammingLanguage] = []
```
brovatten: defaulting to a list seems dangerous because of mutable lists in python. rather use None

agents/diff_analyzer.py
```diff
@@ -57,17 +95,26 @@ def get_diff_data(self) -> List[FileChange]:
...
+
+        Returns:
+            List of file changes with supported extensions
         """
         relevant_files = []
         for change in diff:
-            # For now we are checking just python files
-            if change.filename.endswith(".py"):
+            file_ext = Path(change.filename).suffix
+            if file_ext in self.supported_extensions:
```
ivanmilevtues, line 92: good first step.

I still think the iterative thing will have to be a new task. There are lots of new changes i.e. cluster assigns.

agents/constants.py
```diff
@@ -0,0 +1,15 @@
...
+# See test_diff_analyzer_language_sync for validation
+SUPPORTED_EXTENSIONS_FALLBACK = {
```
ivanmilevtues: Instead of having this and all the logic around extensions_fallback. I would suggest you to just use the https://github.com/CodeBoarding/CodeBoarding/blob/00593183f6e04363f7273f4aad27cacba476bb36/utils.py#L57

And you can look how we use it in: repos.CodeBoarding.static_analyzer.scanner.ProjectScanner.scan

agents/diff_analyzer.py
```diff
@@ -42,6 +53,33 @@ def get_analysis(self) -> AnalysisInsights:
         with open(analysis_file_path, "r") as analysis_file:
             return AnalysisInsights.model_validate_json(analysis_file.read())
 
+    def _get_supported_extensions(self, static_analysis: StaticAnalysisResults) -> set[str]:
```
ivanmilevtues: I don't think we need this after addressing my comment to use get_config.

static_analyzer/analysis_result.py
ivanmilevtues: This probably isn't needed.

agents/constants.py
Svilen-Stefanov: So I was thinking of situations where:
- there's a problem with the LSP or
- there is no LSP for the given language 

It's generally a fallback, so anything unexpected is part of the fallback as well.

agents/constants.py
Svilen-Stefanov: Thanks for the suggestion, this will indeed makes things nicer. We just need to make sure the config is correct.

agents/diff_analyzer.py
Svilen-Stefanov: I think it's helpful, I'd keep it.

agents/diff_analyzer.py
Svilen-Stefanov: Yeah, I think there are many thing that could be improved here. I was thinking we could improve them iteratively but after your latest changes, there will be even more things to improve. :D

agents/diff_analyzer.py
```diff
@@ -22,6 +23,10 @@ def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, proje
         super().__init__(repo_dir, static_analysis, get_system_diff_analysis_message())
         self.project_name = project_name
         self.repo_dir = repo_dir
+        self.supported_extensions = self._get_supported_extensions()
+
+        logger.info(f"[DiffAnalyzingAgent] Monitoring changes for extensions: {sorted(self.supported_extensions)}")
```
Svilen-Stefanov, line 28: pointer

agents/diff_analyzer.py
```diff
@@ -42,6 +47,19 @@ def get_analysis(self) -> AnalysisInsights:
         with open(analysis_file_path, "r") as analysis_file:
             return AnalysisInsights.model_validate_json(analysis_file.read())
 
+    def _get_supported_extensions(self) -> set[str]:
```
ivanmilevtues, line 50: That's what im talking about ;)

agents/diff_analyzer.py
```diff
@@ -57,17 +75,30 @@ def get_diff_data(self) -> List[FileChange]:
...
         relevant_files = []
         for change in diff:
-            # For now we are checking just python files
-            if change.filename.endswith(".py"):
+            file_ext = Path(change.filename).suffix
+            if file_ext in self.supported_extensions:
                 relevant_files.append(change)
+            else:
+                logger.debug(
+                    f"[DiffAnalyzingAgent] Skipping unsupported file: {change.filename} (extension: {file_ext})"
```
Svilen-Stefanov: @brovatten @ivanmilevtues what do you think about this - I added it based on @brovatten's suggestion but now that I'm looking at it, I'm wondering if we could have a lot of logs if we skip multiple files. 
Wouldn't saying that we only look at these files be enough: https://github.com/CodeBoarding/CodeBoarding/pull/143/files#r2693885885

ivanmilevtues: I think you are right, I would rather collect all of them and put them in a single message. Skipped files: [bulk files :D]

ivanmilevtues (approved):

agents/diff_analyzer.py
Svilen-Stefanov: I improved it with some more sophisticated logging.

uv.lock
```diff
@@ -538,7 +538,7 @@ requires-dist = [
     { name = "python-call-graph", specifier = "==2.1.3" },
     { name = "pyyaml", specifier = "==6.0.3" },
     { name = "regex", specifier = "==2024.11.6" },
-    { name = "rich", specifier = ">=13.7.0" },
+    { name = "rich", specifier = "==14.2.0" },
```
Svilen-Stefanov, line 542: @ivanmilevtues I'm adding this here because while testing your recently merged PR, I reallize this is breaking main.

PR #146, opened 2026-01-17
Evaluate accuracy

brovatten:
Hi guys,

This is without the data (ideally atleast 5 datapoints), let's add that as a last step.

This adds a new accuracy evaluation that compares generated `analysis.json` diagrams against a curated ground-truth dataset using an LLM-as-judge approach.

It slightly refactors how we stored the definitions. I did this since the accuracy benchmark became too big, and the definitions folder was starting to get cluttered. Further, it made sense to separate the configs to eval specific ones.

We score with the LLM based on three criterias, and we ask it to provide the reasoning for these. To track across checkpoints, we save previous results to a json, and load these to append to our table. We produce a graph with this.

I took in your guys' feedback from Slack. This removes also the separation of repos by sizes in the eval.

Have a look and see what you guys think. Thanks 🙌

### DIFF
--- a/evals/utils.py
+++ b/evals/utils.py
@@ -35,7 +35,7 @@ def get_git_commit_short() -> str:
 
 
 def generate_header(title: str, timestamp: str | None = None, extra_lines: list[str] | None = None) -> str:
-    ts = timestamp or datetime.now(timezone.utc).isoformat()
+    ts = timestamp or datetime.now(timezone.cet).isoformat()
     lines = [f"# {title}", "", f"**Generated:** {ts}", ""]
     if extra_lines:
         lines.extend(extra_lines)


evals/tasks/accuracy/similarity_judge.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        self.model_override = model_override
+        self._llm = None
+        self._structured_llm = None
+
+    @property
+    def llm(self):
+        """Lazy initialization of base LLM."""
+        if self._llm is None:
+            self._llm, _ = CodeBoardingAgent._static_initialize_llm(
```
Svilen-Stefanov: Let's make `_static_initialize_llm` public (remove the `_`) or create a proper factory/builder pattern for LLM initialization.

evals/base.py
```diff
@@ -98,7 +101,7 @@ def run_pipeline(
         try:
             result = subprocess.run(
                 cmd,
-                capture_output=True,
+                capture_output=False,
```
Svilen-Stefanov, line 107: Is this intentional? 
Doesn't `capture_output=False` mean that stderr/stdout won't be captured in the result? 

If that's the case, on line 112 when we try to access result.stderr[-500:], stderr will be None, causing an `AttributeError` iiuc.

evals/tasks/accuracy/similarity_judge.py
```diff
@@ -0,0 +1,215 @@
...
+                    backoff *= BACKOFF_MULTIPLIER
+                else:
+                    logger.exception(
+                        "LLM call failed after %d attempts: %s",
+                        MAX_RETRIES,
+                        last_exception,
+                    )
+
+        # Return invalid score on any failure
+        return SimilarityScore()
```
Svilen-Stefanov, line 176: When there's a failure, shouldn't we log it before returning an invalid shore? Currently, it seems to me like the method would silently return an empty SimilarityScore() without re-raising the exception or logging an error. This would make debugging failures difficult.

Overall, I don't think it's a good idea to return an empty object and consider this an error state. The current version serves as both a valid score container and an error state (when score=None), which makes the API confusing imo.

evals/tasks/accuracy/score_plotter.py
```diff
@@ -0,0 +1,312 @@
...
+
+import logging
+from pathlib import Path
+from typing import Any
+
+import matplotlib
+
+# Must set backend before importing pyplot to avoid GUI initialization.
+# "Agg" is a non-interactive backend suitable for generating images without a display.
+matplotlib.use("Agg")
```
Svilen-Stefanov, line 7: Setting this here at module import time could break other code using matplotlib. Using context managers or checking if backend is already set is what seems to be the industry standard. 
My main concern here is that when we import this module, this will actually set `matplotlib` everywhere even if we're not running evals. We currently don't need it elsewhere for important things but it seems a bit dangerous for the future.

evals/tasks/accuracy/score_history.py
```diff
@@ -0,0 +1,236 @@
...
+                continue
+
+            size_label = history.project_sizes.get(project, "unknown")
+            if size_label not in size_scores:
+                size_scores[size_label] = []
+            size_scores[size_label].append(score)
+
+        return {size: sum(scores) / len(scores) if scores else None for size, scores in size_scores.items()}
+
+    def migrate_from_report(self, report_content: str) -> None:
```
Svilen-Stefanov: This doesn't seem to be used anywhere.

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
+# Accuracy Evaluation
+
+**Generated:** 2026-01-17T15:03:16.489323+00:00
+
+**Duration:** N/A
```
Svilen-Stefanov, line 5: Why is this N/A?

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+This report tracks the structural accuracy of CodeBoarding's generated diagrams against a curated ground-truth dataset. We measure how well the agent captures system topology, component relationships, and architectural hierarchy.
+
+## Methodology & Scoring
+
+Our evaluation uses a **"Model-as-a-Judge"** approach to compare the `analysis.json` generated by CodeBoarding against a human-verified `train.json`.
```
Svilen-Stefanov, line 12: Hmm, won't we have multiple `analysis.json` files + subcomponents for each project, requiring multiple `train.json` or whatever we call it? 

I would actually call this `<project_name>_analysis.json` and `<project_name>_groundtruth.json` instead of `train.json`

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+## Level 1: Architecture Overview
+
+### Similarity Scores by Code Size
+
+| Commit | markitdown [M] | pytorch_geometric [L] | Avg | User | Cores |
+|---|---|---|---|---|---|
+| a1c84ff | 5.00 | N/A | 5.00 | brovatten | 10 |
+| a1c84ff | 5.00 | N/A | 5.00 | brovatten | 10 |
```
Svilen-Stefanov, line 36: Why are there 2 lines per commit? Applies to both tables.

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
...
+| 84e830c | N/A | N/A | N/A | brovatten | 10 |
+
+
+*Note: To avoid compounding errors from a faulty level 1 generation, we use our ground truth for level 1, and then generate level 2 from that.*
+
+---
+
+### Score History
+
+![Accuracy Score History](accuracy-score-history.png)
```
Svilen-Stefanov, line 60: This doesn't load to me in VSC - I saw the png, it might need the full path from the repo root to load, not sure.

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+### Judge Reasoning
+
+| Commit | Project | Depth | Score | Node Coverage | Relationship Fidelity | Structural Coherence |
+|--------|---------|-------|-------|---------------|----------------------|---------------------|
+| a1c84ff | markitdown | 1 | 5.00 | Both diagrams include a similar set of components related to document conversion, but they differ in naming and specific implementations of converters. | The relationships between nodes in both diagrams show some alignment in purpose, but the specific connections and types of relationships differ significantly. | While both diagrams depict a conversion framework, the overall architecture and flow differ, with the generated diagram introducing additional layers and components not present in the dataset diagram. |
```
Svilen-Stefanov, line 67: Shouldn't there be a single line per commit here as well?

evals/reports/accuracy-report.md
```diff
@@ -0,0 +1,70 @@
...
+| 84e830c | markitdown | 1 | 5.00 | Both diagrams include a similar set of components related to document conversion, but they differ in naming and specific implementations of converters. | The relationships between nodes in both diagrams show some alignment in purpose, but the specific connections and types of relationships differ significantly. | While both diagrams depict a conversion framework, the overall architecture and flow differ, with the generated diagram introducing additional layers and components not present in the dataset diagram. |
+| 84e830c | markitdown | 1 | 5.00 | Both diagrams include a similar set of components related to document conversion, but they differ in naming and specific implementations of converters. | The relationships between nodes in both diagrams show some alignment in purpose, but the specific connections and types of relationships differ significantly. | While both diagrams depict a conversion framework, the overall architecture and flow differ, with the generated diagram introducing additional layers and components not present in the dataset diagram. |
```
Svilen-Stefanov, line 70: To check this is indeed working properly, it would be great to have a single commit for now and maybe have one after @ivanmilevtues merges his depth analysis PR to see the changes.

Svilen-Stefanov (commented):
I haven't finished the review yet but adding some comments to unblock you. This would be quite helpful, thanks for adding it!

evals/reports/accuracy-report.md
ivanmilevtues: that sounds like a good idea

evals/reports/accuracy-report.md
ivanmilevtues: yea ultimately we should have the commit as the Primary Key of the table :D

evals/tasks/accuracy/dataset_manager.py
```diff
@@ -0,0 +1,188 @@
...
+        manager = DatasetManager(project_root)
+        entries = manager.get_entries(
+            graph_id="markitdown",
+            code_sizes=[CodeSizeCategory.SMALL, CodeSizeCategory.MEDIUM],
+            depth_level=1,
+            limit_per_size=5,
+        )
+    """
+
+    def __init__(self, project_root: Path):
```
ivanmilevtues, line 67: Most of the comments of these funcs are useless. The classic LLM story :D

evals/tasks/accuracy/eval.py
```diff
@@ -0,0 +1,395 @@
...
+        super().__init__(name, output_dir)
+        self.include_system_specs_in_footer = False
+
+        # Initialize components
+        self._judge = DiagramSimilarityJudge()
+        self._dataset_manager = DatasetManager(self.project_root)
+        self._history_store = ScoreHistoryStore(output_dir)
+        self._report_only = False
+
+    # =========================================================================
```
ivanmilevtues: I really dislike the comment blocks :D. If you can just prompt to remove them would be nice for my eyes

evals/tasks/accuracy/eval.py
```diff
@@ -0,0 +1,395 @@
...
+                    )
+                )
+
+        return expanded
+
+    # =========================================================================
+    # Data Loading
+    # =========================================================================
+
+    def _load_analysis_json(self, project_name: str) -> dict[str, Any]:
```
ivanmilevtues, line 109: I would probably expect this to be loaded with the funcs from analysis_json.py maybe that thing should be removed but either way, it should contain such functionality

evals/tasks/accuracy/similarity_judge.py
```diff
@@ -0,0 +1,215 @@
...
+                model_override=self.model_override,
+                is_parsing=True,
+            )
+        return self._llm
+
+    @property
+    def structured_llm(self):
+        """Lazy initialization of structured output LLM."""
+        if self._structured_llm is None:
+            self._structured_llm = self.llm.with_structured_output(
```
ivanmilevtues, line 95: I think it is better to stick to the trustcall way of judging, as the structured output we know for a fact is making the LLM worse.

evals/tasks/accuracy/similarity_judge.py
ivanmilevtues: I would suggest here to have a hardcoded LLM as a judge so it is always the same.

We should also use a more powerful LLM - Gemini-3-Pro/GPT 5.2/Claude Opus etc. - as this won't be running all the time + we need as good of a judge as possible (apart from us ofc, we are the best :D)

evals/tasks/end_to_end/config.py
```diff
@@ -0,0 +1,12 @@
+"""
+Configuration for end-to-end evaluation.
+"""
+
+from evals.schemas import ProjectSpec
+
+# Projects to evaluate for end-to-end pipeline execution
+PROJECTS = [
```
ivanmilevtues, line 8: Quite a few configs now, unsure if we need this many, there is an eval/config.py then this one and then scalability/configy.py and also the static_analysis. 

It is okay just mentioning that I already am unsure if the one that is just in the evals is used.

evals/base.py
```diff
@@ -26,10 +26,12 @@
 
 class BaseEval(ABC):
     def __init__(self, name: str, output_dir: Path):
+        load_dotenv()
```
ivanmilevtues: I added this in my change as well, but added it to the cli.py.

I think it makes more sense to be there as this is a orchestrator class not a config one.

evals/tasks/accuracy/similarity_judge.py
```diff
@@ -0,0 +1,215 @@
...
+        examples: str = "",
+    ) -> str:
+        """Build the scoring prompt with diagram JSON."""
+        return SCORING_PROMPT_TEMPLATE.format(
+            examples=examples,
+            dataset_json=json.dumps(expected, indent=2),
+            generated_json=json.dumps(actual, indent=2),
+        )
+
+    def _invoke_structured(self, prompt: str) -> SimilarityScore:
```
ivanmilevtues, line 122: Following on my other comment I would rather reuse the logic of the invocation we have in the agents, instead of having a structured output LLM

evals/tasks/accuracy/datasets/train.json
```diff
@@ -0,0 +1,155 @@
+[{
```
ivanmilevtues: Maybe we should make this less public i.e. download it from our own GDrive link or something else, but I don't think its place is here in the repo.

evals/tasks/accuracy/similarity_judge.py
brovatten: I added it as a factory to `llm_config.py` instead

evals/base.py
brovatten: Forgot this, thanks! Wanted to look at terminal while running it

evals/reports/accuracy-report.md
brovatten: If you are experimenting yourself, you don't want it to overwrite your last result. In this case, I ran it two times for the same commit .We should remove manually to clean it before pushing

evals/tasks/end_to_end/config.py
brovatten: Not sure what you mean here :D but yes I did split them into multiple configs, the main one was getting messy. and I already began separating them with e.g. dataset so I thought this split made sense

brovatten: i can add it back if @Svilen-Stefanov also prefers one config.py

evals/reports/accuracy-report.md
brovatten: for versioning of the dataset I would not do this. Also since we want to upload it to some files storage, it would be much easier to divide the data into one file per set

evals/base.py
brovatten: Makes sense

evals/tasks/accuracy/similarity_judge.py
brovatten: I "copied" the function now rather than inheriting from CodeBoarding agent. lmk if u disagree

evals/tasks/accuracy/datasets/train.json
brovatten: HF it is!

evals/reports/accuracy-report.md
brovatten: implemented it after the run was already made, hence json didnt contain it

brovatten:
I would question again @ivanmilevtues if the "user" and "cores" are really necessary. It will become crowded as we start adding more repos. The user is implied for the commit, and so is the cores from the user. The cores aren't useful for the scoring, only for the E2E timings. Maybe we can have it in the JSON still but not include it in the report? Here's how it looks like now:

image

PR #147, opened 2026-01-21
Add codeboarding ignore file 

Svilen-Stefanov:
Enables users to ignore files and dirs from their analysis.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -326,6 +326,7 @@ class TestProcessLocalRepository(unittest.TestCase):
             repo_path = Path(temp_dir) / "repo"
             repo_path.mkdir()
             output_dir = Path(temp_dir) / "output"
+            output_dir.mkdir(parents=True, exist_ok=True)
 
             process_local_repository(
                 repo_path=repo_path,
@@ -377,6 +378,7 @@ class TestCopyFiles(unittest.TestCase):
             temp_folder = Path(temp_dir) / "temp"
             temp_folder.mkdir()
             output_dir = Path(temp_dir) / "output"
+            output_dir.mkdir(parents=True, exist_ok=True)
 
             # Create test files
             (temp_folder / "test.md").write_text("# Test")
@@ -389,19 +391,6 @@ class TestCopyFiles(unittest.TestCase):
             self.assertTrue((output_dir / "test.md").exists())
             self.assertTrue((output_dir / "data.json").exists())
             self.assertFalse((output_dir / "ignore.txt").exists())
-
-    def test_copy_files_empty_directory(self):
-        # Test with empty directory
-        with tempfile.TemporaryDirectory() as temp_dir:
-            temp_folder = Path(temp_dir) / "temp"
-            temp_folder.mkdir()
-            output_dir = Path(temp_dir) / "output"
-
-            # Should not raise exception
-            copy_files(temp_folder, output_dir)
-
-            # Output directory should still be created
-            self.assertTrue(output_dir.exists())
 
 
 class TestValidateArguments(unittest.TestCase):


repo_utils/ignore.py
```diff
@@ -58,15 +83,21 @@ def _load_gitignore(self) -> pathspec.PathSpec:
...
-        for d in self.DEFAULT_IGNORED_DIRS:
-            patterns.append(f"{d}/\n")
+        return patterns
 
-        # Add default ignored file patterns (build artifacts, minified files, etc.)
-        for pattern in self.DEFAULT_IGNORED_FILE_PATTERNS:
-            patterns.append(f"{pattern}\n")
+    def _load_codeboardingignore_patterns(self) -> list[str]:
+        """Load and parse .codeboardingignore file from .codeboarding directory if it exists."""
+        codeboardingignore_path = self.repo_root / ".codeboarding" / ".codeboardingignore"
```
ivanmilevtues, line 90: .codeboarding and .codeboardingignore should probably be constants somewhere.

This is an ongoing issue that we have too many random constants.

main.py
```diff
@@ -273,6 +278,8 @@ def copy_files(temp_folder: Path, output_dir: Path):
         output_dir.mkdir(parents=True, exist_ok=True)
         logger.info(f"Created output directory: {output_dir}")
 
+    initialize_codeboardingignore(output_dir)
```
ivanmilevtues: I would rather intialize_codeboardingignore in the main, instead of all 3 methods for analysis.
It might be even a better place to add it in the DiagramGenerator in the pre_analysis step, as it does the setup for the analysis.

Svilen-Stefanov: The main reason why this was added there is because it needs to happen after the `.codeboarding` dir is created, which atm is happening everywhere but I think we could add it to the main path as well.

main.py
```diff
@@ -410,17 +401,27 @@ def main():
 
     should_monitor = args.enable_monitoring or monitoring_enabled()
 
+    output_dir = args.output_dir
+    if is_local:
+        output_dir = output_dir or Path("./analysis")
+
+    if not output_dir.exists():
```
Svilen-Stefanov, line 408: The removals and this is needed so that we can do `initialize_codeboardingignore` here as well.

repo_utils/ignore.py
```diff
@@ -43,10 +57,21 @@ def __init__(self, repo_root: Path):
...
+        patterns = self._load_gitignore_patterns()
+        patterns.extend(self._load_codeboardingignore_patterns())
+
+        # Always add default ignored directories as patterns
+        for d in self.DEFAULT_IGNORED_DIRS:
+            patterns.append(f"{d}/\n")
 
-    def _load_gitignore(self) -> pathspec.PathSpec:
+        # Add default ignored file patterns (build artifacts, minified files, etc.)
+        for pattern in self.DEFAULT_IGNORED_FILE_PATTERNS:
```
brovatten, line 69: It seems like this is duplicated in the reload and should_ignore function?

repo_utils/ignore.py
```diff
@@ -58,15 +83,21 @@ def _load_gitignore(self) -> pathspec.PathSpec:
             except Exception as e:
                 logger.warning(f"Failed to read .gitignore at {gitignore_path}: {e}")
 
-        # Always add default ignored directories as patterns
-        for d in self.DEFAULT_IGNORED_DIRS:
```
Svilen-Stefanov, line 62: pointer

repo_utils/ignore.py
Svilen-Stefanov: I only moved this from [here](https://github.com/CodeBoarding/CodeBoarding/pull/147/files#r2721715843), I wasn't sure if there's a reason for the other one and didn't want to break anything for now :D
I can remove this:
```
# in should_ignore
if part in self.DEFAULT_IGNORED_DIRS:
     return True
```

Svilen-Stefanov: We can fix this for the next release, merging as is to reduce the blast radius for this one.

PR #151, opened 2026-01-21
Adding unit test to validate logs outputs

ivanmilevtues:
As we use the logs as communication at the moment, I am adding a test to validate that the structure doesn't change

### DIFF
--- a/tests/agents/test_monitoring.py
+++ b/tests/agents/test_monitoring.py
@@ -95,7 +95,10 @@ class TestMonitoringCallback(unittest.TestCase):
         self.assertEqual(self.callback.stats.output_tokens, 0)
 
     def test_on_llm_end_logs_expected_format(self):
-        # As we use the logs as communication at the moment, I am adding a test to validate that the structure doesn't change
+        """
+        The test ensures the log format stays the same for communication
+        with other processes
+        """
         # Ensure the log format remains stable for downstream parsing
         llm_output = {
             "token_usage": {


tests/agents/test_monitoring.py
```diff
@@ -93,6 +94,31 @@ def test_on_llm_end_no_usage(self):
...
+        }
+        response = LLMResult(generations=[], llm_output=llm_output)
+
+        token = current_step.set("dummyStep")
+        try:
+            self.callback.model_name = "dummyModel"
+            with patch("monitoring.callbacks.logger.info") as mock_info:
+                self.callback.on_llm_end(response)
+
+                expected = (
```
Svilen-Stefanov, line 118: You could add an explanation why this is needed - something like your description in the PR (callers of this executable depend on this string to ...)

Svilen-Stefanov (approved):
One minor suggestion, looks good otherwise. Thank you for adding the test.

tests/agents/test_monitoring.py
```diff
@@ -93,6 +94,32 @@ def test_on_llm_end_no_usage(self):
         self.assertEqual(self.callback.stats.input_tokens, 0)
         self.assertEqual(self.callback.stats.output_tokens, 0)
 
+    def test_on_llm_end_logs_expected_format(self):
+        # As we use the logs as communication at the moment, I am adding a test to validate that the structure doesn't change
+        # Ensure the log format remains stable for downstream parsing
+        llm_output = {
+            "token_usage": {
```
brovatten, line 104: Isn't the namings `input_tokens`, `output_tokens`?
https://github.com/CodeBoarding/CodeBoarding/blob/17ad630e5213bdea21ae96acd69950cde89dc37d/monitoring/writers.py#L105

brovatten (approved):
See my comments. Looks good otherwise! great idea

tests/agents/test_monitoring.py
ivanmilevtues: Double checked it is correct here. What you are refering isn't the corrent thing, this is what the LLMResult is, an internal logic from langchain. It has the format I have added: https://github.com/CodeBoarding/CodeBoarding/blob/d9099124d9958e54cc758ecc1b0e60191dd2cf6a/monitoring/callbacks.py#L113

PR #153, opened 2026-01-22
Interpolate running times

brovatten:
This uses a (too simple but works for now) method of interpolating run time.

It uses a flag so the extension can easily run it and pick up the estimates. Depending on the depth and lines of code, it will imply an estimate that is sent to the extension.


diagram_analysis/estimation.py
```diff
@@ -0,0 +1,59 @@
...
+    for lang, loc in loc_by_language.items():
+        multiplier = multipliers.get(lang.lower(), 1.0)
+        weighted_sum += loc * multiplier
+
+    effective_multiplier = weighted_sum / total_loc
+
+    # Formula: time = a * log10(LOC) + b, where a = 14.8590, b = -43.1970
+    # See branch: estimate-running-times for how interpolation was derived.
+    a, b = 14.8590, -43.1970
+    base_time_minutes = a * math.log10(total_loc) + b
```
ivanmilevtues: Can we name them somehow not a and b :D.
Doesn't feel right this way xDD

diagram_analysis/estimation.py
```diff
@@ -0,0 +1,59 @@
...
+
+    estimated_time_minutes = max(0, base_time_minutes) * effective_multiplier * depth_multiplier
+
+    logger.info(
+        f"Estimated pipeline time: {estimated_time_minutes:.1f} minutes "
+        f"(based on {total_loc:,} LOC, depth level: {depth_level}, effective multiplier: {effective_multiplier:.2f})"
+    )
+
+    if estimate_only:
+        sys.exit(0)
```
ivanmilevtues: This I strongly dislike, the program should always exit via main.

diagram_analysis/estimation.py
```diff
@@ -0,0 +1,59 @@
...
+    # Level 1: 0.5x, Level 2: 1.0x, Level 3: 2.0x
+    depth_multiplier = 1.0
+    if depth_level == 1:
+        depth_multiplier = 0.5
+    elif depth_level == 3:
+        depth_multiplier = 2.0
+
+    estimated_time_minutes = max(0, base_time_minutes) * effective_multiplier * depth_multiplier
+
+    logger.info(
```
ivanmilevtues, line 62: This style of communication is problematic, we are already experiencing problems and have to add specific tests to check if logging is the same I would advice a different style of solution.

diagram_analysis/estimation.py
```diff
@@ -0,0 +1,59 @@
+import json
+import logging
+import math
+import sys
+
+logger = logging.getLogger(__name__)
+
+
+def estimate_pipeline_time(loc_by_language: dict[str, int], depth_level: int, estimate_only: bool = False) -> None:
```
ivanmilevtues: I would rework this function. It takes a depth_level and repository location.

Then you instatiate the Scanner object and get the loc_by_lang:
```python
        scanner = ProjectScanner(self.repo_location)
        loc_by_language = {pl.language: pl.size for pl in scanner.scan()}```

Return the estimate from the function:
````
def estimate_pipeline_time(repo, depth) -> number:```

File: main.py [hunk 5]
```diff
@@ -419,6 +428,7 @@ def main():
             component_name=args.partial_component,
             analysis_name=args.partial_analysis,
             monitoring_enabled=should_monitor,
+            estimate_only=args.estimate_only,
```
ivanmilevtues, line 442: I would much rather have it short and ease already here.
If the argument is present, then do not run the full main. Just return the estimate time, then you print to stdout a json object {'estimated_time': THETime}

This can be captured by the other process and easily parsed. This way we don't have to go through logs and parse them, do not have to worry about log settings and other things.

diagram_analysis/estimation.py
brovatten: Yup this made no sense, it returns now instead.

diagram_analysis/estimation.py
brovatten: Hmmm... I see your point, but I'd rather keep it consistent for now between services. For the others, we have to keep logging or using an API, and cannot return the results in the same fashion

main.py
brovatten: my main pushback is on the logging, see my comment above

diagram_analysis/diagram_generator.py
```diff
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import sys
```
Svilen-Stefanov: not used

diagram_analysis/estimation.py
Svilen-Stefanov: This is my main concern as well, if we add it like this, we should certainly have a test that fails if this ever changes, similar to the test that @ivanmilevtues added for the other log message we use for communication. 

I think there's a very simple way to do this here though - refer to my [other comment](https://github.com/CodeBoarding/CodeBoarding/pull/153#discussion_r2720619010).

main.py
Svilen-Stefanov: Do we need this in `main.py`? Can we have a separate starting point for the estimation so that doesn't need to touch `main.py`? 
This simplifies the communication significantly as well - callers can first call this new entry, returning the estimate, then they call main.

ivanmilevtues: I would keep this in the main.py @Svilen-Stefanov, as if we have a new entry-point that means that we will have to make 2 binaries and to me that doesn't make much sense, it is a bit like --version of most programs.

Callers can still call first the new "entry point", it is just a new flag, and on the communication aspect in the is the same isn't it? Can you give an API example of how it is different proly i don't get it.

diagram_analysis/estimation.py
ivanmilevtues: hmm okay, but at least add a test :D.
I was thinking more on this and proly we should have a pipe for communication and that should be refactored at a later stage. When in pipe mode, we will stream objects i.e. jsons and that should help.

Anyway I think piping is a standart IPC pattern. You can add a test for this and I am fine to continue with it like this.

PR #154, opened 2026-01-23
Refactoring the analysis agents

ivanmilevtues:
As per issues that have started to arise in recent times I refactored the agents to have a simpler logic.

3 main changes in this PR:

1. Now both agents are stateless, so they can be called in parallel easily. The control flow graph is passed as a parameter so that each agent can work with its own subgraph. Generally speakling  I have made the steps to be more similar between the two agents.

2. I have fully removed the validation agent:
- The validation agent is quite out-dated and validates things which we are getting right from the first time most of the time
- MAIN ISSUE: we give feedback without the proper context, keep in mind this isn't ChatGPT chat, memory get's lost, and we ask the model to apply feedback on clusters wihtout the clusters, on relations without the cfg, the only acceptable feedback could be on the files, but that we do now in a completely different way.
- Validation should come in a closer loops, I think a new issue will be open and we will take a closer look on how to revive the validator agent, until then I am just putting it to sleep

3. File assigning happens on 2 steps: First deterministic from the clusters assigned to the component and then for all the files which are not part of those clusters we ask an llm to put them in the existing components.

### DIFF
--- a/static_analyzer/reference_resolve_mixin.py
+++ b/static_analyzer/reference_resolve_mixin.py
@@ -28,6 +28,10 @@ class ReferenceResolverMixin:
                     continue
 
                 self._resolve_single_reference(reference, component.assigned_files)
+
+        # Remove unresolved references
+        self._remove_unresolved_references(analysis)
+
         return self._relative_paths(analysis)
 
     def _resolve_single_reference(self, reference, assigned_files):
@@ -164,6 +168,37 @@ class ReferenceResolverMixin:
             if reference.reference_file is None:
                 logger.error(
                     f"[Reference Resolution] Reference file could not be resolved for {reference.qualified_name} in any language."
+                )
+
+    def _remove_unresolved_references(self, analysis: AnalysisInsights):
+        """Remove references and assigned files that couldn't be resolved to existing files."""
+        for component in analysis.components:
+            # Remove unresolved key_entities
+            original_ref_count = len(component.key_entities)
+            component.key_entities = [
+                ref
+                for ref in component.key_entities
+                if ref.reference_file is not None and os.path.exists(ref.reference_file)
+            ]
+            removed_ref_count = original_ref_count - len(component.key_entities)
+            if removed_ref_count > 0:
+                logger.info(
+                    f"[Reference Resolution] Removed {removed_ref_count} unresolved reference(s) "
+                    f"from component '{component.name}'"
+                )
+
+            # Remove unresolved assigned_files
+            original_file_count = len(component.assigned_files)
+            component.assigned_files = [
+                f
+                for f in component.assigned_files
+                if os.path.exists(os.path.join(self.repo_dir, f)) or os.path.exists(f)
+            ]
+            removed_file_count = original_file_count - len(component.assigned_files)
+            if removed_file_count > 0:
+                logger.info(
+                    f"[Reference Resolution] Removed {removed_file_count} unresolved assigned file(s) "
+                    f"from component '{component.name}'"
                 )
 
     def _relative_paths(self, analysis: AnalysisInsights):
--- a/tests/static_analyzer/test_reference_resolve_mixin.py
+++ b/tests/static_analyzer/test_reference_resolve_mixin.py
@@ -423,6 +423,238 @@ class TestReferenceResolverMixin(unittest.TestCase):
         # Should have attempted both languages before falling back to LLM
         self.assertEqual(self.mock_static_analysis.get_reference.call_count, 2)
 
+    def test_remove_unresolved_references(self):
+        """Test that unresolved references are removed after resolution attempts"""
+        # Create a mix of resolved and unresolved references
+        resolved_ref = SourceCodeReference(
+            qualified_name="test.TestClass",
+            reference_file=str(self.repo_dir / "test.py"),
+            reference_start_line=1,
+            reference_end_line=2,
+        )
+
+        unresolved_ref_none = SourceCodeReference(
+            qualified_name="nonexistent.Class",
+            reference_file=None,  # Never resolved
+            reference_start_line=None,
+            reference_end_line=None,
+        )
+
+        unresolved_ref_invalid = SourceCodeReference(
+            qualified_name="invalid.path",
+            reference_file="nonexistent_file.py",  # Resolved but file doesn't exist
+            reference_start_line=1,
+            reference_end_line=2,
+        )
+
+        component = Component(
+            name="TestComponent",
+            description="Test",
+            key_entities=[resolved_ref, unresolved_ref_none, unresolved_ref_invalid],
+            assigned_files=["test.py"],
+        )
+
+        analysis = AnalysisInsights(description="Test", components=[component], components_relations=[])
+
+        # Call the cleanup method
+        self.resolver._remove_unresolved_references(analysis)
+
+        # Only resolved reference should remain
+        self.assertEqual(len(component.key_entities), 1)
+        self.assertEqual(component.key_entities[0].qualified_name, "test.TestClass")
+        self.assertEqual(component.key_entities[0].reference_file, str(self.repo_dir / "test.py"))
+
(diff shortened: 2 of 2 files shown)


agents/validator_agent.py
```diff
@@ -139,53 +139,19 @@ def validate_component_relations(self, analysis: AnalysisInsights):
...
-                f"Please review the cluster grouping and ensure each cluster is assigned to exactly one component."
-            )
-            logger.warning(f"[ValidatorAgent] {msg}")
-            return ValidationInsights(is_valid=False, additional_info=msg)
-
-        logger.info(f"[ValidatorAgent] All {len(valid_cluster_ids)} cluster IDs are covered in the analysis")
-        return ValidationInsights(is_valid=True, additional_info="All clusters are covered.")
-
-    def run(self, analysis: AnalysisInsights):
+    def run(self, analysis: AnalysisInsights, cluster_results: dict):
```
Svilen-Stefanov: This seems outdated / unused. 

Iiuc, you removed the requirement that all clusters should be part of source_cluster_ids and replaced it by assigning all unassigned files. Conceptually what this means is that the LLM would group the main clusters into new ones and for the ones that it can't decide what to do with them, the agent will look over their files and decide per file which of the new clusters it belongs to.
I think this generally makes sense. That is, I could imagine situations where this makes more sense then assigning all files of an unassigned cluster to a component. This makes me wonder though if this could also be true for some of the main clusters. 
I guess we could use the evals later to evaluate assumptions of this sort and document their results.

agents/agent_responses.py
```diff
@@ -87,28 +87,32 @@ def llm_str(self):
         return f"({self.src_name}, {self.relation}, {self.dst_name})"
 
 
-class ClusterComponent(LLMBaseModel):
-    """A component derived from cluster analysis."""
+class ClustersComponent(LLMBaseModel):
+    """A grouped component from cluster analysis - may contain multiple clusters."""
 
-    cluster_id: int = Field(description="The cluster ID from the CFG analysis (e.g., 1, 2, 3)")
-    name: str = Field(description="Descriptive name for this cluster/component (2-4 words)")
```
Svilen-Stefanov, line 94: Is there a specific reason why you decided to remove the name or is this an oversight?

agents/abstraction_agent.py
```diff
@@ -97,23 +97,24 @@ def apply_feedback(self, analysis: AnalysisInsights, feedback: ValidationInsight
...
-        self.classify_files(analysis)
-
+        # Build full cluster results dict for all languages
+        cluster_results = {}
+        for lang in self.static_analysis.get_languages():
+            cfg = self.static_analysis.get_cfg(lang)
+            cluster_results[lang] = cfg.cluster()
+
+        # Step 1: Group related clusters together into logical components
+        cluster_analysis = self.step_clusters_grouping()
```
Svilen-Stefanov, line 105: The above is almost entirely done in this method as part of `_build_cluster_string`. 
One thing I find a bit hidden is that we need to serialize and deserialize at least three times because we make separate LLM calls as part of `step_clusters_grouping `, `step_final_analysis` and `classify_files`.  
Should we make it more obvious what is deterministic and what is not?

agents/cluster_methods_mixin.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
 from pathlib import Path
+from typing import Dict
```
Svilen-Stefanov: remove

agents/cluster_methods_mixin.py
```diff
@@ -75,18 +74,22 @@ def _build_cluster_string(self, programming_langs: list[str], cluster_ids: set[i
 
         return "".join(cluster_lines)
 
-    def _assign_files_to_component(self, component: Component) -> None:
+    def _assign_files_to_component(self, component: Component, cluster_results: Dict[str, ClusterResult]) -> None:
```
Svilen-Stefanov: nit: this smells a bit - we need to pass something here to pass it down.

agents/cluster_methods_mixin.py
```diff
@@ -169,25 +164,27 @@ def _ensure_unique_key_entities(self, analysis: AnalysisInsights):
...
-            analysis: The analysis to validate
-            valid_cluster_ids: Optional set of valid IDs. If None, fetches from static analysis.
+            analysis: The analysis to sanitize
+            valid_cluster_ids: Optional set of valid IDs. If None, derives from cluster_results.
+            cluster_results: Dict mapping language -> ClusterResult. Required if valid_cluster_ids is None.
         """
         if valid_cluster_ids is None:
-            valid_cluster_ids = self._get_valid_cluster_ids()
+            if cluster_results is None:
+                raise ValueError("Must provide either valid_cluster_ids or cluster_results")
```
Svilen-Stefanov: I see this is raising an exception but we don't catch it, is this intended?

agents/details_agent.py
```diff
@@ -165,10 +172,39 @@ def apply_feedback(self, analysis: AnalysisInsights, feedback: ValidationInsight
...
 
     def run(self, component: Component):
-        logger.info(f"Processing component: {component.name}")
-        self.step_subcfg(component)
-        self.step_cfg(component)
-        self.step_enhance_structure(component)
-        analysis = self.step_analysis(component)
-        self._validate_cluster_ids(analysis)
-        return self.fix_source_code_reference_lines(analysis)
+        """
```
Svilen-Stefanov, line 112: I see a lot of overlap here with the`abstraction_agent.py` (potentially problematic in the future - can get out of sync). Do we want to keep them the processes in sync and how should we do it without overcomplicating everything?

E.g if we find a problem in `step_final_analysis`, we can now miss that we need to update it in both places. 
In `step_final_analysis` it seems like we only need it for the prompt input, I suppose it's similar for other places.

diagram_analysis/diagram_generator.py
```diff
@@ -91,8 +91,8 @@ def process_component(self, component):
                 update_insight = ValidationInsights(is_valid=False, additional_info=update_analysis.feedback)
                 analysis = self.details_agent.apply_feedback(analysis, update_insight)
             else:
-                analysis = self.details_agent.run(component)
-                feedback = self.validator_agent.run(analysis)
+                analysis, subgraph_cluster_results = self.details_agent.run(component)
```
Svilen-Stefanov, line 91: If `subgraph_cluster_results ` is not needed for the `validator_agent`, we should remove it from the return if that's the only place where we would have needed it.

diagram_analysis/diagram_generator.py
```diff
@@ -229,15 +227,22 @@ def generate_analysis(self):
...
+
             if 4 < update_analysis.update_degree < 8:
                 # This is feedback from the diff analyzer, we need to apply it to the abstraction agent
                 update_insight = ValidationInsights(is_valid=False, additional_info=update_analysis.feedback)
                 analysis = self.abstraction_agent.apply_feedback(
                     self.diff_analyzer_agent.get_analysis(), update_insight
                 )
+                # Need to build cluster results for validation
+                cluster_results = {}
+                for lang in self.abstraction_agent.static_analysis.get_languages():
```
Svilen-Stefanov: I see we use this in two places, should we have helpers for common static analysis patterns?

agents/agent_responses.py
ivanmilevtues: Yes, the reasoning is that I didn't want to itroduce bias on the names of the components, but wanted to rather make the LLM focus mostly on coming up what the constraints/reasons for creating the cluster is. As later down the pipeline this can result into another merge and I noticed that with he name it named the clusters X & Y component. To avoid this I removed the name.

agents/cluster_methods_mixin.py
ivanmilevtues: Yes, this shoudl in theory happen never, however now that you say it. It might happen when we have no static analysis result. I will change it log error for now.

PR #155, opened 2026-01-24
Speed up static analysis and ignore more patterns

Svilen-Stefanov:
I tested this by comparing the static analysis results of [mockito](https://github.com/mockito/mockito) before and after this change and making sure that the response is the same. 
I didn't do proper benchmarking and just ran this a couple of times with the changes prior to the change, and the changes after the change. I did all my experiments on low power mode.
On avg, we saved ~6min to analyse it with the new approach (~500 files being analysed). The new analysis needs ~2min to analyze them. 

As mockito uses gradle, I also analysed [JUnit4](https://github.com/junit-team/junit4), which uses maven to make sure I didn't break anything.

### DIFF
--- a/static_analyzer/lsp_client/client.py
+++ b/static_analyzer/lsp_client/client.py
@@ -264,32 +264,16 @@ class LSPClient(ABC):
         response = self._wait_for_response(req_id)
         return response.get("result", [])
 
-    def _prepare_call_hierarchy(self, file_uri: str, line: int, character: int) -> list:
-        """Prepares a call hierarchy at a specific location."""
-        params = {"textDocument": {"uri": file_uri}, "position": {"line": line, "character": character}}
-        req_id = self._send_request("textDocument/prepareCallHierarchy", params)
-        response = self._wait_for_response(req_id)
-        return response.get("result", [])
-
-    def _get_incoming_calls(self, item: dict) -> list:
-        """Gets incoming calls for a call hierarchy item."""
-        req_id = self._send_request("callHierarchy/incomingCalls", {"item": item})
-        response = self._wait_for_response(req_id)
-        return response.get("result", [])
-
-    def _get_outgoing_calls(self, item: dict) -> list:
-        """Gets outgoing calls for a call hierarchy item."""
-        req_id = self._send_request("callHierarchy/outgoingCalls", {"item": item})
-        response = self._wait_for_response(req_id)
-        return response.get("result", [])
-
-    def _prepare_call_hierarchy_batch(self, file_uri: str, positions: list[dict]) -> dict[int, list]:
+    def _prepare_call_hierarchy_batch(
+        self, file_uri: str, positions: list[dict], batch_timeout: int = 120
+    ) -> dict[int, list]:
         """
         Batch prepare call hierarchy for multiple positions.
 
         Args:
             file_uri: The file URI
             positions: List of position dicts with 'line' and 'character' keys
+            batch_timeout: Total timeout for the entire batch in seconds (default 120s)
 
         Returns:
             Dict mapping position index to call hierarchy items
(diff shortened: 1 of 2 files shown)


repo_utils/ignore.py
```diff
@@ -42,36 +28,43 @@ class RepoIgnoreManager:
...
     DEFAULT_IGNORED_FILE_PATTERNS = [
         "*.bundle.js",  # Webpack/bundler output
         "*.bundle.js.map",  # Source maps for bundles
         "*.min.js",  # Minified JavaScript
         "*.min.css",  # Minified CSS
         "*.chunk.js",  # Code-split chunks
         "*.chunk.js.map",  # Source maps for chunks
+        # Test files across multiple languages
+        "*test.py",  # Python test files
+        "*tests.py",  # Python test package
```
ivanmilevtues: I suppose we can do *test*, as this probably doesn't capture files which are named test_lsp_client.py
We can also take advantage of the config we have `utils.get_config` as we have all the file extensions we support.

static_analyzer/lsp_client/client.py
```diff
@@ -610,70 +691,88 @@ def _analyze_single_file(self, file_path: Path, all_classes: list[dict]) -> File
...
-                                            (caller_qualified_name, current_qualified_name)
-                                        )
-                                        logger.debug(
-                                            f"Incoming call: {caller_qualified_name} -> {current_qualified_name}"
-                                        )
-                                except Exception as e:
-                                    logger.debug(f"Error processing incoming call: {e}")
-                    except Exception as e:
-                        logger.debug(f"Error getting incoming calls: {e}")
+            # OPTIMIZED: Batch prepare call hierarchy for all functions at once
```
ivanmilevtues: Such comments might be dangerous for when we are optimizing the thing again as it might make the LLM to skip over it.

repo_utils/ignore.py
Svilen-Stefanov: I've removed these for now. I think we should have a better way to ignore tests - e.g by ignoring the entire test dir. Now that we have the codeboarding_ignore, we can also let the user define these if they don't want to add them to their analysis.

static_analyzer/lsp_client/client.py
```diff
@@ -610,72 +691,90 @@ def _analyze_single_file(self, file_path: Path, all_classes: list[dict]) -> File
...
+                                                result.call_relationships.append(
+                                                    (current_qualified_name, callee_qualified_name)
+                                                )
+                                                logger.debug(
+                                                    f"Outgoing call: {current_qualified_name} -> {callee_qualified_name}"
+                                                )
+                                        except Exception as e:
+                                            logger.debug(f"Error processing outgoing call: {e}")
+                            except Exception as e:
+                                logger.debug(f"Error processing outgoing calls: {e}")
```
brovatten, line 785: Random questions that comes up, when you are checking the logs to look for improvements, do you run it in debug mode or in info mode?

static_analyzer/lsp_client/client.py
```diff
@@ -453,7 +527,14 @@ def build_static_analysis(self) -> dict:
...
         # Get all classes in workspace for hierarchy analysis
-        all_classes = self._get_all_classes_in_workspace()
+        all_symbols = self._retry_workspace_symbol_request(
+            query="",
+            max_attempts=1,
+            request_timeout=360,
+            log_prefix="analysis/workspace/symbol",
+        )
+        # Filter for class symbols (kind 5)
+        all_classes = [s for s in all_symbols if s.get("kind") == 5]
```
brovatten: This 5 needs better justification in the line comment. This is what gemini said:
In Java and TypeScript, Interfaces (Kind 11) are often just as important for class hierarchies and call graphs. By only looking for Kind 5, the "thorough" analysis you aim for will miss half of the architectural structure in many object-oriented projects.

static_analyzer/lsp_client/client.py
```diff
@@ -453,7 +527,14 @@ def build_static_analysis(self) -> dict:
...
         logger.info(f"Found {len(all_classes)} classes in workspace")
 
         cpu_count = os.cpu_count()
```
brovatten, line 576: Gemini says This should default to something since this can give None for some OS'

static_analyzer/lsp_client/client.py
Svilen-Stefanov: I run debug mode when I really need to debug something specific, I mostly run things in info mode otherwise.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: I added kind 11 for now as well - the kind 5 was copied from the previous implementation, I've added a comment explaining this a bit.

PR #156, opened 2026-01-24
Adding deterministic validation after every step

ivanmilevtues:
In this PR I am reintroducing validation.

The problems with our validation agent:
1. We would provide feedback on multiple steps - we cannot expect the llm to be able to do its full analysis based on 2 sentences of feedback
2. We didn't provide the context which was used for the original decision

New solution:
I am introducing validated_invoke in the agent, the validate invoke will invoke the agent with a prompt (same way as we used to do it), after the invocation is done we get a structured output (pydantic object).
Now on this structured output we run all the validators we created (we use only deterministic validator, no llms involved).
If all of those pass, we just return the structured output (pydantic object).
If some of them failed:
We will rerun the original prompt but in a format like:
Produced Output:
The Structured Output
Feedback:
What we found is wrong with it
Original prompt: (so the agent has its original reasoning)

Now in every llm step, we where we are invoking the agent we can provide validators as well as validation context (things we need for the validation).

You might ask yourself why such an explicit way of doing it, yet not in the agent. Alternatives I considered and didn't liked:

1. Validator Object/Agent: To work after every llm call, it will be very tightly coupled with the agent it validates, and we have multiple agents which we want to test (didn't like it, tried implementing it, it was ugly).
2. Decorator: you add decoration around the functions we want to validate, I really romantisized this idea. The problems it has, it is very very difficult to provde stateful context, which our agents need. We cannot afford to use self in the decorator as the agents have to be thread safe (esp. details agent). It is very hard to get the orginal prompt that is used in the llm (for the context) as it is constructed within the function.
3. Just having the validation within the agent. I think it clutters the agent flow and makes it harder to see what exactly are the steps within the agent.

I am quite satisifed with what we have currently, it is slightly bulkier with the ValidationContext, but it is easily extensible and it is very explicit on which funcs are validating which steps (directly in the step) which I think is very nice.

Also all the validators are functions, not objects which helps us stay on track with having stateless agents.

### DIFF
--- a/agents/validation.py
+++ b/agents/validation.py
@@ -12,7 +12,11 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class ValidationContext:
-    """Context for validation functions. Contains all data needed for validation."""
+    """
+    This class is used to provide the necessary context for validating different LLM steps.
+    It encapsulates all relevant information required by validation routines to ensure that each step in the LLM pipeline
+    is checked against the expected criteria.
+    """
 
     cluster_results: dict[str, ClusterResult] = field(default_factory=dict)
     cfg_graphs: dict[str, CallGraph] = field(default_factory=dict)  # For edge checking
@@ -70,7 +74,7 @@ def validate_cluster_coverage(result: ClusterAnalysis, context: ValidationContex
 
 def validate_component_relationships(result: AnalysisInsights, context: ValidationContext) -> ValidationResult:
     """
-    Validate that component relationships have backing edges in the cluster graph.
+    Validate that component relationships have corresponding edges in the cluster graph.
 
     Args:
         result: AnalysisInsights containing components and components_relations


agents/prompts/__init__.py
```diff
@@ -101,4 +103,5 @@ def __getattr__(name: str):
     "META_INFORMATION_PROMPT",
     "FILE_CLASSIFICATION_MESSAGE",
     "UNASSIGNED_FILES_CLASSIFICATION_MESSAGE",
+    "VALIDATION_FEEDBACK_MESSAGE",
```
Svilen-Stefanov, line 89: Now that you removed the `diff_analyzer`, aren't there messages like `get_diff_analysis_message` that are no longer needed? Do we still use old validation messages?

static_analysis_config.yml
```diff
@@ -6,63 +6,43 @@ lsp_servers:
       - --stdio
     languages:
       - python
-    file_extensions:
-      - .py
```
Svilen-Stefanov, line 10: Iirc, we used to use these to filter what files we're looking for in the static analysis. I still see these being referenced in `programming_language.py`. Is this breaking some assumptions we make there? Can you add a test that would verify the behavior?

static_analyzer/graph.py
```diff
@@ -257,20 +257,24 @@ def filter_by_files(self, file_paths: set[str]) -> "CallGraph":
 
         return sub_graph
 
-    def to_cluster_string(self, cluster_ids: Optional[Set[int]] = None) -> str:
+    def to_cluster_string(
+        self, cluster_ids: Optional[Set[int]] = None, cluster_result: Optional[ClusterResult] = None
```
Svilen-Stefanov: Can we remove the use of `Optional` / `Set` as discussed previously? Maybe a good thing to add to `AGENTS.md`

tests/agents/test_abstraction_agent.py
```diff
@@ -108,16 +114,22 @@ def test_step_clusters_grouping_multiple_languages(self, mock_parse_invoke, mock
         mock_response = ClusterAnalysis(
             cluster_components=[],
         )
-        mock_parse_invoke.return_value = mock_response
+        mock_validation_invoke.return_value = mock_response
 
-        result = agent.step_clusters_grouping()
+        # Create mock cluster_results for both languages
+        from static_analyzer.graph import ClusterResult
```
Svilen-Stefanov, line 116: nit: global import instead

agents/agent.py
```diff
@@ -413,17 +460,35 @@ def _classify_unassigned_files_llm(self, analysis, cluster_results: dict) -> Non
...
         prompt = PromptTemplate(
             template=get_unassigned_files_classification_message(), input_variables=["unassigned_files", "components"]
         ).format(unassigned_files="\n".join(unassigned_files), components=components_summary)
 
-        file_classifications = self._parse_invoke(prompt, ComponentFiles)
+        # Get valid component names from the components_summary
+        # Parse component names from the summary (components have format "**Component:** `ComponentName`")
+        valid_component_names = set()
+        for line in components_summary.split("\n"):
+            if line.strip().startswith("**Component:**"):
```
Svilen-Stefanov: I wonder if we can provide the component here instead of the summary to avoid such string manipulations.

agents/validation.py
```diff
@@ -0,0 +1,261 @@
...
+    Args:
+        result: ClusterAnalysis containing cluster_components
+        context: ValidationContext with expected_cluster_ids
+
+    Returns:
+        ValidationResult with feedback for missing clusters
+    """
+    if not context.expected_cluster_ids:
+        logger.warning("[Validation] No expected cluster IDs provided for coverage validation")
+        return ValidationResult(is_valid=True)
```
Svilen-Stefanov, line 50: This means we expect components to have no cluster_ids associated with them? Seems a bit strange but I understand why we do it. Not sure if we want to do it long term.

agents/agent.py
ivanmilevtues: yes we can.

agents/validation.py
ivanmilevtues: Long term probably we don't want to do it, hence I added logging. But this probably means that there were no clusters whatsover, this is done so we don't fail the analysis if the static analysis fails.

static_analysis_config.yml
ivanmilevtues: Absolutely fair point, unsure what happened here :D

agents/prompts/__init__.py
```diff
@@ -22,20 +22,16 @@
     get_system_message,
     get_cluster_grouping_message,
     get_final_analysis_message,
-    get_feedback_message,
```
Svilen-Stefanov, line 25: Now this makes more sense to me

agents/validation.py
```diff
@@ -0,0 +1,261 @@
...
+from static_analyzer.graph import ClusterResult, CallGraph
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ValidationContext:
+    """Context for validation functions. Contains all data needed for validation."""
+
+    cluster_results: dict[str, ClusterResult] = field(default_factory=dict)
```
Svilen-Stefanov, line 21: Can we maybe add a comment here explaining the format here? Or is there a better way to do this / BaseModel?

agents/validation.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Find missing clusters
+    missing_clusters = context.expected_cluster_ids - result_cluster_ids
+
+    if not missing_clusters:
+        logger.info("[Validation] All clusters are represented in the ClusterAnalysis")
+        return ValidationResult(is_valid=True)
+
+    # Build feedback message
+    missing_str = ", ".join(str(cid) for cid in sorted(missing_clusters))
```
Svilen-Stefanov, line 65: This var name is such an oxymoron :D

agents/validation.py
```diff
@@ -0,0 +1,261 @@
...
+        f"Please ensure all clusters are assigned to a component or create new components for them."
+    )
+
+    logger.warning(f"[Validation] Missing clusters: {missing_str}")
+    return ValidationResult(is_valid=False, feedback_messages=[feedback])
+
+
+def validate_component_relationships(result: AnalysisInsights, context: ValidationContext) -> ValidationResult:
+    """
+    Validate that component relationships have backing edges in the cluster graph.
```
Svilen-Stefanov: nit: not sure what's a backing edge, is it a back edge as in pointing to a parent? Without reading the code, I wasn't sure what this is doing.

agents/validation.py
brovatten: 😂

PR #160, opened 2026-01-26
Commit hash as arg

brovatten:
For the eval, and caching the static analysis, we want to be use the same checkpoint of code. So here is the entrypoint with a flag `--commit`.

I set it to only work for remote repositories. The reasoning is that, if the repo is local, it will detach the head of what could be their working repository. This could be dangerous for users. If you guys know a good way to solve this that isn't too difficult, that'd be amazing though ⭐ 

Would be great to have this merged for the evals and the static analysis caching!

edit: I guess we can do what qodo suggests for local runs, if you guys agree

edit: I tried to run this for a certain hash, and then back to the main for markitdown, and it worked!

### DIFF
--- a/main.py
+++ b/main.py
@@ -156,6 +156,7 @@ def generate_docs_remote(
     local_dev: bool = False,
     run_id: str | None = None,
     monitoring_enabled: bool = False,
+    commit_hash: str | None = None,
 ):
     """
     Clone a git repo and generate documentation (backward compatibility wrapper used by local_app).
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -188,13 +191,13 @@ def process_remote_repository(
 
     repo_name = get_repo_name(repo_url)
 
-    # Check cache if enabled
-    if cache_check and caching_enabled() and onboarding_materials_exist(repo_name):
+    # Check cache if enabled (skip if a specific commit is requested)
+    if not commit_hash and cache_check and caching_enabled() and onboarding_materials_exist(repo_name):
```
ivanmilevtues, line 195: This logic is become obsolete here with the onboarding_materials_exist. Nothing to be done, just mentioning it.

PR #172, opened 2026-02-03
Fix static analysis issues

Svilen-Stefanov:
### **User description**
The PR generally fixes the following 2 types of issues:
1. When there are multiple projects for the same language, we analyse all of them but the CFG was overwritten. This is fixed now.
2. Some LSPs actually send requests for configurations etc. right at the beginning. If ignored, it basically skips all other requests we make, making the analysis useless. Now we catch such requests and reply to them with an empty response (we could improve this further later).

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Fix CFG overwriting when analyzing multiple projects by merging instead of replacing

- Handle LSP server-to-client requests to prevent analysis blocking

- Merge class hierarchies, dependencies, and references across multiple subprojects

- Extend source files list instead of overwriting for monorepo support

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Multiple Projects<br/>Same Language"] -->|"Previously Overwritten"| B["Lost Analysis Data"]
  A -->|"Now Merged"| C["Complete Analysis Results"]
  D["LSP Server Requests<br/>workspace/configuration etc."] -->|"Previously Ignored"| E["Analysis Blocked"]
  D -->|"Now Handled"| F["Analysis Continues"]
```

<details><summary><h3>File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>analysis_result.py</strong><dd><code>Merge analysis results instead of overwriting</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

static_analyzer/analysis_result.py

<ul><li>Changed <code>add_class_hierarchy()</code> to merge hierarchies instead of <br>overwriting, supporting dict and list formats<br> <li> Modified <code>add_cfg()</code> to merge CFG nodes and edges from multiple projects <br>instead of replacing<br> <li> Updated <code>add_package_dependencies()</code> to merge dependencies using <br>dict.update()<br> <li> Changed <code>add_references()</code> to merge references by adding to existing <br>dict instead of replacing<br> <li> Modified <code>add_source_files()</code> to extend the list instead of overwriting</ul>

</details>

  </td>
  <td><a href="https://github.com/CodeBoarding/CodeBoarding/pull/172/files#diff-631acff775359e9a36c7b5c1bb69cbcb181978fc46f8f5d33b2fddbe37c1a579">+59/-16</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>client.py</strong><dd><code>Handle LSP server-to-client requests properly</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

static_analyzer/lsp_client/client.py

<ul><li>Added <code>SERVER_REQUEST_METHODS</code> constant defining known LSP <br>server-to-client request types<br> <li> Implemented <code>_handle_server_request()</code> method to handle <br>workspace/configuration, progress, capability, and message requests<br> <li> Added <code>_send_response()</code> method to send JSON-RPC responses back to LSP <br>server<br> <li> Modified <code>_read_messages()</code> to distinguish between server requests (id + <br>method), responses (id only), and notifications (method only)<br> <li> Added detailed logging for server request handling with debug and <br>warning levels</ul>

</details>

  </td>
  <td><a href="https://github.com/CodeBoarding/CodeBoarding/pull/172/files#diff-37cb530b25355b6c06020495b01d893c6f5298e9d5f43b4e3327a8f4d12fa4a9">+82/-2</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tbody></table>

</details>

___

### DIFF
--- a/static_analyzer/lsp_client/client.py
+++ b/static_analyzer/lsp_client/client.py
@@ -127,19 +127,20 @@ class LSPClient(ABC):
             message_id = self._message_id
             self._message_id += 1
 
-        request = {
-            "jsonrpc": "2.0",
-            "id": message_id,
-            "method": method,
-            "params": params,
-        }
+            request = {
+                "jsonrpc": "2.0",
+                "id": message_id,
+                "method": method,
+                "params": params,
+            }
 
-        body = json.dumps(request)
-        message = f"Content-Length: {len(body)}\r\n\r\n{body}"
+            body = json.dumps(request)
+            message = f"Content-Length: {len(body)}\r\n\r\n{body}"
 
-        if self._process and self._process.stdin:
-            self._process.stdin.write(message.encode("utf-8"))
-            self._process.stdin.flush()
+            if self._process and self._process.stdin:
+                self._process.stdin.write(message.encode("utf-8"))
+                self._process.stdin.flush()
+
         return message_id
 
     def _send_notification(self, method: str, params: dict):
@@ -152,9 +153,10 @@ class LSPClient(ABC):
         body = json.dumps(notification)
         message = f"Content-Length: {len(body)}\r\n\r\n{body}"
 
-        if self._process and self._process.stdin:
-            self._process.stdin.write(message.encode("utf-8"))
-            self._process.stdin.flush()
+        with self._lock:
+            if self._process and self._process.stdin:
+                self._process.stdin.write(message.encode("utf-8"))
+                self._process.stdin.flush()
 
     def _read_messages(self):
         """
@@ -266,38 +268,28 @@ class LSPClient(ABC):
         request_id = request.get("id")
         params = request.get("params", {})
 
+        # Per LSP spec, id must be present and is either int or string
+        if request_id is None:
+            logger.warning(f"Received server request without id: {method}")
+            return
+
         if method == "workspace/configuration":
             # Return empty configuration for each requested item
             items = params.get("items", [])
-            result: list[dict] = [{}] * len(items)
-            logger.debug(f"Responding to workspace/configuration request for {len(items)} items")
-            if isinstance(request_id, int):
-                self._send_response(request_id, result)
+            result: list[dict] = [{} for _ in items]
+            self._send_response(request_id, result)
         elif method == "window/workDoneProgress/create":
             # Acknowledge progress token creation
-            logger.debug(f"Acknowledging progress token creation: {params.get('token')}")
-            if isinstance(request_id, int):
-                self._send_response(request_id, None)
+            self._send_response(request_id, None)
         elif method == "client/registerCapability":
             # Acknowledge capability registration
-            logger.debug("Acknowledging capability registration")
-            if isinstance(request_id, int):
-                self._send_response(request_id, None)
+            self._send_response(request_id, None)
         elif method == "window/showMessageRequest":
             # Don't select any action, just acknowledge
-            logger.debug(f"Acknowledging showMessageRequest: {str(params.get('message', ''))[:50]}")
-            if isinstance(request_id, int):
-                self._send_response(request_id, None)
+            self._send_response(request_id, None)
         elif method == "window/showDocument":
             # Acknowledge but indicate we didn't show the document
-            logger.debug("Acknowledging showDocument request")
-            if isinstance(request_id, int):
-                self._send_response(request_id, {"success": False})
-        elif method in self.SERVER_REQUEST_METHODS:
-            # Known method but not explicitly handled - respond with null
-            logger.debug(f"Responding with null to known server request: {method}")
-            if isinstance(request_id, int):
-                self._send_response(request_id, None)
+            self._send_response(request_id, {"success": False})
         else:
             # UNKNOWN server request - log as warning so we can add handling if needed
             logger.warning(
@@ -305,22 +297,29 @@ class LSPClient(ABC):
                 f"Responding with null to unblock server. "
                 f"Consider adding explicit handling for this request type."
             )
-            if isinstance(request_id, int):
-                self._send_response(request_id, None)
+            self._send_response(request_id, None)
 
-    def _send_response(self, request_id: int, result) -> None:
-        """Send a response to a server request."""
+    def _send_response(self, request_id: int | str, result) -> None:
+        """
+        Send a response to a server request.
+
+        Args:
+            request_id: The request ID from the server (int or str per LSP spec)
+            result: The response result to send back to the server
+        """
         response = {
             "jsonrpc": "2.0",
             "id": request_id,
             "result": result,
         }
         body = json.dumps(response)
-        message = f"Content-Length: {len(body)}\r\n\r\n{body}"
+        body_bytes = body.encode("utf-8")
+        header = f"Content-Length: {len(body_bytes)}\r\n\r\n"
 
-        if self._process and self._process.stdin:
-            self._process.stdin.write(message.encode("utf-8"))
-            self._process.stdin.flush()
+        with self._lock:
+            if self._process and self._process.stdin:
+                self._process.stdin.write(header.encode("utf-8") + body_bytes)
+                self._process.stdin.flush()
 
     def _initialize(self):
         """Performs the LSP initialization handshake."""
--- a/tests/static_analyzer/lsp_client/test_lsp_client.py
+++ b/tests/static_analyzer/lsp_client/test_lsp_client.py
@@ -358,6 +358,40 @@ class TestLSPClient(unittest.TestCase):
 
         self.assertEqual(package_name, "module")
 
+    @patch("static_analyzer.lsp_client.client.subprocess.Popen")
+    def test_get_package_name_multiple_root_level(self, mock_popen):
+        # Test that multiple root-level files form separate packages (not false cycles)
+        mock_process = Mock()
+        mock_process.stdin = Mock()
+        mock_process.stdout = Mock()
+        mock_popen.return_value = mock_process
+
+        client = LSPClient(self.project_path, self.mock_language)
+
+        # Create multiple root-level files
+        main_file = self.project_path / "main.py"
+        utils_file = self.project_path / "utils.py"
+        config_file = self.project_path / "config.py"
+
+        main_file.touch()
+        utils_file.touch()
+        config_file.touch()
+
+        # Each should get its own package name based on filename
+        main_package = client._get_package_name(main_file)
+        utils_package = client._get_package_name(utils_file)
+        config_package = client._get_package_name(config_file)
+
+        # They should all be unique (based on stem)
+        self.assertEqual(main_package, "main")
+        self.assertEqual(utils_package, "utils")
+        self.assertEqual(config_package, "config")
+
+        # Verify they are different from each other
+        self.assertNotEqual(main_package, utils_package)
+        self.assertNotEqual(utils_package, config_package)
+        self.assertNotEqual(main_package, config_package)
+
     def test_extract_package_from_import(self):
         self.assertEqual(LSPClient._extract_package_from_import("os.path.join"), "os")
         self.assertEqual(LSPClient._extract_package_from_import("django.http.response"), "django")


static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+        These are distinct from notifications - they have both 'id' and 'method' fields
+        and the server blocks until we respond.
+
+        See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/
+        """
+        method = request.get("method", "")
+        request_id = request.get("id")
+        params = request.get("params", {})
+
+        if method == "workspace/configuration":
```
ivanmilevtues, line 276: Are we doing debug logs by default, this can become a total spam.

static_analyzer/lsp_client/client.py
```diff
@@ -97,7 +110,10 @@ def start(self):
         """Starts the language server process and the message reader thread."""
         logger.info(f"Starting server {' '.join(self.server_start_params)}...")
         self._process = subprocess.Popen(
-            self.server_start_params, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+            self.server_start_params,
```
brovatten, line 113: Was this not black formatted before or why is this changing?

static_analyzer/lsp_client/client.py
```diff
@@ -158,10 +174,18 @@ def _read_messages(self):
                 body = self._process.stdout.read(content_length).decode("utf-8")
                 response = json.loads(body)
 
-                if "id" in response:
+                if "id" in response and "method" in response:
+                    # Server-to-client REQUEST (has both id and method)
+                    # This is distinct from:
+                    # - Responses (have id, no method)
```
brovatten, line 182: Not sure this comments adds that much contexts. If you keep them, I would like to know more

static_analyzer/lsp_client/client.py
brovatten: Add as debug maybe? Codex did the same when I worked on the static analysis

static_analyzer/lsp_client/client.py
```diff
@@ -681,7 +790,10 @@ def _analyze_single_file(self, file_path: Path, all_classes: list[dict]) -> File
 
                                         # Add edge: calling_function -> current_function
                                         result.call_relationships.append(
-                                            (caller_qualified_name, current_qualified_name)
+                                            (
```
brovatten, line 792: Same here, did we change any formatting settings? :D

static_analyzer/lsp_client/client.py
Svilen-Stefanov: No, our default is info, debug logs are not shown normally.

static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+
+    def _send_response(self, request_id: int, result) -> None:
+        """Send a response to a server request."""
+        response = {
+            "jsonrpc": "2.0",
+            "id": request_id,
+            "result": result,
+        }
+        body = json.dumps(response)
+        message = f"Content-Length: {len(body)}\r\n\r\n{body}"
```
brovatten: Claude code: 
len(body) is the character count. LSP requires byte count. For ASCII-only JSON this is equivalent, but if any symbol names or file paths contain non-ASCII characters (common in
  international codebases), this will produce malformed messages. This is a pre-existing bug (all three send methods have it), but _send_response copies the pattern rather than fixing it.

static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+            # Don't select any action, just acknowledge
+            logger.debug(f"Acknowledging showMessageRequest: {str(params.get('message', ''))[:50]}")
+            if isinstance(request_id, int):
+                self._send_response(request_id, None)
+        elif method == "window/showDocument":
+            # Acknowledge but indicate we didn't show the document
+            logger.debug("Acknowledging showDocument request")
+            if isinstance(request_id, int):
+                self._send_response(request_id, {"success": False})
+        elif method in self.SERVER_REQUEST_METHODS:
```
brovatten: Claude code: 
  All five methods in SERVER_REQUEST_METHODS are already handled by the explicit if/elif branches above this. This branch is currently unreachable. It's a defensive pattern for future
  additions, but could be confusing for maintainers who expect it to be reachable.

static_analyzer/analysis_result.py
```diff
@@ -192,14 +228,20 @@ def get_languages(self):
 
     def add_source_files(self, language: str, source_files):
```
brovatten, line 229: Claude code:
  add_source_files can accumulate duplicates

  static_analyzer/analysis_result.py:244:

  self.results[language]["source_files"].extend(source_files)

  Unlike add_references (which uses dict keys for dedup) and add_class_hierarchy (dict update), add_source_files uses extend() with no dedup. If the same file appears in two subprojects,
  it'll be listed twice. Given that the new _ensure_unique_file_assignments in agents suggests duplicate files are a real concern, this seems like an oversight.

static_analyzer/analysis_result.py
```diff
@@ -55,50 +54,87 @@ def __init__(self):
...
+        else:
+            # Merge nodes and edges from the new CFG into the existing one
+            existing_cfg = self.results[language]["cfg"]
+            for node in cfg.nodes.values():
+                existing_cfg.add_node(node)
+            for edge in cfg.edges:
+                try:
+                    existing_cfg.add_edge(edge.get_source(), edge.get_destination())
+                except ValueError:
+                    pass  # Skip duplicate edges
```
brovatten, line 101: Claude code:
  add_edge handles duplicate edges internally by returning early — it never raises for duplicates. The ValueError is raised when a node is missing. Since nodes are added first, this should
  never fire. The comment is misleading; if it does fire, it means a node was referenced in an edge but wasn't in the graph, which would indicate a data integrity issue that's being silently
   swallowed.

tests/static_analyzer/lsp_client/test_lsp_client.py
```diff
@@ -13,35 +13,23 @@
...
-    def test_uri_to_path_unix(self):
-        # Test converting Unix-style file URI to path
-        file_uri = "file:///home/user/project/file.py"
-        result = uri_to_path(file_uri)
-
-        # Result should be a Path object
+    def test_uri_to_path_conversion(self):
+        result = uri_to_path("file:///home/user/project/file.py")
         self.assertIsInstance(result, Path)
         self.assertIn("file.py", str(result))
```
brovatten, line 19: Claude code:
  static_analyzer/lsp_client/client.py:1215-1222: Changing root-level files from "root" to rel_path.stem means every root-level file becomes its own "package." This is a correctness
  improvement (avoids false cycles), but it's a significant behavioral change that could affect downstream consumers of the package dependency graph. The test was updated ("root" →
  "module"), but there's no test verifying the downstream effect when multiple root-level files exist.

static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+            # UNKNOWN server request - log as warning so we can add handling if needed
+            logger.warning(
+                f"Unknown LSP server request received: {method} (id={request_id}). "
+                f"Responding with null to unblock server. "
+                f"Consider adding explicit handling for this request type."
+            )
+            if isinstance(request_id, int):
+                self._send_response(request_id, None)
+
+    def _send_response(self, request_id: int, result) -> None:
```
brovatten: Claude code : 

  _send_response has no locking

  _send_request and _send_notification both write to stdin from the main thread, while _handle_server_request (and thus _send_response) is called from the reader thread. _send_request uses
  self._lock for the message ID increment, but none of the three send methods hold a lock during the stdin.write + flush sequence. If the reader thread writes a response while the main
  thread is writing a request, the two messages could interleave on stdin. This is a pre-existing issue for notifications but is now worse because _send_response actually writes structured
  data that the server parses (unlike notifications which were previously just appended to a list).

static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+        See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/
+        """
+        method = request.get("method", "")
+        request_id = request.get("id")
+        params = request.get("params", {})
+
+        if method == "workspace/configuration":
+            # Return empty configuration for each requested item
+            items = params.get("items", [])
+            result: list[dict] = [{}] * len(items)
```
brovatten: [{} for _ in items] this looks easier

static_analyzer/lsp_client/client.py
```diff
@@ -229,6 +253,75 @@ def handle_notification(self, method: str, params: dict):
...
+            logger.debug(f"Responding to workspace/configuration request for {len(items)} items")
+            if isinstance(request_id, int):
```
brovatten: Claude code:

 Bug: isinstance(request_id, int) silently drops string request IDs

  static_analyzer/lsp_client/client.py:274 and repeated throughout _handle_server_request:

  if isinstance(request_id, int):
      self._send_response(request_id, result)

  The https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage defines id as integer | string. If an LSP server sends a string ID (which is
  valid per spec), the guard silently drops the response — exactly the blocking behavior this PR is trying to fix. The type annotation on _send_response also constrains request_id: int
  unnecessarily.

  This should be if request_id is not None: (or accept int | str).

static_analyzer/analysis_result.py
Svilen-Stefanov: Hm, not sure if we really want to do this. If we identify a file as relevant for multiple languages, maybe it is relevant for both of them, otherwise, we should actually change our config to only look at the files that are needed per language. I think removing duplicates at the analysis layer discards potentially important data, we should remove things during processing later if needed.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: I removed the logs for now but we might want to add them back later. These could help us understand the LSP behaviour better for different languages.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: I don't know exactly, I've started seeing such refactors lately, not sure what's changed. Haven't had time to look into it.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: What is missing here? What the comment says is that there are 3 types of server-client communication:
1. requests (we were not handling this)
2. responses
3. notifications

static_analyzer/analysis_result.py
Svilen-Stefanov: Reasoning: The silent exception swallow is intentional. We need to handle multiple calls to add_cfg() for the same language (monorepo scenario). When merging CFGs from different subprojects, duplicate edges are expected and should be skipped silently. Logging this would spam the logs for normal operation. The ValueError is caught specifically for this use case.

static_analyzer/lsp_client/client.py
brovatten: For me as a reader, ofc having id and no method is distinct from having both. And by reading the code, i realise there are cases where this happens. If this is an edge case, it doesn't tell me why this edge case exists. It also doesn't tell me anything additional.

static_analyzer/lsp_client/client.py
brovatten: Yes, therefore I think a debug makes more sense. It seems like something we wanna track if it goes wrong, but at the same time it fills up the INFO too much. What alternatives are there? Not needed for now

static_analyzer/lsp_client/client.py
Svilen-Stefanov: It's not really an edge case, it's just a communication protocol we never handled so it's a third way of communication. Sure, id and method help you understand they are different, but they don't relly tell us what type of message that is and I think this comments makes it super clear. 
E.g:
- id + no method is a response to a request we made earlier
- id + method is a server request, we don't expect the LSP to talk to us but it did

Without this comment, I think these types of requests are not obvious at all imo.
I actually love the comment :D

PR #174, opened 2026-02-05
Add more stable health status report metrics for unused code

Svilen-Stefanov:
Do we want to remove the previous orphan code analysis completely?

I ran the integration tests to make sure things are not much slower: https://github.com/CodeBoarding/CodeBoarding/actions/runs/21800185135

It seems like the time hasn't changed much. For some reason, for some languages the collected edges are more with after adding the diagnostics (which barely makes any sense), which is why 3 of the tests failed.

### DIFF
--- a/health/checks/unused_code_diagnostics.py
+++ b/health/checks/unused_code_diagnostics.py
@@ -11,8 +11,7 @@ reliable than custom call graph analysis for detecting:
 
 import logging
 from dataclasses import dataclass
-from enum import Enum
-from pathlib import Path
+from enum import StrEnum
 
 from static_analyzer.lsp_client.diagnostics import LSPDiagnostic
 from health.models import (
@@ -26,7 +25,10 @@ from health.models import (
 logger = logging.getLogger(__name__)
 
 
-class DeadCodeCategory(str, Enum):
+ENTITY_NAME_LINE_BREAK = 150
+
+
+class DeadCodeCategory(StrEnum):
     """Categories of dead/unused code detected by LSP diagnostics."""
 
     UNUSED_IMPORT = "unused_import"
@@ -273,8 +275,8 @@ def check_unused_code_diagnostics(
         entities: list[FindingEntity] = []
         for issue in category_issues:
             # Create a descriptive entity name from the message
-            entity_name = f"[{category.value}] {issue.message[:50]}"
-            if len(issue.message) > 50:
+            entity_name = f"[{category.value}] {issue.message[:ENTITY_NAME_LINE_BREAK]}"
+            if len(issue.message) > ENTITY_NAME_LINE_BREAK:
                 entity_name += "..."
 
             entity = FindingEntity(
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -170,17 +170,26 @@ class StaticAnalyzer:
                 results.add_package_dependencies(client.language.language, analysis.get("package_relations", {}))
                 results.add_source_files(client.language.language, analysis.get("source_files", []))
 
-                # Collect diagnostics for health checks (collected during analysis)
-                # Stored on the instance (not in StaticAnalysisResults) since diagnostics
-                # are ephemeral LSP data only consumed by health checks.
-                diagnostics = client.get_collected_diagnostics()
-                if diagnostics:
-                    logger.info(f"Collected {len(diagnostics)} files with diagnostics for {client.language.language}")
-                    total_diags = sum(len(d) for d in diagnostics.values())
-                    logger.info(f"Total diagnostic items: {total_diags}")
-                    self.collected_diagnostics[client.language.language] = diagnostics
+                # Collect diagnostics for health checks.
+                # Prefer diagnostics from the analysis result (populated by incremental cache),
+                # fall back to freshly collected ones from the LSP client.
+                if "diagnostics" in analysis and analysis["diagnostics"]:
+                    cached_diags = analysis["diagnostics"]
+                    logger.info(
+                        f"Using {len(cached_diags)} files with diagnostics from cache for {client.language.language}"
+                    )
+                    self.collected_diagnostics[client.language.language] = cached_diags
                 else:
-                    logger.warning(f"No diagnostics collected for {client.language.language}")
+                    diagnostics = client.get_collected_diagnostics()
+                    if diagnostics:
+                        logger.info(
+                            f"Collected {len(diagnostics)} files with diagnostics for {client.language.language}"
+                        )
+                        total_diags = sum(len(d) for d in diagnostics.values())
+                        logger.info(f"Total diagnostic items: {total_diags}")
+                        self.collected_diagnostics[client.language.language] = diagnostics
+                    else:
+                        logger.warning(f"No diagnostics collected for {client.language.language}")
             except Exception as e:
                 logger.error(f"Error during analysis with {client.language.language}: {e}")
         print(f"Static analysis complete: {results}")
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -140,7 +140,7 @@ class Node:
         LSP servers often report inline callbacks (e.g. `.forEach() callback`,
         `.find() callback`) and anonymous functions (e.g. `<function>`, `<arrow`)
         as separate symbols. These are typically not independently callable and
-        should be excluded from certain health checks like orphan code detection.
+        should be excluded from certain health checks like unused code detection.
         """
         name = self.fully_qualified_name
         return any(pattern in name for pattern in self._CALLBACK_PATTERNS)
--- a/health/config.py
+++ b/health/config.py
@@ -98,7 +98,7 @@ def initialize_health_dir(health_config_dir: Path) -> None:
 
 
 def _load_health_exclude_patterns(health_config_dir: Path) -> list[str]:
-    """Load orphan code exclusion patterns from .healthignore file.
+    """Load health check exclusion patterns from .healthignore file.
 
     Args:
         health_config_dir: Path to the health config directory.
@@ -151,7 +151,7 @@ def load_health_config(health_config_dir: Path | None = None) -> HealthCheckConf
 
     if not config_path.exists():
         logger.debug(f"No health_config.json found at {config_path}, using defaults")
-        return HealthCheckConfig(orphan_exclude_patterns=exclude_patterns)
+        return HealthCheckConfig(health_exclude_patterns=exclude_patterns)
 
     try:
         raw = config_path.read_text(encoding="utf-8")
@@ -164,9 +164,9 @@ def load_health_config(health_config_dir: Path | None = None) -> HealthCheckConf
             else:
                 flat[key] = entry
         config = HealthCheckConfig.model_validate(flat)
-        config.orphan_exclude_patterns = exclude_patterns
+        config.health_exclude_patterns = exclude_patterns
         logger.debug(f"Loaded health config from {config_path}")
         return config
     except Exception as e:
         logger.warning(f"Failed to parse health_config.json at {config_path}: {e}. Using default configuration.")
-        return HealthCheckConfig(orphan_exclude_patterns=exclude_patterns)
+        return HealthCheckConfig(health_exclude_patterns=exclude_patterns)
--- a/health/models.py
+++ b/health/models.py
@@ -172,9 +172,9 @@ class HealthCheckConfig(BaseModel):
         default=50,
         description="Maximum number of circular dependency cycles to include in the report. Range: 10-200. Default: 50.",
     )
-    orphan_exclude_patterns: list[str] = Field(
+    health_exclude_patterns: list[str] = Field(
         default_factory=list,
-        description="Glob patterns for entities excluded from orphan code detection. Managed via .healthignore file.",
+        description="Glob patterns for entities excluded from health checks. Managed via .healthignore file.",
     )
     instability_high: float = Field(
         default=0.8,
--- a/health/runner.py
+++ b/health/runner.py
@@ -1,3 +1,4 @@
+import fnmatch
 import logging
 import os
 from datetime import datetime, timezone
@@ -27,6 +28,38 @@ from static_analyzer.analysis_result import StaticAnalysisResults
 logger = logging.getLogger(__name__)
 
 CheckSummaryList = list[StandardCheckSummary | CircularDependencyCheck]
+
+
+def _matches_exclude_pattern(entity_name: str, file_path: str | None, patterns: list[str]) -> bool:
+    """Check if an entity or file path matches any of the exclusion patterns."""
+    for pattern in patterns:
+        if fnmatch.fnmatch(entity_name, pattern):
+            return True
+        if file_path and fnmatch.fnmatch(file_path, pattern):
+            return True
+    return False
+
+
+def _apply_exclude_patterns(summaries: CheckSummaryList, patterns: list[str]) -> None:
+    """Remove findings that match any exclusion pattern from all check summaries."""
+    if not patterns:
+        return
+    for summary in summaries:
+        if not isinstance(summary, StandardCheckSummary):
+            continue
+        excluded = 0
+        for group in summary.finding_groups:
+            original_count = len(group.entities)
+            group.entities = [
+                entity
+                for entity in group.entities
+                if not _matches_exclude_pattern(entity.entity_name, entity.file_path, patterns)
+            ]
+            excluded += original_count - len(group.entities)
+        summary.finding_groups = [g for g in summary.finding_groups if g.entities]
+        if excluded:
+            summary.findings_count = max(0, summary.findings_count - excluded)
+            logger.debug(f"Excluded {excluded} findings from {summary.check_name} via .healthignore patterns")
 
 
 def _relativize_path(file_path: str, repo_root: str) -> str:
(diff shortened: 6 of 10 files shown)


.codeboarding/health/health_report_old.json
```diff
@@ -0,0 +1,253 @@
+{
```
ivanmilevtues: you can remove the _old. Don't see a reason for it to stay.

static_analyzer/lsp_client/client.py
```diff
@@ -200,33 +207,91 @@ def _read_messages(self):
...
+        if method == "textDocument/publishDiagnostics":
+            self._handle_diagnostics_notification(params)
+
         # Call subclass handler
         try:
             self.handle_notification(method, params)
         except Exception as e:
             logger.debug(f"Error in notification handler for {method}: {e}")
 
+    def _handle_diagnostics_notification(self, params: dict):
```
ivanmilevtues, line 249: Maybe it will be cleaner if we have the unused_code_diagnostics as a listener that subscribes to the client.py

This way we don't have to persist and reparse the diagnostics but directly update the unused_code_diagnostic_listener and in the future all sorts of diagnostic_listeners.

static_analyzer/lsp_client/client.py
```diff
@@ -351,6 +426,107 @@ def _initialize(self):
...
+
+        Returns:
+            Settings dictionary for the language server, or None if not applicable.
+        """
+        lang_id = self.language_id.lower()
+
+        if lang_id == Language.PYTHON:
+            # Pyright/Pylance settings to enable unused code detection
+            return {
+                "python": {
```
ivanmilevtues: Can we create a resource file in the statoc_analyzer which can hold all tehse json configs.
(wanting to get the client.py <1k lines, slowly but surely 🤞 )

ivanmilevtues: You can also consider combining/refactoring it with/within the static_analysis_config.yml

And then we'd not have to have hardcoded yml/config gile for the vscode. You can think of a better/more global config setup for the lsps.

static_analyzer/lsp_client/client.py
```diff
@@ -887,6 +1049,60 @@ def close(self):
                 self._reader_thread.join(timeout=2)
             logger.info("Shutdown complete.")
 
+    def open_all_files_for_diagnostics(self):
```
ivanmilevtues: reconsider this file reopening. We are already opening the files for symbol extraction, why not also perform the diagnostic while the file is open there?

This seems like a low hanging fruite for perfomance improvement.

static_analyzer/__init__.py
```diff
@@ -18,16 +19,21 @@
...
         if not pl.is_supported_lang():
             logger.warning(f"Unsupported programming language: {pl.language}. Skipping.")
             continue
+
+        lang_lower = pl.language.lower()
+
         try:
-            if pl.language.lower() in ["typescript"]:
-                # For TypeScript, scan for multiple project configurations (mono-repo support)
+            if lang_lower in (Language.TYPESCRIPT, Language.JAVASCRIPT):
```
ivanmilevtues, line 36: Maybe not in this PR, but why not 🤷 

It used to be the case that only TS had custom config now we have Java as well.
Maybe it is time to get all Client classes -> 

And then have a class_init method or something like that:
for cls in BaseLSPClient.__subclasses__():
    lsp_client = cls.initialize(pl)
    if lsp_client is not None:
         break

if lsp_client is None:
      continue

And put all the custom config logic into that cls.intialize(pl) method. This way every class knows its stuff and we don't have this method growing too much.

static_analyzer/__init__.py
```diff
@@ -137,6 +172,16 @@ def analyze(self, cache_dir: Path | None = None) -> StaticAnalysisResults:
                 results.add_class_hierarchy(client.language.language, analysis.get("class_hierarchies", {}))
                 results.add_package_dependencies(client.language.language, analysis.get("package_relations", {}))
                 results.add_source_files(client.language.language, analysis.get("source_files", []))
+
+                # Collect diagnostics for health checks (collected during analysis)
+                diagnostics = client.get_collected_diagnostics()
```
ivanmilevtues: I have a strong feeling this should be listener as per the other comment so this won't be part of the StaticAnalysisResult. Just doesn't seem to me as part of this lmk what's your opinion

static_analyzer/analysis_result.py
```diff
@@ -264,3 +264,26 @@ def get_all_source_files(self) -> list[str]:
         for language in self.results:
             all_source_files.extend(self.get_source_files(language))
         return all_source_files
+
+    def add_diagnostics(self, language: str, diagnostics: dict[str, list[dict]]):
```
ivanmilevtues: Check the other comments. unsure if this has to be part of the StatAnaRes

static_analyzer/constants.py
```diff
@@ -0,0 +1,56 @@
...
+    JAVASCRIPT = "javascript"
+    GO = "go"
+    JAVA = "java"
+    PHP = "php"
+    RUST = "rust"
+    RUBY = "ruby"
+    SCALA = "scala"
+    KOTLIN = "kotlin"
+    SWIFT = "swift"
+    CPP = "cpp"
```
ivanmilevtues, line 24: Hmm maybe add only the langs we support?
As now i am not sure if cpp is cpp or c++, we don't support it so it doesn't matter.

static_analyzer/constants.py
```diff
@@ -0,0 +1,56 @@
...
+LANG_JAVA = Language.JAVA.value
+LANG_PHP = Language.PHP.value
+LANG_RUST = Language.RUST.value
+LANG_RUBY = Language.RUBY.value
+LANG_SCALA = Language.SCALA.value
+LANG_KOTLIN = Language.KOTLIN.value
+LANG_SWIFT = Language.SWIFT.value
+LANG_CPP = Language.CPP.value
+LANG_C = Language.C.value
+LANG_CSHARP = Language.CSHARP.value
```
ivanmilevtues: Nothing to be backward compatible with. Break it and keep the new clean thing.

static_analyzer/constants.py
```diff
@@ -0,0 +1,56 @@
...
+
+# Language groups for convenience
+LANGUAGES_WEB = {LANG_TYPESCRIPT, LANG_JAVASCRIPT}
+LANGUAGES_JVM = {LANG_JAVA, LANG_KOTLIN, LANG_SCALA}
+LANGUAGES_NATIVE = {LANG_C, LANG_CPP, LANG_RUST, LANG_GO, LANG_SWIFT}
+
+# All supported languages
+ALL_SUPPORTED_LANGUAGES = {lang.value for lang in Language}
```
ivanmilevtues: Didn't check carefully but 90%+ confidece these are unused. Remove them 🔪

static_analyzer/graph.py
```diff
@@ -64,11 +66,14 @@ class ClusteringConfig:
...
-        "typescript": ".",
-        "javascript": ".",
+        Language.PYTHON: ".",
+        Language.GO: ".",
+        Language.PHP: "\\",  # PHP uses backslash for namespaces
+        Language.TYPESCRIPT: ".",
+        Language.JAVASCRIPT: ".",
+        Language.JAVA: ".",
+        Language.RUST: "::",  # Rust uses double colon
+        Language.CPP: "::",  # C++ uses double colon for namespace
```
ivanmilevtues: Remove unused/unsupported things (CPP and Rust). Also this should probably be in the constants.

I would suggest having a data object not just an enum per programming language. So all relevant config is there and not split around.
Also double check if the map is used, or is it just the default delimieter.

static_analyzer/graph.py
```diff
@@ -170,14 +180,26 @@ class CallGraph:
     CLUSTERING_SEED = 42
 
     def __init__(
-        self, nodes: dict[str, Node] | None = None, edges: list[Edge] | None = None, language: str = "python"
+        self,
+        nodes: dict[str, Node] | None = None,
+        edges: list[Edge] | None = None,
```
ivanmilevtues, line 183: Probably both nodes and edges should not be nullable, in the graph class. maybe empty list/dict but not none. remove default value.

static_analyzer/graph.py
```diff
@@ -170,14 +180,26 @@ class CallGraph:
...
+        language: str = "python",
     ) -> None:
         self.nodes = nodes if nodes is not None else {}
         self.edges = edges if edges is not None else []
         self._edge_set: set[tuple[str, str]] = set()
         self.language = language.lower()
         # Set delimiter based on language for qualified name parsing
-        self.delimiter = ClusteringConfig.DELIMITER_MAP.get(self.language, ClusteringConfig.DEFAULT_DELIMITER)
+        # Convert string language to Language enum for lookup
+        lang_key: Language | None = None
```
ivanmilevtues: do it with list comprehension -> (any) so we don't have a nullable type like this and then for if break for a single assing.
Just a quick thing, the delimiter is always "." for now unsure if we even want the for/if else in the init. please simplify it, I would add teh ClusteringCOnfig.get_delimiter(lang) method or something.

static_analyzer/programming_language.py
```diff
@@ -127,7 +133,7 @@ def build(
             config_suffixes = set(config.get("file_extensions", []))
 
             # Create language-specific config based on the LSP server key
-            if lsp_server_key == "java" and "jdtls_root" in config:
+            if lsp_server_key == Language.JAVA and "jdtls_root" in config:
                 language_specific_config = JavaConfig(jdtls_root=Path(config["jdtls_root"]))
```
ivanmilevtues, line 136: Move to the new initialization of the JavaClient as per my other comment: https://github.com/CodeBoarding/CodeBoarding/pull/174/changes#r2773580654

ivanmilevtues (changes requested):
Mostly remove clutter and possibly redesign the way we collect diagnostics + potential small redesign of client instantiation.

I am putting requst changes as I would like you to ping me on the few Q i had around the design. Ping me on DMs if I am not responding here.

GJ thou, love the LSP usage.

.codeboarding/health/health_report_old.json
Svilen-Stefanov: Yeah, I think I wanted to keep it for myself to track more easily what was the previous state and the current but it shouldn't be added for sure.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: Sounds good. The second part I suggest we do in a separate PR though.

static_analyzer/graph.py
Svilen-Stefanov: Notice this is just a reformat, the behavior didn't change here. If I modify this here, I'll need to change many more methods that don't have to do with the health stuff. I'd prefer to do this in a separate refactor PR.

static_analyzer/lsp_client/client.py
Svilen-Stefanov: I refactored it a bit, let's get back to it again when we refactor the `client.py` overall.

health/checks/unused_code_diagnostics.py
```diff
@@ -0,0 +1,340 @@
...
+# LSP DiagnosticTag values
+DIAGNOSTIC_TAG_UNNECESSARY = 1
+
+# Mapping of diagnostic codes to categories for different LSP servers
+DIAGNOSTIC_CODE_MAPPINGS: dict[str, DeadCodeCategory] = {
+    # Pyright/Pylance (Python)
+    "reportUnusedImport": DeadCodeCategory.UNUSED_IMPORT,
+    "reportUnusedVariable": DeadCodeCategory.UNUSED_VARIABLE,
+    "reportUnusedFunction": DeadCodeCategory.UNUSED_FUNCTION,
+    "reportUnusedClass": DeadCodeCategory.UNUSED_CLASS,
```
brovatten, line 66: Constants.py? same for below and above

repo_utils/__init__.py
```diff
@@ -108,7 +108,10 @@ def clone_repository(repo_url: str, target_dir: Path = Path("./repos")) -> str:
             logger.info(f"Repository {repo_name} has uncommitted changes, skipping pull.")
         else:
             logger.info(f"Repository {repo_name} already exists at {dest}, pulling latest.")
-            repo.remotes.origin.pull()
+            try:
+                repo.remotes.origin.pull()
+            except GitCommandError as e:
+                logger.warning(f"Failed to pull latest changes for {repo_name}: {e}. Continuing with existing code.")
```
brovatten, line 114: I suspect this happens because you are already checked out at existing commit, or? If that's the case, then we should handle this explicitly

static_analyzer/lsp_client/language_settings.py
```diff
@@ -0,0 +1,92 @@
+"""Language-specific LSP server settings.
+
+This module contains configuration settings for various LSP servers to enable
+features like unused code detection, diagnostics, and other language-specific
+capabilities.
+"""
+
+from static_analyzer.constants import Language
+
```
brovatten, line 11: this new class is nice!

static_analyzer/__init__.py
brovatten: I'd have it as an parent class. keep what's in common, create an abstract for the other fucntions (it needs to implement these)

brovatten: i think atleast a match -> case would make this easier to read

static_analyzer/analysis_result.py
```diff
@@ -51,6 +51,7 @@ def save(self, repo_hash: str, result: "StaticAnalysisResults") -> None:
 class StaticAnalysisResults:
     def __init__(self):
         self.results: dict[str, dict] = {}
+        self.diagnostics: dict[str, dict[str, list[dict]]] = {}  # Language -> file_path -> diagnostics
```
brovatten: oufff this typing must suggest a dataclas or something

static_analyzer/programming_language.py
```diff
@@ -87,7 +88,7 @@ def __init__(self, lsp_configs: dict):
                 normalized_ext = ext if ext.startswith(".") else f".{ext}"
                 self._extension_to_lsp[normalized_ext] = lsp_server_key
 
-    def _find_lsp_server_key(self, tokei_language: str, file_suffixes: Set[str]) -> Optional[str]:
+    def _find_lsp_server_key(self, tokei_language: str, file_suffixes: set[str]) -> str | None:
```
brovatten, line 91: good!

health/checks/unused_code_diagnostics.py
Svilen-Stefanov: I'm not sure there sill be used in a different place, they are very specific to the unused code so I think keeping them here is better for now. If we start using them from other modules, we can create a separate file for them.

repo_utils/__init__.py
Svilen-Stefanov: I don't remember exactly if I ran main or something else but I think it's generally good to have this check here.

.codeboarding/health/health_report.json
Svilen-Stefanov, line 1: Please check out the latest `health_report.json` and let me know if we should get rid of the previous approach entirely - I think we should probably delete `orphan_code.py` overall.

health/checks/unused_code_diagnostics.py
```diff
@@ -0,0 +1,342 @@
...
+
+class LSPDiagnosticsCollector:
+    """Collects and categorizes LSP diagnostics for dead code detection."""
+
+    def __init__(self):
+        self.diagnostics: list[FileDiagnostic] = []
+        self.issues: list[DiagnosticIssue] = []
+
+    def add_diagnostic(self, file_path: str, diagnostic: LSPDiagnostic) -> None:
+        """Add an LSP diagnostic.
```
ivanmilevtues: we should add to the agent.md or w.e. that there is no need to add comment like this one.

add_diagnostic -> Add an LSP diagranostic 🤯

static_analyzer/lsp_client/client.py
```diff
@@ -40,6 +42,56 @@ def uri_to_path(file_uri: str) -> Path:
...
+            severity=data.get("severity", 1),
+            tags=data.get("tags", []),
+            range=DiagnosticRange(
+                start=DiagnosticPosition(line=start.get("line", 0), character=start.get("character", 0)),
+                end=DiagnosticPosition(line=end.get("line", 0), character=end.get("character", 0)),
+            ),
+        )
+
+    def dedup_key(self) -> tuple[str, str, int, int]:
+        return (self.code, self.message, self.range.start.line, self.range.start.character)
```
ivanmilevtues: Split these into a different file, this one is already huge.

diagnostic_dto.py or something else, not really sure what it has to be.

static_analyzer/lsp_client/language_settings.py
```diff
@@ -0,0 +1,92 @@
...
+        return {
+            "intelephense": {
+                "diagnostics": {
+                    "unusedSymbols": True,
+                    "unusedUseStatements": True,
+                }
+            }
+        }
+
+    return None
```
ivanmilevtues, line 97: I would log.warning or something so we can see it in the logs whenever someone has issues with this.

static_analyzer/constants.py
ivanmilevtues: This can also be StrEnum, maybe doesn't make sense idk

repo_utils/__init__.py
brovatten: I think we should rather handle this with intent (maybe I did a PR I ended up not merging, cause I remember the detached heading leading to a crash - perhaps for the commit as a cli). Would be super nice if you code add something like this: 

```
from pathlib import Path
from git import Repo
from git.exc import GitCommandError

@require_git_import()
def clone_repository(repo_url: str, target_dir: Path = Path("./repos")) -> str:
    repo_url = sanitize_repo_url(repo_url)
    if not remote_repo_exists(repo_url):
        raise RepoDontExistError()

    dest = target_dir / repo_name

    if not dest.exists(): # repo is not already cloned locally
        logger.info(f"Cloning {repo_url} into {dest}")
        Repo.clone_from(repo_url, dest)
        logger.info("Cloning finished!")
        return repo_name

    repo = Repo(dest)

    if repo.is_dirty(untracked_files=True):
        logger.info(f"Repository {repo_name} has uncommitted changes, skipping pull.")
        logger.info("Cloning finished!")
        return repo_name

    # (2) Explicitly handle detached HEAD / no upstream tracking branch
    if repo.head.is_detached:
        logger.warning(f"Repository {repo_name} is in detached HEAD state, skipping pull.")
        logger.info("Cloning finished!")
        return repo_name

    logger.info(f"Repository {repo_name} already exists at {dest}, pulling latest.")
    try:
        repo.remotes.origin.pull()
    except GitCommandError as e:
        logger.warning(f"Failed to pull latest changes for {repo_name}: {e}. Continuing with existing code.")

    logger.info("Cloning finished!")
    return repo_name


```

.codeboarding/health/health_report.json
brovatten: I love that you call it metric_value to avoid formats of renaming JSONs!!
With the previous approach, I am not sure exactly what this refers to, we can chat tmrw, or feel free to go ahead

PR #177, opened 2026-02-05
Static analysis tests

brovatten:


tests/static_analyzer/test_static_analysis_integration.py
```diff
@@ -0,0 +1,308 @@
...
+
+
+def test_static_analysis_python_e2e(tmp_path: Path) -> None:
+    if not _pyright_available():
+        pytest.skip("pyright-langserver not available")
+
+    repo = tmp_path / "mini_repo"
+    repo.mkdir()
+
+    (repo / "a.py").write_text(
```
Svilen-Stefanov, line 26: Should we maybe create this as a sample project that we can explore a bit easier? It's a bit more difficult to read when the entire code is written as strings, making it more difficult to extend.

tests/static_analyzer/test_static_analysis_integration.py
```diff
@@ -0,0 +1,308 @@
...
+    subprocess.run(
+        ["git", "checkout", CODEBOARDING_COMMIT],
+        cwd=target,
+        check=True,
+        capture_output=True,
+    )
+
+
+@pytest.mark.slow
+def test_large_repo_determinism(tmp_path: Path) -> None:
```
Svilen-Stefanov, line 349: Should we put this under tests/integration/stability(or determinism)? I wonder if it could be combined with the other codeboarding integration test so that we don't download the repo multiple times. 
Also I assume both of them should not use the cache - I'm not sure if you're checking this here already. 
If you use the cache, the test is not really testing much.

PR #178, opened 2026-02-05
Fix typescript

brovatten:
This includes two fixes:
  - TS/JS/TSX/JSX now all use TypeScriptClient (not generic LSP) based on lsp_server_key, and we skip duplicate TS-family entries.
  - Keep one TS file open during workspace/symbol to prevent the TS server dropping the project.

The fixes came from this when I saw this error:
```
  TypeScript Server Error (5.9.3)
  No Project.
  ...
  workspace/symbol failed: {'code': 1, 'message': '<semantic> TypeScript Server Error (5.9.3)\nNo Project.\n...'}
```
 You can reproduce the error on [excalidraw](https://github.com/excalidraw/excalidraw)

edit: easier description is that the TypeScriptClient keeps one bootstrap document open because the TypeScript language server in our analysis flow can drop configured project state and return "No Project" for workspace-level queries when no file is open, while the generic LSPClient path (e.g., Pyright) remains workspace-usable from rootUri without that anchor.

edit2: I updated TypeScript client creation to treat TypeScript and JavaScript as a single TS-family by selecting one preferred entry and merging TS/JS suffixes into one ProgrammingLanguage, so we run a single TypeScriptClient without losing .js/.jsx coverage.

**edit3**: You can see below why it happens:

https://github.com/typescript-language-server/typescript-language-server/issues/705
https://github.com/typescript-language-server/typescript-language-server/issues/636

The latter says : 
```
" Note that a server’s ability to fulfill requests is independent of whether a text document is open or closed. "
But in the code, documentSymbol for example doesn't work for unopened files .
JDT ls (Java language server) and pyright (Python's) do work without having to open the files prior to sending documentSymbol,
```
and they conclude with 
`The decision here is not to support that use case since it's not relevant for editors.`

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -23,8 +23,8 @@ def create_clients(
     clients: list[LSPClient] = []
     ts_family_keys = {"typescript", "javascript"}
     ts_family_candidates = [pl for pl in programming_languages if pl.lsp_server_key in ts_family_keys]
-    ts_family_preferred = "typescript"
-    merged_ts_family_language = None
+    ts_family_preferred: ProgrammingLanguage | None = None
+    merged_ts_family_language: ProgrammingLanguage | None = None
     if ts_family_candidates:
         ts_family_preferred = next((pl for pl in ts_family_candidates if pl.language.lower() == "typescript"), None)
         if ts_family_preferred is None:


static_analyzer/lsp_client/typescript_client.py
```diff
@@ -165,8 +170,13 @@ def _process_config_files(self) -> bool:
     def _bootstrap_project(self, ts_files: list, config_found: bool):
         """Bootstrap TypeScript project by opening files."""
         logger.info("Opening sample files to bootstrap TypeScript project...")
-        # Files are already filtered in _find_typescript_files
-        sample_files = ts_files[:3]
+        priority_by_suffix = {".ts": 0, ".tsx": 1, ".js": 2, ".jsx": 3}
+        sample_files = sorted(
```
ivanmilevtues, line 174: Can you write a comment on why we use the magic numbers [:3] then if sample_files and no self.boostrap_file_path we chose the first file?

And whenever we close bootstrap_files we leave out the sample_files[0]

Also this will fail if samples files is empty? sample-files[1:] index out of range, are we sure this will never fail.

static_analyzer/__init__.py
```diff
@@ -21,12 +21,32 @@ def create_clients(
...
+            ts_family_preferred = max(ts_family_candidates, key=lambda pl: pl.percentage)
+        merged_ts_family_language = ProgrammingLanguage(
+            language=ts_family_preferred.language,
+            size=sum(pl.size for pl in ts_family_candidates),
+            percentage=sum(pl.percentage for pl in ts_family_candidates),
+            suffixes=sorted({suffix for pl in ts_family_candidates for suffix in pl.suffixes}),
+            server_commands=ts_family_preferred.server_commands,
+            lsp_server_key=ts_family_preferred.lsp_server_key,
+            language_specific_config=ts_family_preferred.language_specific_config,
+        )
```
ivanmilevtues, line 40: This deserves its own function with a name which explain what the heck is happening ;d

@Svilen-Stefanov / @brovatten be mindful of merging/rebasing on main, as I also asked Svilen to refactor the building for the clients, so make sure it follows whatever we end up with :)

ivanmilevtues (approved):
Mainly style comments (things are quire random) it would be nice to have some explanation of the numbers. + Make sure we are not throwing index out of boundaries error.

Please fix those, don't want to block with Request changes for these comments :)

static_analyzer/__init__.py
Svilen-Stefanov: I decided to leave that for a pure refactor PR - I'd suggest to merge what we merge beforehand and clean it up afterwards because it can get quite messy otherwise.

static_analyzer/__init__.py
```diff
@@ -21,12 +21,32 @@ def create_clients(
     programming_languages: list[ProgrammingLanguage], repository_path: Path, ignore_manager: RepoIgnoreManager
 ) -> list[LSPClient]:
     clients: list[LSPClient] = []
+    ts_family_keys = {"typescript", "javascript"}
```
Svilen-Stefanov, line 24: Can you run the integration tests and see what's happening. Notice that the values added to the fixtures are as they need to be for the GitHub environment here so you might want to run them directly in GitHub actions here and just paste the link here. 
I just want to understand the impact of this on the analysis of both TS and JS. We might need to potentially change the fixtures if they improve the actual results returned.

brovatten:
Closing this as it's gone stale. While testing this, two weeks ago, I could not find again that this solved it. We have a new pipeline, so i would redo it and use this as a reference

PR #179, opened 2026-02-05
feat: centralize constants into modular files

NevilPatel01:
- Created static_analyzer/constants.py with ClusteringConfig, NodeType, GraphConfig
- Created agents/constants.py with LLMDefaults, FileStructureConfig
- Created health/constants.py with HealthConfig
- Created root constants.py with AppConfig
- Refactored static_analyzer/graph.py to use imported constants
- Refactored agents/tools/read_file_structure.py to use FileStructureConfig
- Refactored agents/llm_config.py to use LLMDefaults
- Refactored local_app.py to use AppConfig
- All constant values preserved, no behavior changes
- All original comments preserved for maintainability

### DIFF
--- a/static_analyzer/constants.py
+++ b/static_analyzer/constants.py
@@ -70,7 +70,6 @@ class ClusteringConfig:
     CLUSTERING_SEED = 42
 
 
-
 class NodeType:
     # LSP SymbolKind constants
     CLASS_TYPE = 5
@@ -95,7 +94,6 @@ class NodeType:
         VARIABLE_TYPE: "Variable",
         CONSTANT_TYPE: "Constant",
     }
-
 
 
 class Node:
@@ -166,4 +164,3 @@ class Node:
 
     def __repr__(self) -> str:
         return f"Node({self.fully_qualified_name}, {self.file_path}, {self.line_start}-{self.line_end})"
-
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -406,7 +406,7 @@ class CallGraph:
     def _cluster_with_algorithm(self, graph: nx.DiGraph, algorithm: str) -> list[set[str]]:
         # Use class-level seed for reproducibility - Louvain/Leiden are non-deterministic without it
         if algorithm == "louvain":
-                return list(nx_comm.louvain_communities(graph, seed=ClusteringConfig.CLUSTERING_SEED))
+            return list(nx_comm.louvain_communities(graph, seed=ClusteringConfig.CLUSTERING_SEED))
         elif algorithm == "greedy_modularity":
             return list(nx.community.greedy_modularity_communities(graph))
         elif algorithm == "leiden":


agents/constants.py
```diff
@@ -0,0 +1,17 @@
+"""Constants for the agents module."""
+
+
+class LLMDefaults:
+    DEFAULT_AGENT_TEMPERATURE = 0.1
+    DEFAULT_PARSING_TEMPERATURE = 0
+    DEFAULT_MAX_RETRIES = 0
+    DEFAULT_TIMEOUT = None
+    DEFAULT_MAX_TOKENS = None
+    ANTHROPIC_MAX_TOKENS = 8192
```
brovatten: I would not keep the model specific defalts here, but rather in the LLM config as they are not general, and we only initialize the model in one place (hence we get no favours of having a reusable variable). 

On having the other defaults here, I kind of like it, at the same time the llm_config.py is already like a constants.py. Wdyt @Svilen-Stefanov ?

static_analyzer/constants.py
```diff
@@ -0,0 +1,53 @@
...
+        "python": ".",
+        "go": ".",
+        "php": "\\",  # PHP uses backslash for namespaces
+        "typescript": ".",
+        "javascript": ".",
+    }
+
+
+class NodeType:
+    # Node type constants
```
brovatten: This could be elaborated by with what the type meaning

brovatten: I see we had this before, but now it is without the context of the class

ivanmilevtues: We already have NodeTypes introduced by @Svilen-Stefanov: Check the Node class in static_analyzer/graph.py

This is obsolete change it with those variables

static_analyzer/graph.py
```diff
@@ -30,73 +32,13 @@ def get_nodes_for_cluster(self, cluster_id: int) -> set[str]:
...
-        "php": "\\",  # PHP uses backslash for namespaces
-        "typescript": ".",
-        "javascript": ".",
-    }
+class NodeType:
+    # Node type constants
+    METHOD_TYPE = 6
 
 
 class Node:
```
Svilen-Stefanov: Can you maybe rebase, it seems like the changes are not up to date, refer to https://github.com/CodeBoarding/CodeBoarding/blob/main/static_analyzer/graph.py#L75-L98 on the main branch.

Svilen-Stefanov: Thank you for the rebase and sorry for the late reply here. Let's move this in the `static_analyzer/constants.py`.

static_analyzer/constants.py
```diff
@@ -22,3 +22,77 @@ class Language(StrEnum):
...
+        FIELD_TYPE: "Field",
+        FUNCTION_TYPE: "Function",
+        VARIABLE_TYPE: "Variable",
+        CONSTANT_TYPE: "Constant",
+    }
+
+
+class GraphConfig:
+    # Deterministic seed for clustering algorithms
+    CLUSTERING_SEED = 42
```
Svilen-Stefanov: Instead of having this in a separate class, I think we can put it under `ClusteringConfig`.

PR #181, opened 2026-02-06
Test for init time of LSP

brovatten:
I considered speeding up the iterative analysis by have the LSP as a daemon (running while we are not analyzing anything).

It depends how quick the iterative analysis is now and if our target is still sub 10s. If so, for java and TS this could be interesting. See below:

image

Would be happy to work on this.


brovatten:
This is a useful test to arrive at : we don't need a daemon. This belongs in the evals repo. no need to merge

PR #183, opened 2026-02-06
Add REVIEW.md

Svilen-Stefanov:


REVIEW.md
```diff
@@ -0,0 +1,360 @@
+# AI Agent Code Review Guidelines
+
+**Purpose:** This document serves as a comprehensive checklist for AI agents to evaluate code changes before considering work complete. Follow these guidelines rigorously to ensure all changes meet production-quality standards.
+
```
brovatten, line 4: I would add context here like:

**Context**: We are developing X that is an open source bla bla. This then communicates with our extension codeboarding-vscode that uses the main.py as an entry point (this context it will probably not have from reading the repo). It can also be used as standalone bla bla

REVIEW.md
```diff
@@ -0,0 +1,360 @@
...
+- **Temporary field:** Instance variables only used in specific methods
+- **Refused bequest:** Subclass ignores inherited methods
+
+### 🚩 Testing Red Flags
+- **Fragile tests:** Tests break on unrelated changes
+- **Slow tests:** Tests taking > 1 second (should be unit tests)
+- **Mystery guest:** Test relies on external data not visible in test
+- **Happy path only:** No error condition tests
+
+---
```
brovatten, line 322: Add one for health to see if it catches it? Adding the same health checks that we have

brovatten (approved):

REVIEW.md
```diff
@@ -0,0 +1,360 @@
...
+## Review Completion Checklist
+
+After making changes, verify:
+
+- [ ] All items in Sections 1-8 reviewed and addressed
+- [ ] No red flags from Section 10
+- [ ] Changes are better than what was there before
+- [ ] You would be comfortable explaining these changes in a code review meeting
+- [ ] Documentation is updated for any new patterns or conventions
+
```
brovatten, line 355: should we add to try the relevant tests and run mypy?

REVIEW.md
```diff
@@ -0,0 +1,360 @@
+# AI Agent Code Review Guidelines
+
+**Purpose:** This document serves as a comprehensive checklist for AI agents to evaluate code changes before considering work complete. Follow these guidelines rigorously to ensure all changes meet production-quality standards.
+
+**When to Use:** After completing any code changes, run through this checklist before marking tasks as complete.
+
+---
```
brovatten, line 7: I always start by asking it to make a git diff between current branch and main. I think we should state this here aswell

REVIEW.md
```diff
@@ -0,0 +1,360 @@
...
+
```
brovatten, line 8: And do we add the cli command to run the main here to make sure it starts the run? Have it run for 10 secs

REVIEW.md
Svilen-Stefanov: I personally started to dislike this - first we have it in our AGENTS.md and then the builds here already check this so I find it a waste of agents resources if this happens all the time. We could remove it for one and keep it only here or the other way around but I think running it always seems unnecessary.

REVIEW.md
Svilen-Stefanov: I personally don't like it because I generally don't want to always run main for every change. For some changes I'd run `health_main.py` for others  tests might be enough etc. I think the main should for sure be run if you're changing the arguments etc but for other functionality, there might be better ways to test it.

PR #184, opened 2026-02-08
Fix for non-npm users

brovatten:
Skips the npm install if it doesn't exist in users path. We already have it in the toml. For a user who don't have npm, java and go will work, and u are very likely not developing in TS/JS then :D 

I added a print to make it easier for users to see what they can and can't do.

image

### DIFF
--- a/install.py
+++ b/install.py
@@ -88,9 +88,11 @@ def install_npm_with_nodeenv() -> bool:
     print("   impact: installs Node.js + npm into the active Python virtual environment")
 
     try:
-        subprocess.run(command, check=True)
+        result = subprocess.run(command, capture_output=True, text=True, check=True)
     except subprocess.CalledProcessError as e:
         print(f"Step: npm remediation finished: failure - {e}")
+        if e.stderr:
+            print(f"   {e.stderr.strip()}")
         print("   You can install Node.js manually from: https://nodejs.org/en/download")
         print("   Then verify with: npm --version")
         return False
@@ -515,7 +517,7 @@ def download_jdtls():
 
         print("  Extracting JDTLS...")
         with tarfile.open(jdtls_archive, "r:gz") as tar:
-            tar.extractall(path=jdtls_dir)
+            tar.extractall(path=jdtls_dir, filter="tar")
 
         jdtls_archive.unlink()
 


static_analyzer/analysis_cache.py
```diff
@@ -18,6 +19,20 @@
 logger = logging.getLogger(__name__)
 
 
+def flush_repository_cache(repo_path: Path) -> None:
```
ivanmilevtues: Unsure if we need this. In main we have the following lines of code:

```
    # Incremental update options
    parser.add_argument(
        "--full",
        action="store_true",
        help="Force full reanalysis, skipping incremental update detection",
    )
```

This will remove (ignore) the cache. What is the purpose of this new flag?

ivanmilevtues (changes requested):
The install is fine for me, the flush_repository_cache is what I don't get.

install.py
```diff
@@ -457,6 +466,62 @@ def install_pre_commit_hooks():
...
+
+    python_ok = py_node_path.exists() or bool(py_env_path)
+    typescript_ok = npm_available and ts_path.exists()
+    javascript_ok = typescript_ok
+    php_ok = npm_available and php_path.exists()
+    go_ok = go_path.exists()
+    java_ok = java_path.exists()
+
+    print(f"  - Python: {'yes' if python_ok else 'no'}")
+    if not python_ok:
```
Svilen-Stefanov: Instead of doing this, I'd have a list of paths and extra checks per language and then do a for loop checking if the path exists and the extra check is satisfied.

main.py
```diff
@@ -380,6 +381,11 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
         action="store_true",
         help="Use smart incremental updates (tries incremental first, falls back to full)",
     )
+    parser.add_argument(
```
Svilen-Stefanov: I don't think we need this as @ivanmilevtues pointed out [here](https://github.com/CodeBoarding/CodeBoarding/pull/184/changes#r2778366620).

install.py
```diff
@@ -457,6 +466,62 @@ def install_pre_commit_hooks():
...
+
+    servers_dir = Path("static_analyzer/servers").resolve()
+    is_win = platform.system() == "Windows"
+    node_ext = ".cmd" if is_win else ""
+
+    ts_path = servers_dir / "node_modules" / ".bin" / f"typescript-language-server{node_ext}"
+    php_path = servers_dir / "node_modules" / ".bin" / f"intelephense{node_ext}"
+    py_node_path = servers_dir / "node_modules" / ".bin" / f"pyright-langserver{node_ext}"
+    py_env_path = shutil.which("pyright-langserver") or shutil.which("pyright-python-langserver")
+    go_path = servers_dir / ("gopls.exe" if is_win else "gopls")
```
Svilen-Stefanov: I don't like that the paths are not consistent with the ones in the extension. Technically these should be under `servers_dir / <OS_dir_name> / <executable>` if we want to follow the same pattern.

install.py
Svilen-Stefanov, line 1: As I user, I'd actually prefer to have more actionable things here. Ideally, when we identify a missing requirement, we'd give the user the option to chose for us to install it for them (tell them what we will run and what's the impact on their system), we could run `subprocess.run([sys.executable, "-m", "nodeenv", "--python-virtualenv"], check=True)` to install npm for them directly. 

If they press n (and they want to install it on their own), we would ideally give them instructions on how to install it (at least point to a website or so). 

Also, this is never performed for the extension, what's the behavior there and how would we fix that?

Svilen-Stefanov (changes requested):
My main concerns are:
1. The new flag in main seems unnecessary
2. The fix is a bit too passive, I'd like to make it more actionable

static_analyzer/analysis_cache.py
brovatten: The idea is to make it easier to use as a CLI. And with the functionality of caching becoming more complicated, I think it would be nice with a simple way to clean it up for the repo.

Also, since the codeboarding manages the runtime saving, if we want to add a button to "clear cache" then having it here would be easiest.

The force "full" analysis only works for full analysis, what about incremental ones?

install.py
brovatten: I really like this idea! it is implemented.

For the extension, do you mean if the install the extension but don't have npm installed?

static_analyzer/analysis_cache.py
Svilen-Stefanov: My thinking is that we don't need both. If you think that makes more sense, you can discuss with @ivanmilevtues and remove the full flag, but having both is redundant and I'm not a fan of. 
I'd either add this flag later or remove the full one here already. 

Also notice that users can remove the cache manually atm so it seems like the flag is not urgent. 
I see your point though that this flag might make more sense than the "full" one.

install.py
Svilen-Stefanov: Yeah, my understanding is that you're trying to solve the problem @ivanmilevtues mentioned a couple days ago that we need to do something when npm is not installed. What I see here makes sense for the core but not for the extension so I was wondering if you thought about that.

install.py
```diff
@@ -457,7 +549,86 @@ def install_pre_commit_hooks():
...
+    ts_path = servers_dir / "node_modules" / ".bin" / f"typescript-language-server{node_ext}"
+    php_path = servers_dir / "node_modules" / ".bin" / f"intelephense{node_ext}"
+    py_node_path = servers_dir / "node_modules" / ".bin" / f"pyright-langserver{node_ext}"
+    py_env_path = shutil.which("pyright-langserver") or shutil.which("pyright-python-langserver")
+    go_path = platform_bin_dir / ("gopls.exe" if is_win else "gopls")
+    java_path = servers_dir / "bin" / "jdtls"
+
+    language_checks = [
+        (
+            "Python",
```
Svilen-Stefanov: Can we make these a dataclass or something, atm it's difficult to say what they refer to.

install.py
```diff
@@ -53,10 +54,89 @@ def check_npm():
...
+def get_platform_bin_subdir() -> str:
+    """Return OS-specific binary folder name used by the extension."""
+    subdirs = {"windows": "win", "darwin": "macos", "linux": "linux"}
+    system = platform.system().lower()
+    if system not in subdirs:
+        raise RuntimeError(f"Unsupported platform: {system}")
+    return subdirs[system]
+
+
+def get_platform_bin_dir(servers_dir: Path) -> Path:
```
Svilen-Stefanov, line 175: I think we need this elsewhere like in `vscode_constants.py` - can we reuse these?

install.py
```diff
@@ -53,10 +54,89 @@ def check_npm():
...
+    if choice in {"y", "yes"}:
+        return install_npm_with_nodeenv()
+
+    print("Step: npm remediation skipped by user")
+    print("   Install Node.js manually from: https://nodejs.org/en/download")
+    print("   Then verify with: npm --version")
+    return False
+
+
+def parse_args() -> argparse.Namespace:
```
Svilen-Stefanov, line 150: What is this used for? I don't think we should ask the user to provide this, this should be the default and we should have a cmd line check when npm is not installed if they want us to install it (press y) and we interrupt for any other key.

static_analyzer/analysis_cache.py
ivanmilevtues: If we remove `--full`, then what is the default behavior? Is it `--iterative`, then why do we have that flag :D.

So if we add this, I would just remove both `--full` and `--iterative`. I think it actually makes sense and makes it easer for our logic, we won't have to check flags around the codebase.

If you want a full, clean analysis you just do `--flush-cache` and that's it. And we'd automatically detect if there is analysis stat analysis cache and so on.

TL;DR, keep the new flag remove both `--full` and `--iterative`

install.py
ivanmilevtues: If this is aiming at addressing that issue, it won't fix it.

install.py
```diff
@@ -311,6 +394,15 @@ def update_static_analysis_config():
                 config[section][key]["command"][0] = str(full_path)
                 updates += 1
 
+    # Minimal fallback: if pyright wasn't installed under node_modules, use active env binary if available
+    node_ext = ".cmd" if is_win else ""
```
ivanmilevtues, line 587: This is wrong. I know that pyright-langserver is infact a js file, which we have to run with node.

You can leave out this issue, I will try to fix it with the help of my windows.

install.py
```diff
@@ -457,7 +549,86 @@ def install_pre_commit_hooks():
...
+    platform_bin_dir = get_platform_bin_dir(servers_dir)
+    is_win = platform.system() == "Windows"
+    node_ext = ".cmd" if is_win else ""
+
+    ts_path = servers_dir / "node_modules" / ".bin" / f"typescript-language-server{node_ext}"
+    php_path = servers_dir / "node_modules" / ".bin" / f"intelephense{node_ext}"
+    py_node_path = servers_dir / "node_modules" / ".bin" / f"pyright-langserver{node_ext}"
+    py_env_path = shutil.which("pyright-langserver") or shutil.which("pyright-python-langserver")
+    go_path = platform_bin_dir / ("gopls.exe" if is_win else "gopls")
+    java_path = servers_dir / "bin" / "jdtls"
```
ivanmilevtues, line 755: I don't think that these are correct.
.cmd is not the case for windows in most cases, I think they are either .exe or nothing at all. I think you cannot really fix/merge this without trying on windows.

main.py
ivanmilevtues: After rereading your stuff, I think it should stay and both `--incremental` and `--full` should go.

install.py
```diff
@@ -1,12 +1,35 @@
...
+from pathlib import Path
+
 import requests
 import tarfile
 import yaml
-from pathlib import Path
+
+
+@dataclass(frozen=True, slots=True)
+class LanguageSupportCheck:
```
Svilen-Stefanov, line 17: My first intuition says this should be in `language_setting.py` but since this is only about requirements, not sure if it belongs there per se.

install.py
Svilen-Stefanov: My previous comment still remains, do we need this arguments? 
It is very atypical to have args in an install script. Can we get rid of them now that we have the interactive prompt?

install.py
Svilen-Stefanov: Did you try this on Windows? Since I don't see a reply here, I'm not sure what's the outcome of it.

install.py
```diff
@@ -457,15 +581,87 @@ def install_pre_commit_hooks():
...
+    print("Step: Language support summary")
+
+    servers_dir = Path("static_analyzer/servers").resolve()
+    platform_bin_dir = get_platform_bin_dir(servers_dir)
+    is_win = platform.system() == "Windows"
+    node_ext = ".cmd" if is_win else ""
+
+    ts_path = servers_dir / "node_modules" / ".bin" / f"typescript-language-server{node_ext}"
+    php_path = servers_dir / "node_modules" / ".bin" / f"intelephense{node_ext}"
+    py_node_path = servers_dir / "node_modules" / ".bin" / f"pyright-langserver{node_ext}"
```
ivanmilevtues, line 752: All these are not .cmd scripts, so this is invalid. They are all js files, you can find more details on what they look like in the vscode constants.

Generally speaking we already have the install_node_servers method check that, there we locate only the  package.

Check the scripts are not bash/cmd scripts they are js
Image

install.py
```diff
@@ -311,6 +426,15 @@ def update_static_analysis_config():
...
                 updates += 1
 
+    # Minimal fallback: if pyright wasn't installed under node_modules, use active env binary if available
+    node_ext = ".cmd" if is_win else ""
+    node_pyright = servers_dir / "node_modules" / ".bin" / f"pyright-langserver{node_ext}"
+    if not node_pyright.exists():
+        env_pyright = shutil.which("pyright-langserver") or shutil.which("pyright-python-langserver")
+        if env_pyright and "lsp_servers" in config and "python" in config["lsp_servers"]:
+            config["lsp_servers"]["python"]["command"][0] = env_pyright
+            updates += 1
```
ivanmilevtues, line 593: This fallback is non existant as per the other comment below

PR #185, opened 2026-02-08
Make meta agent faster

brovatten:
This is an aggresive approach to improving the iterative analysis. First, we cache it - since for iterative analysis we now run it despite often not needing it. I know we might get a cache hit at the provider, but I think there's no guarantee. For this however, we can use local caching with langchain (with a tiny database) as it may not need an update. Here I also think we should look at edit distance of e.g. README.md (but wanted to hear ur guys' opinion) since two characters diff in readme should not re-trigger the meta agent.

Also, in Python I hate having multiple threads that convolutes log messages, but none the less I went ahead. Since meta agent and static analysis is not dependant on eachother, we let them run in parallell. Conceptually makes sense, in practice looks not super nice :D 

Since LLM is biggest part to why iterative is slow I think this is one way to start

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -42,7 +42,9 @@ MONITORING_CALLBACK = MonitoringCallback(stats_container=RunStats())
 class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
     _parsing_llm: Optional[BaseChatModel] = None
 
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str, llm: BaseChatModel):
+    def __init__(
+        self, repo_dir: Path, static_analysis: StaticAnalysisResults | None, system_message: str, llm: BaseChatModel
+    ):
         ReferenceResolverMixin.__init__(self, repo_dir, static_analysis)
         MonitoringMixin.__init__(self)
         self.llm = llm
@@ -526,7 +528,7 @@ class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
 
 
 class LargeModelAgent(CodeBoardingAgent):
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str):
+    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults | None, system_message: str):
         agent_model = os.getenv("AGENT_MODEL")
         llm, model_name = self._static_initialize_llm(model_override=agent_model, is_parsing=False)
         super().__init__(repo_dir, static_analysis, system_message, llm)
--- a/agents/meta_agent.py
+++ b/agents/meta_agent.py
@@ -1,4 +1,8 @@
 import logging
+import hashlib
+import json
+import os
+from datetime import datetime, timezone
 from pathlib import Path
 
 from langchain_core.prompts import PromptTemplate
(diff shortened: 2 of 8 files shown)


agents/agent.py
```diff
@@ -526,7 +528,7 @@ def _classify_unassigned_files_with_llm(
 
 
 class LargeModelAgent(CodeBoardingAgent):
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str):
+    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults | None, system_message: str):
```
ivanmilevtues, line 531: Why is it nullable?

agents/agent.py
```diff
@@ -42,7 +42,9 @@
 class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
     _parsing_llm: Optional[BaseChatModel] = None
 
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults, system_message: str, llm: BaseChatModel):
+    def __init__(
+        self, repo_dir: Path, static_analysis: StaticAnalysisResults | None, system_message: str, llm: BaseChatModel
```
ivanmilevtues, line 46: Why nullable?

agents/meta_agent.py
```diff
@@ -14,10 +18,37 @@
...
+        "uv.lock",
+        "package.json",
+        "package-lock.json",
+        "yarn.lock",
+        "pnpm-lock.yaml",
+        "bun.lockb",
+        "tsconfig.json",
+    }
+    _DEPENDENCY_DIRS = {"requirements", "deps", "dependencies", "env"}
+    _DEPENDENCY_DIR_SUFFIXES = {".txt", ".yml", ".yaml", ".toml"}
```
ivanmilevtues, line 45: This list is incomplete. As different projects have different type of setups, that's why I implented it as an agent.

I.e. this doesn't capture anything for java (maven files) and php (no idea what they use there).

diagram_analysis/diagram_generator.py
```diff
@@ -95,50 +129,52 @@ def process_component(self, component: Component):
...
-            logger.info(f"Health report written to {health_path} (score: {health_report.overall_score:.3f})")
-        else:
-            logger.warning("Health checks skipped: no languages found in static analysis results")
-
-        self.meta_agent = MetaAgent(
-            repo_dir=self.repo_location, project_name=self.repo_name, static_analysis=static_analysis
-        )
-        self._monitoring_agents["MetaAgent"] = self.meta_agent
-        meta_context = self.meta_agent.analyze_project_metadata()
+        self.meta_agent, meta_future, meta_executor = self._start_meta_analysis_background()
```
ivanmilevtues, line 132: This looks very ugly, there are 500 try excepts, and it is very hard to see what is done when.

I think just having the meat_future is okay.

Also we have 300 different try excepts, maybe add the finaly to the awai meta analysis results. Also meta agent i think is not needed in self.

static_analyzer/reference_resolve_mixin.py
```diff
@@ -13,10 +13,15 @@
 class ReferenceResolverMixin:
     _parse_invoke: Any  # Provided by Agent base class
 
-    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults):
+    def __init__(self, repo_dir: Path, static_analysis: StaticAnalysisResults | None):
```
ivanmilevtues, line 16: why nullable? Why not just the normal non-null

The _requrire_stat_analysis seems not needed.

ivanmilevtues (changes requested):
Generally seems very overengineered for what it is.

I think of it:
MetaAgent, collects meta data by reading few files.
If those files/commands haven't changed we should load the cache.

We can collect exactly which files those are as we have access to the read_file tool.

AS now you are doing some complicated hashes that I don't really get:
```Python
        for file_path in meta_content_files:
            rel_path = str(file_path.relative_to(self.repo_dir))
            hasher.update(f"C:{rel_path}\n".encode("utf-8"))
```
I.e. why is it with C:?

diagram_analysis/diagram_generator.py
```diff
@@ -68,6 +68,40 @@ def __init__(
...
+        )
+        self._monitoring_agents["MetaAgent"] = meta_agent
+
+        logger.info("Starting MetaAgent metadata analysis in background")
+        executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="meta-analysis")
+        future = executor.submit(meta_agent.analyze_project_metadata)
+        return meta_agent, future, executor
+
+    @staticmethod
+    def _shutdown_background_executor(executor: ThreadPoolExecutor) -> None:
```
Svilen-Stefanov, line 88: I don't quite get this method, does it need to be static and why if so?

diagram_analysis/diagram_generator.py
```diff
@@ -68,6 +68,40 @@ def __init__(
...
+        try:
+            executor.shutdown(wait=False)
+        except Exception:
+            logger.warning("Failed to shut down background executor cleanly", exc_info=True)
+
+    @staticmethod
+    def _await_meta_analysis_result(meta_future: Any) -> Any:
```
Svilen-Stefanov, line 95: Same for this static method.

static_analyzer/reference_resolve_mixin.py
```diff
@@ -52,10 +58,10 @@ def _resolve_single_reference(self, reference, file_candidates: list[str] | None
         # No resolution found - will be cleaned up later
         logger.warning(f"[Reference Resolution] Could not resolve reference {reference.qualified_name} in any language")
 
-    def _try_exact_match(self, reference, qname, lang):
+    def _try_exact_match(self, reference, qname, lang, static_analysis: StaticAnalysisResults):
         """Attempts exact reference matching."""
         try:
-            node = self.static_analysis.get_reference(lang, qname)
+            node = static_analysis.get_reference(lang, qname)
```
Svilen-Stefanov, line 64: This change seems very strange. Previously we did `self.static_analysis` and now we do `static_analysis`.
What's the idea here? If we want to do this on any static analysis, shouldn't this be a helper function somewhere?

PR #187, opened 2026-02-09
Remove unused code

Svilen-Stefanov:
Fixes based off the new health issue detection. What I noticed while using CodeBoarding to fix these - it would be great if we have a single "Fix issues" button that's not per component - it get's quite annoying to click on all components and fixing their issues one by one.

### DIFF
--- a/agents/tools/toolkit.py
+++ b/agents/tools/toolkit.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, cast
+from typing import cast
 
 from .base import RepoContext, BaseRepoTool
 from .read_source import CodeReferenceReader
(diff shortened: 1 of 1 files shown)


agents/tools/toolkit.py
```diff
@@ -1,6 +1,5 @@
 import logging
-from typing import List, Optional, cast, Any
-from pathlib import Path
+from typing import List, cast
```
ivanmilevtues: Also remove List, and use list.

output_generators/html_template.py
```diff
@@ -119,53 +119,49 @@ def _generate_html_body(project: str, insights: AnalysisInsights, components_htm
     """
 
 
-def _generate_cytoscape_script(cytoscape_json: str) -> str:
-    # Note: uses {{ }} for literal JS braces inside the f-string
-    return f"""
-    <script>
-        // Wait for all scripts to load before initializing
-        document.addEventListener('DOMContentLoaded', function() {{
+def _get_library_checks() -> str:
```
ivanmilevtues, line 122: This method seems dead, you can remove it.

Svilen-Stefanov: It isn't, all of these are called in `_generate_cytoscape_script` like:
```
def _generate_cytoscape_script(cytoscape_json: str) -> str:
    library_checks = _get_library_checks()
    dagre_registration = _get_dagre_registration()
    style = _get_cytoscape_style()
    layout = _get_layout_config()
    event_handlers = _get_event_handlers()
    control_functions = _get_control_functions()
```

PR #188, opened 2026-02-10
Meta agent runs along static analysis

brovatten:
The other branch was getting a mess, and I decided to abandon the caching of meta agent for now.

Instead, let's parallellize the meta agent and static analysis; keeping it simple for now.

### DIFF
--- a/agents/meta_agent.py
+++ b/agents/meta_agent.py
@@ -19,12 +19,11 @@ class MetaAgent(CodeBoardingAgent):
     def __init__(
         self,
         repo_dir: Path,
-        static_analysis: StaticAnalysisResults,
         project_name: str,
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
     ):
-        super().__init__(repo_dir, static_analysis, get_system_meta_analysis_message(), agent_llm, parsing_llm)
+        super().__init__(repo_dir, StaticAnalysisResults(), get_system_meta_analysis_message(), agent_llm, parsing_llm)
         self.project_name = project_name
 
         self.meta_analysis_prompt = PromptTemplate(
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -112,24 +112,36 @@ class DiagramGenerator:
     def pre_analysis(self):
         analysis_start_time = time.time()
 
-        # When force_full is True, skip the cache to perform a full static analysis
-        if self.force_full_analysis:
-            logger.info("Force full analysis: skipping static analysis cache")
-            self.static_analysis = get_static_analysis(self.repo_location, skip_cache=True)
-        else:
-            self.static_analysis = get_static_analysis(self.repo_location)
-        static_analysis = self.static_analysis
+        # Initialize LLMs before spawning threads so both share the same instances
+        agent_llm, parsing_llm = initialize_llms()
+
+        self.meta_agent = MetaAgent(
+            repo_dir=self.repo_location,
+            project_name=self.repo_name,
+            agent_llm=agent_llm,
+            parsing_llm=parsing_llm,
+        )
+        self._monitoring_agents["MetaAgent"] = self.meta_agent
+
+        # Run static analysis and meta analysis in parallel
+        with ThreadPoolExecutor(max_workers=2) as executor:
+            skip_cache = self.force_full_analysis
+            if skip_cache:
+                logger.info("Force full analysis: skipping static analysis cache")
+            static_future = executor.submit(get_static_analysis, self.repo_location, skip_cache=skip_cache)
+            meta_future = executor.submit(self.meta_agent.analyze_project_metadata)
+
+            static_analysis = static_future.result()
+            meta_context = meta_future.result()
+
+        self.static_analysis = static_analysis
 
         # --- Capture Static Analysis Stats ---
         static_stats: dict[str, Any] = {"repo_name": self.repo_name, "languages": {}}
-
-        # Use ProjectScanner to get accurate LOC counts via tokei
         scanner = ProjectScanner(self.repo_location)
         loc_by_language = {pl.language: pl.size for pl in scanner.scan()}
-
         for language in static_analysis.get_languages():
             files = static_analysis.get_source_files(language)
-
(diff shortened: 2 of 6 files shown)


agents/meta_agent.py
```diff
@@ -19,12 +19,11 @@ class MetaAgent(CodeBoardingAgent):
     def __init__(
         self,
         repo_dir: Path,
-        static_analysis: StaticAnalysisResults,
         project_name: str,
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
     ):
-        super().__init__(repo_dir, static_analysis, get_system_meta_analysis_message(), agent_llm, parsing_llm)
+        super().__init__(repo_dir, StaticAnalysisResults(), get_system_meta_analysis_message(), agent_llm, parsing_llm)
```
Svilen-Stefanov, line 26: Hmm, this looks a bit odd. The reason - `CodeBoardingAgent` seems to require a static analysis result but it seems like we don't want the `MetaAgent` to require it iiuc. 
What do you think?

diagram_analysis/diagram_generator.py
```diff
@@ -112,43 +112,43 @@ def _run_health_report(self, static_analysis: StaticAnalysisResults) -> None:
...
-        if self.force_full_analysis:
-            logger.info("Force full analysis: skipping static analysis cache")
-            self.static_analysis = get_static_analysis(self.repo_location, skip_cache=True)
-        else:
-            self.static_analysis = get_static_analysis(self.repo_location)
-        static_analysis = self.static_analysis
+        # Initialize LLMs before spawning threads so both share the same instances
+        agent_llm, parsing_llm = initialize_llms()
+
+        self.meta_agent = MetaAgent(
```
Svilen-Stefanov, line 118: I'm wondering about a use case I saw yesterday - what happens now if there's an issue with the LLM (the setup is incorrect / there's something with the provider breaking things), does that affect the static analysis outcome or the general flow? 
One thing I quite like with the previous setup is that I can actually refresh the analysis and even if the LLM generation fails, all the health stuff is updated as that's based entirely on the static analysis. 

Can you check about the behavior after this change - in particular wrt error handling?

Svilen-Stefanov (commented):
This looks much cleaner now. My main 2 questions are:
1. What's the behavior after this change wrt error handling and what's the general impact (speed improvement)?
2. Is the `MetaAgent` relying on the static analysis? I don't quite like the strange pattern of passing an empty static analysis result to from the `MetaAgent` to the `CodeBoardingAgent`.

diagram_analysis/diagram_generator.py
brovatten: I would that the opposite opinion, and say it's kinda annoying to run the whole static analysis then to realize you have a not proper LLM key. Is it cached? Yes. But expected behaviour if it runs is that your configs are correct.

For the health stuff, it sounds like that should be a separate entrypoint, rather than relying on our main flow?

If the meta agent fails now, it will wait for the results of the static analysis (caching included)

agents/meta_agent.py
brovatten: The CodeBoardingAgent (that all agents inherits from) has the static_analysis attribute. This is the state, and then the ReferenceResolverMixin (that CodeBoardingAgent) inherits from use the static_analysis. 

I couldn't come up with a good way to change this without a big refactor. My first thought was to allow static analysis be None (could work), or to not require static_analysis for the base class (but this is not ideal)

What would u prefer?

brovatten: The speed-up is quite intuitive, the fastest execution time of whichever of the two would be diminished. LLM calls have basically no local CPU impact.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: I agree that it's better to clarify the setup up front but I also think we should be able to provide some benefit in case of errors - notice that as I mentioned, it doesn't have to be setup only, there might be issues with the provider. In such cases, isn't it better to tell ppl - sorry, that didn't work but look, we managed to notify you about health issues (or other static analysis analytics we have).

So my main request is for us not to break that. To rephrase the question - what happens if the LLM thread throws an exception, is the static analysis interrupted?

agents/meta_agent.py
Svilen-Stefanov: Let's chat about the refactor in person, it'll probably be easier. 
About the speedup, I'd like to see some numbers :D

diagram_analysis/diagram_generator.py
brovatten: I disagree, I think to the user, they will not know that static analysis is cached. It will be more seamless for them to get provider error, switch it in settings, and then run it without errors.

On your second questioned, I answered this above : If the meta agent fails now, it will wait for the results of the static analysis (caching included)

diagram_analysis/diagram_generator.py
```diff
@@ -112,43 +112,43 @@ def _run_health_report(self, static_analysis: StaticAnalysisResults) -> None:
...
+            repo_dir=self.repo_location,
+            project_name=self.repo_name,
+            agent_llm=agent_llm,
+            parsing_llm=parsing_llm,
+        )
+        self._monitoring_agents["MetaAgent"] = self.meta_agent
+
+        # Run static analysis and meta analysis in parallel
+        with ThreadPoolExecutor(max_workers=2) as executor:
```
ivanmilevtues, line 127: This is what I wanted to see.

agents/meta_agent.py
ivanmilevtues: We need the CBA for the invoke methods.

Svilen-Stefanov: @brovatten let's keep this for now but can you create a ticket to look into this later?

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Based on our chat, it seems like we are on the same page here. My request here is to really try this and see if it works - a simple way to test this is to not provide an API key, run the analysis and see if a `health_report.json` is generated. 
If it isn't, please let me know and we can see if we want to merge regardless or fix it.

PR #189, opened 2026-02-10
Unify analysis result into a single analysis.json file

Svilen-Stefanov:
This PR merges all of our generated files into a single `analysis.json` file. What this comes with:
1. I needed to modify `main.py` as we no longer need to specify the name of the file we need to modify (now everything will modify `analysis.json`)
2. As changes are quite big, I want to reduce the scope here and not overcomplicate things from the start so for now, I only update the `analysis.json` when the full analysis is complete. 
    - for full analysis this means we need for the entire analysis to complete (we can lose the entire analysis if the final level fails)
    - for the partial and incremental analysis, not much should change
    - I plan to create a follow up that would ideally improve this so that the `analysis.json` is created per level or updated in parallel. Let's discuss this separately if you agree we should make these changes incrementally.
3. Added `component_id` to `analysis.json` to avoid recursion issues where a subcomponent has the same name as the parent.
4. Added an example `analysis.json` and removed the old one. We might want to use better models to actually update it properly because I think the current analysis looks somewhat weak :D 
5. Fixed the integration tests expectations I forgot to fix in my previous PR (just one file, which is unrelated) - I can create a separate PR if you think it's cleaner. 

Performed tests:
- so far I've only tested it on CodeBoarding with full analysis (depth 1 and depth 3)
- partial analysis - ran analysis with depth 1 and then updated a component
- tested both with remote and local setup

Some things/issues I noticed (not necessarily related to the PR):
1. With the weaker models, I see in the currently generated `analysis.json` that it often happens that a component is marked as `canExpand=True` but then the expansion is just a replication of the parent - super useless. It would be cool to fix that.
2. The LLMs also seem to hallucinate relations - I think we resolved assigned_files hallucinations but I saw that some relationships are hallucinated and maybe that can be fixed next as well.

### DIFF
--- a/diagram_analysis/analysis_json.py
+++ b/diagram_analysis/analysis_json.py
@@ -103,14 +103,40 @@ def from_analysis_to_json(
     return json.dumps(data, indent=2)
 
 
+def _compute_depth_level(
+    sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None,
+) -> int:
+    """Compute the maximum depth level from the sub_analyses structure.
+
+    Returns 1 if there are no sub-analyses (root only), 2 if there is one level of
+    sub-analyses, etc.
+    """
+    if not sub_analyses:
+        return 1
+
+    max_child_depth = 1
+    for _, (sub_analysis, _) in sub_analyses.items():
+        # Check if any of this sub-analysis's components themselves have sub-analyses
+        child_depth = 1
+        for comp in sub_analysis.components:
+            if comp.name in sub_analyses:
+                child_depth = 2
+                break
+        max_child_depth = max(max_child_depth, child_depth)
+
+    return 1 + max_child_depth
+
+
 def build_unified_analysis_json(
     analysis: AnalysisInsights,
     expandable_components: list[Component],
     repo_name: str,
-    depth_level: int,
     sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
 ) -> str:
-    """Build the full unified analysis JSON with metadata and nested sub-analyses."""
+    """Build the full unified analysis JSON with metadata and nested sub-analyses.
+
+    The depth_level metadata is computed automatically from the sub_analyses structure.
+    """
     components_json = [
         from_component_to_json_component(c, expandable_components, sub_analyses, None) for c in analysis.components
     ]
@@ -119,7 +145,7 @@ def build_unified_analysis_json(
         metadata=AnalysisMetadata(
             generated_at=datetime.now(timezone.utc).isoformat(),
             repo_name=repo_name,
-            depth_level=depth_level,
+            depth_level=_compute_depth_level(sub_analyses),
         ),
         description=analysis.description,
         components=components_json,
--- a/diagram_analysis/incremental/io_utils.py
+++ b/diagram_analysis/incremental/io_utils.py
@@ -74,7 +74,6 @@ def save_analysis(
     expandable_components: list[str] | None = None,
     sub_analyses: dict[str, AnalysisInsights] | None = None,
     repo_name: str = "",
-    depth_level: int = 1,
 ) -> Path:
     """Save the analysis to a unified analysis.json file.
 
@@ -97,8 +96,6 @@ def save_analysis(
             # Preserve metadata from existing file
             if not repo_name and "metadata" in existing_data:
                 repo_name = existing_data["metadata"].get("repo_name", "")
-            if depth_level == 1 and "metadata" in existing_data:
-                depth_level = existing_data["metadata"].get("depth_level", 1)
 
     # Convert sub_analyses dict to the format expected by build_unified_analysis_json
     sub_analyses_tuples: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None
@@ -115,7 +112,6 @@ def save_analysis(
                 analysis=analysis,
                 expandable_components=expandable,
                 repo_name=repo_name,
-                depth_level=depth_level,
                 sub_analyses=sub_analyses_tuples,
             )
         )
@@ -158,12 +154,10 @@ def save_sub_analysis(
     # Update the sub-analysis for this component
     sub_analyses[component_name] = sub_analysis
 
-    # Determine repo_name and depth_level from existing metadata
+    # Determine repo_name from existing metadata
     repo_name = ""
-    depth_level = 1
     if "metadata" in raw_data:
         repo_name = raw_data["metadata"].get("repo_name", "")
-        depth_level = raw_data["metadata"].get("depth_level", 1)
 
     # Determine which root components are expandable
     all_expandable = expandable_components or list(sub_analyses.keys())
@@ -174,7 +168,6 @@ def save_sub_analysis(
         expandable_components=all_expandable,
         sub_analyses=sub_analyses,
         repo_name=repo_name,
-        depth_level=depth_level,
     )
 
 
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -271,7 +271,6 @@ class DiagramGenerator:
                         analysis=analysis,
                         expandable_components=expanded_components,
                         repo_name=self.repo_name,
-                        depth_level=self.depth_level,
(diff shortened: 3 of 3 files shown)


diagram_analysis/incremental/component_checker.py
```diff
@@ -9,29 +9,25 @@
 from pathlib import Path
 
 from agents.agent_responses import AnalysisInsights
-from diagram_analysis.analysis_json import from_analysis_to_json
 from diagram_analysis.incremental.io_utils import load_sub_analysis
 from diagram_analysis.incremental.models import ChangeImpact
 from diagram_analysis.manifest import AnalysisManifest
-from output_generators.markdown import sanitize
```
ivanmilevtues, line 16: nice, this was causing problems with & sometimes.

diagram_analysis/incremental/updater.py
```diff
@@ -382,21 +382,16 @@ def _execute_update_components(self) -> bool:
...
                 continue
 
-            safe_name = sanitize(component_name)
-            sub_analysis_path = self.output_dir / f"{safe_name}.json"
-
-            if sub_analysis_path.exists():
-                from diagram_analysis.incremental.io_utils import load_sub_analysis, save_sub_analysis
+            sub_analysis = load_sub_analysis(self.output_dir, component_name)
+            if sub_analysis:
                 from diagram_analysis.incremental.path_patching import patch_sub_analysis
```
ivanmilevtues: local import

diagram_analysis/incremental/scoped_analysis.py
```diff
@@ -92,13 +91,7 @@ def handle_scoped_component_update(
...
     """
 
-    # Ensure this component is expanded (has a sub-analysis file)
-    safe_name = sanitize(component_name)
-    sub_path = output_dir / f"{safe_name}.json"
-    if not sub_path.exists():
-        return
-
-    # Load sub-analysis
+    # Load sub-analysis from the unified file
```
ivanmilevtues: this comment seems useless, but it is w.e.

diagram_analysis/analysis_json.py
```diff
@@ -1,41 +1,222 @@
...
+    depth_level: int,
+    sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
+) -> str:
+    """Build the full unified analysis JSON with metadata and nested sub-analyses."""
+    components_json = [
+        from_component_to_json_component(c, expandable_components, sub_analyses, None) for c in analysis.components
+    ]
+
+    unified = UnifiedAnalysisJson(
+        version=2,
```
ivanmilevtues: Why version 2 ;d, do we need this or it will be again like the current version never used :D

I suppose we can also remove the Version from `diagram_analysis/version.py`

diagram_analysis/analysis_json.py
```diff
@@ -1,41 +1,222 @@
...
+
+            for sub_comp_data in comp_data["components"]:
+                sub_component = Component(
+                    name=sub_comp_data["name"],
+                    description=sub_comp_data["description"],
+                    key_entities=sub_comp_data.get("key_entities", []),
+                    assigned_files=sub_comp_data.get("assigned_files", []),
+                    source_cluster_ids=sub_comp_data.get("source_cluster_ids", []),
+                )
+                sub_components.append(sub_component)
```
ivanmilevtues: This is the exact same code fragment as _extract_sub_analysis

Svilen-Stefanov: If you take a closer look at the previous changes, you can see that they are quite primitive - it was hardcoding the depth and processing to only lvl 3. Now I fixed it to be a recursive check for any level. I think there's still a lot of potential for improvement though... 
One of the issues that we still have here is that we refer to components by name, which is not unique and I think that this can potentially have many side effects. I won't fix this in this PR though.

output_generators/html.py
```diff
@@ -126,26 +131,28 @@ def generate_html_file(
...
 
-def component_header_html(component_name: str, link_files: List[Path]) -> str:
+def component_header_html(component_name: str, expanded_components: set[str]) -> str:
     """
     Generate an HTML header for a component with its name and a link to its details.
     """
     sanitized_name = sanitize(component_name)
-    if contains_json(sanitized_name, link_files):
+    if component_name in expanded_components:
         return f'<h3 id="{sanitized_name}">{component_name} <a href="./{sanitized_name}.html">[Expand]</a></h3>'
```
brovatten, line 156: Did you try yourself to look at the MD outputs?

diagram_analysis/analysis_json.py
```diff
@@ -1,41 +1,178 @@
-from typing import List
+import logging
+from datetime import datetime, timezone
```
brovatten, line 3: analysis_json.py  - we can not end a filename with fileextension

diagram_analysis/incremental/reexpansion.py
```diff
@@ -56,54 +54,39 @@ def reexpand_single_component(
...
+                logger.info(
+                    f"Component '{component_name}' sub-analysis has only renames, patching instead of re-expanding"
+                )
+
+                # Patch the sub-analysis
+                if patch_sub_analysis(
                     existing_sub_analysis,
-                    context.impact,
+                    context.impact.deleted_files if context.impact else [],
+                    context.impact.renames if context.impact else {},
```
brovatten, line 74: why did this get added here?

diagram_analysis/incremental/io_utils.py
```diff
@@ -3,64 +3,138 @@
...
+    return result[0]
+
+
+def save_analysis(
+    analysis: AnalysisInsights,
+    output_dir: Path,
+    expandable_components: list[str] | None = None,
+    sub_analyses: dict[str, AnalysisInsights] | None = None,
+    repo_name: str = "",
+    depth_level: int = 1,
```
brovatten: never do default args for this imo

.codeboarding/analysis.json
```diff
@@ -1,305 +1,2140 @@
...
+            }
+          ],
+          "components_relations": [
+            {
+              "relation": "requests code structure from",
+              "src_name": "UnusedCodeDiagnostics",
+              "dst_name": "CodebaseAnalyzer"
+            },
+            {
+              "relation": "provides code structure to",
```
brovatten, line 2040: holllyyyy this is now a long file... i hope there's another extension to visualize jsons in easier :D

diagram_analysis/analysis_json.py
Svilen-Stefanov: I didn't create the file, it already exists. I can rename it but this will introduce more diff in the PR. 
Generally, I suggest to separate improvements and refactoring as it reduces blast radius and makes it easier to track how files evolved over time.

diagram_analysis/incremental/reexpansion.py
```diff
@@ -56,54 +54,39 @@ def reexpand_single_component(
...
                 ):
-                    logger.info(
-                        f"Component '{component_name}' sub-analysis has only renames, patching instead of re-expanding"
-                    )
-
-                    # Patch the sub-analysis
-                    if patch_sub_analysis(
-                        existing_sub_analysis,
-                        context.impact.deleted_files if context.impact else [],
-                        context.impact.renames if context.impact else {},
```
Svilen-Stefanov, line 79: pointer

diagram_analysis/incremental/reexpansion.py
Svilen-Stefanov: It wasn't added, it was just moved - here's the reference: https://github.com/CodeBoarding/CodeBoarding/pull/189/changes#r2802904264

output_generators/html.py
Svilen-Stefanov: I looked at them but didn't spend time verifying much. I think the behavior is the same as before there, not sure if we still need these but I just checked they are still generated.

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: Hmm, good point, I don't like this much either. I think this was originally added to add the depth level to the `analysis.json` file but the depth is visible from the file already and is probably even incorrect in some cases.
I'll remove it.

PR #190, opened 2026-02-11
Using vscode/vscode-forks process as node

ivanmilevtues:
Using the vscode process as node instead of requiring our users to have node.
Fallback is still node.

### DIFF
--- a/vscode_constants.py
+++ b/vscode_constants.py
@@ -55,8 +55,11 @@ def update_command_paths(bin_dir):
                     cmd[0] = os.path.join(bin_path, cmd[0])
 
             # Apply Windows-specific node prefix for specified languages
+            # Use VSCode's bundled Node.js (passed via CODEBOARDING_NODE_PATH) so users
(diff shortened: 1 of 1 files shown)


vscode_constants.py
```diff
@@ -55,8 +55,11 @@ def update_command_paths(bin_dir):
                     cmd[0] = os.path.join(bin_path, cmd[0])
 
             # Apply Windows-specific node prefix for specified languages
+            # Use VSCode's bundled Node.js (passed via CODEBOARDING_NODE_PATH) so users
+            # don't need a separate Node.js/npm installation on Windows.
             if is_windows and key in node_languages:
-                cmd.insert(0, "node")
+                node_path = os.environ.get("CODEBOARDING_NODE_PATH", "node")
```
Svilen-Stefanov, line 61: Nice! For the local setup, are we relying on @brovatten's PR? I can't wait to unify this a bit and don't have these 2 things to support :D

PR #191, opened 2026-02-11
Add file summary for analyzed and not analyzed files

Svilen-Stefanov:


output_generators/html.py
ivanmilevtues, line 113: I would suggest to use the ids here instead of the names. I am romantisizing those ids quite a lot atm.

diagram_analysis/analysis_json.py
```diff
@@ -103,49 +129,26 @@ def from_analysis_to_json(
...
-    return 1 + max_child_depth
-
-
 def build_unified_analysis_json(
     analysis: AnalysisInsights,
     expandable_components: list[Component],
     repo_name: str,
+    depth_level: int,
     sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
+    file_coverage_summary: FileCoverageSummary | None = None,
```
ivanmilevtues, line 163: This shouldn't be nullable, the file_coverage_summary.  I think generally we should stop having the nullable types as they make it easier for errors and skipping things. I've noticed the agents are not as good at dealing with those.

If there are no files for the summary then the summary is all files are in gj all ;D

diagram_analysis/diagram_generator.py
```diff
@@ -104,46 +112,120 @@ def _run_health_report(self, static_analysis: StaticAnalysisResults) -> None:
...
+            if rel_str in analyzed_paths:
+                continue
+
+            reason = ignore_manager.categorize_file(Path(rel_str))
+            entry: dict = {"path": rel_str}
+            if reason:
+                entry["reason"] = reason
+                reason_counts[reason] = reason_counts.get(reason, 0) + 1
+            else:
+                reason_counts["other"] = reason_counts.get("other", 0) + 1
```
ivanmilevtues: sorry to be so nit picky, but I am getting to hate Agent code :D.

To me the categorize_file, should give a reason "other" rather than after categorizing the file we do a second categorization here for the other.

I don't really get the entry + reason_coutns, to me you can use DictCounter or something else afterwards.

This method should be very simple to understand but to me is overcomplicated for no reason :D.

Also for the reason_counts I think you can just use a default dict:
```python
ht = defaultdict(lambda:0, ht)
```
and then have just a reason_counts[reason] += 1
https://stackoverflow.com/questions/30356892/defaultdict-with-default-value-1


Genuinely sorry for the crash-out but this is a simple method how is it somehow hard to undertands☠️ 

To me it has to be something like:
for file in all_files:
    if unalized:

diagram_analysis/diagram_generator.py
```diff
@@ -104,46 +112,120 @@ def _run_health_report(self, static_analysis: StaticAnalysisResults) -> None:
...
+
+        analyzed_paths: set[str] = set()
+        for src_file in static_analysis.get_all_source_files():
+            try:
+                p = Path(src_file)
+                if p.is_absolute():
+                    p = p.relative_to(self.repo_location)
+                analyzed_paths.add(str(p))
+            except (ValueError, TypeError):
+                analyzed_paths.add(str(src_file))
```
ivanmilevtues: Is there a need to have the paths as set[str] and not set[Path]?
If we can do just set of paths we can do directly analyzed_files = set(stat_analysis.get_all_src_files)
and then we do all_files = {Path(file_path) for file _path in scanner.all_text_files}

set intersection to get the unused files.

Then:
for file in unused_files:
  report.append(file, ignore_manager.categorize_file(file)) 
return report.

I think it is probably better to have the report to build itself, the diagram generator is getting really heavy and containing logic for report building and other things.

If we want to be super clean we can have a file report builder idk.

diagram_analysis/diagram_generator.py
```diff
@@ -104,46 +112,120 @@ def _run_health_report(self, static_analysis: StaticAnalysisResults) -> None:
...
+                "not_analyzed_by_reason": reason_counts,
+            },
+        }
+
+    def _write_file_coverage(self) -> None:
+        """Write file_coverage.json to output directory."""
+        if not self.file_coverage_data:
+            return
+
+        report = FileCoverageReport(
```
ivanmilevtues, line 135: Interesting to see how does this interact with partial analysis?

output_generators/html.py
Svilen-Stefanov: I love the idea but our infrastructure is not ready for IDs yet - I think I'll open a separate PR that will just refactor this because I assume it's gonna require some changes and testing. 
Consider that we still use the component name as an argument to main and this is used everywhere as identifier. I just didn't want to do this here / or in the other unification PRs already.

ivanmilevtues: yea I understand, but we should scrap that, and do the ids :D

ivanmilevtues: you can put it in next PR.

PR #192, opened 2026-02-12
Cache metaagent & rework static caching

brovatten:
This PR introduces a unified caching architecture and adds SQLite-backed caching for Meta agent analysis. SQLite was chosen because it is lightweight, embedded, zero-ops, and provides safe atomic persistence for local/repeated runs.

  - Centralized cache layer under cache/ with clear public APIs (meta_cache, static_cache).
  - Private cache internals moved to underscored modules (_paths, _sqlite_store) to keep boundaries clean.
  - Kept orchestration in domain resolvers (diagram_analysis/meta_context_resolver.py, static_analyzer/cache_resolver.py) so cache classes do not run workflows.
  - Added stable cache identity rules:
      - Meta cache invalidates on model/prompt/dependency/tree changes.
      - Static cache keys include language + project path hash to avoid
        monorepo collisions.
  - Added typo-tolerant metadata invalidation (small README/doc edits
    don’t force recompute).

### DIFF
--- a/cache/meta_cache.py
+++ b/cache/meta_cache.py
@@ -15,7 +15,7 @@ from repo_utils.project_manifests import (
     COMMON_DEPENDENCY_GLOBS,
     COMMON_DEPENDENCY_SUBDIRS,
 )
-from utils import sha256_hexdigest
+from utils import safe_read_text, sha256_hexdigest
 
 logger = logging.getLogger(__name__)
 
@@ -120,15 +120,6 @@ class MetaCacheIdentity:
         }
 
 
-def _safe_read_text(path: Path) -> str:
-    """Read file text safely and return empty text on failure."""
-    try:
-        return path.read_text(encoding="utf-8", errors="ignore")
-    except (OSError, UnicodeDecodeError) as e:
-        logger.debug("Failed to read text for meta cache snapshot from %s: %s", path, e)
-        return ""
-
-
 def _normalize_model_id(agent_llm: BaseChatModel) -> str:
     """Extract a stable model identifier from the configured LLM."""
     for attr in ("model_name", "model", "model_id"):
@@ -174,7 +165,7 @@ def _hash_dependency_inputs(repo_dir: Path, ignore_manager: RepoIgnoreManager) -
     payload_parts: list[str] = []
     for dep_path in _collect_dependency_paths(repo_dir, ignore_manager):
         rel = dep_path.relative_to(repo_dir).as_posix()
-        payload_parts.append(f"{rel}\n{_safe_read_text(dep_path)}")
+        payload_parts.append(f"{rel}\n{safe_read_text(dep_path)}")
     return sha256_hexdigest("\n---\n".join(payload_parts))
 
 
@@ -205,7 +196,7 @@ def _build_docs_manifest(repo_dir: Path, doc_paths: list[Path]) -> dict[str, obj
     priority_files: list[str] = []
     for doc in doc_paths:
         rel = doc.relative_to(repo_dir).as_posix()
-        text = _safe_read_text(doc)
+        text = safe_read_text(doc)
         file_hashes[rel] = sha256_hexdigest(text)
 
         rel_lower = rel.lower()
--- a/utils.py
+++ b/utils.py
@@ -106,6 +106,15 @@ def sanitize(name: str) -> str:
     return re.sub(r"\W+", "_", name)
 
 
+def safe_read_text(path: Path) -> str:
+    """Read file text safely, returning empty string on failure."""
+    try:
+        return path.read_text(encoding="utf-8", errors="ignore")
+    except (OSError, UnicodeDecodeError) as e:
+        logger.debug("Failed to read %s: %s", path, e)
+        return ""
+
+
 def sha256_hexdigest(text: str) -> str:
     """Return SHA-256 hex digest for a UTF-8 string."""
     return hashlib.sha256(text.encode("utf-8")).hexdigest()


cache/static_cache.py
```diff
@@ -0,0 +1,58 @@
...
+    """Public static-analysis cache facade built on top of AnalysisCacheManager."""
+
+    def __init__(
+        self,
+        repo_path: Path,
+        cache_dir: Path | None = None,
+        manager: AnalysisCacheManager | None = None,
+    ):
+        self.repo_path = Path(repo_path)
+        self.cache_dir = Path(cache_dir) if cache_dir is not None else get_repo_cache_dir(self.repo_path)
```
Svilen-Stefanov, line 13: Shouldn't this just call `ensure_repo_cache_dir` instead? 

Edit: I suggested even better refactoring. You can check what that would look like first.

cache/_paths.py
```diff
@@ -0,0 +1,45 @@
...
+CACHE_DIRNAME = "cache"
+META_CACHE_DB_FILENAME = "meta_cache.sqlite"
+
+
+def get_repo_cache_dir(repo_dir: Path) -> Path:
+    """Return the standard repository-local cache directory."""
+    return repo_dir / CACHE_ROOT_DIRNAME / CACHE_DIRNAME
+
+
+def ensure_repo_cache_dir(repo_dir: Path) -> Path:
```
Svilen-Stefanov, line 16: Can we have this funtion only and get rid of `get_repo_cache_dir`? Is there a use case where we want to only get the path but not check if the path exists? 
I think we will always need the path to exists if we use it so maybe just keep this function?

cache/_paths.py
```diff
@@ -0,0 +1,45 @@
...
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    return cache_dir
+
+
+def get_meta_cache_db_path(repo_dir: Path) -> Path:
+    """Return the SQLite file path used by metadata cache."""
+    return ensure_repo_cache_dir(repo_dir) / META_CACHE_DB_FILENAME
+
+
+def build_static_client_cache_id(language: str, project_path: Path) -> str:
```
Svilen-Stefanov, line 28: Instead of `language: str`, can't we use the Enum here? Same for the function below.

cache/_paths.py
```diff
@@ -0,0 +1,45 @@
...
+    - project_path hash: isolates monorepo subprojects sharing one language.
+    """
+    normalized_path = str(project_path.resolve())
+    path_hash = hashlib.sha1(normalized_path.encode("utf-8")).hexdigest()[:12]
+    return f"{language.lower()}_{path_hash}"
+
+
+def get_static_incremental_cache_path(cache_dir: Path, language: str, project_path: Path) -> Path:
+    """Return per-client incremental static cache path under the given cache directory."""
+    cache_dir.mkdir(parents=True, exist_ok=True)
```
Svilen-Stefanov, line 43: It looks strange we're doing this here. Overall, the module here seems a bit suboptimal. 
For this function, the user should first get the cache dir and then pass it to this function (so an external caller would call one function from this module only to pass the result again to another function in this module). 

I think the right design for this would be to have:
```
from abc import ABC, abstractmethod
from pathlib import Path
from typing import TypeVar, Generic

T = TypeVar('T')

class BaseCache(ABC, Generic[T]):
    DEFAULT_CACHE_DIR = Path(".codeboarding/.cache")

    def __init__(self, filename: str, custom_dir: Path = None): 
        # we can even get rid of the custom_dir
        self.cache_dir = custom_dir or self.DEFAULT_CACHE_DIR
        self.file_path = self.cache_dir / filename
        
        # Ensure the directory exists immediately upon initialization
        self.cache_dir.mkdir(parents=True, exist_ok=True)

    @abstractmethod
    def load(self) -> T:
        pass

    # Note: store is implemented in subclasses due to varying arguments


class MetaCache(BaseCache[MetaAnalysisInsights]):
    def __init__(self):
        # The subclass only cares about its specific file
        super().__init__(filename="meta_cache.sqlite")

    def load(self) -> MetaAnalysisInsights | None:
        # Implementation...
        return {}

   def store(self, ...):
        ...
```

Then we only ever use it like:
```
meta_cache = MetaCache()
meta_cache.store()
# or
meta_cache_result = meta_cache.load()
```

Everything else seems a bit redundant / misplaced

cache/_sqlite_store.py
```diff
@@ -0,0 +1,77 @@
...
+class SQLiteCacheStore:
+    """Minimal SQLite key-value store for scoped cache records."""
+
+    def __init__(self, db_path: Path, table_name: str):
+        self.db_path = db_path
+        self.table_name = _validate_identifier(table_name)
+        self.db_path.parent.mkdir(parents=True, exist_ok=True)
+        self._init_db()
+
+    def _conn(self) -> sqlite3.Connection:
```
Svilen-Stefanov, line 26: The first question I have here is how often are we querying this? Based on the code, it seems we will be calling this every time for `load_latest`, which basically creates a new connection on each call. 

The best practice here is to have a somewhat persistent connection and reuse that instead of creating a new one every time when a request is made. I think this has to change - the AI agents can do better :D

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+                self.tree_hash,
+                self.model_id,
+                self.prompt_version,
+                hashlib.sha256(self.docs_text.encode("utf-8")).hexdigest(),
+            ]
+        )
+        return hashlib.sha256(raw.encode("utf-8")).hexdigest()
+
+
+def _sha256(text: str) -> str:
```
Svilen-Stefanov: This seems like a utils function - move it to a common place and remove the `_`, maybe modify the name to str_to_sha256 or smth.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+    return hashlib.sha256(text.encode("utf-8")).hexdigest()
+
+
+def _safe_read_text(path: Path) -> str:
```
Svilen-Stefanov: Same here.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+    try:
+        return path.read_text(encoding="utf-8", errors="ignore")
+    except Exception:
+        return ""
```
Svilen-Stefanov: Let's log something in case of error.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+        remaining = max_total_chars - used
+        clipped = text[:remaining]
+        chunks.append(f"## {rel}\n{clipped}")
+        used += len(clipped)
+    return "\n\n".join(chunks)
+
+
+def build_meta_snapshot(repo_dir: Path, agent_llm: object) -> MetaSnapshot:
+    repo_root = repo_dir.resolve()
+    ignore_manager = RepoIgnoreManager(repo_root)
```
Svilen-Stefanov: It's a bit strange that we create a separate `RepoIgnoreManager` here. Seems like this might need to be a singleton somewhere.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+        rel = doc.relative_to(repo_dir).as_posix()
+        text = _safe_read_text(doc)
+        remaining = max_total_chars - used
+        clipped = text[:remaining]
+        chunks.append(f"## {rel}\n{clipped}")
+        used += len(clipped)
+    return "\n\n".join(chunks)
+
+
+def build_meta_snapshot(repo_dir: Path, agent_llm: object) -> MetaSnapshot:
```
Svilen-Stefanov: All of these functions seem to belong to either `MetaSnapshot` or `MetaCache`. They kind of don't make sense outside of the classes I think. Why do we need this method for example instead of just instantiating `MetaSnapshot` directly?

diagram_analysis/meta_context_resolver.py
```diff
@@ -0,0 +1,30 @@
...
+from pathlib import Path
+from typing import Any
+
+from agents.meta_agent import MetaAgent
+from cache.meta_cache import MetaCache, build_meta_snapshot
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_meta_context(repo_dir: Path, meta_agent: MetaAgent, agent_llm: object) -> Any:
```
Svilen-Stefanov: This seems to belong in `agents.meta_agent` / under `MetaAgent` directly. I'd even say, this should probably replace the current implementation of `analyze_project_metadata` (when you analyze it, you take the cache if the cache exists). We can add an attribute that flushes the cache if provided explicitly.

This will probably reduce the changes overall as I see that you've mostly replaced all callers of `analyze_project_metadata` to now use `resolve_meta_context`. If you instead change the implementation of `analyze_project_metadata` and make it call another internal function, things will become much simpler I think.

I don't like the return type of `Any`.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+META_CACHE_TABLE_NAME = "meta_cache"
+
+
+@dataclass(frozen=True)
+class MetaSnapshot:
+    """
+    Snapshot of metadata inputs used to decide whether meta-analysis can be reused.
+
+    Identity keys:
+    - scope: isolates cache by repository path.
```
Svilen-Stefanov, line 34: Do we need this? Shouldn't the scope/path always be the same? We seem to always pass `str(repo_root)`, which seems unnecessary.

Also most of the comments here are a bit too vague. It would be better to explain more (e.g what exactly do we get from the docs / deps). I suggested also that we should probably add some doc strings to the functions that implement these.

cache/meta_cache.py
```diff
@@ -0,0 +1,242 @@
...
+    for path in repo_dir.rglob("*"):
+        if ignore_manager.should_ignore(path):
+            continue
+        if path.is_file():
+            rel_paths.append(path.relative_to(repo_dir).as_posix())
+    rel_paths.sort()
+    return _sha256("\n".join(rel_paths))
+
+
+def _compute_docs_text(repo_dir: Path, ignore_manager: RepoIgnoreManager, max_total_chars: int = 400_000) -> str:
```
Svilen-Stefanov: I think these functions can be documented - here we seem to do some magic clipping etc. It would be good to know the assumptions we're making here - aka, we can document briefly in what cases we consider the docs the same and in what different. 
From what I see, we take all the content of the first couple of files and  only consider the first `400_000` chars. So imagine there's 10 huge files, each with `400_000` chars. This means that we effectively only consider changes to the first file. 
This seems strange. It almost feel better to take the first X lines of each file and check those. That's kind of what I mean by assumptions. 

Applies to all compute functions.

cache/meta_cache.py
brovatten: Here we have error message that's specific to the meta agent

cache/meta_cache.py
brovatten: It's not used anywhere else, so should we really move it to a utils before that?

brovatten: That's like moving a constant to a constant file before multi file access

Svilen-Stefanov: It is - refer to `repo_utils/__init__.py` - `get_repo_state_hash`.

cache/meta_cache.py
Svilen-Stefanov: What do you mean? You're gonna add a logging message that's specific to the meta agent? Do we need it to be specific? 

Look for `.read_text(encoding="utf-8")` we use it in so many different places in the repo already.

cache/meta_cache.py
brovatten: It's not that clean to pass it. Now instead, both instances of RepoIgnoreManager relies on the same config classes

Svilen-Stefanov: I'm not saying to necessarily pass it. Take a look at what I mentioned above and how Python implements the Singleton pattern. It is the better way to reuse objects and reduce memory space.

cache/meta_cache.py
brovatten: Done

cache/meta_cache.py
brovatten: Done

cache/meta_cache.py
brovatten: Done, altho memory space is tiny here

cache/meta_cache.py
brovatten: Done

cache/_sqlite_store.py
brovatten: Overoptimization imo, we do very few reads

cache/_paths.py
brovatten: We discussed this in chat

diagram_analysis/meta_context_resolver.py
brovatten: Done

agents/tools/external_deps.py
```diff
@@ -23,30 +28,8 @@ class ExternalDepsTool(BaseRepoTool):
...
-        # Node.js / TypeScript specific
-        "package.json",
-        "package-lock.json",
-        "yarn.lock",
-        "pnpm-lock.yaml",
-        "bun.lockb",
-        "tsconfig.json",  # TypeScript compiler configuration (not dependencies, but relevant)
-    ]
+    # Shared source of truth for dependency discovery patterns.
+    DEPENDENCY_FILES: List[str] = list(COMMON_DEPENDENCY_FILES)
```
ivanmilevtues, line 32: we use list and |None instead of Optional for typing this file is outdated so please repormpt for the types.

The DEPENDENCY_FILES variable is no useles, use direcctly the common depdency files?

Which seems to be the case anyway, just remove it as it is unsued.

cache/_sqlite_store.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Any
+
+
+def _validate_identifier(identifier: str) -> str:
+    if not identifier.replace("_", "").isalnum():
+        raise ValueError(f"Invalid SQLite identifier: {identifier}")
+    return identifier
+
+
+class SQLiteCacheStore:
```
ivanmilevtues, line 16: I still think it is nicer if we use sqlalchemy and have model classes for the tables.

repo_utils/project_manifests.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import annotations
+
+# Shared manifest/build-configuration filenames used across:
+# - dependency discovery (ExternalDepsTool + meta cache invalidation)
+# - language-specific project scanners (TypeScript/Java)
+#
+# Keep these constants centralized to avoid drift between scanners and cache/tools.
+
+# Python ecosystem
+PYTHON_DEPENDENCY_FILES: tuple[str, ...] = (
```
ivanmilevtues, line 10: the proejct-manifests.py doesn't explain to me what this is I'd name this liek project_build_file_constants or something liek that idk. It is a nit pick

agents/meta_agent.py
```diff
@@ -45,3 +47,30 @@ def analyze_project_metadata(self) -> MetaAnalysisInsights:
...
         logger.info(f"[MetaAgent] Completed metadata analysis for project: {analysis.llm_str()}")
         return analysis
+
+    def get_meta_context(
+        self,
+        force_refresh: bool = False,
+        agent_llm: BaseChatModel | None = None,
+    ) -> MetaAnalysisInsights:
+        """Return cached metadata context or recompute and persist it."""
+        effective_llm = agent_llm or self.agent_llm
```
ivanmilevtues, line 58: why do we need effective_llm?
I don't get it add a comment with explanation, to me it is alwasy self.agent_llm as that is what is passed to the CodeBoarding agent in super.

agents/tools/dependency_patterns.py
```diff
@@ -0,0 +1,12 @@
...
+
+from repo_utils.project_manifests import COMMON_DEPENDENCY_FILES
+
+# Shared dependency-discovery patterns used by both:
+# - ExternalDepsTool (tooling output for agents)
+# - MetaAgentCache dependency fingerprinting (cache invalidation inputs)
+#
+# Keeping a single source of truth prevents drift where one consumer
+# updates dependency file patterns but the other does not.
+COMMON_DEPENDENCY_SUBDIRS: tuple[str, ...] = ("requirements", "deps", "dependencies", "env")
```
ivanmilevtues: thise are not for all tech stacks, but rather for python?

agents/tools/dependency_patterns.py
```diff
@@ -0,0 +1,12 @@
...
+COMMON_DEPENDENCY_GLOBS: tuple[str, ...] = ("*.txt", "*.yml", "*.yaml", "*.toml")
```
ivanmilevtues: xml should be part of these i think and maybe possibly json idk, (package.json)

PR #193, opened 2026-02-12
Coordinate concurrent access to analysis.json and update it per level

Svilen-Stefanov:
This PR extends https://github.com/CodeBoarding/CodeBoarding/pull/189 and supports:
1. Parallel component analysis while still storing everything under a single file `analysis.json`
2. Update the `analysis.json` iteratively and not just at the end of the analysis

### DIFF
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -17,13 +17,12 @@ from agents.meta_agent import MetaAgent
 from agents.planner_agent import plan_analysis
 from agents.llm_config import initialize_llms
 from diagram_analysis.analysis_json import (
-    build_unified_analysis_json,
     FileCoverageReport,
     FileCoverageSummary,
     NotAnalyzedFile,
 )
 from diagram_analysis.file_coverage import FileCoverage
-from diagram_analysis.incremental.io_utils import AnalysisFileStore, save_analysis
+from diagram_analysis.incremental.io_utils import save_analysis
 from diagram_analysis.manifest import (
     build_manifest_from_analysis,
     save_manifest,
@@ -335,17 +334,15 @@ class DiagramGenerator:
                     not_analyzed_by_reason=s["not_analyzed_by_reason"],
                 )
 
-            # Final write of unified analysis.json (uses build directly for full expandable fidelity)
-            store = AnalysisFileStore(Path(self.output_dir))
+            # Final write of unified analysis.json
             analysis_path = str(
-                store.write_raw(
-                    build_unified_analysis_json(
-                        analysis=analysis,
-                        expandable_components=expanded_components,
-                        repo_name=self.repo_name,
-                        sub_analyses=all_sub_analyses,
-                        file_coverage_summary=file_coverage_summary,
-                    )
+                save_analysis(
+                    analysis=analysis,
+                    output_dir=Path(self.output_dir),
+                    expandable_components=[c.name for c in expanded_components],
+                    sub_analyses={name: sub for name, (sub, _) in all_sub_analyses.items()},
+                    repo_name=self.repo_name,
+                    file_coverage_summary=file_coverage_summary,
                 )
             )
 
--- a/diagram_analysis/incremental/io_utils.py
+++ b/diagram_analysis/incremental/io_utils.py
@@ -1,9 +1,14 @@
 """
 I/O utilities for incremental analysis.
 
-This module provides the ``AnalysisFileStore`` class – a coordinated
-reader/writer for the unified ``analysis.json`` file – along with thin
-free-function wrappers that preserve the original public API.
+This module provides coordinated read/write access to the unified
+``analysis.json`` file through **free functions** (``save_analysis``,
+``load_analysis``, etc.).
+
+Internally a singleton ``_AnalysisFileStore`` per ``output_dir`` owns the
+``FileLock`` and in-memory cache.  External code should **never** instantiate
+``_AnalysisFileStore`` directly – always use the free functions which route
+through the module-level registry.
 
 The unified format stores all analysis data (root + sub-analyses) in a single
 analysis.json file with nested components.
@@ -17,6 +22,7 @@ from filelock import FileLock
 
 from agents.agent_responses import AnalysisInsights, Component
 from diagram_analysis.analysis_json import (
+    FileCoverageSummary,
     build_unified_analysis_json,
     parse_unified_analysis,
 )
@@ -24,15 +30,13 @@ from diagram_analysis.analysis_json import (
 logger = logging.getLogger(__name__)
 
 
-class AnalysisFileStore:
-    """Coordinated reader/writer for ``analysis.json`` with file locking and caching.
+class _AnalysisFileStore:
+    """Coordinated reader/writer for ``analysis.json`` with file locking.
 
     All concurrent access to a given ``analysis.json`` should go through the
-    same ``AnalysisFileStore`` instance (or the module-level free functions
-    which share instances via ``_get_store``).  The store owns:
-
-    * the ``FileLock`` that serialises writes,
-    * an in-memory cache that is invalidated on every write.
+    same ``_AnalysisFileStore`` instance (or the module-level free functions
+    which share instances via ``_get_store``).  The store owns the
+    ``FileLock`` that serialises writes across threads and processes.
     """
 
     def __init__(self, output_dir: Path) -> None:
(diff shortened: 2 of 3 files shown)


diagram_analysis/incremental/io_utils.py
ivanmilevtues: Generally speaking shouldn't we have a write_file_thread_safe method so that we don't forget lock/unlock the file. It should probably also be a with block unsure

diagram_analysis/incremental/io_utils.py
```diff
@@ -120,6 +131,27 @@ def save_analysis(
...
+
+    If sub_analyses is not provided, attempts to preserve existing sub-analyses
+    from the current file on disk.
+    """
+    lock = FileLock(analysis_lock_path(output_dir), timeout=120)
+    with lock:
+        _invalidate_cache(output_dir)
+        return _save_analysis_unlocked(
+            analysis, output_dir, expandable_components, sub_analyses, repo_name, depth_level
+        )
```
ivanmilevtues: Exactly this lock, I would love to have as the method itself with the overload of the:
```python
import threading

class ThreadSafeFileWriter:
    def __init__(self, filename, mode="a"):
        self.filename = filename
        self.mode = mode
        self.lock = threading.Lock()
        self.file = None

    def __enter__(self):
        # acquire lock when entering the with block
        self.lock.acquire()
        self.file = open(self.filename, self.mode)
        print("File opened safely")
        return self.file

    def __exit__(self, exc_type, exc_val, exc_tb):
        # always close file and release lock
        if self.file:
            self.file.close()
            print("File closed safely")
        self.lock.release()

        # return False to propagate exceptions (recommended)
        return False
```
^^ ChatGpt prompt

ivanmilevtues: Proly has to be like the thing you did here, but you get the idea :D, as we want not just thread safety but rather process safety as well.

diagram_analysis/incremental/io_utils.py
ivanmilevtues: This read is without the lock, and we are writting to the file and not appending. Unsure what will happen if we start reading a file while someone finishes their partial analysis (so we write the update), what are the odds of things crashing as the file is lcoked by other process/reading invalid state.

I think proly we need also a thread safe thing for reading. Generally I think we need a custom open for our thing. open_analysis_locked. or w.e.

diagram_analysis/incremental/io_utils.py
```diff
@@ -141,34 +173,41 @@ def save_sub_analysis(
...
-    all_expandable = expandable_components or list(sub_analyses.keys())
-
-    return save_analysis(
-        root_analysis,
-        output_dir,
-        expandable_components=all_expandable,
-        sub_analyses=sub_analyses,
-        repo_name=repo_name,
-    )
+    lock = FileLock(analysis_lock_path(output_dir), timeout=120)
```
ivanmilevtues: Yea this thing we can see at many placese and it also makes the functions quite long, so I think that if we do our custom open block it will be prettier 😊

diagram_analysis/diagram_generator.py
```diff
@@ -263,17 +266,31 @@ def generate_analysis(self):
...
+            # Final write of unified analysis.json (uses build directly for full expandable fidelity)
             analysis_path = os.path.join(self.output_dir, "analysis.json")
-            with open(analysis_path, "w") as f:
-                f.write(
-                    build_unified_analysis_json(
-                        analysis=analysis,
-                        expandable_components=expanded_components,
-                        repo_name=self.repo_name,
-                        sub_analyses=all_sub_analyses,
+            lock = FileLock(analysis_lock_path(Path(self.output_dir)), timeout=120)
```
ivanmilevtues: comment for attention ;d, so you don't miss it

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
 
 logger = logging.getLogger(__name__)
 
-# Module-level cache for loaded unified data
-_unified_cache: dict[str, tuple[AnalysisInsights, dict[str, AnalysisInsights], dict]] = {}
 
+# ---------------------------------------------------------------------------
+# AnalysisFileStore – single point of coordination for analysis.json
+# ---------------------------------------------------------------------------
```
ivanmilevtues: No need for block comments over the name of the class + a class comment under the class name :D

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
-def _load_unified_data(output_dir: Path) -> tuple[AnalysisInsights, dict[str, AnalysisInsights], dict] | None:
-    """Load and cache the unified analysis.json file.
+    def __init__(self, output_dir: Path) -> None:
+        output_dir.mkdir(parents=True, exist_ok=True)
+        self._output_dir = output_dir
+        self._analysis_path = output_dir / "analysis.json"
+        self._lock = FileLock(output_dir / "analysis.json.lock", timeout=120)
+        self._cache: tuple[AnalysisInsights, dict[str, AnalysisInsights], dict] | None = None
+
+    # -- public readers -----------------------------------------------------
```
ivanmilevtues: this is not CC, I think it was something esle which is doing such comments :D

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+    * an in-memory cache that is invalidated on every write.
+    """
 
-def _load_unified_data(output_dir: Path) -> tuple[AnalysisInsights, dict[str, AnalysisInsights], dict] | None:
-    """Load and cache the unified analysis.json file.
+    def __init__(self, output_dir: Path) -> None:
+        output_dir.mkdir(parents=True, exist_ok=True)
+        self._output_dir = output_dir
+        self._analysis_path = output_dir / "analysis.json"
+        self._lock = FileLock(output_dir / "analysis.json.lock", timeout=120)
```
ivanmilevtues: 120 seems too big for a file read/write timeout. Maybe 10 secs?

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+                result = (root_analysis, sub_analyses, data)
+                self._cache = result
+                return result
+            except Exception as e:
+                logger.error(f"Failed to load unified analysis: {e}")
+                return None
+
+    def read_root(self) -> AnalysisInsights | None:
+        """Load just the root analysis."""
+        with self._lock:
```
ivanmilevtues: the read itself has a lock. pretty sure this is a bug or not?
with lock -> read -> with lock?

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+            result = self.read()
+            return result[0] if result else None
+
+    def read_sub(self, component_name: str) -> AnalysisInsights | None:
+        """Load a sub-analysis for a specific component."""
+        with self._lock:
```
ivanmilevtues: same double locking.

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+        repo_name: str = "",
+    ) -> Path:
+        """Write the full analysis to ``analysis.json`` with file locking.
+
+        If *sub_analyses* is not provided, existing sub-analyses on disk are
+        preserved.
+        """
+        with self._lock:
+            self._invalidate_cache()
+            return self._write_unlocked(analysis, expandable_components, sub_analyses, repo_name)
```
ivanmilevtues: it seems a bit odd to do:
with lock:
    self.write_unlocked :D

If something write locked?

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+    ) -> Path:
+        """Update a single sub-analysis within ``analysis.json``.
+
+        Acquires the file lock, loads the existing unified file, replaces the
+        sub-analysis for *component_name*, and re-writes the whole file.
+        """
+        with self._lock:
+            self._invalidate_cache()
+
+            existing = self.read()
```
ivanmilevtues, line 115: double lcoking.

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
+
+        Acquires the file lock and invalidates the cache.  Used by callers
+        that build the JSON payload themselves (e.g. the final write in
+        ``DiagramGenerator.generate_analysis``).
+        """
+        with self._lock:
+            self._invalidate_cache()
+            with open(self._analysis_path, "w") as f:
+                f.write(content)
+            return self._analysis_path
```
ivanmilevtues: I feel like the locks should be only in the read_raw/write_raw methods then everything that wraps them shouldn't care right?

diagram_analysis/incremental/io_utils.py
```diff
@@ -20,52 +23,224 @@
...
-            data = json.load(f)
 
-        root_analysis, sub_analyses = parse_unified_analysis(data)
-        result = (root_analysis, sub_analyses, data)
-        _unified_cache[key] = result
-        return result
-    except Exception as e:
-        logger.error(f"Failed to load unified analysis: {e}")
-        return None
+# ---------------------------------------------------------------------------
```
ivanmilevtues, line 204: I would reprompt the agent, to create io_utils package and put the block comment related methods in different files.

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: For the writes, the logs are correct. I think I did some of these myself, I'll fix the reads, take a look again when you can.

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: Hm, I think this is a bit unnecessary as these are only  ~30-40 lines.

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: This is technically not a problem as the same problem is reading / writing. The lock is mostly used for this to be thread safe so the double locking here is fine.

PR #195, opened 2026-02-13
Dont overwrite env

brovatten:
This is so much nicer to avoid having to copy paste the .env before rerunning install.

So u now need to delete your .env file to generate a new one (we rarely update the .env file so I think this is justified).

### DIFF
--- a/install.py
+++ b/install.py
@@ -329,6 +329,9 @@ def init_dot_env_file():
     print("Step: .env file creation started")
 
     env_file_path = Path(".env")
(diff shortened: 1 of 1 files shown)


install.py
```diff
@@ -329,6 +329,9 @@ def init_dot_env_file():
     print("Step: .env file creation started")
 
     env_file_path = Path(".env")
+    if env_file_path.exists():
```
ivanmilevtues, line 332: good one.

PR #197, opened 2026-02-14
Replace component name with component

Svilen-Stefanov:
Currently, we can't handle the case well when a a parent and a child have the same name. Prior to the unification of analysis.json, we had the issue that a component could be opened infinitely, now this actually breaks the behavior as the analysis fails with an infinite loop. 
This PR fixes this and other redundancies by properly making use of unique component ids for almost everything in the code base.

### DIFF
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -1,9 +1,12 @@
 import abc
 import hashlib
+import logging
 from abc import abstractmethod
 from typing import get_origin, Optional
 
 from pydantic import BaseModel, Field
+
+logger = logging.getLogger(__name__)
 
 ROOT_PARENT_ID = "ROOT"
 COMPONENT_ID_BYTES = 8
@@ -183,8 +186,8 @@ class AnalysisInsights(LLMBaseModel):
         return title + body + relations
 
 
-def compute_component_id(parent_id: str, name: str, sibling_index: int = 0) -> str:
-    """Compute a deterministic component ID from parent ID, name, and sibling index.
+def hash_component_id(parent_id: str, name: str, sibling_index: int = 0) -> str:
+    """Hash a deterministic component ID from parent ID, name, and sibling index.
 
     Note:
         The ID is a compact, 64-bit prefix of SHA-256 (8 bytes -> 16 hex chars).
@@ -202,11 +205,19 @@ def assign_component_ids(analysis: AnalysisInsights, parent_id: str = ROOT_PAREN
     name_counts: dict[str, int] = {}
     for component in analysis.components:
         count = name_counts.get(component.name, 0)
-        component.component_id = compute_component_id(parent_id, component.name, count)
+        component.component_id = hash_component_id(parent_id, component.name, count)
         name_counts[component.name] = count + 1
 
-    # Assign relation IDs by looking up component names
-    name_to_id: dict[str, str] = {c.name: c.component_id for c in analysis.components}
+    # Assign relation IDs by looking up component names (first occurrence wins for duplicates)
+    name_to_id: dict[str, str] = {}
+    for c in analysis.components:
+        if c.name in name_to_id:
+            logger.warning(
+                f"Duplicate component name '{c.name}' found during ID assignment; "
+                f"relation lookup will use the first occurrence (ID: {name_to_id[c.name]})"
+            )
+        else:
+            name_to_id[c.name] = c.component_id
     for relation in analysis.components_relations:
         relation.src_id = name_to_id.get(relation.src_name, "")
         relation.dst_id = name_to_id.get(relation.dst_name, "")
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -263,14 +263,13 @@ class DiagramGenerator:
 
             # Get the initial components to analyze (deterministic, no LLM)
             current_level_components = plan_analysis(analysis)
-            root_expandable_components = list(current_level_components)
             logger.info(f"Found {len(current_level_components)} components to analyze at level 0")
 
             level = 0
             max_workers = min(os.cpu_count() or 4, 8)  # Limit to 8 workers max
 
             # Track components that were actually expanded (have sub-analysis)
-            expanded_components: list[Component] = []
+            actually_expanded: list[Component] = []
 
             # Collect all sub-analyses: component_id -> (AnalysisInsights, expandable_sub_components)
             all_sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] = {}
@@ -302,7 +301,7 @@ class DiagramGenerator:
                             comp_id, sub_analysis, new_components = future.result()
                             if comp_id and sub_analysis:
                                 all_sub_analyses[comp_id] = (sub_analysis, new_components)
-                                expanded_components.append(component)
+                                actually_expanded.append(component)
                             if new_components:
                                 next_level_components.extend(new_components)
                         except Exception as exc:
@@ -316,7 +315,6 @@ class DiagramGenerator:
                 save_analysis(
                     analysis=analysis,
                     output_dir=Path(self.output_dir),
-                    expandable_components=[c.component_id for c in root_expandable_components],
                     sub_analyses={cid: sub for cid, (sub, _) in all_sub_analyses.items()},
                     repo_name=self.repo_name,
                 )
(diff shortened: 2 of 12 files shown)


agents/agent_responses.py
```diff
@@ -177,6 +182,30 @@ def llm_str(self):
         return title + body + relations
 
 
+def compute_component_id(parent_id: str, name: str, sibling_index: int = 0) -> str:
+    """Compute a deterministic component ID from parent ID, name, and sibling index."""
+    raw = f"{parent_id}:{name}:{sibling_index}"
+    return hashlib.sha256(raw.encode()).hexdigest()[:16]
```
ivanmilevtues: We should be careful with teh 16 :D

diagram_analysis/incremental/io_utils.py
```diff
@@ -30,6 +31,28 @@
 logger = logging.getLogger(__name__)
 
 
+def _compute_expandable_components(
```
ivanmilevtues: we should proly rename what the planner agent does, as can expand makes more sense in this context here.

The planner agent gives us if it makes sense to analyze further so something like analyzable idk.

diagram_analysis/analysis_json.py
```diff
@@ -3,12 +3,21 @@
 
 from pydantic import BaseModel, Field
 
-from agents.agent_responses import Component, Relation, AnalysisInsights
+from agents.agent_responses import Component, Relation, AnalysisInsights, assign_component_ids
 
 logger = logging.getLogger(__name__)
 
 
+class RelationJson(Relation):
```
ivanmilevtues, line 11: Now with the excluded fields of the AnalysisInsights maybe these are obsolete, back in the day I didn't know that those exist.

I also didn't check can we double check if exclude=True, also excludes them from the prompt not only from the validation. (Intuitively it should skip it as pydantic is also an ai company now ;d)

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: Yeah, let's discuss this today, I wasn't entirely sure where / how we want to handle this - do we want the planner agent to look at the component and say if it could be expanded further (as a separate attribute) or can we rely on the source_cluster_ids only (clearly determine if something should be expanded or not)?

agents/agent_responses.py
Svilen-Stefanov: Yup, I refactored this to make sure that these are indeed unique.

diagram_analysis/analysis_json.py
Svilen-Stefanov: It doesn't exclude them from the prompt because we iterate over attributes manually using `extractor_str` and not using `model_dump()` and `model_dump_json()`, which are apparently where these would be excluded. 
We probably could write some custom logic in the `LLMBaseModel.extractor_str` method to make sure these are excluded but I'm not sure of the impact. Should we explore this further in a separate PR?

agents/agent_responses.py
```diff
@@ -177,6 +183,35 @@ def llm_str(self):
         return title + body + relations
 
 
+def compute_component_id(parent_id: str, name: str, sibling_index: int = 0) -> str:
```
brovatten: I think hash_component_id is a better name, despite that it also produces the "raw". Compute is good but hash is more descripitive of the actual function

diagram_analysis/incremental/io_utils.py
```diff
@@ -30,6 +31,28 @@
...
+    analysis: AnalysisInsights,
+    parent_had_clusters: bool,
+) -> list[Component]:
+    """Compute expandable components deterministically for one analysis level."""
+    return [c for c in analysis.components if should_expand_component(c, parent_had_clusters=parent_had_clusters)]
+
+
+def _build_component_lookup(
```
brovatten: doesn't private functions (_) here seem weird for a utils folder?

Also, since this only needs the analysisinights (and the func below), this seems like it should be a private function to that class, or a new class we define.

@ivanmilevtues what do u think?

diagram_analysis/incremental/io_utils.py
```diff
@@ -150,10 +173,12 @@ def _write_with_lock_held(
         file_coverage_summary: FileCoverageSummary | None = None,
     ) -> Path:
         """Write ``analysis.json`` — caller must already hold ``self._lock``."""
-        # Build expandable component list
-        expandable: list[Component] = []
-        if expandable_components:
-            expandable = [c for c in analysis.components if c.name in expandable_components]
+        # Keep caller-provided root expandables, but also preserve deterministic planner eligibility.
+        root_expandable_ids = set(expandable_components or [])
```
brovatten: Root expandables. Is it clear to u guys what root means in this context?

diagram_analysis/analysis_json.py
brovatten: JSON should be capslock?

agents/agent_responses.py
```diff
@@ -177,6 +183,35 @@ def llm_str(self):
...
+
+    Handles same-named siblings by using a sibling index tiebreaker.
+    """
+    name_counts: dict[str, int] = {}
+    for component in analysis.components:
+        count = name_counts.get(component.name, 0)
+        component.component_id = compute_component_id(parent_id, component.name, count)
+        name_counts[component.name] = count + 1
+
+    # Assign relation IDs by looking up component names
```
brovatten: Can we log here that duplicate components are found? Can be good if we want to avoid this later on

agents/agent_responses.py
```diff
@@ -177,6 +183,35 @@ def llm_str(self):
...
+    name_to_id: dict[str, str] = {c.name: c.component_id for c in analysis.components}
```
brovatten: This silently drops entries if you have multiple components with the same name..

main.py
```diff
@@ -124,9 +124,13 @@ def generate_markdown_docs(
         demo=demo_mode,
     )
 
+    # Build id-to-name mapping for file naming
+    id_to_name = {c.component_id: c.name for c in root_analysis.components}
```
brovatten, line 128: Wait, so the subcomponents are not getting an ID assigned? if only root_anaylsis is used

output_generators/markdown.py
```diff
@@ -26,7 +26,7 @@ def generated_mermaid_str(
     # Linking to other files.
     for comp in analysis.components:
         node_id = sanitize(comp.name)
-        if comp.name in expanded_components:
+        if comp.component_id in expanded_components:
```
brovatten, line 29: On line 15, the node_id still uses comp.name, is this a problem? Also the same for the other output generators

diagram_analysis/diagram_generator.py
```diff
@@ -318,8 +316,8 @@ def generate_analysis(self):
                 save_analysis(
                     analysis=analysis,
                     output_dir=Path(self.output_dir),
-                    expandable_components=[c.name for c in expanded_components],
-                    sub_analyses={name: sub for name, (sub, _) in all_sub_analyses.items()},
+                    expandable_components=[c.component_id for c in root_expandable_components],
```
brovatten: Not sure I like the list comprehension in the arg parsing.

Also, shouldn't this on 319 and 340 use expanded_components and not root_expandable_components?

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: I think these got placed here before this class existed and I'll move them to the class.

diagram_analysis/incremental/io_utils.py
Svilen-Stefanov: I renamed it.

agents/agent_responses.py
Svilen-Stefanov: We need to rework this overall, the problem is that the LLMs currently only work with src_name and dst_name, which makes it impossible to actually understand which component this refers to. 
We definitely need to fix this but probably not here.

main.py
Svilen-Stefanov: They are, you can check the `analysis.json` in the PR and also the code right below this to see where the component ids for the subcomponents are used.

output_generators/markdown.py
Svilen-Stefanov: This is intentional. The node_id here (in Mermaid/HTML diagrams) is a display identifier for the diagram, not the internal component ID so that it still produces human-readable node labels in the Mermaid graph (e.g., AuthenticationModule instead of a3f2b1c9d4e5f678). 
I do agree this is super confusing though and renamed it - currently node_key, we can think of a better name later as well.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: I like the list comprehensions but I agree this is super confusing... 
The current method is generally confusing, I removed these as they only lead to confusion.

PR #199, opened 2026-02-15
Add first version of plugin architecture

Svilen-Stefanov:
Summary of the changes:
The changes introduce a plugin architecture that allows external packages (wrappers) to extend CodeBoarding's functionality without modifying its source code. Here's how it works:
1. Protocols (core/protocols.py) - function as contracts
These are type aliases that define what shape a plugin extension must have. Think of them as function descriptions, e.g:
     - HealthCheckFunc - "If you want to add a health check, your function must accept (StaticAnalysisResults, str, HealthCheckConfig) and return a list of check summaries"

2. Registry (core/registry.py) - list of predefined plugin extensions
A generic typed dictionary where plugins register their extensions by name. It prevents duplicates and provides lookup. I added two registries to show how this could work:
     - health_checks - for custom health check functions
     - tools - for custom agent tool factories

3. Plugin Loader (core/plugin_loader.py) - registers extensions
Uses Python's `entry_points` mechanism (codeboarding.plugins group) to auto-discover installed plugins at startup. Each plugin provides an `init(registries)` function that registers its extensions.

How wrappers use it? A wrapper package would:
- Define its extensions (e.g. a custom health check)
- Declare an entry point in its pyproject.toml under [project.entry-points."codeboarding.plugins"]
- Implement an init(registries) function that calls registries.health_checks.register("my_check", my_func)
- the core would then load these and execute them where they're defined

## Summary by CodeRabbit

* **New Features**
  * Plugin system: discover/loadable plugins that add tools and health checks at runtime.
  * Declarative tool registry for centralized tool discovery and installation.

* **Refactor**
  * Unified, registry-driven installation and configuration of external tools.
  * Startup now initializes plugins and integrates their health checks.

* **Enhancements**
  * Configuration override API for testing and runtime tweaks.

* **Tests**
  * New tests covering plugin loading and registry behaviors.

### DIFF
--- a/agents/tools/toolkit.py
+++ b/agents/tools/toolkit.py
@@ -11,7 +11,7 @@ from .get_method_invocations import MethodInvocationsTool
 from .read_file import ReadFileTool
 from .read_docs import ReadDocsTool
 from .external_deps import ExternalDepsTool
-from core import get_registries
+from core import load_plugin_tools
 from .read_git_diff import ReadDiffTool, FileChange
 
 logger = logging.getLogger(__name__)
@@ -115,10 +115,5 @@ class CodeBoardingToolkit:
             self.read_docs,
             self.external_deps,
         ]
-        for name, factory in get_registries().tools.all().items():
-            try:
-                plugin_tools = factory(self.context)
-                tools.extend(plugin_tools)
-            except Exception:
-                logger.warning(f"Plugin tool factory '{name}' failed", exc_info=True)
+        tools.extend(load_plugin_tools(self.context))
         return tools
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -11,9 +11,19 @@ Type aliases for plugin contracts are available in core.protocols:
     from core.protocols import HealthCheckFunc, ToolFactory
 """
 
+import logging
+from typing import TYPE_CHECKING
+
 from core.plugin_loader import load_plugins
 from core.protocols import HealthCheckFunc, ToolFactory
 from core.registry import Registry
+
+if TYPE_CHECKING:
+    from agents.tools.base import BaseRepoTool, RepoContext
+    from health.models import HealthCheckConfig, StandardCheckSummary, CircularDependencyCheck
+    from static_analyzer.analysis_result import StaticAnalysisResults
+
+logger = logging.getLogger(__name__)
 
 
 class Registries:
@@ -44,10 +54,48 @@ def reset_registries() -> None:
     _registries = None
 
 
+def run_plugin_health_checks(
+    static_analysis: "StaticAnalysisResults",
+    language: str,
+    config: "HealthCheckConfig",
+) -> "list[StandardCheckSummary | CircularDependencyCheck]":
+    """Run all registered plugin health checks for a language and return their summaries.
+
+    Failures in individual plugins are logged and skipped so a broken plugin
+    cannot take down the entire health check pipeline.
+    """
+    summaries: list = []
+    for name, check_func in get_registries().health_checks.all().items():
+        try:
+            summaries.extend(check_func(static_analysis, language, config))
+        except Exception:
+            logger.exception(f"Plugin health check '{name}' failed for language '{language}'")
+    return summaries
+
+
+def load_plugin_tools(context: "RepoContext") -> "list[BaseRepoTool]":
+    """Instantiate and return all tools registered by plugins for the given repo context.
+
+    Failures in individual plugin factories are logged and skipped so a broken
+    plugin cannot prevent other tools from loading.
+    """
+    tools: list = []
+    for name, factory in get_registries().tools.all().items():
+        try:
+            tools.extend(factory(context))
+        except Exception:
+            logger.warning(f"Plugin tool factory '{name}' failed", exc_info=True)
+    return tools
+
+
 __all__ = [
-    "Registry",
+    "HealthCheckFunc",
     "Registries",
+    "Registry",
+    "ToolFactory",
     "get_registries",
-    "reset_registries",
+    "load_plugin_tools",
     "load_plugins",
+    "reset_registries",
+    "run_plugin_health_checks",
 ]
--- a/core/protocols.py
+++ b/core/protocols.py
@@ -3,11 +3,12 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from static_analyzer.analysis_result import StaticAnalysisResults
 
 if TYPE_CHECKING:
+    from agents.tools.base import BaseRepoTool, RepoContext
     from health.models import (
         CircularDependencyCheck,
         HealthCheckConfig,
@@ -22,6 +23,5 @@ HealthCheckFunc = Callable[
 ]
 
 # A tool factory: given a RepoContext, returns a list of BaseRepoTool instances.
-# Uses Any to avoid circular dependency with agents/ package.
 # Plugin authors import RepoContext and BaseRepoTool from agents.tools.base directly.
-ToolFactory = Callable[[Any], list[Any]]
+ToolFactory = Callable[["RepoContext"], "list[BaseRepoTool]"]
--- a/health/runner.py
+++ b/health/runner.py
@@ -4,7 +4,7 @@ import os
 from datetime import datetime, timezone
 from pathlib import Path
 
-from core import get_registries
+from core import run_plugin_health_checks
 from health.checks.circular_deps import check_circular_dependencies
 from health.checks.cohesion import check_component_cohesion
 from health.checks.coupling import check_fan_in, check_fan_out
@@ -117,12 +117,7 @@ def _collect_checks_for_language(
     summaries.append(check_unused_code_diagnostics(collector, config))
 
     # Run plugin-provided health checks
-    for name, check_func in get_registries().health_checks.all().items():
-        try:
-            plugin_summaries = check_func(static_analysis, language, config)
-            summaries.extend(plugin_summaries)
-        except Exception:
-            logger.exception(f"Plugin health check '{name}' failed for language '{language}'")
+    summaries.extend(run_plugin_health_checks(static_analysis, language, config))
 
     # Apply .healthignore exclusion patterns across all check findings
     _apply_exclude_patterns(summaries, exclude_patterns)
--- a/tool_registry.py
+++ b/tool_registry.py
@@ -25,6 +25,8 @@ from typing import Any, cast
 
 import requests
 
+from vscode_constants import VSCODE_CONFIG
+
 logger = logging.getLogger(__name__)
 
 GITHUB_REPO = "CodeBoarding/CodeBoarding"
@@ -108,7 +110,7 @@ TOOL_REGISTRY: list[ToolDependency] = [
         npm_packages=["pyright"],
     ),
     ToolDependency(
-        key="typescript",
+        key="typescript",  # javascript uses the same LSP as typescript
         binary_name="typescript-language-server",
         kind=ToolKind.NODE,
         config_section=ConfigSection.LSP_SERVERS,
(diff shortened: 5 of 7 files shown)


tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+                cmd = cast(list[str], config[dep.config_section][dep.key]["command"])
+                cmd[0] = str(binary_path)
+
+        elif dep.kind is ToolKind.NODE:
+            binary_path = base_dir / "node_modules" / ".bin" / f"{dep.binary_name}{node_ext}"
+            if binary_path.exists():
+                cmd = cast(list[str], config[dep.config_section][dep.key]["command"])
+                cmd[0] = str(binary_path)
+
+        elif dep.kind is ToolKind.ARCHIVE and dep.archive_subdir:
```
brovatten, line 186: Is there any smart way to assure the Tools of kind ARCHIVE has a archive_subdir? I guess this is an assumption atm

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+            archive_dir = base_dir / "bin" / dep.archive_subdir
+            if archive_dir.is_dir() and (archive_dir / "plugins").is_dir():
+                config[dep.config_section][dep.key]["jdtls_root"] = str(archive_dir)
+
+    return config
+
+
+def resolve_config_from_path() -> dict[str, Any]:
+    """Discover tools on the system PATH and return a config dict."""
+    from vscode_constants import VSCODE_CONFIG
```
brovatten: How come this import is local?

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+
+    config = deepcopy(VSCODE_CONFIG)
+
+    for dep in TOOL_REGISTRY:
+        if dep.kind in (ToolKind.NATIVE, ToolKind.NODE):
+            path = shutil.which(dep.binary_name)
+        if path:
+            cmd = cast(list[str], config[dep.config_section][dep.key]["command"])
+            cmd[0] = path
```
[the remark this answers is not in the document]
brovatten: I agree with this. Maybe define path=None at the top, or just boolean AND statement in first if

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+    return config
+
+
+def has_required_tools(base_dir: Path) -> bool:
+    """Check if the minimum required tools (tokei) are installed."""
+    if not base_dir.exists():
+        return False
+    bin_dir = platform_bin_dir(base_dir)
+    is_win = platform.system() == "Windows"
+    tokei = bin_dir / ("tokei.exe" if is_win else "tokei")
```
brovatten: I see this now in three places. Maybe add a helper function that adds either ".exe" or "" to the str and returns it (works well since platform is not a local variable either and doesn't need to be passed)

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+    if subdir is None:
+        raise RuntimeError(f"Unsupported platform: {system}")
+    return base / "bin" / subdir
+
+
+def get_latest_release_tag() -> str:
+    """Fetch the latest release tag from the GitHub repository."""
+    response = requests.get(f"https://api.github.com/repos/{GITHUB_REPO}/releases/latest", timeout=30)
+    response.raise_for_status()
+    return response.json()["tag_name"]
```
brovatten, line 239: maybe use .get("tag_name", None)

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+
+    logger.info("Installing Node.js packages: %s", all_packages)
+    original_cwd = os.getcwd()
+    try:
+        os.chdir(target_dir)
+        if not Path("package.json").exists():
+            subprocess.run([npm_path, "init", "-y"], check=True, capture_output=True, text=True)
+        subprocess.run(
+            [npm_path, "install", *all_packages],
+            check=True,
```
brovatten, line 315: switching working dirs seems like it could add risks. rather add cwd here instead:
```
subprocess.run(
                [npm_path, "init", "-y"], 
                cwd=target_dir,
                check=True, 
                capture_output=True, 
                text=True
            )
```

utils.py
```diff
@@ -48,7 +48,31 @@ def monitoring_enabled():
...
+
+
+def set_config(config: dict) -> None:
+    """Programmatically set the tool/LSP configuration.
+
+    When set, get_config() uses this dict instead of reading
+    STATIC_ANALYSIS_CONFIG or falling back to VSCODE_CONFIG.
+    """
+    global _config_override
+    _config_override = config
```
brovatten, line 61: Naive question mby, but can we check the format of this config somehow?

agents/tools/toolkit.py
```diff
@@ -114,3 +115,10 @@ def get_all_tools(self) -> list[BaseRepoTool]:
             self.read_docs,
             self.external_deps,
         ]
+        for name, factory in get_registries().tools.all().items():
+            try:
```
brovatten: These are the tools used by the agents, quiet sure the LLM should not have this?

brovatten: or is it future proofing perhaps? can u can give an example of a current or upcoming tool. for example, health checks i dont think should be here

core/__init__.py
```diff
@@ -0,0 +1,53 @@
+"""
+CodeBoarding plugin infrastructure.
+
```
brovatten, line 3: I like this core module, nice!

pyproject.toml
```diff
@@ -76,7 +76,8 @@ dependencies = [
 ]
 
 [tool.setuptools]
-packages = ["agents", "diagram_analysis", "health", "health.checks", "monitoring", "output_generators", "repo_utils", "static_analyzer"]
+packages = ["agents", "agents.prompts", "agents.tools", "core", "diagram_analysis", "diagram_analysis.incremental", "health", "health.checks", "monitoring", "output_generators", "repo_utils", "static_analyzer", "static_analyzer.lsp_client"]
```
brovatten, line 79: Since "agents" is already added, do we need to add "agents.prompts" too?

tool_registry.py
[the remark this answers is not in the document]
brovatten: java is an example of an archive tool where this may happen for

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+def download_asset(tag: str, asset_name: str, destination: Path) -> bool:
+    """Download a GitHub release asset to destination. Returns True on success."""
+    destination.parent.mkdir(parents=True, exist_ok=True)
+    url = f"https://github.com/{GITHUB_REPO}/releases/download/{tag}/{asset_name}"
+    response = requests.get(url, stream=True, timeout=300, allow_redirects=True)
+    response.raise_for_status()
+    with open(destination, "wb") as f:
+        for chunk in response.iter_content(chunk_size=32768):
+            if chunk:
+                f.write(chunk)
```
brovatten, line 251: Does this need a cleanup on case of failure?

core/protocols.py
```diff
@@ -0,0 +1,27 @@
...
+# Receives (static_analysis, language, config) and returns a list of check summaries.
+HealthCheckFunc = Callable[
+    [StaticAnalysisResults, str, "HealthCheckConfig"],
+    "list[StandardCheckSummary | CircularDependencyCheck]",
+]
+
+# A tool factory: given a RepoContext, returns a list of BaseRepoTool instances.
+# Uses Any to avoid circular dependency with agents/ package.
+# Plugin authors import RepoContext and BaseRepoTool from agents.tools.base directly.
+ToolFactory = Callable[[Any], list[Any]]
```
brovatten: Should we make the Any more concrete here?

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+    return destination.exists() and destination.stat().st_size > 0
+
+
+def install_native_tools(target_dir: Path, deps: list[ToolDependency]) -> None:
```
brovatten, line 255: Is there a reason this always redownloads binaries, when install_archive_tool checks for existence?

tool_registry.py
```diff
@@ -0,0 +1,352 @@
...
+        config_section=ConfigSection.LSP_SERVERS,
+        github_asset_template="gopls-{platform_suffix}",
+    ),
+    ToolDependency(
+        key="python",
+        binary_name="pyright-langserver",
+        kind=ToolKind.NODE,
+        config_section=ConfigSection.LSP_SERVERS,
+        npm_packages=["pyright"],
+    ),
```
brovatten, line 111: where is javascript?

agents/tools/toolkit.py
```diff
@@ -11,6 +11,7 @@
 from .read_file import ReadFileTool
 from .read_docs import ReadDocsTool
 from .external_deps import ExternalDepsTool
+from core import get_registries
```
ivanmilevtues: unsure if this package has to be called core.

Core in the core 😅 , maybe integration idk

agents/tools/toolkit.py
```diff
@@ -114,3 +115,10 @@ def get_all_tools(self) -> list[BaseRepoTool]:
             self.read_docs,
             self.external_deps,
         ]
+        for name, factory in get_registries().tools.all().items():
```
ivanmilevtues: unsure if we need this but it can be:
core_tools = [......] 
loaded_tools = load_registry_tools()
reutrn core_tools + loaded_tools

health/runner.py
```diff
@@ -115,6 +116,14 @@ def _collect_checks_for_language(
     # Always add the check summary, even if empty
     summaries.append(check_unused_code_diagnostics(collector, config))
 
+    # Run plugin-provided health checks
+    for name, check_func in get_registries().health_checks.all().items():
```
ivanmilevtues: load_external_plugins or something like that. Just to be clear that this is a new thing :)

ivanmilevtues (approved):
Very minor stylistic comments.

pyproject.toml
Svilen-Stefanov: We could but we might want to keep this in the core only. I think we're almost exposing too much of the library atm, I think we should be exposing it gradually as we know exactly what we want to make extensible and how.

tool_registry.py
Svilen-Stefanov: We already enforce this via `assert dep.archive_subdir` inside `install_archive_tool()`

tool_registry.py
Svilen-Stefanov: This is intentional. A missing key here is a real API contract violation that should raise an error instead of silently return `None`.

tool_registry.py
Svilen-Stefanov: We clean it up in the caller via `archive_path.unlink(missing_ok=True)`.

utils.py
Svilen-Stefanov: We could define a basemodel for it I assume but let's not do this in this PR.

agents/tools/toolkit.py
Svilen-Stefanov: This is just an example that could be extended. We can use or remove it. Such plugins will become clearer down the line, for now it's just a demo.

tool_registry.py
Svilen-Stefanov: Javascript uses the same LSP as typescript, which is why we don't need it as a separate registry. Added a comment to make this more explicit.

tool_registry.py
Svilen-Stefanov: Fixed

agents/tools/toolkit.py
Svilen-Stefanov: @brovatten I saw you had a comment here too, I just wanted to gave examples. Now that you've seen it, I can remove this or even both of them but I thought it might be helpful to have at least one to refer to later on. 
Let me know what you think and I can implement it accordingly.

agents/tools/toolkit.py
Svilen-Stefanov: Haha, I think we might need a larger refactoring of the code overall to make it nice in the future. I can rename it but I think we should restructure thing overall and I tend to prefer to do this all together later.

brovatten (approved):

PR #203, opened 2026-02-17
Change default to gemini-3-flash

Svilen-Stefanov:
I ran a full analysis and it seems like this is working now.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -1,19 +1,18 @@
-import os
 import logging
+import os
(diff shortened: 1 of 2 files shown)


ivanmilevtues (approved):
nice, this should be mergable.

Interesting to see if someone can run some sorf of eval to see how it compare to 2.5-flash I am sure it is better but still.

@Svilen-Stefanov @brovatten

brovatten:
I agree, feel like a change in models deserves an eval run

PR #205, opened 2026-02-18
Enable meta agent read file

brovatten:
I realised while working on the `MetaAgent` we are not providing the `readFile` tool that it uses.

Also, claude and gpt prompts are prompting it to look for the `packageDependencies` tool which it shouldn't use - and now can't use since static analysis is done in parallell with the `MetaAgent`.

and I upgraded the deprecated langchain

## Summary by CodeRabbit

* **Improvements**
  * Agents now use documentation and external dependency analysis tools for more accurate project understanding.
  * Enhanced consistency of agent configuration and behavior across supported AI models.

* **Tests**
  * Added verification of agent tool consistency across AI model types.

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -11,6 +11,7 @@ from langchain_core.messages import SystemMessage, HumanMessage, AIMessage
 from langchain_core.output_parsers import PydanticOutputParser
 from langchain_core.prompts import PromptTemplate
 from langchain.agents import create_agent
+from langgraph.graph.state import CompiledStateGraph
 from pydantic import ValidationError
 from trustcall import create_extractor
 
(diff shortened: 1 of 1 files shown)


agents/agent.py
```diff
@@ -10,7 +10,7 @@
 from langchain_core.messages import SystemMessage, HumanMessage, AIMessage
 from langchain_core.output_parsers import PydanticOutputParser
 from langchain_core.prompts import PromptTemplate
-from langgraph.prebuilt import create_react_agent
+from langchain.agents import create_agent
```
ivanmilevtues, line 13: Double check if you have properly migrated as the create_agent is not exactly the same as create_react_agent:

https://docs.langchain.com/oss/python/migrate/langchain-v1#migrate-to-create_agent. SHouldn't be too big of a deal as we are using our own system prompts but double check.

Svilen-Stefanov (approved):
LGTM, haven't run it and because the tests are not running, I'm not sure if this needs a dependency upgrade but I think the tests should catch this, hence approving.

agents/agent.py
brovatten: We only ever feed it "model" or "tools" to handle automatically. This is the appropriate wrapper for our langchain version.
For the latter checkbox u made, I meant the wrapper and not the dependency.
I also ran it and everything worked! There was this thing:
image
that potentially could have broken backwards compat, but I tried with vercel (that uses openai langchain) and all good.

PR #207, opened 2026-02-18
Don't log API keys

brovatten:
It does not feel secure as a user to have my API keys exposed in the logs

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -33,6 +33,13 @@ class TestValidateEnvVars(unittest.TestCase):
             validate_env_vars()
         self.assertEqual(cm.exception.code, 1)
 
+    @patch.dict(os.environ, {"OPENAI_API_KEY": "   "}, clear=True)
+    def test_validate_env_vars_empty_key_value(self):
+        # Should treat empty/whitespace key values as unset
+        with self.assertRaises(SystemExit) as cm:
+            validate_env_vars()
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -47,7 +47,7 @@ def validate_env_vars():
         "CEREBRAS_API_KEY",
         "VERCEL_API_KEY",
     ]
-    api_env_keys = [(key, os.getenv(key)) for key in api_provider_keys if os.getenv(key) is not None]
+    api_env_keys = [key for key in api_provider_keys if os.getenv(key) is not None]
 
     if len(api_env_keys) == 0:
```
ivanmilevtues, line 52: This might get false positives for empty keys as ("" is not None) is True:
Also I like better the following syntax :P 
```python
if not any(os.getenv(key) for key in api_provider_keys):
```

Svilen-Stefanov: Yup, I second that.

Svilen-Stefanov (approved):
LGTM. An alternative to the current solution would be to just print the keys in the error while keeping both the key and its value in the list here. 

I can imagine that at some point we might want to fail early if the key is invalid (send a quick try first) for which keeping the the value somewhere might be useful.

main.py
brovatten: To make it consistent for the case of multiple api keys, I just changed the api_env_keys definition, and I also look for keys with empty spaces

PR #209, opened 2026-02-20
Parallellize the levels

brovatten:
This lets avoids having levels wait for eachother in the processing. So instead of a loop by level, we now have a `ThreadPoolExecutor` acting as a queue. 
I chose it over `asyncio` as its less code and the bottleneck is API bound (rather than CPU - in which case asyncio might be more interesting). Although you know this better @Svilen-Stefanov , so maybe you have advice on [thread/multiprocessing/concurrent.futures/asyncio](https://stackoverflow.com/questions/61351844/difference-between-multiprocessing-asyncio-threading-and-concurrency-futures-i). 

PS: I will run the benchmarks once we agreed on the approach

## Summary by CodeRabbit

* **New Features**
  * Frontier-based processing that starts child analyses as soon as parents complete, speeding overall analysis.
  * Live progress updates showing submissions, completions, in-flight tasks, saves, and errors.

* **Improvements**
  * Frequent intermediate saves after each component to improve crash resilience.
  * Enhanced per-component logging and metrics for clearer diagnostics.

* **Tests**
  * New test verifying child components can start before slower siblings finish.


diagram_analysis/diagram_generator.py
```diff
@@ -262,63 +263,83 @@ def generate_analysis(self):
...
-                    output_dir=Path(self.output_dir),
-                    sub_analyses={cid: sub for cid, (sub, _) in all_sub_analyses.items()},
-                    repo_name=self.repo_name,
+            expanded_components: list[Component] = []
+
+            # Track all discovered sub-analyses for checkpoints/final write.
+            sub_analyses: dict[str, AnalysisInsights] = {}
+
+            # Process components using a frontier queue: submit children as soon as parent finishes.
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
```
Svilen-Stefanov: I'm leaving my thoughts here:
1. The static analysis is ready at this point, so this won't have impact on the LSPs, right?
2. For the LLMs, this could potentially lead to throttling (no component analysis might finish if there's too much of it). So that depends a lot on the API key people use, right? 
3. Number of workers should probably be the min of max_workers and component count?

Based on 2., I wonder if we should have a config var that could set this (e.g by default it is enabled but you could disable it if it's causing you issues).

tests/diagram_analysis/test_diagram_analysis.py
```diff
@@ -422,106 +395,74 @@ def test_process_component_with_exception(self):
         self.assertIsNone(result_analysis)
         self.assertEqual(new_components, [])
 
-    def test_generate_analysis_uses_root_expandables_for_can_expand(self):
+    @patch("diagram_analysis.diagram_generator.save_analysis")
+    @patch("diagram_analysis.diagram_generator.plan_analysis")
+    def test_generate_analysis_frontier_submits_child_before_slow_sibling_finishes(
```
Svilen-Stefanov, line 428: Apart from the unit test, I suppose it's sufficient to run a full analysis with a higher depth to test the impact here, right? Is there anything else we need to do?

Svilen-Stefanov (commented):
Generally lgtm. Leaving a comment as I'd like to hear what you think about the potential LLM agent throttling and see if we need to fix it.

diagram_analysis/diagram_generator.py
brovatten: 1. That's correct.
2. It had the same max concurrency now as before. Now however, we will consistently have higher concurrency. On our side, given my vercel experiments I did, this wont impact us.
For users, I think the 5 attempts we do with exponential backoff will handle this just fine. If we need it fancier (backpressure, loadbalancing API keys, etc.) at some point we can use LiteLLM.
This is out of scope for repo, and is not something we need.
3. It is better to set it to the maximum the system can handle and let it manage itself. We also can't predict for deeper levels than 1 how many components will be processed concurrently.

tests/diagram_analysis/test_diagram_analysis.py
brovatten: This PR does not impacts the outcome of the agents. So I would just run it two times with and without the new change, and average the time and then compare. Setting the depth at 3 (since 1 and 2 is not impacted by this change, and 4 we basically never do).

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Yeah, I guess the only difference for 2. is that we might be using the max parallelism of 8 for a longer period of time (or could reach it more often) as mostly we have 4-6 components in a level and when they finish now, we will automatically continue with their subcomponents. If this is an issue, we will see it earlier than later so that should be fine. 

Ignore my comment on 3, I think the current approach should be fine.

tests/diagram_analysis/test_diagram_analysis.py
Svilen-Stefanov: Sounds great, thanks!

diagram_analysis/diagram_generator.py
```diff
@@ -262,63 +263,83 @@ def generate_analysis(self):
...
+                save_count = 0
+                error_count = 0
+
+                if self.depth_level > 1:
+                    for component in root_components:
+                        future = executor.submit(self.process_component, component)
+                        future_to_task[future] = (component, 1)
+                        submitted_components += 1
+
+                pbar = tqdm(
```
ivanmilevtues: dynamic progress bar how does this look, i feel like it is probably not super useful :D.

diagram_analysis/diagram_generator.py
```diff
@@ -262,63 +263,83 @@ def generate_analysis(self):
...
+                            pbar.set_postfix(
+                                in_flight=len(future_to_task),
+                                saves=save_count,
+                                errors=error_count,
+                                refresh=False,
+                            )
+                            pbar.refresh()
+                finally:
+                    pbar.close()
+
```
ivanmilevtues: put this into a new funciton/method (L278-L342)

diagram_analysis/diagram_generator.py
```diff
@@ -262,63 +263,83 @@ def generate_analysis(self):
...
+                                    )
+                                    save_count += 1
+                                if new_components and level + 1 < self.depth_level:
+                                    for child_component in new_components:
+                                        child_future = executor.submit(self.process_component, child_component)
+                                        future_to_task[child_future] = (child_component, level + 1)
+                                        submitted_components += 1
+                            except Exception as exc:
+                                error_count += 1
+                                logging.error(f"Component {component.name} generated an exception: {exc}")
```
ivanmilevtues: put this into a new method.

PR #214, opened 2026-02-22
Improve codeboarding entry points for codeboarding package

Svilen-Stefanov:
Current version of the documentation in Test PyPI: https://test.pypi.org/project/codeboarding/0.9.0.dev2/

## Summary by CodeRabbit

* **New Features**
  * New setup command for automated language-server installation and automatic npm install.
  * CLI: added model-selection and incremental/full run options; new binary-location, upload, and monitoring flags.
  * Per-repo outputs now use .codeboarding directories and a shared ~/.codeboarding/servers for tool binaries.
  * Added user config support with an on-disk template and env injection.

* **Documentation**
  * Revised quick-start, API examples, and CLI reference to reflect new workflows and defaults.

* **Chores**
  * Version bumped to 0.9.0.

### DIFF
--- a/install.py
+++ b/install.py
@@ -13,6 +13,7 @@ import requests
 from tool_registry import (
     TOOL_REGISTRY,
     ToolKind,
+    _write_manifest,
     get_servers_dir,
     install_archive_tool,
     install_native_tools,
@@ -505,14 +506,12 @@ def main() -> None:
 
     run_install(auto_install_npm=args.auto_install_npm, auto_install_vcpp=args.auto_install_vcpp)
 
-    from tool_registry import _write_manifest
-
     _write_manifest()
 
     print("\n" + "=" * 40)
     print("Setup complete!")
     print("Configure your LLM provider key in ~/.codeboarding/config.toml, then run:")
-    print("  codeboarding --local /path/to/repo --project-name MyProject")
+    print("  codeboarding --local /path/to/repo")
 
 
 if __name__ == "__main__":


diagram_analysis/diagram_generator.py
```diff
@@ -53,6 +53,8 @@ def __init__(
         project_name: str | None = None,
         run_id: str | None = None,
         monitoring_enabled: bool = False,
+        agent_model: str | None = None,
+        parsing_model: str | None = None,
```
ivanmilevtues: This is never used, within the main repo. and we just propagate it left and right.

I can see that the ffort here is to make the core less dependent on env vars, if you ask me it is better to have a config file like the static_analyzer_config, than having these kind of propagations everywhere, also refering to repo_location, temp_folder, repo_name, output_dir, depth_level etc.

main.py
```diff
@@ -487,44 +452,55 @@ def main():
     is_local = args.local is not None
     validate_arguments(args, parser, is_local)
 
-    # Setup logging first, before any operations that might log
-    log_dir: Path | None = args.output_dir if args.output_dir else None
-    setup_logging(log_dir=log_dir)
+    # Derive output directory from repo path
+    if is_local:
+        output_dir = args.local.resolve() / ".codeboarding"
```
ivanmilevtues, line 430: This requires a change in the vscode extension as well

main.py
```diff
@@ -450,6 +400,20 @@ def define_cli_arguments(parser: argparse.ArgumentParser):
...
+        type=str,
+        default=None,
+        help="Model name for the agent LLM (e.g. gpt-4o, claude-3-7-sonnet-20250219)",
+    )
+    parser.add_argument(
+        "--parsing-model",
+        type=str,
+        default=None,
+        help="Model name for the parsing LLM (e.g. gpt-4o-mini, claude-3-haiku-20240307)",
+    )
```
ivanmilevtues: I understand the change, but tbh, we should probably consider a config yml, for these instead of propagating them 6 layers deep. Something similar to the static_analysis_config.yml, if not extending that to a full config file.

main.py
```diff
@@ -292,6 +270,8 @@ def process_local_repository(
     monitoring_enabled: bool = False,
     incremental: bool = False,
     force_full: bool = False,
+    agent_model: str | None = None,
+    parsing_model: str | None = None,
```
ivanmilevtues: I would in fact make them non-nullable and just do the os.env retrieval at the top level (main.py)
Or as I said add that configuration to an yml file instead of propagating it.

output_generators/sphinx.py
```diff
@@ -142,6 +141,7 @@ def generate_rst_file(
     expanded_components: set[str],
     temp_dir: Path,
     demo: bool = False,
+    repo_path: Path | None = None,
```
ivanmilevtues: This shouldn't be nullable in my opinion.

output_generators/mdx.py
```diff
@@ -82,8 +82,7 @@ def generate_mdx(
 
     detail_lines = ["\n### Details\n", f"{insights.description}\n"]
 
-    repo_root = os.getenv("REPO_ROOT")
-    root_dir = os.path.join(repo_root, project) if repo_root else project
+    root_dir = str(repo_path / project) if repo_path else project
```
ivanmilevtues: instead of nullable just make this if whenever we call the generators.

agents/llm_config.py
```diff
@@ -256,7 +266,10 @@ def initialize_parsing_llm(model_override: str | None = None) -> BaseChatModel:
...
 
 
-def initialize_llms() -> tuple[BaseChatModel, BaseChatModel]:
-    agent_llm = initialize_agent_llm(os.getenv("AGENT_MODEL"))
-    parsing_llm = initialize_parsing_llm(os.getenv("PARSING_MODEL"))
+def initialize_llms(
+    agent_model: str | None = None,
+    parsing_model: str | None = None,
+) -> tuple[BaseChatModel, BaseChatModel]:
+    agent_llm = initialize_agent_llm(agent_model or os.getenv("AGENT_MODEL"))
```
ivanmilevtues: instead of doing this here, we should either to it at top level, or use the config that I mentioned directly here.

main.py
Svilen-Stefanov: Created a `config.toml`.

PR #219, opened 2026-02-27
Add basic separation of concerns for the static analysis

Svilen-Stefanov:
Minor change to make the static analysis of the core more of a library than a single orchestrated engine. The functionality hasn't changed.

## Summary by CodeRabbit

* **New Features**
  * Optional injectable long-lived static analyzer for faster, consistent analysis and improved incremental updates.
  * Start/stop lifecycle and file-change notifications for language analysis tools to keep diagnostics current.

* **Refactor**
  * Analysis, diagnostics and caching flows unified to surface analyzer diagnostics through health and coverage reporting.

* **Performance**
  * Improved monorepo multi-project analysis and parallel static/metadata analysis for quicker diagram generation.

### DIFF
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -68,7 +68,7 @@ class DiagramGenerator:
         # Optional pre-started StaticAnalyzer injected by long-lived callers (e.g. the
         # wrapper). When set, pre_analysis() uses it directly instead of creating a new
         # one-shot analyzer via get_static_analysis().
-        self._static_analyzer: StaticAnalyzer | None = static_analyzer
+        self._static_analyzer = static_analyzer
 
         self.details_agent: DetailsAgent | None = None
         self.static_analysis: StaticAnalysisResults | None = None  # Cache static analysis for reuse
@@ -170,23 +170,23 @@ class DiagramGenerator:
         )
         self._monitoring_agents["MetaAgent"] = self.meta_agent
 
-        # Decide how to obtain static analysis results, then run it in parallel
-        # with the meta-context computation so neither blocks the other.
-        if self._static_analyzer is not None:
-            # Long-lived caller (e.g. wrapper): clients already running, skip lifecycle.
-            logger.info("Using injected StaticAnalyzer (clients already running)")
+        def get_static_with_injected_analyzer() -> StaticAnalysisResults:
+            cache_dir = None if self.force_full_analysis else get_cache_dir(self.repo_location)
+            return self._get_static_from_injected_analyzer(cache_dir)
 
-            cache_dir = get_cache_dir(self.repo_location)
-            static_callable = lambda: self._get_static_from_injected_analyzer(
-                None if self.force_full_analysis else cache_dir
-            )
-
-        else:
+        def get_static_with_new_analyzer() -> StaticAnalysisResults:
             skip_cache = self.force_full_analysis
             if skip_cache:
                 logger.info("Force full analysis: skipping static analysis cache")
+            return get_static_analysis(self.repo_location, skip_cache=skip_cache)
 
-            static_callable = lambda: get_static_analysis(self.repo_location, skip_cache=skip_cache)
+        # Decide how to obtain static analysis results, then run it in parallel
+        # with the meta-context computation so neither blocks the other.
+        if self._static_analyzer is not None:
+            logger.info("Using injected StaticAnalyzer (clients already running)")
+            static_callable = get_static_with_injected_analyzer
+        else:
+            static_callable = get_static_with_new_analyzer
 
         with ThreadPoolExecutor(max_workers=2) as executor:
             static_future = executor.submit(static_callable)
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -132,15 +132,28 @@ class StaticAnalyzer:
         Idempotent — safe to call even if clients are already running.
         """
         if self._clients_started:
+            logger.info(f"Clients already started for {self.repository_path}, skipping start.")
             return
+
+        started_clients: list[LSPClient] = []
         for client in self.clients:
             try:
                 logger.info(f"Starting LSP client for {client.language.language}")
                 client.start()
+                started_clients.append(client)
                 if isinstance(client, JavaClient):
                     client.wait_for_import()  # timeout auto-computed based on project size
             except Exception as e:
-                logger.error(f"Failed to start LSP client for {client.language.language}: {e}")
+                logger.exception(f"Failed to start LSP client for {client.language.language}")
+                # Clean up already-started clients
+                for started in reversed(started_clients):
+                    try:
+                        started.close()
+                    except Exception:
+                        logger.exception(f"Error closing client for {started.language.language}")
+                self._clients_started = False
+                raise RuntimeError(f"Failed to start LSP client for {client.language.language}") from e
+
         self._clients_started = True
 
     def stop_clients(self) -> None:
@@ -174,10 +187,12 @@ class StaticAnalyzer:
             if file_path.suffix.lstrip(".") not in handled_suffixes:
                 continue
             file_uri = file_path.as_uri()
+            next_version = client._document_versions.get(file_uri, 1) + 1
+            client._document_versions[file_uri] = next_version
             client._send_notification(
                 "textDocument/didChange",
                 {
-                    "textDocument": {"uri": file_uri, "version": 2},
+                    "textDocument": {"uri": file_uri, "version": next_version},
                     "contentChanges": [{"text": content}],
                 },
             )
@@ -187,7 +202,8 @@ class StaticAnalyzer:
         """
         Analyze the repository using LSP clients.
 
-        Assumes clients are already running (call start_clients() first).
+        Clients must be running before calling this method. Use start_clients() or
+        the context manager (``with StaticAnalyzer(...) as sa:``) to start them.
         get_static_analysis() does this automatically for CLI callers.
 
         Args:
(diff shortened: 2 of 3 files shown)


diagram_analysis/diagram_generator.py
```diff
@@ -157,14 +163,32 @@ def pre_analysis(self):
...
-        with ThreadPoolExecutor(max_workers=2) as executor:
+        # Decide how to obtain static analysis results, then run it in parallel
+        # with the meta-context computation so neither blocks the other.
+        if self._static_analyzer is not None:
+            # Long-lived caller (e.g. wrapper): clients already running, skip lifecycle.
+            logger.info("Using injected StaticAnalyzer (clients already running)")
+
+            cache_dir = get_cache_dir(self.repo_location)
+
+            def _get_static() -> StaticAnalysisResults:
```
ivanmilevtues: I am not a big fan of annonymous function overrides. Can't we make it into two funcs.
And then in the threadpool have the if _self._static_analyzer is not None, submit the one else the other
now we have get_static_analysis makes sense.
_get_static(), which is in fact two methods.

static_analyzer/__init__.py
```diff
@@ -312,24 +367,16 @@ def get_static_analysis(
...
-        # Default behavior: use standard cache location
-        actual_cache_dir = get_cache_dir(repo_path)
-    else:
-        actual_cache_dir = cache_dir
-
-    # Use incremental analysis - it handles cache internally via IncrementalAnalysisOrchestrator
     analyzer = StaticAnalyzer(repo_path)
-    results = analyzer.analyze(cache_dir=actual_cache_dir)
-    # Attach diagnostics to results for convenient access
+    analyzer.start_clients()
```
ivanmilevtues: A random suggestion, just wondering what you think of the following, it comes from the fact that it really gives me the ick when we have try/except/finaly everywhere. Pure suggestion!
```python
class Analyzer:
    def __enter__(self):
        self.start_clients()
        return self

    def __exit__(self, exc_type, exc, tb):
        self.stop_clients()
```

```python
actual_cache_dir = None if skip_cache else (
    cache_dir if cache_dir is not None else get_cache_dir(repo_path)
)

with analyzer:
    results = analyzer.analyze(cache_dir=actual_cache_dir)
```

Svilen-Stefanov: Great suggestion, love it.

diagram_analysis/diagram_generator.py
```diff
@@ -63,6 +65,10 @@ def __init__(
         self.run_id = run_id
         self.monitoring_enabled = monitoring_enabled
         self.force_full_analysis = False  # Set to True to skip incremental updates
+        # Optional pre-started StaticAnalyzer injected by long-lived callers (e.g. the
+        # wrapper). When set, pre_analysis() uses it directly instead of creating a new
+        # one-shot analyzer via get_static_analysis().
+        self._static_analyzer: StaticAnalyzer | None = static_analyzer
```
brovatten: Do we need this typing?

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
         programming_langs = ProjectScanner(self.repository_path).scan()
         self.clients = create_clients(programming_langs, self.repository_path, self.ignore_manager)
         self.collected_diagnostics: dict[str, FileDiagnosticsMap] = {}
+        self._clients_started: bool = False
+
+    def __enter__(self) -> "StaticAnalyzer":
+        self.start_clients()
+        return self
+
+    def __exit__(self, _exc_type: type | None, _exc_val: Exception | None, _exc_tb: object | None) -> None:
```
brovatten, line 125: What are all of these input params for?

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
...
+        self.stop_clients()
+
+    def start_clients(self) -> None:
+        """Start all LSP server processes.
+
+        Call once before invoking analyze() or analyze_with_cluster_changes().
+        Idempotent — safe to call even if clients are already running.
+        """
+        if self._clients_started:
```
brovatten, line 134: Maybe add a warning log, i'm assuming this is not expected to be called while occupied

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
...
+            return
+        for client in self.clients:
+            try:
+                logger.info(f"Starting LSP client for {client.language.language}")
```
brovatten, line 141: client.language.language 😅  this we should rename

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
         programming_langs = ProjectScanner(self.repository_path).scan()
         self.clients = create_clients(programming_langs, self.repository_path, self.ignore_manager)
         self.collected_diagnostics: dict[str, FileDiagnosticsMap] = {}
+        self._clients_started: bool = False
```
brovatten, line 119: I guess it's not a problem for now, but is this way of doing it thread safe?

static_analyzer/__init__.py
```diff
@@ -312,24 +374,9 @@ def get_static_analysis(
...
-
-    # Determine actual cache directory to use
-    if cache_dir is None:
-        # Default behavior: use standard cache location
-        actual_cache_dir = get_cache_dir(repo_path)
-    else:
-        actual_cache_dir = cache_dir
-
-    # Use incremental analysis - it handles cache internally via IncrementalAnalysisOrchestrator
+    actual_cache_dir = None if skip_cache else (cache_dir if cache_dir is not None else get_cache_dir(repo_path))
```
brovatten, line 398: Separate this, too much for one line

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
...
+            return
+        for client in self.clients:
+            try:
+                logger.info(f"Starting LSP client for {client.language.language}")
+                client.start()
+                if isinstance(client, JavaClient):
+                    client.wait_for_import()  # timeout auto-computed based on project size
+            except Exception as e:
+                logger.error(f"Failed to start LSP client for {client.language.language}: {e}")
+        self._clients_started = True
```
brovatten: This is set to true no if started or not

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
...
+        """
+        for client in self.clients:
+            handled_suffixes = {s.lstrip("*").lstrip(".") for s in client.language_suffix_pattern}
+            if file_path.suffix.lstrip(".") not in handled_suffixes:
+                continue
+            file_uri = file_path.as_uri()
+            client._send_notification(
+                "textDocument/didChange",
+                {
+                    "textDocument": {"uri": file_uri, "version": 2},
```
brovatten: I don't really get this one, but perhaps you do:

```
 notify_file_changed() LSP protocol risk in static_analyzer/init.py:161
     Suggestion: manage per-document open/version state correctly.
     Concrete fix:

  - Maintain self._doc_versions: dict[str, int] and self._open_docs: set[str].
  - First change for a URI: send didOpen with version 1.
  - Subsequent changes: increment version and send didChange.
  - Optionally send didClose when session ends.
    Add tests for repeated changes (v1, v2, v3 monotonicity).
```

static_analyzer/__init__.py
```diff
@@ -116,11 +116,80 @@ def __init__(self, repository_path: Path):
...
     def analyze(self, cache_dir: Path | None = None) -> StaticAnalysisResults:
         """
         Analyze the repository using LSP clients.
 
+        Assumes clients are already running (call start_clients() first).
+        get_static_analysis() does this automatically for CLI callers.
+
         Args:
             cache_dir: Optional cache directory for incremental analysis.
                       If provided, uses git-based incremental analysis per client.
```
brovatten, line 211: Isn't it a bit strange for this to "assume" lsp clients are started? 
Maybe call start_clients inside here and tank the warning, or raise an error here if it's not already started. Or check if it's started, and if not start it. Or send them as a parameter

static_analyzer/__init__.py
Svilen-Stefanov: Since we are overwriting the __exit__ method, these are needed to properly define this as a context manager.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: I don't think so, I'll remove it

static_analyzer/__init__.py
Svilen-Stefanov: haha, yes, won't change this here though.

static_analyzer/__init__.py
Svilen-Stefanov: It isn't but that's not too important for now as we define the callers of this function. Generally, I think if something is not explicitly marked as thread safe, it isn't :D

static_analyzer/__init__.py
Svilen-Stefanov: If I try to, black fixes it like this :X At least for now, that's what our formatter likes.

static_analyzer/__init__.py
Svilen-Stefanov: Yeah, this is actually a minor bug, which might be a larger bug for some LSPs. I'll introduce a straightforward fix.

static_analyzer/__init__.py
```diff
@@ -116,11 +116,88 @@ def __init__(self, repository_path: Path):
...
+            next_version = client._document_versions.get(file_uri, 1) + 1
+            client._document_versions[file_uri] = next_version
+            client._send_notification(
+                "textDocument/didChange",
+                {
+                    "textDocument": {"uri": file_uri, "version": next_version},
+                    "contentChanges": [{"text": content}],
+                },
+            )
+            logger.debug(f"Sent didChange for {file_path} to {client.language.language} LSP")
```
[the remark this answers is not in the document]
Svilen-Stefanov: This is an actual concern fixed in a follow up here: https://github.com/CodeBoarding/CodeBoarding/pull/220/changes#diff-2d8405a64d2092f08b6fc3aa619c3194c28cacf1fc9cf0fada9000be6c2fc881

PR #220, opened 2026-02-28
Remove previously reported warnings if no longer relevant and remove double file closure

Svilen-Stefanov:
Previously reported warnings might not be removed from the health report if we don't rerun a full analysis again. 
This PR fixes this and makes the warnings somewhat more consistent. 

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable diagnostics by overlaying live client results on cached data and evicting stale entries to reduce incorrect warnings.
  * Analyzer now waits for diagnostics before finishing, preventing incomplete or missing reports.

* **New Features**
  * File-change notifications use an explicit open-then-change sequence so analysis sees current file content and versions.
  * Diagnostics are collected and processed in batches for coherent, up-to-date reporting.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -174,13 +174,13 @@ class StaticAnalyzer:
     def notify_file_changed(self, file_path: Path, content: str) -> None:
         """Notify the LSP server that the editor has saved new content for a file.
 
-        Sends textDocument/didOpen (so pyright registers the file if it is not
-        already open) followed by textDocument/didChange (so pyright re-type-checks
-        with the new content and emits a fresh publishDiagnostics notification).
+        If the file is not already open in the server, sends textDocument/didOpen
+        followed by textDocument/didChange.  If it is already open, sends only
+        textDocument/didChange with a monotonically incremented version number.
 
         The LSP spec requires a file to be open before didChange is sent.  After
         the analysis loop closes all files, a bare didChange would be silently
-        ignored by pyright, which is why we always re-open first.
+        ignored by pyright, which is why we re-open when needed.
 
         The resulting publishDiagnostics notification is captured by the reader
         thread into ``client.diagnostics`` and will be picked up by the next
@@ -195,31 +195,48 @@ class StaticAnalyzer:
             if file_path.suffix.lstrip(".") not in handled_suffixes:
                 continue
             file_uri = file_path.as_uri()
-            # Re-open the file so pyright tracks it.  Strictly, LSP 3.17 says a
-            # second didOpen for the same URI is an error, but pyright tolerates
-            # it in practice (it simply resets its internal state for the file).
-            # A proper fix would track open-file state and skip the didOpen when
-            # the file is already open, but for now this works with pyright.
-            client._send_notification(
-                "textDocument/didOpen",
-                {
-                    "textDocument": {
-                        "uri": file_uri,
-                        "languageId": client.language_id,
-                        "version": 1,
-                        "text": content,
-                    }
-                },
-            )
-            # Send the change so pyright re-type-checks with the new content.
+
+            with client._lock:
+                current_version = client._open_documents.get(file_uri)
+
+            if current_version is None:
+                # File not open yet — send didOpen so the server tracks it.
+                client._send_notification(
+                    "textDocument/didOpen",
+                    {
+                        "textDocument": {
+                            "uri": file_uri,
+                            "languageId": client.language_id,
+                            "version": 1,
+                            "text": content,
+                        }
+                    },
+                )
+                new_version = 2
+            else:
+                new_version = current_version + 1
+
+            # Send the change so the server re-type-checks with the new content.
             client._send_notification(
                 "textDocument/didChange",
                 {
-                    "textDocument": {"uri": file_uri, "version": 2},
+                    "textDocument": {"uri": file_uri, "version": new_version},
                     "contentChanges": [{"text": content}],
                 },
             )
-            logger.debug(f"Sent didOpen+didChange for {file_path} to {client.language.language} LSP")
+            with client._lock:
+                client._open_documents[file_uri] = new_version
+                # Invalidate the diagnostics snapshot so get_collected_diagnostics()
+                # returns the live diagnostics dict (which will be updated by the
+                # publishDiagnostics notification pyright sends in response).
+                # Without this, the stale snapshot from build_static_analysis()
+                # would mask any diagnostics received after notify_file_changed.
+                client._diagnostics_snapshot = None
+
+            logger.debug(
+                f"Sent {'didOpen+' if current_version is None else ''}didChange "
+                f"(v{new_version}) for {file_path} to {client.language.language} LSP"
+            )
 
     def analyze(self, cache_dir: Path | None = None) -> StaticAnalysisResults:
         """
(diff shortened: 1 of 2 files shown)


static_analyzer/lsp_client/client.py
```diff
@@ -645,6 +636,56 @@ def _create_qualified_name(self, file_path: Path, symbol_name: str) -> str:
...
+            timeout: Maximum seconds to wait (default 10 s).
+        """
+        if not expected_files:
+            return
+
+        expected_paths = {str(f) for f in expected_files}
+        deadline = time.monotonic() + timeout
+        poll_interval = 0.2
+
+        while True:
```
ivanmilevtues: That seems scary :D.

Can't we have await with timout and then an async function?

The answer can be no, we cannot. :D

static_analyzer/lsp_client/client.py
```diff
@@ -720,12 +774,15 @@ def build_static_analysis(self, source_files_override: list[Path] | None = None)
...
 
-        # Allow time for final diagnostics to arrive from the LSP server
-        # before closing files. Files were kept open during analysis to
-        # give the server time to produce diagnostics.
-        time.sleep(2)
+        # Wait for pyright to finish type-checking and emit publishDiagnostics
+        # for every open file.  Files with no issues never emit a notification,
+        # so _wait_for_diagnostics falls back to the timeout for those.
+        # 10 s is generous — pyright typically responds within 2-5 s.
+        self._wait_for_diagnostics(src_files, timeout=10.0)
```
brovatten, line 793: wait_for_diagnostics always hits the 10s timeout for any project with clean files. Line 271 pops files from self.diagnostics when the server reports an empty diagnostics list. So self.diagnostics.keys() only ever contains files with errors. The check on L681 (covered >= expected_paths) can never pass if any expected file is clean → guaranteed 10-second wall-clock stall on every run. Fix: track a separate self._files_with_diagnostics_received: set that is updated in _handle_diagnostics_notification regardless of whether the diagnostics list is empty or not, and check against that set.

static_analyzer/__init__.py
```diff
@@ -187,16 +195,31 @@ def notify_file_changed(self, file_path: Path, content: str) -> None:
...
                 {
-                    "textDocument": {"uri": file_uri, "version": next_version},
+                    "textDocument": {"uri": file_uri, "version": 2},
                     "contentChanges": [{"text": content}],
                 },
             )
-            logger.debug(f"Sent didChange for {file_path} to {client.language.language} LSP")
+            logger.debug(f"Sent didOpen+didChange for {file_path} to {client.language.language} LSP")
 
     def analyze(self, cache_dir: Path | None = None) -> StaticAnalysisResults:
```
brovatten, line 241: bug (high) - __init__.py: Ghost diagnostics survive indefinitely on the first analyze() call. _previous_live_diagnostics defaults to set() on a fresh client. If the cache has stale errors for file.py but file.py is now clean, it won't appear in live_diags (popped at L271) and the eviction loop at L318-320 won't fire (because previously_live is empty). Stale error persists in merged_diags. Fix: use the set of files actually opened by build_static_analysis / notify_file_changed as the eviction set, rather than relying on the previous live snapshot.

static_analyzer/__init__.py
```diff
@@ -187,16 +195,31 @@ def notify_file_changed(self, file_path: Path, content: str) -> None:
...
+                        "text": content,
+                    }
+                },
+            )
+            # Send the change so pyright re-type-checks with the new content.
             client._send_notification(
                 "textDocument/didChange",
                 {
-                    "textDocument": {"uri": file_uri, "version": next_version},
+                    "textDocument": {"uri": file_uri, "version": 2},
```
brovatten: bug (medium) - __init__.py: Sending didOpen + didChange(version=2) unconditionally for every notify_file_changed violates the LSP spec if the file is already open. Pyright tolerates it, but other language servers (gopls, jdtls) may reject or desync. Track open-file state and send only didChange for already-open documents.

brovatten: High: notify_file_changed now
     violates LSP document lifecycle/
     versioning and can cause dropped
     updates after the first save.
      - It always sends didOpen with
        version: 1 and didChange with
        version: 2 on every call, even
        for already-open docs
        (static_analyzer/
        __init__.py:200,
        static_analyzer/
        __init__.py:215).
      - There is no per-document
        monotonic version tracking in
        this path anymore, so repeated
        saves reuse stale versions and
        can be ignored by strict
        servers.
      - Practical effect: diagnostics
        can become stale or stop
        updating in long-lived sessions.

static_analyzer/__init__.py
```diff
@@ -257,25 +280,60 @@ def analyze(self, cache_dir: Path | None = None) -> StaticAnalysisResults:
...
+                # previous live run) but are now absent from live_diags.
+                # Use the previous live snapshot stored on the client to identify
+                # which files have been actively cleared vs simply not seen.
+                previously_live = client._previous_live_diagnostics
+                for file_path in previously_live:
+                    if file_path not in live_diags:
+                        merged_diags.pop(file_path, None)
+
+                # Remember which files the client tracked this cycle.
+                client._previous_live_diagnostics = set(live_diags.keys()) | previously_live
```
brovatten: Bug 1 — Critical:
  _previous_live_diagnostics accumulates
  via union, defeating the cache

  File: static_analyzer/__init__.py:323

  # BUG: union means this set only grows,
  never shrinks
  client._previous_live_diagnostics =
  set(live_diags.keys()) | previously_live

  The eviction loop at lines 317–320 reads
   this set and pops any file in it that
  is absent from the current live_diags.
  Because the set is updated with |
  (union), it accumulates every file that
  has ever had live diagnostics across all
   analyze() calls. On subsequent calls,
  any file that was live in a previous
  session but isn't in the current
  live_diags (e.g., because it's being
  served from cache and wasn't re-opened)
  gets evicted.

  Worse: if pyright sends
  publishDiagnostics([]) on didClose
  (which is exactly what the new
  _handle_diagnostics_notification handles
   — the PR explicitly supports this),
  then after build_static_analysis closes
  all files, client.diagnostics is
  emptied. On the next analyze() call,
  live_diags = {}, so every file in
  _previous_live_diagnostics is evicted
  from merged_diags. The entire diagnostic
   cache is nuked on every second call.

  # Should be:
  client._previous_live_diagnostics =
  set(live_diags.keys())  # replace, not
  union

static_analyzer/lsp_client/client.py
brovatten: Bug 3 — Medium: Race condition between
  didClose notifications and
  get_collected_diagnostics

  File: static_analyzer/__init__.py:298,
  static_analyzer/lsp_client/client.py:773
  –778

  The flow in analyze():
  analysis =
  client.build_static_analysis()   # (1)
  opens all files, waits for diagnostics,
  closes all files
  ...
  live_diags =
  client.get_collected_diagnostics()  #
  (2) reads client.diagnostics snapshot

  In build_static_analysis,
  _wait_for_diagnostics correctly waits
  for pyright to emit its type-check
  results, then sends didClose for all
  files. If pyright sends
  publishDiagnostics([]) responses to
  those close events (asynchronously via
  the reader thread), they can arrive and
  zero-out entries in client.diagnostics
  either before or after step (2). The
  diagnostic snapshot is therefore
  non-deterministically partial — some
  files may or may not be present
  depending on scheduling.

  The old time.sleep(2) was blunt but at
  least came before the close loop. Now
  there's a window between the batch
  didClose sends and the
  get_collected_diagnostics call.

PR #222, opened 2026-03-01
Fix warning lines and improve deduplication

Svilen-Stefanov:
Tiny PR to fix the wrong warning lines, as well as remove duplicate warnings. 

## Release Notes

* **Bug Fixes**
  * Improved deduplication of unused code diagnostics to reduce duplicate reports when multiple diagnostic sources identify the same issue.
  * Fixed line number reporting to use 1-based numbering (matching standard editor conventions) instead of internal indexing for improved clarity.
  * Enhanced logging for tracking and reporting duplicate diagnostics.

### DIFF
--- a/health/checks/unused_code_diagnostics.py
+++ b/health/checks/unused_code_diagnostics.py
@@ -136,13 +136,17 @@ class LSPDiagnosticsCollector:
         """Process all collected diagnostics and convert to issues."""
         self.issues = []
         skipped = 0
-        seen_issues: set[tuple[str, str, int]] = set()  # (file_path, code, line_start)
+        seen_issues: set[tuple[str, str, int]] = set()  # (file_path, category, line_start)
 
         for item in self.diagnostics:
             issue = self._convert_to_issue(item.file_path, item.diagnostic)
             if issue:
-                # Deduplicate: same file, code, and line
-                key = (issue.file_path, issue.code or "", issue.line_start)
+                # Deduplicate: same file, category, and line.
+                # Using category instead of code prevents duplicates from the same
+                # LSP server reporting the same issue under different diagnostic codes
+                # (e.g. Pyright reports unused variables both via reportUnusedVariable
+                # and via the "Unnecessary" tag with different messages).
+                key = (issue.file_path, issue.category.value, issue.line_start)
                 if key not in seen_issues:
                     seen_issues.add(key)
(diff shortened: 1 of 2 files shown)


health/checks/unused_code_diagnostics.py
```diff
@@ -187,8 +191,8 @@ def _convert_to_issue(self, file_path: str, diagnostic: LSPDiagnostic) -> Diagno
 
         return DiagnosticIssue(
             file_path=file_path,
-            line_start=diagnostic.range.start.line,
-            line_end=diagnostic.range.end.line,
+            line_start=diagnostic.range.start.line + 1,  # LSP lines are 0-based, convert to 1-based
+            line_end=diagnostic.range.end.line + 1,  # LSP lines are 0-based, convert to 1-based
```
ivanmilevtues, line 195: nice, that is really annoying in vscode ;d

brovatten: nice!!

health/checks/unused_code_diagnostics.py
```diff
@@ -136,13 +136,17 @@ def process_diagnostics(self) -> list[DiagnosticIssue]:
...
             issue = self._convert_to_issue(item.file_path, item.diagnostic)
             if issue:
-                # Deduplicate: same file, code, and line
-                key = (issue.file_path, issue.code or "", issue.line_start)
+                # Deduplicate: same file, category, and line.
+                # Using category instead of code prevents duplicates from the same
+                # LSP server reporting the same issue under different diagnostic codes
+                # (e.g. Pyright reports unused variables both via reportUnusedVariable
+                # and via the "Unnecessary" tag with different messages).
+                key = (issue.file_path, issue.category.value, issue.line_start)
```
brovatten, line 149: category.value is not a great naming. i can sort of understand category is the type of issue? but then what is value?
also, shouldn't issues be named "warnings"? i guess this is what the user sees

Svilen-Stefanov: `.value` is the pythonic way to take the value of an enum so we can't change that. 
The original idea of this being called issues was that a flagged "issue" can have different severity (`info`, `warning`, `critical` etc). I think this still makes sense even though we decided to have a single severity atm. Let's refactor this separately if we want to do so.

PR #223, opened 2026-03-05
Use component cluster ids as component ids

Svilen-Stefanov:
This PR moves away from random component ids to structural ones. Example:
- lvl 1 components have names (1, 2, 3 ..., <number_of_components>)
- The names of all subcomponents of 1 are called 1.1, 1.2, 1.3 etc.
- With each next level, we add a new suffix to the component name

The main benefit of this is that it enables us to provide detailed context in relationships. If we have a relationship between 1.3  and 2.6.9, this also means that 1 and 2 are connected and 1.3 and 2.6 are connected, which enables us to have the "right" level of relationship granularity when we need it.

So far, I haven't updated the .codeboarding directory for 2 reasons:
1. Reduce the diff to make it easier to review
2. This branch is based on another unmerged one

Currently, this won't change too much atm as relationships are still per level. I'm going to create a second separate PR that will make relationships global / cross level.

Tested both with full analysis and partial analysis. @ivanmilevtues I'm not checking incremental analysis as you mentioned you'll change it.

## Summary by CodeRabbit

* **New Features**
  * Component IDs now use hierarchical dotted notation (e.g., "1", "1.2", "1.2.1").
  * Static CFG-based inter-component analysis added; relations now include edge counts and a static flag and are merged with detected relations.

* **Documentation**
  * CLI examples and README updated to show the new component ID format.

* **Tests**
  * New and updated tests covering static relation extraction, merging behavior, and hierarchical ID assignment.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/223" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/integration/fixtures/mockito_java.json
+++ b/tests/integration/fixtures/mockito_java.json
@@ -17,7 +17,6 @@
   },
   "sample_references": [
     "mockito-core.src.main.java.org.mockito.additionalanswers.additionalanswers",
-    "mockito-core.src.main.java.org.mockito.additionalanswers.additionalanswers()",
     "mockito-core.src.main.java.org.mockito.additionalanswers.answer(answer1<t, a>) <t, a>",
     "mockito-core.src.main.java.org.mockito.additionalanswers.answer(answer2<t, a, b>) <t, a, b>",
     "mockito-core.src.main.java.org.mockito.additionalanswers.answer(answer3<t, a, b, c>) <t, a, b, c>",
--- a/tests/integration/fixtures/prometheus_go.json
+++ b/tests/integration/fixtures/prometheus_go.json
@@ -13,7 +13,7 @@
     "call_graph_nodes": 5377,
     "call_graph_edges": 9244,
     "source_files_count": 350,
-    "execution_time_seconds": 270
+    "execution_time_seconds": 395
   },
   "sample_references": [
     "cmd.prometheus.main.(*flagconfig).setfeaturelistoptions",
--- a/tests/integration/fixtures/wordpress_php.json
+++ b/tests/integration/fixtures/wordpress_php.json
@@ -13,7 +13,7 @@
     "call_graph_nodes": 8328,
     "call_graph_edges": 28088,
     "source_files_count": 1644,
-    "execution_time_seconds": 400
+    "execution_time_seconds": 106
   },
   "sample_references": [
     "index.wp_use_themes",


agents/cluster_methods_mixin.py
```diff
@@ -430,3 +431,23 @@ def populate_file_methods(self, analysis: AnalysisInsights, cluster_results: dic
...
+    def build_static_relations(
+        self,
+        analysis: AnalysisInsights,
+        cfg_graphs: dict[str, CallGraph] | None = None,
+    ) -> None:
+        """Build inter-component relations from CFG edges and merge with LLM relations.
+
+        Replaces LLM-only relations with statically-backed ones:
+        - LLM + static match: keep LLM label, attach edge_count
+        - LLM only (no static backing): drop
```
ivanmilevtues, line 458: this is dangerous, it might be that for some reason we drop all relations of a component and it becomes disconnected for some odd reason.

Hmm, i am not sure how this should look. Just I think within the level itself, the LLM relations should exist.
When it is cross level then that is where you might want to actually add the static edges, unsure. My concern comes from having a nasty insane network of edges, making everything hard to understand/read.

diagram_analysis/analysis_json.py
```diff
@@ -16,10 +16,12 @@
 
 
 class RelationJson(Relation):
-    """Relation subclass that includes src_id/dst_id in JSON serialization."""
+    """Relation subclass that includes src_id/dst_id and static analysis evidence in JSON serialization."""
 
     src_id: str = Field(default="", description="Component ID of the source.")
     dst_id: str = Field(default="", description="Component ID of the destination.")
+    edge_count: int = Field(default=0, description="Number of CFG edges backing this relation.")
+    is_static: bool = Field(default=False, description="True if derived from static CFG analysis.")
```
ivanmilevtues, line 24: Is the idea of this to sort it out in UI?

agents/abstraction_agent.py
```diff
@@ -138,16 +138,19 @@ def run(self):
 
         # Step 2: Generate abstract components from grouped clusters
         analysis = self.step_final_analysis(cluster_analysis, cluster_results)
-        # Step 3: Assign deterministic component IDs (must happen before methods that key on component_id)
+        # Step 3: Assign hierarchical component IDs ("1", "2", "3", ...)
```
brovatten, line 144:
```suggestion
        # Step 3: Assign hierarchical component IDs ("1", "2", "3", "1.1", "1.2", ...)
```

agents/cluster_methods_mixin.py
```diff
@@ -440,3 +445,24 @@ def populate_file_methods(self, analysis: AnalysisInsights, cluster_results: dic
...
+        - Static only: add with auto-label "calls"
+
+        If cfg_graphs is not provided, builds them from self.static_analysis.
+        """
```
brovatten, line 462: have you seen this be the case? not relevant for this PR prolly:

- High: Nested detail analyses leak full-repo nodes into subcomponents. `agents/cluster_methods_mixin.py:204` pulls `cfg.nodes` from the full language CFG, and `agents/cluster_methods_mixin.py:421` then assigns every collected node during `populate_file_methods()`. When `agents/details_agent.py:182` calls that path for a filtered subgraph, unrelated files become graph-distance or fallback orphans and are written into the sub-analysis. Confirmed locally: a subgraph containing only `a.py` still assigned `b.py` and fallback `c.py`.

agents/abstraction_agent.py
ivanmilevtues: 1.1 and 1.2 are for lovel levle so details agent if i understand correctly

agents/abstraction_agent.py
```diff
@@ -138,16 +138,19 @@ def run(self):
 
         # Step 2: Generate abstract components from grouped clusters
         analysis = self.step_final_analysis(cluster_analysis, cluster_results)
-        # Step 3: Assign deterministic component IDs (must happen before methods that key on component_id)
+        # Step 3: Assign hierarchical component IDs ("1", "2", "3", "1.1", "1.2", ...)
```
ivanmilevtues: I think this should be without the 1.1 as this is the abstraction agent :D

Doesn't matter that much.

static_analyzer/cluster_relations.py
```diff
@@ -0,0 +1,168 @@
...
+def merge_relations(
+    llm_relations: list[Relation],
+    static_relations: list[ClusterRelation],
+    analysis: AnalysisInsights,
+) -> list[Relation]:
+    """Merge LLM-generated relations with static analysis evidence.
+
+    Strategy (Supplement):
+    - Static + matching LLM: keep LLM's human-readable label, attach edge_count, is_static=True
+    - LLM only (no static backing): drop it
```
ivanmilevtues: I wouldn't drop the LLM only, as it might be the case that the Static analysis failed.
Just keep it and then somehow in the UI display the issue (not backed), we have the var anyway.

static_analyzer/cluster_relations.py
```diff
@@ -0,0 +1,168 @@
...
+    - Static only (no LLM label): add with auto-generated label "calls" + edge_count
```
ivanmilevtues, line 96: this is okay.

ivanmilevtues (approved):
The only real comment is on the dropping of the edges, please take a look at that.

I think stat analysis is not good enough yet to drop llm made edges i.e. we won't find back and front end communication through stat analysis.

PR #224, opened 2026-03-06
Fix static analysis issues

Svilen-Stefanov:
The PR is the first step to improving the static analysis, which currently misses a lot of cross-file edges.

## Summary by CodeRabbit

* **New Features**
  * Multi-language analysis engine with first-class adapters (Go, Java, PHP, Python, TypeScript/JavaScript), richer call-graph and class-hierarchy construction, and broader symbol-kind coverage.

* **Bug Fixes**
  * More robust reference matching (case-insensitive, generics/receiver-aware), improved incremental caching per language/project, and fewer dropped/misattributed edges.

* **Refactor**
  * Major engine rearchitecture to an engine-driven, per-project analysis pipeline enabling mono-repo support.

* **Tests**
  * Expanded unit/integration tests, updated snapshots, and new diagnostics scripts.

* **Chores**
  * CI now exposes stdout and INFO-level logs during test runs.

### DIFF
--- a/static_analyzer/engine/adapters/typescript_adapter.py
+++ b/static_analyzer/engine/adapters/typescript_adapter.py
@@ -15,7 +15,7 @@ class TypeScriptAdapter(LanguageAdapter):
 
     @property
     def file_extensions(self) -> tuple[str, ...]:
-        return (".ts", ".tsx")
+        return (".ts", ".tsx", ".mts", ".cts")
 
     @property
     def lsp_command(self) -> list[str]:
--- a/static_analyzer/engine/language_adapter.py
+++ b/static_analyzer/engine/language_adapter.py
@@ -214,7 +214,7 @@ class LanguageAdapter(ABC):
         detail: str = "",
     ) -> str:
         """Build the name used in call graph edges. Defaults to qualified_name."""
-        return self.build_qualified_name(file_path, symbol_name, symbol_kind, parent_chain, project_root)
+        return self.build_qualified_name(file_path, symbol_name, symbol_kind, parent_chain, project_root, detail)
 
     def get_all_packages(self, source_files: list[Path], project_root: Path) -> set[str]:
         """Get all packages that should appear in package dependencies.
--- a/repo_utils/ignore.py
+++ b/repo_utils/ignore.py
@@ -14,7 +14,7 @@ CODEBOARDINGIGNORE_TEMPLATE = """# CodeBoarding Ignore File
 # NOTE: The following are ALWAYS excluded (not configurable):
 #   - Hidden directories (starting with .)
 #   - .git/, .codeboarding/, node_modules/, __pycache__/
-#   - Build output: build/, dist/, target/, bin/, out/, coverage/
+#   - Build output: build/, dist/, coverage/
 #
 # This file is automatically loaded by CodeBoarding analysis tools to exclude
 # specified paths from code analysis, architecture generation, and other processing.
--- a/static_analyzer/engine/lsp_client.py
+++ b/static_analyzer/engine/lsp_client.py
@@ -83,7 +83,7 @@ class LSPClient:
             self._command,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
             env=env,
             cwd=str(self._project_root),
         )
@@ -476,7 +476,7 @@ class LSPClient:
             message = self._msg_queue.get(timeout=min(remaining, 1.0))
         except queue.Empty:
             if self._process and self._process.poll() is not None:
-                raise RuntimeError(f"LSP server process exited with code {self._process.returncode}")
+                raise RuntimeError(f"LSP server process exited with code {self._process.returncode}") from None
             return None
 
         # Handle server-initiated requests (e.g. workspace/configuration)
--- a/tests/static_analyzer/test_lsp_client.py
+++ b/tests/static_analyzer/test_lsp_client.py
@@ -399,7 +399,7 @@ class TestCollectBatchResponses:
         import logging
 
         with caplog.at_level(logging.WARNING):
-            results, _, error_ids = client._collect_batch_responses([1, 2, 3], timeout=5)
+            _, _, error_ids = client._collect_batch_responses([1, 2, 3], timeout=5)
 
         assert error_ids == {1, 2, 3}
         # Should have a single deduplicated warning, not 3 separate ones
@@ -428,7 +428,7 @@ class TestHandleNotification:
 
         diags = client.get_collected_diagnostics()
         assert len(diags) == 1
-        file_key = list(diags.keys())[0]
+        file_key = next(iter(diags))
         assert len(diags[file_key]) == 1
         assert diags[file_key][0].message == "Unused import"
 


static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+        if root_str := java_entry.get("jdtls_root"):
+            root = Path(root_str)
+            if root.is_dir() and (root / "plugins").is_dir():
+                return root
+
+        # 2. Fallback to well-known locations
+        for location in [
+            Path.home() / ".jdtls",
+            Path.home() / ".codeboarding" / "servers" / "bin" / "jdtls",
+            Path("/opt/jdtls"),
```
ivanmilevtues: this fallback doesn't address windows default well-know locations

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+        file_count = len(jvm_files)
+        if file_count < 100:
+            return "1G"
+        elif file_count < 500:
+            return "2G"
+        elif file_count < 2000:
+            return "4G"
+        elif file_count < 5000:
+            return "6G"
+        return "8G"
```
ivanmilevtues: Better to check how much ram does the system have, as 8Gigs is substential.

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+        for param in JavaAdapter._split_params(params_raw):
+            param = param.strip()
+            if not param:
+                continue
+            params.append(JavaAdapter._strip_generics(param))
+
+        result = f"{method_name}({', '.join(params)})"
+        if suffix:
+            result += " " + suffix
+        return result
```
ivanmilevtues, line 194: Can't this be done with 1 simple regex?

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+        result = []
+        depth = 0
+        for ch in param:
+            if ch == "<":
+                depth += 1
+            elif ch == ">":
+                depth -= 1
+            elif depth == 0:
+                result.append(ch)
+        return "".join(result).strip()
```
ivanmilevtues, line 208: This also feels like a job for a regex

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+                effective_parents = effective_parents[1:]
+
+            if effective_parents:
+                clean_parents = ".".join(self._clean_symbol_name(name) for name, _ in effective_parents)
+                return f"{module}.{clean_parents}.{clean_name}"
+        return f"{module}.{clean_name}"
+
+    @staticmethod
+    def _clean_symbol_name(name: str) -> str:
+        """Clean a JDTLS symbol name by stripping generics from params."""
```
ivanmilevtues, line 157: Is there a reason to strip out these even?

I was reading all the things that do those weird for loops, which remove generics.

I think generics are good information, unsure if the LLM would pick it up, but it might be good. Wdyt?

Also generally I don't like this stripping, proly it added it because of fixtures idk?

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+
+        JDTLS serializes references requests (~1-10s each), making the default
+        references-based approach impractical for large projects.
+        """
+        return build_edges_via_definitions(self, ctx, source_files)
+
+    def should_track_for_edges(self, symbol_kind: int) -> bool:
+        return symbol_kind in (CALLABLE_KINDS | CLASS_LIKE_KINDS | {SYMBOL_KIND_VARIABLE, SYMBOL_KIND_CONSTANT})
+
+    def is_test_file(self, file_path: Path) -> bool:
```
ivanmilevtues: I think it is a better approach to skip the testfiles at a file ignore level and not per technology.

static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -0,0 +1,277 @@
...
+        name = file_path.name
+        # Exclude Java metadata files that contain no analyzable symbols
+        if name in ("package-info.java", "module-info.java"):
+            return True
+        return super().is_test_file(file_path)
+
+    def get_excluded_dirs(self) -> set[str]:
+        return super().get_excluded_dirs() | {
```
ivanmilevtues: maybe here you skip the test, or even better in the .codeboarding directory you add it as a default in the .codeboardingignore.

I can see a world where I want to see how my complicated tests are organized because they use a bunch of abstractions and mocking setup.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+            "build",
+            ".next",
+            "coverage",
+        }
+
+    def get_all_packages(self, source_files: list[Path], project_root: Path) -> set[str]:
+        return self._get_hierarchical_packages(source_files, project_root)
+
+
+class JavaScriptAdapter(TypeScriptAdapter):
```
ivanmilevtues, line 35: It seems weird to extend TS for the JS, I would expect it to be the other way around. But it makes sense as we are using the ts lsp.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+from static_analyzer.engine.language_adapter import LanguageAdapter
+from static_analyzer.engine.lsp_client import LSPClient
+from static_analyzer.engine.models import CallFlowGraph, EdgeBuildContext, LanguageAnalysisResult
+from static_analyzer.engine.source_inspector import SourceInspector
+from static_analyzer.engine.symbol_table import SymbolTable
+
+logger = logging.getLogger(__name__)
+
+# Batch size for did_open to avoid overwhelming LSP servers
+DID_OPEN_BATCH_SIZE = 50
```
ivanmilevtues: move to constants file for the static_analysis.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+        ctx = EdgeBuildContext(self._lsp, self._symbol_table, self._source_inspector)
+        edge_set = self._adapter.build_edges(ctx, source_files)
+        edge_set = self._postprocess_edges(edge_set)
+        t_edges_done = time.monotonic()
+        logger.info("Phase 2 total (build edges): %.1fs, %d edges", t_edges_done - t_indices_done, len(edge_set))
+
+        hierarchy_builder = HierarchyBuilder(self._lsp, self._symbol_table, self._source_inspector, self._adapter)
+        hierarchy = hierarchy_builder.build()
+        t_hierarchy_done = time.monotonic()
+        logger.info("Phase 3 (hierarchy): %.1fs", t_hierarchy_done - t_edges_done)
```
ivanmilevtues: We can actually skip phase 3, we don't leverage hierarchy. Unsure how big of a help is (maybe we use it for health unsure). But for the agents we don't use it.

It used to be something that we have in order to "force" indexing.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+
+        package_deps = self._build_package_deps(edge_set, source_files)
```
ivanmilevtues, line 79: Same as hierachy. We don't really use this. We should probably use it thou unsure how 🤷

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def _discover_symbols(self, source_files: list[Path]) -> None:
+        """Phase 0+1: Open files, wait for indexing, then extract all symbols."""
+        total = len(source_files)
+        t_open_start = time.monotonic()
+        logger.info("Phase 0 (open): opening %d files for indexing", total)
+        for i in range(0, total, DID_OPEN_BATCH_SIZE):
+            batch = source_files[i : i + DID_OPEN_BATCH_SIZE]
+            for file_path in batch:
+                self._lsp.did_open(file_path, self._adapter.language_id)
```
ivanmilevtues, line 136: it sounds as this might return false and we are more interested in counting the trues instead of adding the batchsize.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+            opened = min(i + DID_OPEN_BATCH_SIZE, total)
+            logger.info("Phase 0 (open): %d/%d files opened", opened, total)
+            if i + DID_OPEN_BATCH_SIZE < total:
+                time.sleep(0.1)
```
ivanmilevtues: I think this if is useless? It basically sleeps on every iteration between batches, I think you can just have the sleep on its own. Unsure if we need 0.1 sleep as well. I suppose it gives time for the lsp to process stuff. (interestingly we might just want to have a blocking/semi-blocking idk, await on the did_open?)

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+        logger.info(
+            "Sync probe: %.1fs (%d symbols)",
+            time.monotonic() - t_probe,
+            len(probe_result) if probe_result else 0,
+        )
+
+        total = len(source_files)
+        for idx, file_path in enumerate(source_files, 1):
+            if idx == 1 and probe_result is not None:
+                symbols = probe_result
```
ivanmilevtues, line 179: Do we need this if?
It feels like if it has indexed the document_symbol should give almost instant result either way. The first file optimization seems like it won't make a big difference but makes me wunder on the code, so an LLM would also like to use that if for other stuff :D

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+        # Warmup probe: trigger the LSP server's cross-reference index build
+        # by sending a single references request with a long timeout.
+        if source_files and self._adapter.references_per_query_timeout > 0:
+            logger.info("Phase 1.5 (warmup): triggering LSP index build with a single references request...")
+            t_warmup = time.monotonic()
+            try:
+                self._lsp.references(source_files[0], 0, 0)
+            except (TimeoutError, Exception) as e:
+                logger.warning("Warmup probe failed (non-fatal): %s", e)
+            logger.info("Phase 1.5 (warmup): completed in %.1fs", time.monotonic() - t_warmup)
```
ivanmilevtues: This has to be a different function warmup_references or something and it should be called in the orchestrator func rather than here imo

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+                    alias_self_edges += 1
+                    continue
+                edge_key = (src_sym.definition_location, dst_sym.definition_location)
+                existing = pos_to_edge.get(edge_key)
+                if existing is None or (len(src) + len(dst), src, dst) > (
+                    len(existing[0]) + len(existing[1]),
+                    existing[0],
+                    existing[1],
+                ):
+                    pos_to_edge[edge_key] = (src, dst)
```
ivanmilevtues, line 243: add a comment that this replaces an edge with the longest qualified names possible ;d.

It took me some head scratching to figure it out :d

static_analyzer/engine/call_graph_builder.py
```diff
@@ -0,0 +1,241 @@
...
+        if source_files and self._adapter.references_per_query_timeout > 0:
+            logger.info("Phase 1.5 (warmup): triggering LSP index build with a single references request...")
+            t_warmup = time.monotonic()
+            try:
+                self._lsp.references(source_files[0], 0, 0)
+            except (TimeoutError, Exception) as e:
+                logger.warning("Warmup probe failed (non-fatal): %s", e)
+            logger.info("Phase 1.5 (warmup): completed in %.1fs", time.monotonic() - t_warmup)
+
+    def _postprocess_edges(self, edge_set: set[tuple[str, str]]) -> set[tuple[str, str]]:
```
ivanmilevtues, line 208: I really don't get the aliasing in this method.
Alias is when you say: funcb = funca (right?)

Then if 2 symbols src and dst are in the same file, we say they have an alias edge?

That i really don't understand.
Can you elaborate on this, why do we need it?

I suppose without it tests fail, but to me this is an indication of not using the lsp properly OR our qualified name is not working properly.

static_analyzer/engine/edge_builder.py
```diff
@@ -0,0 +1,441 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# References-based strategy (default)
+# ---------------------------------------------------------------------------
+
+
+def build_edges_via_references(
```
ivanmilevtues, line 33: 140 loc for a func, i think this deserves some splits ;d

static_analyzer/engine/edge_builder.py
```diff
@@ -0,0 +1,441 @@
...
+                    ref_file = uri_to_path(ref_uri)
+                    if ref_file is None:
+                        continue
+                    ref_loc = (str(ref_file), ref_line, ref_char)
+                    if ref_loc == sym_def_loc:
+                        continue
+
+                    refs_total += 1
+
+                    if adapter.is_class_like(sym.kind):
```
ivanmilevtues: maybe transfor the double ifs into if A and B, i think it'd be easier to read.

static_analyzer/engine/edge_builder.py
```diff
@@ -0,0 +1,441 @@
...
+    )
+    return edge_set
+
+
+# ---------------------------------------------------------------------------
+# Definition-based strategy (Java / JDTLS)
+# ---------------------------------------------------------------------------
+
+
+def build_edges_via_definitions(
```
ivanmilevtues, line 226: again quite a long func, maybe split. At least the phases ;d

static_analyzer/engine/language_adapter.py
```diff
@@ -0,0 +1,268 @@
+"""Abstract base class for language-specific LSP adapters."""
+
+from __future__ import annotations
+
+import logging
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+from static_analyzer.engine.edge_builder import build_edges_via_references
+from static_analyzer.engine.lsp_constants import (  # noqa: F401 — re-exported for backward compatibility
```
ivanmilevtues: No need to have backward compatibility!

Clean and new is the way to go :)

static_analyzer/engine/models.py
```diff
@@ -0,0 +1,116 @@
+"""Data models for static analysis results."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
ivanmilevtues: that is smelly ;d
remove it, it might mean there is circular dep idk.

Also __future__ import annotations, i hate this have no idea what it does, only agents use it ;f

static_analyzer/engine/result_converter.py
```diff
@@ -0,0 +1,133 @@
...
+        "class_hierarchies": result.hierarchy,
+        "package_relations": result.package_dependencies,
+        "references": references,
+        "source_files": result.source_files,
+        "diagnostics": {},
+    }
+
+
+def _map_symbol_kind(kind: int) -> NodeType:
+    """Map an LSP SymbolKind integer to CodeBoarding's NodeType.
```
ivanmilevtues, line 125: use only one type imo. No need to have 2 (this proly comes from backward compat)

static_analyzer/engine/scanner.py
```diff
@@ -0,0 +1,57 @@
...
+
+import logging
+from pathlib import Path
+
+from static_analyzer.engine.language_adapter import LanguageAdapter
+
+logger = logging.getLogger(__name__)
+
+
+class EngineProjectScanner:
```
ivanmilevtues: I think there might be a duplication on the scanner can you check what happened with: `static_analyzer/scanner.py` 
I don't see it being removed or modified, so unsure ;d

static_analyzer/engine/source_inspector.py
```diff
@@ -0,0 +1,196 @@
...
+        before = line[:ref_start_char]
+        depth = 0
+        for ch in reversed(before):
+            if ch == ")":
+                depth += 1
+            elif ch == "(":
+                if depth == 0:
+                    # We found an unmatched '(' — this reference is inside a call
+                    return True
+                depth -= 1
```
ivanmilevtues: Regex, or at least a new func ;d

static_analyzer/engine/source_inspector.py
```diff
@@ -0,0 +1,196 @@
...
+                "void",
+                "assert",
+                "synchronized",
+                "try",
+                "instanceof",
+                "typeof",
+                "sizeof",
+                "default",
+            }
+        )
```
ivanmilevtues, line 170: These seem to be python specific. i.e. I see import but not include

static_analyzer/engine/symbol_table.py
```diff
@@ -0,0 +1,332 @@
...
+                child_kinds = {c.get("kind", 0) for c in children}
+                if child_kinds & CALLABLE_KINDS:
+                    kind = SYMBOL_KIND_CLASS
+
+            range_info = sym.get("range", sym.get("location", {}).get("range", {}))
+            sel_range = sym.get("selectionRange", range_info)
+
+            start = range_info.get("start", {})
+            end = range_info.get("end", {})
+            sel_start = sel_range.get("start", start)
```
ivanmilevtues, line 89: does nothing, reuse L93

static_analyzer/engine/symbol_table.py
```diff
@@ -0,0 +1,332 @@
...
+                if not self._adapter.is_callable(sym.kind):
+                    continue
+                if not sym.qualified_name.startswith(best.qualified_name + "."):
+                    continue
+                gap = sym.start_line - line
+                if 0 < gap < nearest_gap:
+                    nearest_child = sym
+                    nearest_gap = gap
+            if nearest_child is not None:
+                best = nearest_child
```
ivanmilevtues, line 239: This seems python specific, proly should go into the adapter somehow unsure.

static_analyzer/__init__.py
```diff
@@ -20,101 +27,100 @@
...
+
         try:
             if lang_lower in (Language.TYPESCRIPT, Language.JAVASCRIPT):
-                # For TypeScript/JS, scan for multiple project configurations (mono-repo support)
                 ts_config_scanner = TypeScriptConfigScanner(repository_path, ignore_manager=ignore_manager)
                 typescript_projects = ts_config_scanner.find_typescript_projects()
 
                 if typescript_projects:
-                    # Create a separate client for each TypeScript project found
                     for project_path in typescript_projects:
```
ivanmilevtues, line 66: nested ts configs are a problem. unsure how to handle just mentioning it.

static_analyzer/__init__.py
```diff
@@ -20,101 +27,100 @@
...
+        "python": "Python",
+        "typescript": "TypeScript",
+        "javascript": "JavaScript",
+        "tsx": "TypeScript",
+        "jsx": "JavaScript",
+        "go": "Go",
+        "java": "Java",
+        "php": "PHP",
+    }
+    return mapping.get(language.lower())
```
ivanmilevtues, line 111: I don't reeally like this, i would rather have the adapter know what it can do. I.e. JS/TS adapters should have the can_handle method working with tsx, typescript, ts. Same for JS.

Just another place to maintain this data that's why i don't like it

static_analyzer/__init__.py
```diff
@@ -1,17 +1,24 @@
...
 from repo_utils import get_git_commit_hash
 from repo_utils.ignore import RepoIgnoreManager
+from static_analyzer.analysis_cache import AnalysisCacheManager
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.cluster_change_analyzer import ChangeClassification
 from static_analyzer.constants import Language
+from static_analyzer.engine.adapters import get_adapter
+from static_analyzer.engine.call_graph_builder import CallGraphBuilder
+from static_analyzer.engine.language_adapter import LanguageAdapter
+from static_analyzer.engine.lsp_client import LSPClient as EngineLSPClient
```
ivanmilevtues: Why the alias? I got confused ;d

static_analyzer/__init__.py
```diff
@@ -135,116 +141,77 @@ def start_clients(self) -> None:
...
-                logger.info(f"Starting LSP client for {client.language.language}")
-                client.start()
-                started_clients.append(client)
-                if isinstance(client, JavaClient):
-                    client.wait_for_import()  # timeout auto-computed based on project size
+                logger.info(f"Starting engine LSP client for {adapter.language} at {project_path}")
+                t_start = time.monotonic()
+                command = adapter.get_lsp_command(project_path)
+                init_options = adapter.get_lsp_init_options()
+                engine_client = EngineLSPClient(
```
ivanmilevtues: Can't we use the with block that you created instead of hacing to have the try except?

The lsp has the overrides for a with block

static_analyzer/incremental_orchestrator.py
```diff
@@ -16,9 +16,13 @@
     analyze_cluster_changes_for_languages,
```
ivanmilevtues, line 16: this will need rework as we discussed, not in this pr.

I am not reading through it :)

tests/static_analyzer/scripts/diagnose_relations.py
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env python3
+"""Diagnose static analysis edge distribution without running LLMs.
```
ivanmilevtues, line 2: do we need this script, seems like a one time thing for you to test your change.

If we want this I think maybe its better to just have the tests?

tests/static_analyzer/scripts/dropped_edges_diagnostic.py
```diff
@@ -0,0 +1,82 @@
+"""Diagnostic script to analyze dropped edges from CFG construction."""
```
ivanmilevtues, line 1: Same comment as the other scripts scripts

tests/static_analyzer/scripts/edge_recovery_diagnostic.py
```diff
@@ -0,0 +1,143 @@
+"""Diagnostic script to identify recoverable dropped edges."""
```
ivanmilevtues, line 1: same comment as diagnose_relations.py script

tests/test_windows_compatibility.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 import os
 
-from static_analyzer.lsp_client.client import uri_to_path
+from static_analyzer.engine.utils import uri_to_path
```
ivanmilevtues, line 10: now I am convinced that there are 2 scanners, i didn't see a test where it was modified, and in the integrations I think we use it ;d

ivanmilevtues (approved):
I am not blocking the PR, I added lots of comments, would like to address them, but don't want to stop the merge.

Thanks for the change, it makes things much easier to understand esp around the lsp!

Awesome to see!

static_analyzer/engine/scanner.py
```diff
@@ -0,0 +1,57 @@
...
+    """Scans a project directory to detect languages and discover source files."""
```
brovatten: Is the .gitignore .codeboarding .codeboardingignore still ignore here? I think it used to consult RepoIgnoreManager but doesn't anymore

static_analyzer/engine/call_graph_builder.py
Svilen-Stefanov: Yeah, I was thinking if I should remove this and the hierarchy stuff overall or not. We can remove them for now and explore adding them later? I'd prefer to focus on stabilizing what we have / use atm and then extend it to using hierarchy/package deps.

tests/static_analyzer/scripts/diagnose_relations.py
Svilen-Stefanov: We don't need it, I used this for testing / verifying edges.

PR #230, opened 2026-03-13
Fix Windows encoding

Svilen-Stefanov:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved console logging robustness to gracefully handle non-encodable Unicode characters, preventing crashes when special characters appear in log output.

### DIFF
--- a/logging_config.py
+++ b/logging_config.py
@@ -1,3 +1,4 @@
+import io
 import logging
 import logging.config
 import os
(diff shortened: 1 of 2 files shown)


logging_config.py
```diff
@@ -87,6 +88,19 @@ def setup_logging(
...
     logging.config.dictConfig(config)
 
+    # Reconfigure the console handler's stream to use 'replace' error handling
+    # so that non-encodable Unicode characters (e.g. \u2011 on Windows cp1251)
+    # don't crash the logging system.
+    for handler in logging.root.handlers:
+        if isinstance(handler, logging.StreamHandler) and not isinstance(handler, logging.FileHandler):
+            stream = handler.stream
+            if hasattr(stream, "reconfigure"):
+                stream.reconfigure(errors="replace")
```
brovatten, line 98: Good! Good to know: this will replace the stream in all handlers. handler.stream that we are configuring holds a reference of the stream object that e.g. print (and all stdout uses). So a good side-effect I think

brovatten (approved):
Thanks for fixing this!

PR #238, opened 2026-03-18
Expose run context

brovatten:
Since the wrapper now wants to generate run_id and log_path when calling the DiagramGenerator, it no longer makes sense to keep them private. We make them public instead, and expose and easy to use lifecycle manager that the wrapper can enjoy.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/238" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added run-scoped tracking and caching for faster repeated analysis sessions.
  * Implemented persistent cache storage for cluster and final analysis results, reducing recomputation time.

* **Refactor**
  * Enhanced type safety with explicit type annotations across agent interfaces.
  * Migrated cache system to SQLite-backed storage for improved reliability and query performance.

### DIFF
--- a/tests/agents/test_details_agent.py
+++ b/tests/agents/test_details_agent.py
@@ -1,7 +1,6 @@
-from typing import cast
 import unittest
 from pathlib import Path
-from unittest.mock import MagicMock, Mock, patch, PropertyMock
+from unittest.mock import MagicMock, patch
 
 from agents.details_agent import DetailsAgent
 from agents.agent_responses import (
@@ -14,10 +13,10 @@ from agents.agent_responses import (
     SourceCodeReference,
 )
 
-from static_analyzer.graph import ClusterResult
-from static_analyzer.node import Node
-from static_analyzer.constants import NodeType
 from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.constants import NodeType
+from static_analyzer.graph import CallGraph, ClusterResult
+from static_analyzer.node import Node
 
 
 class TestDetailsAgent(unittest.TestCase):
@@ -325,8 +324,8 @@ class TestDetailsAgent(unittest.TestCase):
         self.assertEqual(mock_validation_invoke.call_count, 2)
         mock_fix_ref.assert_called_once()
 
-    def test_classify_files(self):
-        # Test populate_file_methods assigns files from cluster results via CFG nodes
+    def test_populate_file_methods(self):
+        # Test deterministic file population from cluster results
         mock_llm = MagicMock()
         mock_parsing_llm = MagicMock()
         agent = DetailsAgent(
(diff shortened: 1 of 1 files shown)


diagram_analysis/run_context.py
```diff
@@ -0,0 +1,35 @@
...
+    if run_id is None:
+        run_id = generate_run_id()
+
+    return RunContext(
+        run_id=run_id,
+        log_path=generate_log_path(project_name),
+    )
+
+
+def finalize_run_context(repo_dir: Path, run_id: str) -> None:
```
ivanmilevtues: Maybe these should be methods in the class. resolve_run_context is the ctor. The finalize reuses self.repo_dir and self.run_id?

caching/details_cache.py
```diff
@@ -58,7 +58,7 @@ def prune_details_caches(repo_dir: Path, only_keep_run_id: str) -> None:
     ClusterCache(repo_dir).clear(keep_run_ids=[only_keep_run_id])
 
 
-def _load_existing_run_id(repo_dir: Path) -> str | None:
+def load_existing_run_id(repo_dir: Path) -> str | None:
```
ivanmilevtues: This should be part of the run_context file I thinkr right?

diagram_analysis/run_context.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass(frozen=True, slots=True)
+class RunContext:
+    """Identifiers associated with a single analysis execution."""
+
+    run_id: str
+    log_path: str
+
+
+def resolve_run_context(
```
ivanmilevtues: Should this method be the same as the load, it is weird that it is in the core but never used here. Can you write a brief description of why we need it. I saw it in the wrapper. But shouldn't that work the same way it works here -> meaning no need for specific methods for it?

diagram_analysis/run_context.py
```diff
@@ -0,0 +1,35 @@
...
+
+    return RunContext(
+        run_id=run_id,
+        log_path=generate_log_path(project_name),
+    )
+
+
+def finalize_run_context(repo_dir: Path, run_id: str) -> None:
+    """Prune detail caches so only the current run id remains."""
+    prune_details_caches(repo_dir=repo_dir, only_keep_run_id=run_id)
```
ivanmilevtues: These can be methods of the RunContext class.

I am unsure why do we need these methods. They are never used in the core instead the core uses something else to get/use the run_id? This shouldn't be the case right, as I don't see a point of having wrapper specific methods here.

More importantly I don't see why are there different ways to make the run_ids for wrapper and core.

caching/details_cache.py
ivanmilevtues: I think this method should go into the run_context if I understand correctly. Please read my comments there.

tests/agents/test_details_agent.py
```diff
@@ -341,17 +336,27 @@ def test_classify_files(self, mock_relpath, mock_exists, mock_get_files_for_clus
             run_id="test-run-id",
         )
 
-        key_entity = SourceCodeReference(
-            qualified_name="test.TestClass",
-            reference_file=str(self.repo_dir / "test_file.py"),
-            reference_start_line=1,
-            reference_end_line=10,
+        from static_analyzer.graph import ClusterResult
```
Svilen-Stefanov: Function-level imports, move to top

diagram_analysis/run_context.py
brovatten: Thanks for noticing, I agree with this. I updated it now

tests/agents/test_details_agent.py
```diff
@@ -341,17 +336,27 @@ def test_classify_files(self, mock_relpath, mock_exists, mock_get_files_for_clus
...
 
+        mock_cfg = MagicMock()
+        mock_cfg.nodes = {"node1": node}
+        self.mock_static_analysis.get_cfg.return_value = mock_cfg
+
         sub_component = Component(
             name="SubComponent",
             description="Sub component",
-            key_entities=[key_entity],
+            key_entities=[],
```
Svilen-Stefanov, line 344: Is there a reason for this change?

brovatten: The test was rewritten because the underlying method changed from classify_files (which assigned files from both clusters and key_entities) to populate_file_methods (which assigns files from cluster CFG nodes), so key_entities are no longer relevant to the test.

PR #239, opened 2026-03-18
Lock caches

brovatten:
Below is the AI summary. I think I'd need your help here @Svilen-Stefanov to figure out if this is the best way to manage concurrency for the database. Happy to chat about this tmrw.

Previously, coordination only existed between threads within a single Python process. Separate CodeBoarding runs did not share any locking, so cross-process access to the same cache DB could still race.

With the per-DB file lock added in `caching/cache.py:50`, those cross-process cases are now covered:

- **Two runs start against the same cache DB at the same time**  
  Previously, both processes could try to create or open the SQLite schema simultaneously, and one could fail with `database is locked`.  
  Now, one process initializes the DB while the other waits.

- **Two runs write to the same cache DB at the same time**  
  Previously, both processes could issue writes concurrently and race in SQLite.  
  Now, writes are serialized, so one completes before the other begins.

- **One run clears or prunes the cache while another run writes to it**  
  Previously, `clear()` and `store()` could overlap and collide.  
  Now, `clear()` and `_upsert()` are both protected by the same file lock.

- **One run asks for the latest run ID while another run is initializing or clearing the DB**  
  Previously, `load_most_recent_run()` could race with schema initialization or write/clear operations.  
  Now, that query waits on the same lock.

- **Short-lived SQLite contention during normal overlap**  
  Previously, SQLite could fail fast under transient contention.  
  Now, the engine also uses a connection timeout, `PRAGMA busy_timeout`, and WAL mode in `caching/cache.py:75`, making brief overlap much less likely to fail immediately.

### What this does *not* cover

This change does **not** eliminate higher-level logic races. For example, one run could still prune data that another run wanted to keep. The goal here is to prevent lock collisions around SQLite access, not to make all cache lifecycle decisions fully concurrency-safe.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/239" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added caching system for analysis results to improve performance on repeated operations.
  * Introduced run tracking and logging paths for better traceability of analysis executions.

* **Improvements**
  * Enhanced type safety with parameter annotations across agents and analysis tools.
  * Expanded metadata discovery to better identify project dependencies and configuration files.


caching/cache.py
```diff
@@ -122,25 +157,26 @@ def _lookup(self, key_sig: str) -> str | None:
         return str(value_json) if value_json is not None else None
 
     def _upsert(self, key_sig: str, value_json: str, run_id: str) -> None:
```
Svilen-Stefanov: Iiuc `store()` calls `self.clear()` then `self._upsert()`, each acquiring/releasing `_db_lock` independently. Another thread could write between clear and upsert, losing that write.
We should probably add a method, e.g `_clear_and_upsert()` that holds `_db_lock` once and executes them in a single transaction?

I'd also add try/except to a central place, catching all potentials issues coming from the cache. What would be the best place for it - we would need to make sure that `_upsert` and all other methods are covered by that path while not catching other issues that are blocking.

PR #240, opened 2026-03-18
Allow console only logging

Svilen-Stefanov:
In order not to create a separate log file when only a session is started, I've introduced these changes to support console logging only (without the creation of a new file). 
Now callers will not create a separate file if `log_dir` is not provided. 

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/240" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Optional file logging capability that can be independently enabled after the initial console logging setup.

* **Improvements**
  * Logging system now defaults to console-only mode; file logging is entirely optional and configurable on-demand.
  * Enhanced handling of concurrent log files to automatically prevent naming collisions.
  * Improved centralized logging directory structure for streamlined organization.

### DIFF
--- a/logging_config.py
+++ b/logging_config.py
@@ -108,7 +108,16 @@ def _resolve_log_path(log_dir: Path, log_filename: str | None) -> Path:
     else:
         timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
         filename = f"{timestamp}.log"
-    return logs_dir / filename
+
+    path = logs_dir / filename
+    if path.exists():
+        stem = path.stem
+        suffix = path.suffix
+        counter = 2
+        while path.exists():
+            path = logs_dir / f"{stem}_{counter}{suffix}"
+            counter += 1
+    return path
 
 
 def _fix_console_encoding() -> None:
--- a/main.py
+++ b/main.py
@@ -426,8 +426,9 @@ Examples:
     if is_local:
         output_dir = args.local.resolve() / ".codeboarding"
     else:
-        # Remote: will be set per-repo inside the loop below
-        output_dir = None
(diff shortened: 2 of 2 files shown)


logging_config.py
```diff
@@ -62,32 +43,79 @@ def setup_logging(
...
+        logs_dir = log_dir / "logs"
+    logs_dir.mkdir(parents=True, exist_ok=True)
 
-    # Reconfigure the console handler's stream to use 'replace' error handling
-    # so that non-encodable Unicode characters (e.g. \u2011 on Windows cp1251)
-    # don't crash the logging system.
+    if log_filename:
+        filename = log_filename
+    else:
+        timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
```
brovatten, line 109: This timestamp only has second-level precision. If add_file_handler() is called twice within the same second, both handlers resolve to the same filename, so we end up with two RotatingFileHandlers writing to one file instead of “a new file each call”. 

Should we assume this doesn't happen for now?

logging_config.py
```diff
@@ -62,32 +43,79 @@ def setup_logging(
...
         "root": {
             "level": default_level,
             "handlers": handlers,
         },
         "loggers": {
-            # quiet down verbose libraries
             "git": {"level": "WARNING"},
             "urllib3": {"level": "WARNING"},
         },
     }
```
brovatten, line 56: It looks like the remote run now doesn't have a logging path, which I think it should have.

logging_config.py
Svilen-Stefanov: Good point! I'll add a suffix if a file with this name already exists (I think milliseconds would be an overkill).

PR #242, opened 2026-03-18
Overwrite old codeboardingignore, compact LLM content and add adaptive LLM timeouts

Svilen-Stefanov:
This PR fixes a couple of different things:
- now that we rely more on the .codeboardingignore to ignore significant parts of the code, we need to overwrite it for existing users
- added adaptive LLM timeout based on the LLM prompt size with the hope to hit fewer timeouts (I saw a couple of timeouts that were mostly for large prompts - I could remove this though if we think it's a bit hacky)
- compacting of the LLM content for the abstract agent - listing all methods results in a huge prompt, Claude suggested adding just the counts (given that this is only for the abstract agent, I thought this might be okay). @ivanmilevtues what do you think?
- added autofixing qualified names to reduce the iterations on the validation step - @ivanmilevtues didn't we have this at some point, seems it got lost at some point?

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/242" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Release Notes

* **Bug Fixes**
  * Added automatic correction of invalid qualified names with logging feedback.

* **Improvements**
  * Dynamic timeout calculation based on prompt length and retry attempts for improved reliability.
  * Compact cluster summaries to optimize processing efficiency and reduce overhead.
  * Enhanced ignore file management with automatic updates for outdated configurations.

### DIFF
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -72,7 +72,10 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         programming_langs = self.static_analysis.get_languages()
 
         # Build cluster string using the pre-computed cluster results
-        cluster_str = self._build_cluster_string(programming_langs, cluster_results)
+        # Use compact mode: summarize methods as counts instead of listing each one.
+        # The grouping step only needs to understand what each cluster is about (files + classes),
+        # not individual method names. This significantly reduces prompt size.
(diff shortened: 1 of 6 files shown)


brovatten:
We may revisit similar ideas in the future, but it will likely look different, and probably not in the next month or two. Hence I'm closing this

PR #246, opened 2026-03-19
Add support for cross-component relationships

Svilen-Stefanov:
The current PR introduces cross-component relationships - this means, when we expand a component, we want to see how subcomponents relate to other existing components, regardless of their depth. 
Example: 
If we have an overview of 3 components (1, 2, 3) and we expand component 1 to 4 subcomponents (1.1, 1.2, 1.3, 1.4), the preexisting relationships are updated to use the leaf node relationships. This means, we remove connections from 1 to other components if there's a connection between any subcomponent and that component - e.g remove 1 ->2 (prev) if 1.2->2 or 1.3 -> 2. If we then expand 2, the subcomponents do the same - e.g 1.2 -> 2.3 etc. 

Refer to the `test_global_relations.py` and the `analysis.json`, which shows how this looks like. In particular, I tested:
```
Tested level combinations (src -> dst):
    lvl3 -> lvl3  cross-root              1.1.1 -> 2.1.2, 1.1.1 -> 2.1.1, 1.1.2 -> 2.2.1, 1.2.1 -> 2.1.1
    lvl3 -> lvl3  same root, diff parent  1.2.1 -> 1.1.1
    lvl3 -> lvl3  sibling                 2.2.2 -> 2.2.1, 2.1.1 -> 2.1.2
    lvl3 -> lvl2                          1.1.1 -> 4.1, 2.2.2 -> 4.2
    lvl3 -> lvl1                          1.1.1 -> 3, 2.1.1 -> 3, 2.2.2 -> 3
    lvl2 -> lvl3                          4.1 -> 2.1.1, 4.2 -> 2.2.2
    lvl2 -> lvl2  sibling                 4.1 -> 4.2
    lvl2 -> lvl1                          4.1 -> 3
    lvl1 -> lvl3                          3 -> 2.1.1
    lvl1 -> lvl2                          3 -> 4.2
```

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/246" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Incremental and partial component expansion now automatically rebuild global cross-boundary relations.
  * Relations are projected per visible hierarchy level to match what’s rendered.

* **Bug Fixes**
  * Improved cross-hierarchy relation labeling and edge inheritance; defaults are applied when labels are missing.
  * Rendering no longer produces phantom nodes or redundant/duplicated edges after roll-up.

* **Tests**
  * Added coverage for global relation construction, label inheritance, deterministic output, incremental refinement, and rendering projection.
  * Updated partial-update tests to validate the unified analysis refresh/persistence behavior.


codeboarding_workflows/analysis.py
```diff
@@ -152,13 +152,24 @@ def run_partial(
         logger.error(f"Component with ID '{component_id}' not found in analysis")
         return
 
-    _, sub_analysis, __ = generator.process_component(component_to_analyze)
-
-    if sub_analysis:
-        save_sub_analysis(sub_analysis, output_dir, component_id)
-        logger.info(f"Updated component '{component_id}' in analysis.json")
-    else:
+    _comp_id, sub_analysis, _new_components = generator.process_component(component_to_analyze)
```
ivanmilevtues: _new_components is unused.

codeboarding_workflows/analysis.py
```diff
@@ -152,13 +152,24 @@ def run_partial(
...
+    if sub_analysis is None:
         logger.error(f"Failed to generate sub-analysis for component '{component_id}'")
+        return
+
+    # Mutate the in-memory tree and rebuild global cross-boundary relations.
+    # Why: we work from memory (no save -> reload) because per-sub-analysis
+    # ``components_relations`` is not serialized; a round-trip would drop the
+    # freshly produced LLM labels needed by ``rebuild_global_relations``.
+    sub_analyses[component_id] = sub_analysis
```
ivanmilevtues, line 163: I am a bit confused about this line, generator.process_component(component_to_analyze)
produceses a sub_analysis for component with _comp_id, what is _comp_id?
component_id is the id for the component for which we want to expand the anlaysis.

So we are adding (unsure what are we mutating) the new sub_analysis to the map of sub_analyses.

I think the comment makes it hard to understand, what we do is we have a map which has sub_analyses and we are adding the new thing to it. 

Mutate the in-memory tree and rebuild global cross-boundary relations seems too much and I don't get it.

I think the comment is: add the new analysis (unsure if we need that as comment)
Rebuild relations which is the method below, so we also don't need that comment.

diagram_analysis/analysis_json.py
```diff
@@ -47,10 +47,8 @@ class ComponentJson(Component):
     components: list["ComponentJson"] | None = Field(
         description="Sub-components if expanded, None otherwise.", default=None
     )
-    components_relations: list[RelationJson] | None = Field(
-        description="Relations among sub-components if expanded, None otherwise.",
-        default=None,
-    )
+    # No per-level ``components_relations`` field: all relations live once on the
+    # root analysis as a global leaf-only set, projected to each level at render time.
```
ivanmilevtues: I think this will, might break the markdown renderers. Can you double check if it is the case?

I don't see changes in the renderers and they do use the relations.

Drop the comment I think

codeboarding_workflows/rendering.py
```diff
@@ -21,6 +21,60 @@
...
+) -> list[Relation]:
+    """Roll up a global leaf-only relation set onto the components visible at a level.
+
+    Each leaf relation's endpoints are projected to the ancestor in
+    ``level_component_ids``; edges that collapse to a self-loop or whose
+    endpoint isn't a descendant of any level component are dropped.
+    Duplicates after roll-up are merged by summing ``edge_count`` and keeping
+    the first label seen.
+
+    Why: cross-boundary relations are stored once at the root as the deepest
```
ivanmilevtues: We can drop the why ;d

diagram_analysis/diagram_generator.py
```diff
@@ -489,6 +490,9 @@ def generate_analysis(self) -> Path:
             # Process components using a frontier queue: submit children as soon as parent finishes.
             expanded_components, sub_analyses = self._generate_subcomponents(analysis, root_components)
 
+            if sub_analyses:
+                self.rebuild_global_relations(analysis, sub_analyses)
```
ivanmilevtues: This seems like patchwork, I think the rebuild_gloabl_relations should be baked in either the save, or somehow after generation.

We do it for partial and we also do it in the diagram_generator general analysis, but we have missed the incremental i.e.

That's why I think it has to be done in one of the existing steps for the agents or something like that. It could also be pre-save of the analysis, and it should be fine to be ran with empty list of sub_analysis (shouldn't do anytning in that case -> drop the if as it is creating branching)

static_analyzer/cluster_relations.py
```diff
@@ -83,6 +83,165 @@ def build_component_relations(
...
+                edge_count=sr.edge_count,
+                is_static=True,
+            )
+        )
+
+    # Also include LLM-only relations (no static backing) — these are architectural
+    # relations the LLM identified that may not have direct CFG edges.
+    #
+    # Only keep those whose BOTH endpoints resolve to a live component id. This
+    # drops relations to components deleted in a deep sub-scope (the incremental
```
ivanmilevtues: This should never happen in theory we are validating for it while running the agents (validators.py)

static_analyzer/cluster_relations.py
```diff
@@ -83,6 +83,165 @@ def build_component_relations(
...
+
+    # Collect all LLM-generated relations from every level for label matching
+    all_llm_relations: list[Relation] = list(root_analysis.components_relations)
+    for sub_analysis in sub_analyses.values():
+        all_llm_relations.extend(sub_analysis.components_relations)
+
+    # Build id-to-name map across all levels
+    id_to_name: dict[str, str] = {}
+
+    def collect_names(analysis: AnalysisInsights) -> None:
```
ivanmilevtues: can we make this into _collect_names or something so the build_gloabal_relations is a bit smaller.

static_analyzer/cluster_relations.py
```diff
@@ -83,6 +83,165 @@ def build_component_relations(
...
+
+    # Index LLM relations by (src_id, dst_id) for label matching.
+    # A parent-level relation like "1" -> "2" should provide labels for
+    # child relations like "1.1" -> "2.3". We match by checking if the
+    # static relation's src/dst are descendants of the LLM relation's src/dst.
+    llm_by_ids: dict[tuple[str, str], Relation] = {}
+    for rel in all_llm_relations:
+        llm_by_ids[(rel.src_id, rel.dst_id)] = rel
+
+    def find_llm_label(src_id: str, dst_id: str) -> str | None:
```
ivanmilevtues: can we get find_llm_label outside of this, maybe add it to the llm class for the relations (which puts the labels) so that we don't have the anonymous function here.

ivanmilevtues (changes requested):
Thanks for adding this, looks really cool in the UI.

Main comment is the one that the rebuild_global_relations might not be in the right place (diagram_genrator.py) and is called "manually" when we do different analysis. I feel like it should be done in a more "global" place as it has to be done after any sort of analysis.

I think we should add in the AGENTS.md to stop adding anonymous func as it extends other funcs and it become hard to read (and impossible to reuse). Wrote few comments, I think some of these fucns can be in-fact methods to relevant types.

codeboarding_workflows/rendering.py
```diff
@@ -21,6 +21,60 @@
...
+
+
+def project_relations_to_level(
+    global_relations: list[Relation],
+    level_component_ids: set[str],
+    id_to_name: dict[str, str],
+) -> list[Relation]:
+    """Roll up a global leaf-only relation set onto the components visible at a level.
+
+    Each leaf relation's endpoints are projected to the ancestor in
```
ivanmilevtues: I don't really get this comment it is hard to digest.

Can we have an example of what the roll-up is instead of explaining it, so that we can quickly get what is the output, list of relations. But I am not sure I get it:

It is a src_id to dst_id, so we will collect things like:
1.1 -> 2.3
2 -> 1.2.3
and so on.

I don't get the aggregration here.

codeboarding_workflows/rendering.py
```diff
@@ -21,6 +21,60 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _ancestor_in_level(component_id: str, level_ids: set[str]) -> str | None:
```
ivanmilevtues, line 26: reference for duplicated code.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Sooo, I tried to combine all common things accross all analysis types but this is still not a unified place, the unified function is now just called in all 3 paths. This should probably be modified in a separate refactor where we have a clear pre- and post-analysis hooks of some sort.

ivanmilevtues (approved):

PR #263, opened 2026-04-01
Validation fixes now it passes via first validation

ivanmilevtues:
Ivan's comments and observations:
LLMs cannot figure out things if we send too much information, so the first step doesn't perform well with many clusters.

To fix this I am introducing super-clusters, after we run the normal clustering based on edge connectivity we ourselves further super-cluster the result to get to less than 50. If there are disconnected parts of the graph we cluster them by file name as a fall-back mechanism (this is relatively okay as it is mostly on the highest level of abstraction).
As a final prompting technique, we list all cluster ids [1 to ~50] as a constraint for the LLM to know it has to include all of them.

Second observation from @brovatten: readfile goes crazy due to unresolved key_entities.
That is true, and happens a lot esp in big codebases -> 20 bad key_entities -> leads to >20 readfiles (at least one for each bad_entry)
This happens because we have 2 steps:
1. Use the CFG and clusters (code info) to extract cluster groups, combinations of clusters with a natural language description and name
2. Use those groups to create the components. Here the LLM has no information on the code itself so it hallucinates methods often...

Fix: I have explicitly asked for step 1, for key methods/classes to be added in order for us to have more information on step 2 for the llm so it can actually find what are the key methods/classes. Then instead of saying that a key entity is wrong I just drop everything unresolved (we use partial name resolution as well), and give feedback for empty key entities rather than making the llm search through the codebase.

Third observation: In similar fashion to the first observation sometimes not all cluster groups are added to the component in the AnalysisInsight. So I added a list of all the groups which have to be assigned, same solution as the first observation.

Enjoy the change.

Now CodeBoarding takes 10-15mins for level 3.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/263" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Improvements**
  * Prompts now require listing the most important classes/methods by exact qualified name in each component/group.
  * Final analysis and cluster outputs now include explicit group and "all cluster IDs" sections that require each listed name/ID to be assigned exactly once.
  * Cluster budgeting/merging across languages prevents oversized cluster sets and avoids ID collisions.
  * Validation drops out-of-scope/missing entities silently and emits consolidated warnings.

* **Tests**
  * Added unit tests for method-level change detection.

### DIFF
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -13,9 +13,6 @@ logger = logging.getLogger(__name__)
 class LLMBaseModel(BaseModel, abc.ABC):
     """Base model for LLM-parseable response types."""
 
-    def _new_methods():
-        pass
-
     @abstractmethod
     def llm_str(self):
         raise NotImplementedError("LLM String has to be implemented.")


static_analyzer/graph.py
```diff
@@ -416,10 +416,9 @@ def __cluster_str(communities: list[set[str]], cfg_graph_x: nx.DiGraph) -> str:
         valid_communities = [c for c in communities if len(c) >= 2]
         top_communities = sorted(valid_communities, key=len, reverse=True)
 
-        # Limit display to avoid overwhelming output
-        display_communities = top_communities[: ClusteringConfig.MAX_DISPLAY_CLUSTERS]
+        display_communities = top_communities
```
Svilen-Stefanov: You don't need this here if we don't cut them, you can just use `top_communities`, right?
Iiuc, we don't need the limit here because we do it in the `cluster_helpers` via `MAX_LLM_CLUSTERS`, right?

agents/cluster_methods_mixin.py
```diff
@@ -71,6 +71,7 @@ def _build_cluster_string(
             Formatted cluster string with headers per language
         """
         cluster_lines = []
+        all_ids: set[int] = set()
```
Svilen-Stefanov: nit:
```suggestion
        all_cluster_ids: set[int] = set()
```

agents/cluster_methods_mixin.py
```diff
@@ -83,6 +84,17 @@ def _build_cluster_string(
                 cluster_lines.append(f"\n## {lang.capitalize()} - {header}\n")
                 cluster_lines.append(cluster_str)
                 cluster_lines.append("\n")
+                if cluster_result:
+                    lang_ids = cluster_ids if cluster_ids else cluster_result.get_cluster_ids()
+                    all_ids.update(lang_ids)
+
+        # Add explicit ID checklist so the LLM knows exactly which IDs to assign
+        if all_ids and not cluster_ids:
+            sorted_ids = sorted(all_ids)
```
Svilen-Stefanov:
```suggestion
            sorted_cluster_ids = sorted(all_ids)
```

agents/cluster_methods_mixin.py
```diff
@@ -83,6 +84,17 @@ def _build_cluster_string(
...
+            cluster_lines.append(
+                f"\n## All Cluster IDs ({len(sorted_ids)} total)\n"
+                f"Every one of these IDs must appear in exactly one group: {sorted_ids}\n"
```
Svilen-Stefanov: I'm not sure I understand this - at the end we pass the `sorted_ids` but we last mention groups of clusters. Does it make more sense to add the ids first like:
```
Every one of these IDs: {sorted_ids}, must appear in exactly one group.
```

agents/details_agent.py
```diff
@@ -146,6 +148,12 @@ def step_final_analysis(
             project_type=project_type,
         )
 
+        if group_names:
+            prompt += (
+                f"\n\n## All Group Names ({len(group_names)} total)\n"
+                f"Every one of these names must appear in exactly one component's source_group_names: {group_names}\n"
```
Svilen-Stefanov: Same comment from `agents/cluster_methods_mixin.py` applies here.

agents/validation.py
```diff
@@ -322,38 +322,18 @@ def validate_group_name_coverage(result: AnalysisInsights, context: ValidationCo
...
     """
     Validate key_entities on every component:
-    1. Every component must have at least one key_entity.
-    2. Key entities must reference code within the component's cluster scope.
-       If cluster_results are provided, validates strictly against the scope.
-       Otherwise, validates that qualified names exist in the static analysis.
-       Uses loose matching to auto-correct shortened paths in-place.
+    1. Auto-correct qualified names via loose matching.
+    2. Silently drop invalid key entities (out of scope or not found).
+    3. Only fail if a component ends up with zero key_entities after dropping.
```
Svilen-Stefanov, line 327: Makes sense.

agents/cluster_methods_mixin.py
ivanmilevtues: Okay.

PR #273, opened 2026-04-04
Fix raise condition when multiple processes try to setup codeboarding

Svilen-Stefanov:
---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/273" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Improvements**
  * Cross-process file locking to prevent concurrent installs and reduce corruption risk.
  * Atomic, validated downloads with better cleanup on failure to avoid partial/corrupt files.
* **New Features**
  * Unified, step-based progress reporting across native tools, Node packages, and archive installs.
  * Installer can skip up-to-date work and may run at startup to ensure required tools are present.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "codeboarding"
-version = "0.10.1"
+version = "0.10.2"
 description = "Interactive Diagrams for Code"
 readme = "PYPI.md"
 license = "MIT"
--- a/uv.lock
+++ b/uv.lock
@@ -327,7 +327,7 @@ wheels = [
 
 [[package]]
 name = "codeboarding"
-version = "0.10.1"
+version = "0.10.2"
 source = { editable = "." }
 dependencies = [
     { name = "docker" },


tool_registry.py
```diff
@@ -491,13 +542,14 @@ def install_native_tools(target_dir: Path, deps: list[ToolDependency]) -> None:
...
-    for dep in deps:
-        assert dep.github_asset_template, f"{dep.key}: github_asset_template required for native tools"
+    downloadable = [d for d in deps if d.github_asset_template]
+    for i, dep in enumerate(downloadable, 1):
         binary_path = bin_dir / f"{dep.binary_name}{exe_suffix()}"
         if binary_path.exists():
             logger.info("  %s: already installed, skipping", dep.binary_name)
             continue
         asset_name = dep.github_asset_template.format(platform_suffix=suffix)
+        print(f"[download] {dep.binary_name} ({i}/{len(downloadable)})", flush=True)
```
ivanmilevtues: let's use logger.info instead of print here

tool_registry.py
```diff
@@ -558,6 +611,7 @@ def install_archive_tool(target_dir: Path, dep: ToolDependency) -> None:
         logger.info("%s already installed", dep.key)
         return
 
+    print(f"[download] {dep.key}", flush=True)
```
ivanmilevtues: This duplicates teh thing below it.

tool_registry.py
```diff
@@ -527,6 +579,7 @@ def install_node_tools(target_dir: Path, deps: list[ToolDependency]) -> None:
         return
 
     env = npm_subprocess_env(target_dir)
+    print(f"[download] npm packages: {', '.join(all_packages)}", flush=True)
```
ivanmilevtues: this duplicates the thing below it.

tool_registry.py
```diff
@@ -461,16 +499,29 @@ def get_latest_release_tag() -> str:
...
+            temp_dest.rename(destination)
+            return True
+        temp_dest.unlink(missing_ok=True)
+        return False
+    except Exception:
+        temp_dest.unlink(missing_ok=True)
+        raise
 
 
 def install_native_tools(target_dir: Path, deps: list[ToolDependency]) -> None:
```
ivanmilevtues: i suppose we can have a progress monitor which will give the wrapper information of the progress in the for i, dep in ...:
tool.call (the same way we register to other palces like the static anlaysis.

tool_registry.py
```diff
@@ -293,17 +299,49 @@ def needs_install() -> bool:
...
+        except OSError:
+            logger.info("Another instance is downloading tools, waiting...")
+            print("Waiting for another instance to finish downloading tools...", flush=True, file=sys.stderr)
+            msvcrt.locking(lock_fd.fileno(), msvcrt.LK_LOCK, 1)
+    else:
+        try:
+            fcntl.flock(lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError:
+            logger.info("Another instance is downloading tools, waiting...")
+            print("Waiting for another instance to finish downloading tools...", flush=True, file=sys.stderr)
```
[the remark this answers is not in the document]
ivanmilevtues: yea we should use logger.

tool_registry.py
```diff
@@ -18,13 +18,19 @@
...
 import subprocess
+import sys
 import tarfile
 from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import StrEnum
 from pathlib import Path
 from typing import Any, cast
 
+if sys.platform == "win32":
```
ivanmilevtues, line 31: It seems as we might want to update uv spec?
Are fcntl or msvcrt a system stdlib.

ivanmilevtues: We need to update uv

Svilen-Stefanov: They are part of the standard Python library so we don't. However, you cannot import the lib from the other platform (e.g Windows if you're on Unix)

Issue #280, opened 2026-04-05
LSP client failure for one language crashes entire polyglot analysis


cameronsjo:
## Summary

When analyzing a polyglot repository (e.g., JS/Python/TS/C#), if any single language's LSP server fails to start, `start_clients()` tears down all successfully started clients and raises `RuntimeError`, killing the entire analysis pipeline.

## Location

```python
# static_analyzer/__init__.py:188-196
except Exception as e:
    logger.exception(f"Failed to start engine LSP client for {adapter.language}")
    for _, _, client in reversed(started):
        try:
            client.shutdown()
        except Exception:
            logger.exception("Error shutting down engine client during cleanup")
    self._clients_started = False
    raise RuntimeError(f"Failed to start engine LSP client for {adapter.language}") from e
```

## Observed Behavior

Testing C# support (#274) on a polyglot monorepo (JS/Python/TS/C#, 1615+ files). OmniSharp LSP times out during initialization, which crashes the entire pipeline — Python, TypeScript, and JavaScript analysis never runs despite their LSP servers being available and healthy.

## Question

Is the all-or-nothing behavior intentional? Two possible approaches:

**Option A (graceful degradation):** Log the failure, skip the failing language, continue with the remaining LSP clients. This is how VS Code handles language server failures — one broken extension doesn't kill the others.

**Option B (keep current behavior):** If the design intent is "complete analysis or nothing", then keep as-is but consider making the error message more actionable (e.g., suggest removing the language from analysis config).

## Environment

- CodeBoarding main + C# support branch (#277)
- OmniSharp LSP timing out on macOS (dotnet@6 + .NET 10 SDK)
- Other LSP clients (pyright, typescript-language-server) start successfully

chris-okuda:
I have a similar issue on a primarily Go codebase, where the JS LSP fails to start and so nothing runs. I am using the extension (version `0.10.0`, released `2026-04-04, 21:25:41`) in Cursor (version `3.0.12 `). 

Here's my log output for one failure loop:

```
CodeBoarding: Error running analysis: Traceback (most recent call last):
  File "static_analyzer/__init__.py", line 161, in start_clients
  File "static_analyzer/engine/lsp_client.py", line 89, in start
  File "subprocess.py", line 1026, in __init__
  File "subprocess.py", line 1955, in _execute_child
FileNotFoundError: [Errno 2] No such file or directory: 'cli.mjs'

The above exception was the direct cause of the following exception:

[...]

PR #288, opened 2026-04-10
Fix LSP launch failures due to node missing

Svilen-Stefanov:
This PR reproduces and fixes the missing node issue reported in #280. 

Adding a Node.js bootstrap path so users without Node installed can run language-server-backed analysis: `install.py` downloads a pinned Node v18 into `~/.codeboarding/servers/nodeenv/` via the `nodeenv` Python package called in-process (safe inside the PyInstaller-frozen executable), with a version sentinel and strict health check (non-empty + executable + sentinel matches pin) that recovers from partial installs, zero-byte binaries, chmod failures, and version bumps.

I modified the CI integration tests workflow to mask system Node on the existing integration-test matrix so the real bootstrap path executes end-to-end on every PR touching the install or static-analysis code.

Testing:
- I created a separate PR with the same pipeline changes that validate node is indeed not found in the environments we run these tests on without the changes in this PR: https://github.com/CodeBoarding/CodeBoarding/pull/289

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/288" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Auto-provisions a pinned Node.js runtime and ensures npm is available; adds safer download locking and PATH adjustment for Node-based LSPs.

* **Chores**
  * Reorganized tool-installation into a new tool_registry package (paths/manifest/installers), added nodeenv dependency, and adjusted packaging.

* **Tests**
  * Extensive unit and integration tests for embedded Node install, health/recovery, version probing, and CLI lock behavior.

* **CI**
  * Integration tests now mask system Node/npm and verify the embedded Node is used.

* **Documentation**
  * Updated setup docs to describe embedded Node behavior.


static_analyzer/engine/adapters/java_adapter.py
```diff
@@ -114,14 +115,11 @@ def _calculate_heap_size(project_root: Path) -> str:
...
 
-        # Cap at 50% of available physical memory
-        try:
-            total_ram_bytes = os.sysconf("SC_PAGE_SIZE") * os.sysconf("SC_PHYS_PAGES")
-            total_ram_gb = total_ram_bytes / (1024**3)
-            max_heap_gb = max(1, int(total_ram_gb * 0.5))
+        # Cap at 50% of available physical memory.
+        ram_gb = total_ram_gb()
+        if ram_gb is not None:
+            max_heap_gb = max(1, int(ram_gb * 0.5))
```
ivanmilevtues: nit. This max is weird, if 1GB is your full ram then we take that over 1/2 of the ram, I think just int(ram_gb * 0.5) is good enough.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -139,24 +139,15 @@ def _discover_symbols(self, source_files: list[Path]) -> None:
...
-            probe_timeout = _PROBE_BASE_TIMEOUT
-        probe_timeout = int(probe_timeout)
+        # Synchronization probe — blocks until the LSP server has indexed
+        # the project. Scales linearly with file count (no OS branching);
+        # the per-file ceiling is picked conservatively to cover gopls on
+        # AV-heavy Windows and APFS macOS without being loose enough to
+        # let a hung LSP waste CI minutes.
+        _PROBE_STARTUP_BASE = 60  # seconds — LSP startup independent of file count
+        _PROBE_PER_FILE = 2.0  # seconds — ceiling across OSes (Linux observed ~0.35s/file)
+        _PROBE_MAX_TIMEOUT = 1800  # seconds — hard cap
```
ivanmilevtues, line 149: Let's ping Svetli to tell us if this is enough for his use-cases.

static_analyzer/engine/utils.py
```diff
@@ -2,21 +2,67 @@
...
     try:
         parsed = urlparse(uri)
-        if parsed.scheme == "file":
-            return Path(unquote(parsed.path))
-        return None
+        if parsed.scheme != "file":
+            return None
+        path = unquote(parsed.path)
+        # Strip the leading slash on Windows-style URIs (``/C:/foo`` -> ``C:/foo``).
+        if len(path) >= 3 and path[0] == "/" and path[2] == ":":
```
ivanmilevtues: The parth[0] == "/" and the others seem scetchy i.e. : at path[2] what if my disc name is DE:/ (you can do that, unsure why, but it will be impossible to track down to this :D)

ivanmilevtues: also this seems like a winodws fix but we never check if the os is windows?

ivanmilevtues (commented):
Thanks for wokring on this, my comments are nits, only one is somewhat real (the C:/ parsing)

ivanmilevtues (approved):
Thanks for wokring on this, my comments are nits, only one is somewhat real (the C:/ parsing, proly unrealistic)

static_analyzer/engine/utils.py
Svilen-Stefanov: You are right - gated the whole strip on `_IS_WINDOWS` and dropped the cross-platform "Windows URI on Linux" test assertions (they were testing a shape that never occurs in production, since LSP URIs always come from the local machine).

PR #289, opened 2026-04-11
Verify: run integration tests on Linux/macOS/Windows without Node fixes

Svilen-Stefanov:
Branches off main with ONLY the multi-OS integration-tests workflow from fix-lsp-launch-failures-due-to-node-missing. No production code, test helpers, or fixture changes.

The ``Verify embedded Node was installed`` steps are also removed: on main there is no ``ensure_node_runtime`` path, so ``codeboarding-setup`` silently skips Node bootstrap (it just warns and returns 0). Leaving those guards in would fail the job at the verify step and never reach pytest — which is where the actual signal lives.

Intended purpose: push this branch to trigger the scheduled-style matrix (16 jobs) with Node deliberately uninstalled before ``codeboarding-setup`` runs, so the Node-based LSPs fail on every machine, documenting which integration tests actually catch the regression.

DO NOT MERGE. This branch is temporary.

## Chores
* Improved integration test workflow with enhanced pull request trigger conditions
* Optimized test matrix generation for more efficient OS/language combinations
* Added comprehensive environment setup and cleanup verification steps across platforms
* Enhanced artifact collection for better cross-platform support

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,7 +1,6 @@
 name: Integration Tests
 
 on:
-  # Manual trigger with optional language filter
   workflow_dispatch:
     inputs:
       language:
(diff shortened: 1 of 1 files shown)


Svilen-Stefanov:
Closing this, it was only needed to verify the changes to the integration tests indeed lead to failures without the changes fixing #288.

PR #295, opened 2026-04-12
Improve the installation setup and requirement errors for C# and Go

Svilen-Stefanov:
I noticed C# was not properly installing the csharp-ls and I'm fixing this here. For go, we also didn't have a clear message when go is not installed, which I added here too. 

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/295" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Package-manager based installation for language servers with per-tool install locations and orchestrated installer steps.

* **Improvements**
  * Runtime preflight checks for Go and .NET toolchains with clear guidance when missing.
  * Installer-driven language-support reporting and data-driven setup summary.
  * Human-friendly language labels and updated installation guidance for C# tooling.

* **Tests**
  * Broad test suite additions covering registry/install parity, package-manager installs, adapters, and end-to-end invariants.

### DIFF
--- a/static_analyzer/constants.py
+++ b/static_analyzer/constants.py
@@ -29,8 +29,8 @@ class Language(StrEnum):
     def display_label(self) -> str:
         """Human-readable label for UI / setup summary output.
 
-        Defaults to ``value.capitalize()``; overridden for acronyms
-        (PHP) and mixed-case names (CSharp, TypeScript, JavaScript)
+        Defaults to ``value.capitalize()``; overridden for acronyms (PHP),
+        punctuated names (C#), and mixed-case names (TypeScript, JavaScript)
         that ``capitalize()`` would mangle.
         """
         return _LANGUAGE_DISPLAY_LABELS.get(self, self.value.capitalize())
@@ -39,7 +39,7 @@ class Language(StrEnum):
 # Populated after the class body because values reference Language members.
 _LANGUAGE_DISPLAY_LABELS: dict["Language", str] = {
     Language.PHP: "PHP",
-    Language.CSHARP: "CSharp",
+    Language.CSHARP: "C#",
     Language.TYPESCRIPT: "TypeScript",
     Language.JAVASCRIPT: "JavaScript",
 }
--- a/static_analyzer/constants.py
+++ b/static_analyzer/constants.py
@@ -25,6 +25,25 @@ class Language(StrEnum):
     CSHARP = "csharp"
     CPP = "cpp"
 
+    @property
+    def display_label(self) -> str:
+        """Human-readable label for UI / setup summary output.
+
+        Defaults to ``value.capitalize()``; overridden for acronyms
+        (PHP) and mixed-case names (CSharp, TypeScript, JavaScript)
+        that ``capitalize()`` would mangle.
+        """
+        return _LANGUAGE_DISPLAY_LABELS.get(self, self.value.capitalize())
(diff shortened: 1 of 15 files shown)


cameronsjo:
Some insights from our testing of C# support on a large polyglot monorepo (1879 files: 725 C#, 559 TS, 43 Python, 11 JS):

### 1. `DOTNET_ROOT` resolution still needed

The `get_lsp_command` change checks for `dotnet` on PATH, but on macOS with Homebrew-installed .NET, `csharp-ls` also needs `DOTNET_ROOT` set or it fails at startup with "You must install .NET to run this application." Our `get_lsp_env()` handles this — just flagging that the new `get_lsp_command` RuntimeError check alone may not catch this case (dotnet is on PATH but `DOTNET_ROOT` isn't set, so the apphost can't find the runtime).

### 2. `csharp-ls` version pinning

The CI workflow being removed had a pin to `csharp-ls --version 0.20.0` with a note about 0.21.0+ shipping a malformed NuGet package. Is that still an issue with the current version (0.22.0)? We tested with 0.22.0 successfully on macOS, but it might behave differently in CI.

### 3. `display_label` for CSharp

The `_LANGUAGE_DISPLAY_LABELS` mapping uses `Language.CSHARP: "CSharp"` — should this be `"C#"` for user-facing output? "CSharp" is the internal adapter name, but the setup summary is user-facing.

### 4. Go adapter check — gopls vs go

The Go `get_lsp_command` checks for `go` on PATH, which makes sense. However, `gopls` can be installed independently of the Go toolchain (e.g., `go install golang.org/x/tools/gopls@latest` puts it in `$GOPATH/bin`). If someone has gopls but not go, this check would block them — though in practice gopls without go is indeed useless since it needs the module system.

### 5. Registry coverage tests

The `test_registry_coverage.py` is a great pattern — catches the exact class of bugs we hit (Tokei key mismatch, missing adapter mapping, `lang_lower == Language.CSHARP` comparison failure). These would have caught several of our issues earlier.

Svilen-Stefanov:
@cameronsjo thank you for helping us out here! Let me know if the current PR works for you - we're planning to merge it later today and make a PYPI release. Let us know if you see any issues and we can continue iterating on it.

static_analyzer/constants.py
```diff
@@ -25,6 +25,25 @@ class Language(StrEnum):
...
+
+
+# Populated after the class body because values reference Language members.
+_LANGUAGE_DISPLAY_LABELS: dict["Language", str] = {
+    Language.PHP: "PHP",
+    Language.CSHARP: "C#",
+    Language.TYPESCRIPT: "TypeScript",
+    Language.JAVASCRIPT: "JavaScript",
+}
+
```
ivanmilevtues: you can use the string enum directly right? as it is strenum

tests/integration/test_static_analysis_consistency.py
```diff
@@ -320,19 +327,30 @@ def _check_metric_within_tolerance(
         expected: int | float,
         tolerance: float,
         min_absolute: int | float = MIN_ABSOLUTE_TOLERANCE,
+        upper_only: bool = False,
```
ivanmilevtues, line 330: nice

ivanmilevtues (approved):
nice well done

PR #306, opened 2026-04-16
Fix scope leakage: filter subgraph by exact method set instead of files

ivanmilevtues:
_create_strict_component_subgraph used filter_by_files which included ALL methods from a file even when the parent component only owned a subset. Re-clustering then assigned extra methods to children, violating the parent's scope boundary.

Added CallGraph.filter_by_nodes() and switched the subgraph creation to use it, ensuring children can only see methods the parent actually owns.

Amp-Thread-ID: https://ampcode.com/threads/T-019d9817-8ee7-724a-aef5-3bd014a00ecf

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/306" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Release Notes

* **Refactor**
  * Enhanced component analysis precision by switching to granular method-level filtering, reducing scope leakage during subgraph extraction.

* **Tests**
  * Updated test cases to align with new filtering methodology.

### DIFF
--- a/agents/cluster_methods_mixin.py
+++ b/agents/cluster_methods_mixin.py
@@ -247,11 +247,11 @@ class ClusterMethodsMixin:
             logger.warning(f"Component {component.name} has no assigned files")
             return "No assigned files found for this component.", {}, {}
 
-        # Convert files to absolute paths for comparison
-        assigned_file_set = set()
-        for f in component_files:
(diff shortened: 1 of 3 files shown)


Svilen-Stefanov (approved):
Nice, thank you!

PR #307, opened 2026-04-16
Keep file and method changes in memory

Svilen-Stefanov:
This PR introduces 2 main changes:
1. Removes the status from the analysis.json overall.
2. Modifies the analysis.json format to reduce redundancies in the format (files only refer to the methods but do not inline all of their info) - leads to a 3x + reduction in the file size.  

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/307" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Refactor**
  * Removed embedded per-file and per-method diff-status fields; status tracking is now external and computed separately.
  * Analysis output now serializes files as method references and reconstructs full details on load.
  * Generation and incremental updates no longer mutate status fields during analysis or final output.

* **Tests**
  * Updated tests to use the new status-computation API and align with the refactored behavior.


agents/agent_responses.py
```diff
@@ -136,16 +134,16 @@ def llm_str(self):
 
 
 class MethodEntry(BaseModel):
-    """A single method/function within a file, with its location and identity."""
+    """A single method/function within a file, with its location and identity.
+
+    Pure architecture — no change status. Status lives in the wrapper session's
+    ``StatusIndex`` and is transported via ``IncrementalDelta`` notifications.
```
ivanmilevtues: We don't need this comment, this is a backwards compatibility comment, please keep it clean. This will bias the agent in the future that this is a great idea and should be considered, let it be free and unbiased.

diagram_analysis/analysis_json.py
```diff
@@ -70,23 +67,28 @@ def _build_methods_index_from_files(files_index: dict[str, FileEntry]) -> dict[s
                 start_line=method.start_line,
                 end_line=method.end_line,
                 type=method.node_type,
-                status=method.status,
             )
     return methods_index
 
 
+def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict[str, "FileEntryJson"]:
```
ivanmilevtues: The FileEntryJson can be a normal type and not this hacky thing. Add to agent.md so we can wip the agent to know

diagram_analysis/analysis_json.py
```diff
@@ -190,27 +191,34 @@ class MethodIndexEntry(BaseModel):
...
-        default="unchanged",
-        description="Diff status of this file: added, modified, deleted, renamed, or unchanged.",
-    )
     methods: list[str] = Field(
         default_factory=list,
         description="Qualified method/function names assigned to this component in this file.",
     )
 
 
+class FileEntryJson(BaseModel):
```
ivanmilevtues, line 107: move up, maybe it is time for `agents/output_dto/different_classes`

diagram_analysis/analysis_json.py
```diff
@@ -400,25 +409,68 @@ def parse_unified_analysis(
...
 
+def _reconstruct_files_index(
+    files_raw: dict,
+    methods_index: dict[str, "MethodIndexEntry"],
+) -> dict[str, FileEntry]:
+    """Rebuild in-memory ``FileEntry`` objects from persisted ``method_keys``.
+
+    Supports the new persisted schema (``{"method_keys": [...]}``). For legacy
+    files still carrying inline ``{"methods": [...]}`` we fall back to parsing
+    those entries directly, so reading existing analyses keeps working.
```
ivanmilevtues: Please no backwards compatibility, this we can skip and just invalidate the analysis.json
```
    Supports the new persisted schema (``{"method_keys": [...]}``). For legacy
    files still carrying inline ``{"methods": [...]}`` we fall back to parsing
    those entries directly, so reading existing analyses keeps working.
```

My suggestion is to discuss if we want to have some migrations in the future to have a separete project which is used by the wrapper to migrate stuff from one format to another so we don't bloat the clean, crisp and minimal core.

diagram_analysis/analysis_json.py
```diff
@@ -400,25 +409,68 @@ def parse_unified_analysis(
...
+            legacy_methods = [
+                MethodEntry(
+                    qualified_name=m["qualified_name"],
+                    start_line=m.get("start_line", 0),
+                    end_line=m.get("end_line", 0),
+                    node_type=m.get("node_type", "METHOD"),
+                )
+                for m in (entry_raw.get("methods", []) if isinstance(entry_raw, dict) else [])
+            ]
+            files_index[file_path] = FileEntry(methods=legacy_methods)
```
ivanmilevtues: 🤬 no BaCKWarDS compAT Mr. AgeNT

diagram_analysis/incremental_types.py
```diff
@@ -68,3 +68,67 @@ def to_dict(self) -> dict[str, Any]:
...
+
+@dataclass
+class StatusIndex:
+    """In-memory status cache owned by the wrapper session.
+
+    Status is never persisted to analysis.json — it lives only in this index
+    (wrapper side) and its wire-format twin ``IncrementalDelta`` (sent to the
+    extension). Both ``.file_statuses`` and ``.method_statuses`` default to
+    ``UNCHANGED`` on miss.
+    """
```
ivanmilevtues: hmm if something is for the wrapper maybe it should go there and not in the core, the core should provide only functionality for it but then the wrapper call it. I am unsure if what I say makes sense as I have to get the `full context`, `full picture`

diagram_analysis/incremental_types.py
```diff
@@ -68,3 +68,67 @@ def to_dict(self) -> dict[str, Any]:
...
+
+    file_statuses: dict[str, ChangeStatus] = field(default_factory=dict)
+    method_statuses: dict[str, ChangeStatus] = field(default_factory=dict)
```
ivanmilevtues: It feels like we need only method statuses. Why you might ask?

Because in the diagram representaiton we don't want to highlight on a file level but on a method level example:
Component Y: fileA with methods A and B
Component Z: file A with methods C and D
I modify method D -> only component Z has to be highlighted so the need for the file status is not there.

diagram_analysis/io_utils.py
```diff
@@ -83,6 +83,18 @@ def read(self) -> tuple[AnalysisInsights, dict[str, AnalysisInsights], dict] | N
                     data = json.load(f)
 
                 root_analysis, sub_analyses = parse_unified_analysis(data)
+
+                # BACKWARDS COMPAT — delete after 2026-05-17.
```
ivanmilevtues: Delete now, don't wayt for 17 of may

diagram_analysis/io_utils.py
```diff
@@ -232,6 +244,20 @@ def _write_with_lock_held(
         return self._analysis_path
 
 
+# ---------------------------------------------------------------------------
+# Legacy format detection — BACKWARDS COMPAT, delete after 2026-05-17.
+# ---------------------------------------------------------------------------
+
+
+def _is_legacy_format(data: dict) -> bool:
```
ivanmilevtues: Delete now don't wayt for 17th of May

repo_utils/method_diff.py
```diff
@@ -100,78 +98,67 @@ def _resolve_file_status(file_path: str, changes: ChangeSet) -> ChangeStatus:
...
+                method_statuses[method.qualified_name] = ChangeStatus.MODIFIED
+            elif _method_fully_inside_ranges(method, added_ranges):
+                method_statuses[method.qualified_name] = ChangeStatus.ADDED
+            elif _method_overlaps_ranges(method, added_ranges):
+                method_statuses[method.qualified_name] = ChangeStatus.MODIFIED
+            elif _method_overlaps_ranges(method, deletion_points):
+                method_statuses[method.qualified_name] = ChangeStatus.MODIFIED
+            else:
+                method_statuses[method.qualified_name] = ChangeStatus.UNCHANGED
+        return file_status, method_statuses
```
ivanmilevtues: This, will have to be reworkd with the AST thing that Alex was talking about. WE can use treesitter. So it is okay for now, but for sure we should rework we have added 1 million test cases because of edge cases, ast parse would help tremendeously!

Thanks @brovatten for trying ASTs/tree-sitter so that we can come back to 21st century

repo_utils/method_diff.py
```diff
@@ -100,78 +98,67 @@ def _resolve_file_status(file_path: str, changes: ChangeSet) -> ChangeStatus:
...
-def get_method_statuses_for_file(
+def compute_method_statuses_for_file(
     methods: list[MethodEntry],
     file_path: str,
     changes: ChangeSet,
     repo_dir: Path,
-) -> ChangeStatus:
-    """Update method statuses for one file and return the file status."""
-    file_status = _resolve_file_status(file_path, changes)
+) -> tuple[ChangeStatus, dict[str, ChangeStatus]]:
```
ivanmilevtues: Do not return a tuple, return only the dict, we never iuse the ChangeStatus (Note: add this as a health check)

ivanmilevtues (changes requested):
An amazing change making us fly, before we take off I have few comments :).
I would love to take a second look after you do those especially around if somethings are misplaced here and should be in the wrapper

diagram_analysis/incremental_types.py
Svilen-Stefanov: We do need to track file level because not everything is stored in a method - good example of this are constants in a `constants.py` file. @ivanmilevtues and I discussed this and we're gonna keep this for now

Svilen-Stefanov:
We decided to remove the backwards compatibility changes and let people re-generate existing analysis to keep the code clean.

diagram_analysis/analysis_json.py
```diff
@@ -222,6 +133,103 @@ class UnifiedAnalysisJson(BaseModel):
     components_relations: list[RelationJson] = Field(description="List of relations among the components.")
 
 
+def _build_files_index_from_analysis(analysis: AnalysisInsights) -> dict[str, FileEntry]:
+    """Build a top-level files index from analysis."""
+    return {file_path: entry.model_copy(deep=True) for file_path, entry in analysis.files.items()}
+
+
+def _method_key(file_path: str, qualified_name: str) -> str:
```
brovatten, line 141: Can we use Path everywhere? I tried to introduce this, but str always slips in from claude. Since we expect a filepath. maybe also comment if it's expected to be relative or full path

diagram_analysis/analysis_json.py
```diff
@@ -222,6 +133,103 @@ class UnifiedAnalysisJson(BaseModel):
...
+    return f"{file_path}|{qualified_name}"
+
+
+def _to_method_qualified_name(method: MethodEntry) -> str:
```
brovatten, line 145: Do we need a method for this?

agents/agent_responses.py
```diff
@@ -156,24 +150,19 @@ def __eq__(self, other: object) -> bool:
         return self.qualified_name == other.qualified_name
 
     @classmethod
-    def from_method_change(cls, method_change, *, status_override: ChangeStatus | None = None) -> "MethodEntry":
+    def from_method_change(cls, method_change) -> "MethodEntry":
```
brovatten: What does "MethodEntry" return mean here?

diagram_analysis/incremental_updater.py
```diff
@@ -19,19 +16,27 @@
...
+# Resolves a repo-relative file path to its current MethodEntry list.
 SymbolResolver = Callable[[str], list[MethodEntry]]
 
+# Returns the caller's last-known status for ``(file_path, qualified_name)``.
+# ``UNCHANGED`` is the neutral default when the caller holds no record.
+MethodStatusLookup = Callable[[str, str], ChangeStatus]
+
+
+def _always_unchanged(_file_path: str, _qualified_name: str) -> ChangeStatus:
+    return ChangeStatus.UNCHANGED
```
brovatten: huh? naming of this function, and returning it like this

diagram_analysis/incremental_updater.py
```diff
@@ -19,19 +16,27 @@
 from agents.change_status import ChangeStatus
 from diagram_analysis.incremental_types import FileDelta, IncrementalDelta, MethodChange
 from repo_utils.change_detector import ChangeSet
-from repo_utils.method_diff import get_method_statuses_for_file
+from repo_utils.method_diff import compute_method_statuses_for_file
```
brovatten, line 19: did this have to be renamed before the mama cita iterative merge comin up?

brovatten: i see now why, nice 👍

brovatten (approved):
Great job! Looks clean

The only time i would have liked to see the analysis.json and there is none haha 😢

diagram_analysis/analysis_json.py
Svilen-Stefanov: These are methods from before, just moved down because of: https://github.com/CodeBoarding/CodeBoarding/pull/307#discussion_r3104019167

This requires a larger refactor so it's probably better to leave it for later.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Same as above, these are methods from before, just moved down because of: #307 (comment)

agents/agent_responses.py
Svilen-Stefanov: The reason is that `MethodEntry` isn't bound yet at class-body evaluation time. We can add `from __future__ import annotations` to remove them.

PR #309, opened 2026-04-18
Add context-window resolver for LLM models

brovatten:
Adds a small helper, get_context_window(provider, model), that tells you the input/output token limits for any LLM we use.

Right now we don't have anything like this — the context size is either hardcoded or we just hope for the best. This gives us one function to ask.

How it works: it checks a few public catalogs in order and returns the first real answer it finds.

- models.dev is the main source. It's the only one that separates the total context window from the usable input cap, which matters for models like GPT-5 (400K context, but you can only send 272K of input).
- LiteLLM is the backup. It covers more AWS Bedrock IDs than models.dev, especially the regional ones (us./eu./apac. prefixes).
- OpenRouter catches everything else — the long tail of models you can only really find through the aggregator (x-ai, perplexity, cohere, nvidia, some mistralai variants).

If all three miss, we fall back to 128K / 8K and log a warning.

There's also an Ollama path that calls the local /api/show endpoint, and an env var override (CB_CTX_<PROVIDER>_<MODEL>) for when a catalog is wrong or the model is private.

Results are cached for 24 hours in ~/.codeboarding/cache so we only hit the network once a day.

Tests:
- 18 unit tests with the catalogs mocked out — no network, run by default.
- 64 live integration tests hitting the real catalogs across openai, anthropic, aws, google, openrouter, kimi, glm, cerebras, deepseek. These are under tests/integration/ and skipped by the default pytest run; opt in with `pytest tests/integration/...`.

Not wired into agents/llm_config.py yet — that's a follow-up. This PR is just the resolver and its tests.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/309" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a model context-window lookup API that returns resolved input/output token limits across providers, honoring environment overrides, provider-specific name normalization, cached remote catalog lookups with TTL, and sensible fallbacks.
  * Added optional user-configurable context-window setting in the app config template.

* **Tests**
  * Added unit and integration tests covering resolver precedence, parsing rules, caching behavior, and live-catalog smoke checks.

### DIFF
--- a/agents/model_capabilities.py
+++ b/agents/model_capabilities.py
@@ -160,7 +160,7 @@ def _load(source: str) -> dict:
     try:
         # Why: models.dev rejects urllib's default UA with 403.
         req = urllib.request.Request(ModelCapabilities.SOURCES[source], headers={"User-Agent": "codeboarding/1.0"})
-        with urllib.request.urlopen(req, timeout=5) as r:
+        with urllib.request.urlopen(req, timeout=2) as r:
             raw = json.load(r)
         data = _normalize(source, raw)
         path.parent.mkdir(parents=True, exist_ok=True)
--- a/agents/model_capabilities.py
+++ b/agents/model_capabilities.py
@@ -20,6 +20,10 @@ _OVERRIDES: dict[tuple[str, str], tuple[int, int]] = {}
 
 _BEDROCK_REGION = re.compile(r"^(us|eu|apac|global|au|ca|us-gov)\.")
 
+# Why: cached by hand (not via @lru_cache) so a transient Ollama outage -- user starts the
+# app before `ollama serve` is up -- doesn't memoize None for the rest of the process.
+_OLLAMA_CACHE: dict[tuple[str, str], tuple[int, int]] = {}
+
 
 @dataclass(frozen=True)
 class ContextWindow:
@@ -50,8 +54,14 @@ def _resolve_env(provider: str, model_name: str) -> tuple[int, int] | None:
     if not val:
         return None
     parts = val.split(",")
-    inp = int(parts[0])
-    out = int(parts[1]) if len(parts) > 1 else ModelCapabilities.FALLBACK_OUTPUT
+    try:
+        inp = int(parts[0])
+        out = int(parts[1]) if len(parts) > 1 else ModelCapabilities.FALLBACK_OUTPUT
+    except ValueError as e:
+        # Why: env layer is an escape hatch; a typo like `500k` must fall through
+        # to the next resolver, not crash the whole call out of get_context_window.
+        logger.warning(f"Ignoring malformed {key}={val!r} ({e})")
+        return None
     return inp, out
 
 
@@ -68,8 +78,11 @@ def _resolve_ollama(provider: str, model_name: str) -> tuple[int, int] | None:
     return _ollama_show(model_name, base.rstrip("/"))
 
 
-@lru_cache(maxsize=64)
 def _ollama_show(model_name: str, base_url: str) -> tuple[int, int] | None:
+    key = (model_name, base_url)
+    cached = _OLLAMA_CACHE.get(key)
+    if cached is not None:
+        return cached
     try:
         req = urllib.request.Request(
             f"{base_url}/api/show",
@@ -93,7 +106,9 @@ def _ollama_show(model_name: str, base_url: str) -> tuple[int, int] | None:
     # Tell the user so they can bump PARAMETER num_ctx in their Modelfile.
     if num_ctx and arch_max and num_ctx < arch_max:
         logger.info(f"{model_name}: num_ctx={num_ctx} < arch_max={arch_max}; Ollama will truncate")
-    return ctx, ModelCapabilities.FALLBACK_OUTPUT
+    result = (ctx, ModelCapabilities.FALLBACK_OUTPUT)
+    _OLLAMA_CACHE[key] = result
+    return result
 
 
 def _parse_num_ctx(params: str) -> int | None:
@@ -158,7 +173,9 @@ def _load(source: str) -> dict:
     # parses of the same multi-MB JSON. On-disk TTL still applies on first hit per process.
     path = get_cache_dir(Path.cwd()) / f"{source}.json"
     if path.exists() and time.time() - path.stat().st_mtime < ModelCapabilities.CACHE_TTL_SECONDS:
-        return json.loads(path.read_text())
+        cached = _read_cache(path)
+        if cached is not None:
+            return cached
     try:
         # Why: models.dev rejects urllib's default UA with 403.
         req = urllib.request.Request(ModelCapabilities.SOURCES[source], headers={"User-Agent": "codeboarding/1.0"})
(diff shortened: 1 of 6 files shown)


agents/model_capabilities.py
```diff
@@ -0,0 +1,197 @@
...
+_CACHE_DIR = Path.home() / ".codeboarding" / "cache"
+_TTL_SECONDS = 24 * 3600
+_FALLBACK_INPUT = 128_000
+_FALLBACK_OUTPUT = 8_192
+
+_SOURCES = {
+    "litellm": "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json",
+    "modelsdev": "https://models.dev/api.json",
+    "openrouter": "https://openrouter.ai/api/v1/models",
+}
```
ivanmilevtues: Put in a constants file

agents/model_capabilities.py
```diff
@@ -0,0 +1,197 @@
...
+    # Why: num_ctx < arch_max means Ollama silently truncates beyond num_ctx tokens.
+    # Tell the user so they can bump PARAMETER num_ctx in their Modelfile.
+    if num_ctx and arch_max and num_ctx < arch_max:
+        logger.info(f"{model_name}: num_ctx={num_ctx} < arch_max={arch_max}; Ollama will truncate")
+    return ctx, _FALLBACK_OUTPUT
+
+
+def _parse_num_ctx(params: str) -> int | None:
+    m = re.search(r"^num_ctx\s+(\d+)", params, re.MULTILINE)
+    return int(m.group(1)) if m else None
```
ivanmilevtues, line 126: mmkay i guess

agents/model_capabilities.py
```diff
@@ -0,0 +1,197 @@
...
+def _openrouter_id(provider: str, model_name: str) -> str:
+    if provider == "aws" and "anthropic." in model_name:
+        stripped = _BEDROCK_REGION.sub("", model_name).removeprefix("anthropic.").removesuffix("-v1:0")
+        return f"anthropic/{stripped}"
+    if "/" in model_name:
+        return model_name
+    return f"{_OPENROUTER_PREFIX.get(provider, provider)}/{model_name}"
+
+
+def _load(source: str) -> dict:
```
ivanmilevtues, line 180: source is model_name or probably model_provider from what I see. Better name it like that.

agents/model_capabilities.py
```diff
@@ -0,0 +1,197 @@
...
+    if path.exists() and time.time() - path.stat().st_mtime < _TTL_SECONDS:
+        return json.loads(path.read_text())
+    try:
+        # Why: models.dev rejects urllib's default UA with 403.
+        req = urllib.request.Request(_SOURCES[source], headers={"User-Agent": "codeboarding/1.0"})
+        with urllib.request.urlopen(req, timeout=5) as r:
+            raw = json.load(r)
+        data = _normalize(source, raw)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(json.dumps(data))
```
ivanmilevtues, line 195: This is super dangerours, if I install it from pypi and run it in my main dir I will get a parent of my main dir. We should use the full paths if possible.

agents/model_capabilities.py
```diff
@@ -0,0 +1,197 @@
...
+import re
+import time
+import urllib.request
+from dataclasses import dataclass
+from functools import lru_cache
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_CACHE_DIR = Path.home() / ".codeboarding" / "cache"
```
ivanmilevtues: It feels like this should be part of the config.toml no? And not per project rather per installation

ivanmilevtues: I think we shoud alos have the cache probably as part of the `config.toml` with some default values and then override them if we have a newer data if not keep it to the defaults.

ivanmilevtues (approved):
I have few comsetic details and my main comment is taht this should be part of config.toml instead of adding a new config file per workspace. I think having it globally is fine. THis should be linekd probably not with the agents but rather in the `install.py` and then we directly read it in llm_configs.

agents/model_capabilities.py
brovatten: Hahha let me add a short comment here

agents/model_capabilities.py
brovatten: it's not, it's what HTTPS app we use to where we look up the context window size

agents/model_capabilities.py
brovatten: let's use get_cache_dir from utils, good point

brovatten: the latter we spoke about irl

agents/model_capabilities.py
brovatten: this is mostly relevant here for now, premature to add it to constants

ivanmilevtues (approved):
gut

agents/model_capabilities.py
```diff
@@ -0,0 +1,208 @@
...
+        return None
+    return inp, out
+
+
+def _resolve_override(provider: str, model_name: str) -> tuple[int, int] | None:
+    return _OVERRIDES.get((provider, model_name))
+
+
+def _resolve_ollama(provider: str, model_name: str) -> tuple[int, int] | None:
+    if provider != "ollama":
```
Svilen-Stefanov, line 82: Enum?

agents/model_capabilities.py
```diff
@@ -0,0 +1,208 @@
...
+_BEDROCK_REGION = re.compile(r"^(us|eu|apac|global|au|ca|us-gov)\.")
+
+# Why: cached by hand (not via @lru_cache) so a transient Ollama outage -- user starts the
+# app before `ollama serve` is up -- doesn't memoize None for the rest of the process.
+_OLLAMA_CACHE: dict[tuple[str, str], tuple[int, int]] = {}
+
+
+@dataclass(frozen=True)
+class ContextWindow:
+    input: int
```
Svilen-Stefanov:
```suggestion
    input_token: int
```

agents/model_capabilities.py
```diff
@@ -0,0 +1,208 @@
...
+
+from agents.constants import ModelCapabilities
+from utils import get_cache_dir
+
+logger = logging.getLogger(__name__)
+
+# Escape hatch for catalog bugs or private IDs no catalog covers. Empty by design —
+# models.dev resolves the previously-suspected overrides (gpt-5, kimi-k2.5, glm-4.6,
+# gpt-oss-120b) correctly when we prefer limit.input over limit.context.
+_OVERRIDES: dict[tuple[str, str], tuple[int, int]] = {}
```
Svilen-Stefanov: The name is a bit generic

Svilen-Stefanov (approved):
LGTM. Left a couple of nits.

agents/model_capabilities.py
brovatten: we said we wouldnt bcs of llm_config

agents/model_capabilities.py
brovatten: done

agents/model_capabilities.py
brovatten: this was outdated, fixed this

PR #311, opened 2026-04-18
Fix static type errors in graph

Tsvetan888:
- graph.py: narrow node_type to NodeType before ENTITY_LABELS.get so the dict[NodeType, str] key type holds when node data lacks a "type" attr.
- test_tool_registry.py: annotate fake_download closures to match download_asset's real signature so gzip.open resolves to the binary-mode overload (fixes Pyright bytes-vs-str error).
- Add test_cluster_str_handles_missing_node_type covering the graph fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cluster rendering is more robust: members missing from the graph are now omitted, and nodes lacking type metadata display a safe "[Function]" label to avoid incorrect or stale labels.

* **Tests**
  * Added tests covering cluster rendering with incomplete or missing node metadata.
  * Test helpers updated with explicit type annotations for clarity and consistency.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/311" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -524,8 +524,7 @@ class CallGraph:
             files_in_cluster: set[str] = set()
 
             for node_name in sorted(community):
-                # Communities may reference nodes no longer in the graph when an
-                # externally supplied ClusterResult is passed in.
+                # Externally supplied ClusterResults can reference stale nodes.
                 if node_name not in cfg_graph_x.nodes:
                     continue
                 node_data = cfg_graph_x.nodes[node_name]
@@ -533,8 +532,6 @@ class CallGraph:
                 node_type = node_data.get("type")
                 files_in_cluster.add(file_path)
 
-                # ``isinstance`` narrows ``Any`` from networkx attribute stubs; it is
-                # not defending against missing data (the guard above handles that).
                 type_label = node_type.label() if isinstance(node_type, NodeType) else "Function"
                 parts = node_name.split(".")
 
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -505,7 +505,7 @@ class CallGraph:
                 node_type = node_data.get("type")
                 files_in_cluster.add(file_path)
 
-                type_label = ENTITY_LABELS.get(node_type, "Function")
+                type_label = ENTITY_LABELS.get(node_type, "Function") if isinstance(node_type, NodeType) else "Function"
                 parts = node_name.split(".")
 
                 if node_type == NodeType.CLASS:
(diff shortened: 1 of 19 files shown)


static_analyzer/graph.py
```diff
@@ -505,7 +505,7 @@ def __cluster_str(communities: list[set[str]], cfg_graph_x: nx.DiGraph) -> str:
                 node_type = node_data.get("type")
                 files_in_cluster.add(file_path)
 
-                type_label = ENTITY_LABELS.get(node_type, "Function")
+                type_label = ENTITY_LABELS.get(node_type, "Function") if isinstance(node_type, NodeType) else "Function"
```
Svilen-Stefanov: I wonder if this is treating the symptoms of the problem instead of resolving the root cause - have you seen an issue with this overall? Are we adding things to `ENTITY_LABELS` that are not hash-able that we shouldn't add?

tests/test_tool_registry.py
```diff
@@ -1317,7 +1317,7 @@ def test_gz_asset_is_decompressed_and_chmod_set(self):
         binary_payload = b"\x7fELF" + b"fake-elf-bytes" * 200
 
         # The fake "download" writes a gzipped payload to the destination path.
-        def fake_download(url, destination, expected_sha256=None):
+        def fake_download(url: str, destination: Path, expected_sha256: str | None = None) -> bool:
```
Svilen-Stefanov, line 1320: Thanks for improving the typing here!

Svilen-Stefanov (commented):
I have one question about this: https://github.com/CodeBoarding/CodeBoarding/pull/311/changes#r3105825503

The rest looks good to me.

static_analyzer/graph.py
Tsvetan888: Good point - you’re right that the isinstance check was working around a symptom.

ENTITY_LABELS itself is fine - it’s a static dict[NodeType, str], and all keys are NodeType (hashable IntEnum). Nothing unhashable is getting in.

The real root cause was in __cluster_str: it was doing cfg_graph_x.nodes.get(node_name, {}), which silently returns {} when a community references a node that isn’t in the current graph (can happen when an externally supplied ClusterResult is used). Then .get("type") becomes None, and the downstream dict lookup blows up typing.

I’ve pushed a follow-up that mirrors the existing guard in _build_result:

1. if node_name not in cfg_graph_x.nodes: continue up front, so stale community entries are skipped instead of rendered with placeholder data.
2. Switched the label lookup to NodeType.label() (the canonical accessor on the enum) — the remaining isinstance now only narrows networkx's Any attribute stubs, which is a genuine concern rather than a data-shape workaround.
3. Added two targeted tests: one for a node present-but-missing "type", and one for a stale node reference not in the graph.

Svilen-Stefanov (approved):
LGTM. Thanks!

ivanmilevtues:
Thanks @Tsvetan888

PR #312, opened 2026-04-18
Add CFG skip planner to fit cluster prompts in context window

brovatten:
## Summary

- When a repo's call graph is too large to fit in the LLM's input window, the cluster-string prompt is pruned by dropping low-degree (peel-safe) methods — preserving cluster identity and graph structure.
- Single-stage design: keep at least `min_keep_per_cluster=5` members per cluster, respect a global `max_peel_frac=0.5` cap, binary-search the smallest peel-order prefix that fits the budget. Warn and return best-effort if nothing fits.
- Filtering is **serialization-only**: the graph, cluster results, `file_methods`, and inter-component relations are all computed from the unmodified data, so every node still ends up in the final output.
- Budget derived from the `get_context_window` resolver (PR #309): subtracts an 8K-token output headroom and the measured prompt overhead (system message + rendered template), then keeps a 10% safety margin.

## Known limitation — heuristic, not formal safeguard

Peel order is a valid prefix-safe removal sequence, but the final skip set is a *subsequence* of peel order (per-cluster caps and cluster-membership filter can create gaps). A subsequence of a peel-safe order is not itself peel-safe. In pathological cases — e.g. a bridge node `A` with `B — A — C`, where `B` and `C` are in a size-2 cluster (cap = 0) and `A` is in a larger cluster (cap > 0) — the planner can drop `A` while keeping `B` and `C`. The rendered view is still a correct induced subgraph (no false edges), but the LLM loses the signal that `B` and `C` were bridged through `A`.

Why it rarely bites in practice:
- Bridge nodes like `A` are usually in the 2-core and not peel-safe at all.
- `max_peel_frac=0.5` prevents concentration.
- Real clusters are typically well above the size-5 floor, so cap-induced gaps are uncommon.
- DetailsAgent method-level expansion produces cap=0 everywhere → planner no-ops harmlessly.

If we ever observe bad LLM grouping tied to this, the fix is to either (a) restrict skip sets to strict prefixes of the original peel order (drop per-cluster caps), or (b) recompute peel order on the live subgraph after each greedy addition (correct but more expensive).

## Test plan

- [x] `pytest tests/static_analyzer/ tests/agents/ --no-cov` — 910 passed, 24 skipped
- [x] `mypy` clean on all modified files
- [x] `black --check` clean
- [ ] Smoke-test on a large repo that previously tripped the context window

## Summary by CodeRabbit

* **Improvements**
  * Smarter context budgeting for multi-language cluster summaries so analyses better fit model windows and avoid unexpected truncation.
  * Better error reporting when prompt overhead leaves insufficient room for content.

* **New Features**
  * Automatic per-language trimming of cluster details to keep outputs within budget; explicit failure when trimming can't satisfy limits.

* **Tests**
  * Expanded tests covering budgeting, trimming behavior, and failure modes.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -11,7 +11,8 @@ from langchain_google_genai import ChatGoogleGenerativeAI
 from langchain_ollama import ChatOllama
 from langchain_openai import ChatOpenAI
 
-from agents.constants import LLMDefaults
+from agents.constants import LLMDefaults, ModelCapabilities
+from agents.model_capabilities import ContextWindow, get_context_window
 from agents.prompts.prompt_factory import LLMType, initialize_global_factory
 from monitoring.callbacks import MonitoringCallback
 
@@ -311,17 +312,13 @@ def initialize_agent_llm(model_override: str | None = None) -> BaseChatModel:
     return model
 
 
-def get_current_agent_context_window():
-    """Context window for the currently configured agent LLM.
+def get_current_agent_context_window() -> ContextWindow:
+    """Context window for the currently active agent provider/model.
 
-    Resolves the first active provider (same rule as ``_initialize_llm``) and
-    the effective agent model name, then delegates to ``get_context_window``.
-    Returns a ``ContextWindow`` with fallback values when no provider is active
-    (which would only happen if this is called before any LLM is initialized).
+    Resolves the first active provider (same rule as ``_initialize_llm``) on
+    every call. ``get_context_window`` handles its own caching, so this is
+    cheap enough to call without a module-level cache.
     """
-    from agents.constants import ModelCapabilities
-    from agents.model_capabilities import ContextWindow, get_context_window
-
     for name, config in LLM_PROVIDERS.items():
         if not config.is_active():
             continue
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -259,41 +259,53 @@ def _initialize_llm(
     log_prefix: str,
     init_factory: bool = False,
 ) -> tuple[BaseChatModel, str]:
+    resolved = _resolve_active_provider(model_override, model_attr)
+    if resolved is None:
+        required_vars = []
+        for config in LLM_PROVIDERS.values():
+            required_vars.append(config.api_key_env)
+            required_vars.extend(config.alt_env_vars)
+
+        raise ValueError(
+            f"No valid LLM configuration found. Please set one of: {', '.join(sorted(set(required_vars)))}"
+        )
+
+    name, config, model_name = resolved
+
+    if init_factory:
+        detected_llm_type = LLMType.from_model_name(model_name)
+        initialize_global_factory(detected_llm_type)
+        logger.info(
+            f"Initialized prompt factory for {name} provider with model '{model_name}' "
+            f"-> {detected_llm_type.value} prompt factory"
+        )
+
+    logger.info(f"Using {name.title()} {log_prefix}LLM with model: {model_name}")
+
+    kwargs = {
+        "model": model_name,
+        "temperature": getattr(config, temperature_attr),
+    }
+    kwargs.update(config.get_resolved_extra_args())
+
+    if name not in ["aws", "ollama"]:
+        api_key = config.get_api_key()
+        kwargs["api_key"] = api_key or "no-key-required"
+
+    model = config.chat_class(**kwargs)  # type: ignore[call-arg, arg-type]
+    return model, model_name
+
+
+def _resolve_active_provider(
+    model_override: str | None,
+    model_attr: str,
+) -> tuple[str, LLMConfig, str] | None:
+    """Return the active provider, config, and resolved model name."""
     for name, config in LLM_PROVIDERS.items():
         if not config.is_active():
             continue
-
-        model_name = model_override or getattr(config, model_attr)
-
-        if init_factory:
-            detected_llm_type = LLMType.from_model_name(model_name)
-            initialize_global_factory(detected_llm_type)
-            logger.info(
-                f"Initialized prompt factory for {name} provider with model '{model_name}' "
-                f"-> {detected_llm_type.value} prompt factory"
-            )
-
-        logger.info(f"Using {name.title()} {log_prefix}LLM with model: {model_name}")
-
-        kwargs = {
-            "model": model_name,
-            "temperature": getattr(config, temperature_attr),
-        }
-        kwargs.update(config.get_resolved_extra_args())
-
-        if name not in ["aws", "ollama"]:
-            api_key = config.get_api_key()
-            kwargs["api_key"] = api_key or "no-key-required"
-
-        model = config.chat_class(**kwargs)  # type: ignore[call-arg, arg-type]
-        return model, model_name
-
-    required_vars = []
-    for config in LLM_PROVIDERS.values():
-        required_vars.append(config.api_key_env)
-        required_vars.extend(config.alt_env_vars)
-
-    raise ValueError(f"No valid LLM configuration found. Please set one of: {', '.join(sorted(set(required_vars)))}")
+        return name, config, model_override or getattr(config, model_attr)
+    return None
 
 
 def validate_api_key_provided() -> None:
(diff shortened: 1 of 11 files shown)


agents/llm_config.py
```diff
@@ -311,6 +311,25 @@ def initialize_agent_llm(model_override: str | None = None) -> BaseChatModel:
...
+def get_current_agent_context_window():
+    """Context window for the currently configured agent LLM.
+
+    Resolves the first active provider (same rule as ``_initialize_llm``) and
+    the effective agent model name, then delegates to ``get_context_window``.
+    Returns a ``ContextWindow`` with fallback values when no provider is active
+    (which would only happen if this is called before any LLM is initialized).
+    """
+    from agents.constants import ModelCapabilities
+    from agents.model_capabilities import ContextWindow, get_context_window
```
ivanmilevtues: No local imports plis :)

agents/llm_config.py
```diff
@@ -311,6 +311,25 @@ def initialize_agent_llm(model_override: str | None = None) -> BaseChatModel:
     return model
 
 
+def get_current_agent_context_window():
```
ivanmilevtues: It feels like we should get that once in the LLM_config and then just get it from the LLM_PROVIDERS.

I think it is cleaner if we get config.is_active() -> config.get_context_window()

The way it is now we'd ask multiple times, I think its fine to have it once when we pick up the LLM_Config at the begining of the analysis :)

static_analyzer/cfg_skip_planner.py
```diff
@@ -0,0 +1,172 @@
...
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Callable
+
+import networkx as nx
+
+if TYPE_CHECKING:
```
ivanmilevtues: Please noe if TYPE_CHECKING ;d, models for some reason love these, they make no sense, we are always type_checking, no need to work around mypy...

God Damn it agents!

static_analyzer/cfg_skip_planner.py
```diff
@@ -0,0 +1,172 @@
...
+    from static_analyzer.graph import CallGraph, ClusterResult
+
+logger = logging.getLogger(__name__)
+
+# Community consensus middle ground: prose ≈ 4 chars/token, code ≈ 3 chars/token.
+# 3.5 keeps us conservative without leaving too much slack.
+CHARS_PER_TOKEN = 3.5
```
ivanmilevtues: maybe in llm_config?

static_analyzer/cfg_skip_planner.py
```diff
@@ -0,0 +1,172 @@
...
+    """Return peel-safe nodes in iterative-leaf order.
+
+    Repeatedly removes every node with undirected degree ≤ 1 from the live
+    graph, sorted within each round by ascending degree so degree-0 isolates
+    go first. Nodes that remain at the end of peeling are in the 2-core and
+    are NOT returned — they are unsafe to drop.
+    """
+    live = cfg_nx.to_undirected()
+    order: list[str] = []
+    while True:
```
ivanmilevtues: Is there a way to rework this so it is not while True.

While True gives me the ick and is dangerous 👻

static_analyzer/cfg_skip_planner.py
```diff
@@ -0,0 +1,172 @@
...
+) -> list[str]:
+    """Greedy pass over peel_order, keeping each node only if its cluster has room."""
+    skipped_per_cluster: dict[int, int] = {cid: 0 for cid in max_skip_per_cluster}
+    allowed: list[str] = []
+    for name in peel_order:
+        if len(allowed) >= global_cap:
+            break
+        cid = node_to_cluster.get(name)
+        if cid is None:
+            # Node isn't in any cluster — safe to skip (only affects unclustered edges)
```
ivanmilevtues, line 66: Hmm I think this is fine, but just to make sure that the codebase analysis is not incomplete. 
My concern: Once removed nodes which are not part of a cluster they will never be part of the analysis. (this I think will be fine but double check please).

Current behavior helps and fixes this because we retroactively try to add the nodes somewhere i.e. in a component, can you validate that this still works if they are dropped here.
```
  No, unclustered nodes are not dropped. They go through a multi-step orphan assignment process to ensure 100% node coverage.                                                                                                                       
                                                                                                                                                                                                                                                    
  Here's what happens:                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  1. Clustering (`_build_result` in `graph.py`) — When the call graph is clustered, communities smaller than min_cluster_size are filtered out. Nodes in those tiny communities (or nodes not assigned to any community at all) are simply not      
  present in the ClusterResult.clusters dict. They're "unclustered" at this point.                                                                                                                                                                  
                                                                                                                                                                                                                                                    
  2. The LLM sees unclustered edges — The __non_cluster_str method builds a summary of edges involving unclustered nodes and includes it in the prompt sent to the LLM, so the model is at least aware they exist.                                  
                                                                                                                                                                                                                                                    
  3. `populate_file_methods` guarantees full coverage — This is the key step. It collects all CFG nodes (from cfg.nodes, not from clusters), then tries to assign each one to a component:                                                          
                                                                                                                                                                                                                                                    
  - Clustered nodes: assigned directly via cluster_id → component mapping.                                                                                                                                                                          
  - Orphan nodes (not in any cluster) go through a 3-tier fallback:                                                                                                                                                                                 
    1. File co-location: if the node's file already has other nodes assigned to a component, the orphan joins that component.                                                                                                                       
    2. Graph distance: find the nearest cluster in the call graph (using BFS on an undirected version) and assign to that cluster's component.                                                                                                      
    3. Last resort fallback: assigned to the first component (analysis.components[0]).                                                                                                                                                              
                                                                                                                                                                                                                                                    
  The method even logs coverage stats at the end (_log_node_coverage) to confirm all nodes were placed.                                                                                                                                             
                                                                                                                                                                                                                                                    
  TL;DR: Unclustered nodes are never lost. The clustering step may not assign them to a cluster, but populate_file_methods picks them all up afterward using file proximity, graph distance, or a fallback component. Every node in the call graph  
  ends up in some component.                                                                                                                                                                                                                        
```

This is still the case right?: `The clustering step may not assign them to a cluster, but populate_file_methods picks them all up afterward using file proximity, graph distance, or a fallback component`

static_analyzer/cfg_skip_planner.py
```diff
@@ -0,0 +1,172 @@
...
+        for cid, members in cluster_result.clusters.items()
+    }
+    total_nodes = len(node_to_cluster)
+    global_cap = int(total_nodes * max_peel_frac)
+    allowed = _build_allowed_skip_list(peel_order, node_to_cluster, max_skip_per_cluster, global_cap)
+
+    def render(skip: set[str]) -> int:
+        return len(cfg.to_cluster_string(cluster_result=cluster_result, skip_nodes=skip))
+
+    skip, fits = _binary_search_smallest_fit(allowed, render, char_budget)
```
ivanmilevtues: I would have the logging directly in the binary_search_smallst_fit, returning fits just for logging is not too useful log directly there and return the smallest_fit (skip) only. Makes the function cleaner.

static_analyzer/graph.py
```diff
@@ -279,6 +279,7 @@ def to_cluster_string(
         self,
         cluster_ids: set[int] | None = None,
         cluster_result: ClusterResult | None = None,
+        skip_nodes: set[str] | None = None,
```
ivanmilevtues, line 282: Instead of optional make it by default to empty set?

(hate optionals, thats why ppl use rust :D)

static_analyzer/graph.py
```diff
@@ -300,6 +305,7 @@ def to_cluster_string(
             return cluster_result.strategy if cluster_result.strategy in ("empty", "none") else "No clusters found."
 
         cfg_graph_x = self.to_networkx()
+        skip = skip_nodes or set()
```
ivanmilevtues, line 308: do directly in the func def

static_analyzer/graph.py
```diff
@@ -489,7 +498,8 @@ def _build_result(
         )
 
     @staticmethod
-    def __cluster_str(communities: list[set[str]], cfg_graph_x: nx.DiGraph) -> str:
+    def __cluster_str(communities: list[set[str]], cfg_graph_x: nx.DiGraph, skip: set[str] | None = None) -> str:
```
ivanmilevtues: don't have set[str] as optional jus tdefault it to empty

ivanmilevtues (approved):
Looks good thanks, add the numbers of saving to the PR so ppl know how many tokens we are saving with these.

Few minor comments on style + one ask just ot make sure we are not breaking stuff.

static_analyzer/graph.py
brovatten: for the same reason as above, avoid this

PR #313, opened 2026-04-18
Factor shared dotted prefix per file in cluster string

brovatten:
## Summary

When `CallGraph.to_cluster_string()` lists nodes in a cluster, every class and standalone function is printed with its full dotted qualified name (e.g. `agents.llm_config.LLMConfig`). Within a single file block, that prefix is shared by every identifier, so on large polyglot repos this becomes the largest source of redundancy in the grouping-prompt payload fed to the LLM.

This change hoists the shared prefix once per file block:

**Before**
```
  agents/llm_config.py:
    agents.llm_config.LLMConfig [Class]
      .get_api_key [Method]
    agents.llm_config._initialize_llm [Function]
    agents.llm_config.initialize_llms [Function]
```

**After**
```
  agents/llm_config.py (identifiers below prefixed with "agents.llm_config."):
    LLMConfig [Class]
      .get_api_key [Method]
    _initialize_llm [Function]
    initialize_llms [Function]
```

Only triggers when the prefix has at least 2 dotted segments **and** the file block has at least 2 identifiers — so short names are never ambiguous and the header overhead is always net-positive. Method lines were already short-form so they are unchanged.

Measured on the CodeBoarding self-analysis (1,293 nodes / 87 clusters): this lever alone accounts for roughly the dominant share of per-file FQN bytes in the cluster string; the header format is adjacent to the TOON-style header-then-rows pattern that benchmarks show LLMs parse reliably.

## Test plan

- [x] New unit tests cover both the factoring path and the `_common_dot_prefix` helper (single-name and all-names-identical edge cases).
- [x] `pytest tests/static_analyzer/test_graph.py` → 46 passed.
- [x] `pytest --ignore=tests/integration` → only 2 failures, both pre-existing on `main` and unrelated (non-cp1252 char in an untracked script; unrelated registry coverage test).
- [x] `black --check` and `mypy static_analyzer/graph.py` clean.

## Summary by CodeRabbit

* **New Features**
  * Call graph output now factors shared dotted-name prefixes across identifiers (including standalone functions and classes), showing shortened names with a "prefixed with…" header when applicable and preserving original label suffixes.

* **Tests**
  * Added unit tests covering prefix factoring behavior and dot-notation prefix computation.

### DIFF
--- a/static_analyzer/graph.py
+++ b/static_analyzer/graph.py
@@ -489,6 +489,22 @@ class CallGraph:
         )
 
     @staticmethod
+    def _common_dot_prefix(names: list[str]) -> str:
+        """Longest dotted-segment prefix shared by all names, leaving at least one trailing segment per name."""
+        if len(names) < 2:
+            return ""
+        parts_list = [n.split(".") for n in names]
+        min_len = min(len(p) for p in parts_list)
+        common: list[str] = []
+        for i in range(min_len - 1):
+            seg = parts_list[0][i]
+            if all(p[i] == seg for p in parts_list):
+                common.append(seg)
+            else:
+                break
+        return ".".join(common)
+
+    @staticmethod
     def __cluster_str(communities: list[set[str]], cfg_graph_x: nx.DiGraph) -> str:
         valid_communities = [c for c in communities if len(c) >= 2]
         top_communities = sorted(valid_communities, key=len, reverse=True)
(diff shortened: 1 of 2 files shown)


static_analyzer/graph.py
```diff
@@ -488,6 +488,22 @@ def _build_result(
             strategy=strategy,
         )
 
+    @staticmethod
+    def _common_dot_prefix(names: list[str]) -> str:
```
ivanmilevtues: maybe rename to qualified_names, as I got confused thought this is file paths...

ivanmilevtues (approved):
Thanks looks good to me!

PR #317, opened 2026-04-22
Add semantic incremental analysis pipeline

brovatten:
## Summary
Introduces a semantic incremental mode that reuses a prior `analysis.json` when only a subset of methods changed, avoiding a full re-clustering pass. The pipeline classifies the git diff into added/modified/deleted methods, traces semantic impact through per-language call graphs with a bounded LLM loop, and applies EASE-encoded JSON Patches to the affected sub-analyses.

Structurally, this extracts CLI/workflow logic out of `main.py` into two new packages (`codeboarding_cli/`, `codeboarding_workflows/`) and adds supporting modules under `diagram_analysis/` (`incremental_tracer`, `incremental_pipeline`, `analysis_patcher`, `semantic_diff`, `ease`, `run_metadata`, `incremental_models`) and `repo_utils/` (`parsed_diff`, `change_detector`).

## Caveats / known limitations
- **No cross-language tracing.** Call graphs are built per-language and the neighbor index does not create edges across languages. Changes that ripple across a Python ↔ TypeScript boundary — backend/frontend splits, IPC layers — will not be detected; a full reanalysis is needed to pick them up.
- **Renames and copies force a full run.** `compute_delta` has no rename path, so any rename/copy in the diff triggers `REQUIRES_FULL_ANALYSIS`.
- **`--target-ref` must match the current checkout on a clean worktree.** Otherwise rejected — the tracer reads sources from the worktree and would patch from the wrong revision.
- **Failed component patches do not advance the baseline.** All-failed returns `REQUIRES_FULL_ANALYSIS` and skips the save; partial-failed saves what was patched but still requires a full run before the baseline can advance.
- **Bounded LLM trace.** Hard caps at `max_hops=3`, `max_fetched_methods=30`. Exceeding either — or any LLM retry/parse failure — ends the trace as `UNCERTAIN` and requires full analysis.
- **Syntax errors abort.** Any file in the delta with parse errors aborts incremental and falls back to full.
- **Requires a prior \`analysis.json\` + metadata.** No baseline ⇒ full analysis required.
- **LLM-dependent.** The tracer has no heuristic fallback; prompt caching is Anthropic-only.

## Test plan
- [ ] \`uv run pytest --ignore=tests/integration\`
- [ ] Manual: run \`codeboarding --local <repo> --incremental\` against a repo with an existing baseline; verify material / cosmetic / rename / target-ref paths behave as documented.

## Summary by CodeRabbit

* **New Features**
  * Explicit CLI subcommands: full and incremental; local/remote workflows and targeted partial-component runs.
  * Incremental analysis with base/target refs, semantic impact tracing, and partial component patching to speed updates.
  * Automated generation of onboarding markdown from analysis output.

* **Enhancements**
  * Structured JSON error/output for integrations and improved logging behavior.
  * More robust change detection (renames, untracked files) and safer incremental metadata handling.
  * Parallelized patching for faster incremental runs.

* **Documentation**
  * Updated CLI docs and README examples to reflect subcommand-based usage.

* **Tests**
  * Expanded test coverage for incremental flows, patching, and diff parsing.

---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/317" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


codeboarding_cli/parser.py
```diff
@@ -0,0 +1,42 @@
...
+  # Use custom binary location (e.g. VS Code extension)
+  codeboarding --local /path/to/repo --binary-location /path/to/binaries
+        """,
+    )
+    full.add_arguments(parser)
+    incremental.add_arguments(parser)
+    return parser
+
+
+def main(argv: list[str] | None = None) -> None:
```
ivanmilevtues: feels like this should be in the main, and then we import the build_parser from there.

codeboarding_cli/commands/incremental.py
```diff
@@ -0,0 +1,118 @@
...
+
+    repo_path = args.local.resolve()
+    output_dir = args.output_dir.resolve() if args.output_dir else repo_path / ".codeboarding"
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    try:
+        bootstrap_environment(output_dir, args.binary_location)
+    except ValueError as exc:
+        logger.warning("Incremental bootstrap failed: %s", exc)
+        _emit_payload(api_key_missing_payload(str(exc)), args.json)
```
ivanmilevtues: How come this api key missing payload is only in the incremental. The API KEY should be present for both, full and incremental analysis so it is better to have this as part of th bootstrap_envronment method I suppose, or something else. which preps things.

codeboarding_workflows/incremental.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Any
+
+from diagram_analysis import DiagramGenerator
+from diagram_analysis.incremental_pipeline import run_incremental_pipeline
+from monitoring.paths import generate_log_path
+from utils import generate_run_id
+
+
+def run_incremental_analysis(
+    *,
```
ivanmilevtues: Why do we have this * at the front? What does it do :D

also probably better to return a stronger type than dict[str, Any]

codeboarding_cli/commands/full.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        try:
+            process_local_repository(
+                repo_path=local_repo_path,
+                output_dir=output_dir,
+                project_name=project_name,
+                depth_level=args.depth_level,
+                component_id=args.partial_component_id,
+                monitoring_enabled=should_monitor,
+                incremental=False,
```
ivanmilevtues: hmm we shoudl probably remove the incremental flag, it is never called with true and you have introduced a new method which is doing the incremental now. We are calling that method directly (run_incremental_analysis)

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+    payload = result.to_dict()
+    payload["mode"] = "incremental"
+    payload["requiresFullAnalysis"] = True
+    return payload
+
+
+def run_incremental_pipeline(
+    generator: DiagramGenerator,
+    *,
+    base_ref: str | None = None,
```
ivanmilevtues: These can't be none right? if they are non we cannot do an analysis also why do we have a star, just saw what it is you cannot pass follow-on params as args, only as kwargs. Didn't know that still hate nullable types avoid at any cost :D

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+    (with whatever ``StaticAnalyzer`` attachment is appropriate for its
+    context) and hands it in. This function owns the orchestration —
+    loading the prior analysis, resolving refs, running the updater,
+    invoking ``generator.generate_analysis_incremental``, and writing
+    ``incremental_run_metadata.json`` on success.
+
+    Returns the wire-format dict that both Core's CLI and the wrapper's
+    JSON-RPC layer surface unchanged to their callers.
+    """
+    repo_path = Path(generator.repo_location).resolve()
```
ivanmilevtues: Yea this is a classic GIGO, we used to do this and now it copied it over, we should try to be type strcit, so probably better to do the resolve as soon as we get to the repo_location.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+            capture_output=True,
+            text=True,
+            check=True,
+        ).stdout.strip()
+    except (OSError, subprocess.CalledProcessError):
+        return False
+    return bool(head) and head == target
+
+
+def _full_required_payload(message: str) -> dict[str, Any]:
```
ivanmilevtues: Here we can return the incrementalRunResult right? I don't see why we put it back to dict so we are not strongly typed.

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
...
+            patch_lines.append(line)
+
+    finalized = _finalize_file_diff(current_file, patch_lines)
+    if finalized is not None:
+        files.append(finalized)
+
+    return ParsedGitDiff(base_ref=base_ref, target_ref=target_ref, files=files)
+
+
+def load_parsed_git_diff(
```
ivanmilevtues: I feel like this is not loading but getting as we are running the command now, load I though that we are caching it somewhere ;d.

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
...
+    repo_dir: Path,
+    base_ref: str,
+    target_ref: str = "HEAD",
+    context_lines: int = 3,
```
ivanmilevtues: parameters context_lines, fetch_missing_refs and exclude_patterns are never passed so maybe we don't need em

ivanmilevtues: Knowing that fetching_missing_refs is True and exclude_patterns is None should also simplify the method by removing few ifs.

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
...
+        "--raw",
+        f"-U{context_lines}",
+        "-M",
+        "-C",
+        "--find-renames=50%",
+        base_ref,
+    ]
+    if target_ref:
+        cmd.append(target_ref)
+    if exclude_patterns:
```
ivanmilevtues: this is always true

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
...
+    exclude_patterns: list[str],
+) -> None:
+    """Inject untracked worktree files as ADDED entries.
+
+    `git diff` does not list files unknown to the index, so a user editing
+    against the current worktree would otherwise get ``no_changes`` for a
+    freshly created file until they ``git add`` it. Only applied for
+    worktree diffs (empty target_ref).
+    """
+    cmd = ["git", "ls-files", "--others", "--exclude-standard", "-z", "--", "."]
```
ivanmilevtues: here we should append only files which are interesting i.e. files like .md/.rst (any PL that we don't support) should be ingored.

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
...
+
+    existing = {file_diff.file_path for file_diff in parsed.files}
+    for path in result.stdout.split("\0"):
+        if not path or path in existing:
+            continue
+        parsed.files.append(ParsedDiffFile(status_code="A", file_path=path))
+        existing.add(path)
+
+
+def classify_new_file_ranges(
```
ivanmilevtues: dead code

repo_utils/parsed_diff.py
```diff
@@ -0,0 +1,328 @@
+"""Shared git diff parsing for incremental analysis.
```
ivanmilevtues: When I read this I thought there is a duplication with something I introduced some time ago:
method_diff.py 
Asked the agent to check:
```
Yes, there's significant overlap. Here's the breakdown:
Duplicate _parse_hunk_side
Identical logic in both files:
- method_diff.py:21-28 — inline re.match(r"^[+-](\d+)(?:,(\d+))?$", side)
- parsed_diff.py:17,59-66 — pre-compiled _HUNK_SIDE_RE + same match logic
Duplicate hunk header parsing
Both parse @@ -old,count +new,count @@ headers:
- method_diff.py:47-59 — _parse_diff_hunks: iterates git stdout lines, finds @@, returns list[tuple[int,int,int,int]]
- parsed_diff.py:104-129 — _parse_patch_text: same logic, returns list[ParsedDiffHunk] (with body lines)
Duplicate range classification
Both convert hunk metadata into (added_ranges, changed_ranges, deletion_points):
- method_diff.py:132-144 — classifies from raw (old_count, new_count) tuples using the heuristic: old_count==0 → added, else → changed
- parsed_diff.py:275-328 — classify_new_file_ranges: same classification but from line-by-line +/-/  content inside each hunk
Both run git diff independently
- method_diff.py:37-38 — git diff -U0 base_ref -- file_path (per-file, no context, no renames)
- parsed_diff.py:178-186 — git diff --raw -U3 -M -C base_ref [target] (bulk, rename-aware, 3 lines context)
What should happen
method_diff.py is the older file. It runs its own git diff per file and re-implements all the parsing. It should instead consume the already-loaded ParsedGitDiff / ParsedDiffHunk objects from parsed_diff.py — eliminating the duplicate _parse_hunk_side, the duplicate hunk parsing, and the redundant per-file git diff subprocess call. Something like:
def compute_method_statuses_for_file(
    methods, file_path, parsed_diff, repo_dir
) -> dict[str, ChangeStatus]:
    file_diff = parsed_diff.get_file(file_path)
    hunks = file_diff.hunks if file_diff else []
    # use classify_new_file_ranges or walk hunks directly
This would remove ~30 lines of duplication and eliminate N redundant git diff subprocess calls (one per modified file).```

File: repo_utils/change_detector.py [hunk 14]
```diff
@@ -279,6 +296,53 @@ def _parse_status_line(line: str) -> DetectedChange | None:
...
+    parsed_diff: ParsedGitDiff,
+    exclude_patterns: list[str] | None = None,
+) -> ChangeSet:
+    """Build a ``ChangeSet`` from an already-loaded ``ParsedGitDiff``.
+
+    Allows callers that need both file-level status and per-file hunks to share
+    a single ``git diff`` invocation instead of running two.
+    """
+    if exclude_patterns is None:
+        exclude_patterns = [".codeboarding/", ".codeboarding\\"]
```
ivanmilevtues: here the exclude_patterns are different than other places. We need to unify them.

repo_utils/change_detector.py
```diff
@@ -279,6 +296,53 @@ def _parse_status_line(line: str) -> DetectedChange | None:
...
+        change_type=change_type,
+        file_path=file_diff.file_path,
+        old_path=file_diff.old_path,
+        similarity=file_diff.similarity,
+    )
+
+
+def detect_changes_from_parsed_diff(
+    parsed_diff: ParsedGitDiff,
+    exclude_patterns: list[str] | None = None,
```
ivanmilevtues: This param is never passed we porly don't need it then, just use the value at 326

repo_utils/change_detector.py
```diff
@@ -279,6 +296,53 @@ def _parse_status_line(line: str) -> DetectedChange | None:
...
+    if exclude_patterns is None:
+        exclude_patterns = [".codeboarding/", ".codeboarding\\"]
+
+    changes: list[DetectedChange] = []
+    for file_diff in parsed_diff.files:
+        change = _detected_change_from_parsed_diff(file_diff)
+        if change is None:
+            continue
+
+        should_skip = any(
```
ivanmilevtues: nit. we are iterating all of them either way we can filter at the end it is easier to understand than using any at every file imo.

repo_utils/change_detector.py
```diff
@@ -279,6 +296,53 @@ def _parse_status_line(line: str) -> DetectedChange | None:
...
+def _detected_change_from_parsed_diff(file_diff: ParsedDiffFile) -> DetectedChange | None:
+    """Convert one ``ParsedDiffFile`` into a ``DetectedChange``."""
+    status = file_diff.status_code.upper()
+    try:
+        change_type = ChangeType(status)
+    except ValueError:
+        logger.warning("Unknown git status: %s", status)
+        return None
+
+    return DetectedChange(
```
ivanmilevtues: detected change is the same as file_diff, I don't see the point of having to objects with the same data, change_type is file_diff.status

for simplicity I think we should keep just one data type, as now we are doing this which I was expecting to do something more but it is again git diff with an extra filter?

repo_utils/change_detector.py
```diff
@@ -279,6 +296,53 @@ def _parse_status_line(line: str) -> DetectedChange | None:
...
+        should_skip = any(
+            change.file_path.startswith(p) or (change.old_path and change.old_path.startswith(p))
+            for p in exclude_patterns
+        )
+        if should_skip:
+            continue
+
+        changes.append(change)
+
+    return ChangeSet(changes=changes, base_ref=parsed_diff.base_ref, target_ref=parsed_diff.target_ref)
```
ivanmilevtues: (Random comment) - unsure if it is from this commit but lots of methods in this file are no longer used, if you are prompting around it ask the LLM to clean up.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+        if worktree_has_changes(repo_path):
+            return _full_required_payload(
+                "--target-ref cannot be combined with a dirty worktree; "
+                "commit or stash local changes, or omit --target-ref."
+            )
+    parsed_diff = load_parsed_git_diff(repo_path, resolved_base_ref, resolved_target_ref)
+    if parsed_diff.error:
+        return _full_required_payload(f"Git diff failed for incremental analysis: {parsed_diff.error}")
+
+    changes = detect_changes_from_parsed_diff(parsed_diff)
```
ivanmilevtues: As I wrote in the other file if i understand this, it is not really needed changes is the same as parsed diff.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+    if any(c.change_type in (ChangeType.RENAMED, ChangeType.COPIED) for c in changes.changes):
```
ivanmilevtues: This case seems somewhat trivial to be handled (maybe it isn't), but shouldn't renames be just renames in the analysis.json. I might be missing something and for it to be more complicated.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+
+    return resolve
+
+
+# ---------------------------------------------------------------------------
+# Orchestration entry point
+# ---------------------------------------------------------------------------
+
+
+def _resolve_source_identity(repo_dir: Path, ref: str | None) -> str:
```
ivanmilevtues, line 92: can we add a comment of what does this do, i have no idea what is a source identity ;d/

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+    analysis_path = output_dir / "analysis.json"
+    if changes.is_empty():
+        source_identity = _resolve_source_identity(repo_path, resolved_target_ref)
+        write_last_run_metadata(
+            output_dir,
+            repo_path,
+            mode="incremental",
+            analysis_path=analysis_path,
+            commit_hash=source_identity,
```
ivanmilevtues: commit_hash and source_identity are always the same probably we don't need to send both.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+        return _full_required_payload(f"Git diff failed for incremental analysis: {parsed_diff.error}")
+
+    changes = detect_changes_from_parsed_diff(parsed_diff)
+    if any(c.change_type in (ChangeType.RENAMED, ChangeType.COPIED) for c in changes.changes):
+        return _full_required_payload(
+            "Rename/copy changes detected; full analysis required until rename handling is implemented."
+        )
+    analysis_path = output_dir / "analysis.json"
+    if changes.is_empty():
+        source_identity = _resolve_source_identity(repo_path, resolved_target_ref)
```
ivanmilevtues: this method get kinda lenghty maybe i'd extract the if statements into methods so it reads easier:
```python
if changes.is_empty():
    return update_time_no_analysis_needed()
```

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+        return payload
+
+    root_analysis, _sub_analyses = existing
+    if root_analysis is None:
+        return _full_required_payload("No root analysis present; full analysis required.")
+
+    with contextlib.redirect_stdout(io.StringIO()):
+        generator.pre_analysis()
+
+    if generator.static_analysis is None:
```
ivanmilevtues, line 225: I don't think this can ever happen. pre_analysis is also part of the full analysis, so unsure of this if ;d

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+        return _full_required_payload("Static analysis could not be initialized; full analysis required.")
+
+    symbol_resolver = build_symbol_resolver(generator.static_analysis, repo_path)
```
ivanmilevtues: This seems like it should be it's own class with a resolve method, instead of populating the object state with anonymous functions.

diagram_analysis/incremental/pipeline.py
```diff
@@ -0,0 +1,313 @@
...
+            )
+
+    for file_path in list(methods_by_file.keys()):
+        methods_by_file[file_path] = sorted(
+            methods_by_file[file_path],
+            key=lambda method: (method.start_line, method.end_line, method.qualified_name),
+        )
+
+    return methods_by_file
+
```
ivanmilevtues, line 61: We have a very similar code snippet, I just remembered and asked an agent to pull it out:
```
Yes — collect_method_entries in incremental_pipeline.py:47-84 is a near-duplicate of the existing method indexing in analysis_json.py.
The overlap
incremental_pipeline.py:47-84 — walks static_analysis.results[lang]["references"], extracts Node objects, converts to MethodEntry, groups by file, sorts by (start_line, end_line, qualified_name).
analysis_json.py:176-187 (_build_methods_index_from_files) — walks analysis.files entries, extracts MethodEntry objects into a methods_index dict keyed by file_path|qualified_name.
The real source of truth
The analysis.files index on AnalysisInsights already contains exactly this data — MethodEntry objects grouped by file, with qualified_name, start_line, end_line, node_type. It's populated during full analysis via the same Node → MethodEntry conversion in the details/abstraction agents.
collect_method_entries re-derives the same information from the raw StaticAnalysisResults instead of just reading analysis.files from the already-loaded analysis.json.
What should happen
build_symbol_resolver (which calls collect_method_entries) should consume the existing analysis.files index rather than re-scanning StaticAnalysisResults. The MethodEntry objects are already there after load_full_analysis() deserializes analysis.json. Something like:
def build_symbol_resolver(root_analysis: AnalysisInsights, repo_dir: Path) -> SymbolResolver:
    def resolve(file_path: str) -> list[MethodEntry]:
        normalized = normalize_repo_path(file_path, repo_dir)
        entry = root_analysis.files.get(normalized)
        return list(entry.methods) if entry else []
    return resolve
This would eliminate collect_method_entries and normalize_repo_path entirely (~45 lines), and avoid the fragile isinstance(references, dict) branching that duplicates logic from StaticAnalysisResults.add_references.````

Comment (ivanmilevtues) [reply diagram_analysis/incremental/pipeline.py:61]: it is better to unify both so whenever something changes we have a single source of truth

File: diagram_analysis/incremental/pipeline.py [hunk 27]
```diff
@@ -0,0 +1,313 @@
...
+            text=True,
+            check=True,
+        )
+        return result.stdout.strip() or ref
+    except subprocess.CalledProcessError:
+        logger.warning("Could not resolve source ref '%s'; preserving original value", ref)
+        return ref
+
+
+def _diff_base_for_successful_target(repo_dir: Path, target_ref: str | None, source_identity: str) -> str | None:
```
ivanmilevtues, line 117: probably a repo_util?

diagram_analysis/incremental/updater.py
```diff
@@ -51,7 +78,7 @@ def __init__(
         self._symbol_resolver = symbol_resolver
```
ivanmilevtues, line 72: method_status_lookup: MethodStatusLookup | None = None,
        component_resolver: ComponentResolver | None = None,
These two parameters are never used we should delete them as parameters and then just use the `resolve_component_id_by_path_prefix`

ivanmilevtues: I would generally double check this class as method_status_lookup will always be None (except tests maybe?)

ivanmilevtues:
```Python
    def compute_delta(
        self,
        added_files: list[str],
        modified_files: list[str],
        deleted_files: list[str],
        changes: ChangeSet,
        reset_files: list[str] | None = None,
    ) -> IncrementalDelta:
```

The reset_files param is never used we should delete it

diagram_analysis/incremental/models.py
```diff
@@ -0,0 +1,217 @@
...
+@dataclass
+class TraceConfig:
+    """Budget constraints for the semantic tracing loop."""
+
+    max_hops: int = 3
+    max_fetched_methods: int = 30
+    max_parallel_regions: int = 4
+    max_neighbor_preview: int = 8
+    max_cached_turns: int = 4
+    llm_max_retries: int = 2
```
ivanmilevtues, line 25: The llm config stuff should be unified with the LLMConfig so we don't have them in multiple places.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+                    patched[comp_id] = result
+        return patched, failed
+
+    def generate_analysis_incremental(
+        self,
+        delta: IncrementalDelta,
+        base_ref: str,
+        parsed_diff: ParsedGitDiff | None = None,
+        baseline_cfgs: dict[str, CallGraph] | None = None,
+        config: TraceConfig | None = None,
```
ivanmilevtues: make these non-nullable they are always passed.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+
+        Applies the architectural delta in place, traces semantic impact of
+        changed methods through the call graph, patches the affected
+        sub-analyses via EASE-encoded JSON Patch, and saves the result.
+
+        The caller owns mode selection (cosmetic-only fast paths live in the
+        wrapper); this method handles the semantic-tracing path and its
+        deterministic fallbacks.
+        """
+        if self.details_agent is None or self.abstraction_agent is None:
```
ivanmilevtues: just assert them we have already ran pre_analysis.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+            self.pre_analysis()
+        assert self.static_analysis is not None
+
+        existing = load_full_analysis(Path(self.output_dir))
```
ivanmilevtues, line 497: This logic we have already done, we shoudl pass the analysis.json here, this is again teh same logic and I am getting puzzled 😵

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+        if existing is None:
```
ivanmilevtues, line 498: run_incremental_pipeline does the same checks, but not really. I think we should skip em or idk. Ich bin confusion.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+        if delta.needs_reanalysis:
+            return IncrementalRunResult(
+                summary=IncrementalSummary(
+                    kind=IncrementalSummaryKind.REQUIRES_FULL_ANALYSIS,
+                    message="Delta references files without a known component; full analysis required.",
+                    requires_full_analysis=True,
+                ),
+            )
+
+        apply_delta(root_analysis, sub_analyses, delta)
```
ivanmilevtues: we shoudl name this method better:
def apply_delta(
    root: AnalysisInsights,
    delta: IncrementalDelta,
) -> None:
    """Apply ``IncrementalDelta`` architectural changes in place.

    Only architectural facts are mutated here — added/removed methods and
    files. Status tracking is the caller's responsibility.
    """

I suggest apply_method_delta or something like that as we are just updating the methods adding the new ones, and removing oldies.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+            return IncrementalRunResult(
+                summary=IncrementalSummary(
+                    kind=IncrementalSummaryKind.ADDITIVE_ONLY,
+                    message="Changes are purely additive; no semantic patching needed.",
+                ),
+                analysis_path=analysis_path,
+            )
+
+        assert self.agent_llm is not None and self.parsing_llm is not None
+        agent_llm = self.agent_llm
```
ivanmilevtues: no need to do a = a @Svilen-Stefanov favorito burrito

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+    config: TraceConfig | None = None,
+    baseline_cfgs: dict[str, CallGraph] | None = None,
+    parsed_diff: ParsedGitDiff | None = None,
+) -> TraceResult:
+    """Run the semantic tracing loop over changed methods.
+
+    Returns a TraceResult with impacted methods. Scope classification
+    (impacted_components) happens in the orchestrator via ``classify_scope``.
+    """
+    del agent_llm  # reserved for future routing; tracing currently uses parsing_llm
```
ivanmilevtues: ?? this makes 0 sense ;d, we pass in an llm and delete it

diagram_analysis/diagram_generator.py
```diff
@@ -160,6 +176,8 @@ def pre_analysis(self):
 
         # Initialize LLMs before spawning threads so both share the same instances
         agent_llm, parsing_llm = initialize_llms()
+        self.agent_llm = agent_llm
```
ivanmilevtues, line 186: you can do self.agent_llm, self.parsing_llm = intialize_llms()

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+        return list(self._unresolved)
+
+
+# ---------------------------------------------------------------------------
+# Neighbor extraction from CFG
+# ---------------------------------------------------------------------------
+NeighborIndex = tuple[dict[str, list[str]], dict[str, list[str]]]
+
+
+def _build_neighbor_indexes(*cfg_dicts: dict[str, CallGraph]) -> NeighborIndex:
```
ivanmilevtues: rename to:
Build upstream/downstream adjacency maps

ivanmilevtues (commented):
Putting out some comments.

diagram_analysis/incremental_tracer.py
```diff
@@ -0,0 +1,925 @@
...
+    cosmetic_skipped = 0
+    extension_skipped = 0
+    fast_path_impacted_methods: set[str] = set()
+    saw_fallback_region = False
+
+    for file_delta in delta.file_deltas:
+        fp = file_delta.file_path
+
+        ext = Path(fp).suffix.lower()
+        if ext not in EXTENSION_TO_LANGUAGE:
```
ivanmilevtues: I think thi is already filtered 6 times ;d, by the time we come here

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+
+        if (
+            file_delta.file_status == ChangeStatus.MODIFIED
+            and file_delta.modified_methods
+            and not file_delta.added_methods
+            and not file_delta.deleted_methods
+            and is_file_cosmetic(repo_dir, base_ref, fp)
+        ):
+            cosmetic_skipped += 1
+            logger.info("Skipping cosmetic-only file: %s", fp)
```
ivanmilevtues: hmm interesting, how do we know that by modifying a file we haven't added it to a new component if we have a bunch of additions and they are connected to ti.

diagram_analysis/semantic_diff.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+def _get_parser(language: Language, file_ext: str) -> Any | None:
+    """Return a cached tree-sitter Parser for *language*, or None on failure."""
+    cache_key = f"{language}:{file_ext}" if language == Language.TYPESCRIPT else str(language)
+    if cache_key in _parser_cache:
+        return _parser_cache[cache_key]
+
+    try:
+        from tree_sitter import Language as TSLanguage, Parser
```
ivanmilevtues: lets not use local imporst + I feel like tree_sitter should be in static_analyzer, similar to our LSP clients.


Just a note: The constructor for class "Language" is deprecated
  int argument support is deprecatedPylance

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+
+        diff_text = (
+            _get_diff_hunks(repo_dir, base_ref, fp, parsed_diff) if file_delta.file_status != ChangeStatus.ADDED else ""
+        )
+
+        for mc in all_methods:
+            body = _read_method_body(repo_dir, fp, mc.start_line, mc.end_line)
+            if body is not None:
+                body = strip_comments_from_source(fp, body)
+            up, down = _get_neighbors(upstream_index, downstream_index, mc.qualified_name)
```
ivanmilevtues: please quadriple check that the names are properly found, we have had so many silent issues with method qualified_name being not the thing that we expect. I.e. add a logger if something is not matched so we can see if it explodes.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+
+            if mc.change_type == ChangeStatus.MODIFIED:
```
ivanmilevtues: mc sounds like method_call, maybe name just method (minor)

codeboarding_cli/parser.py
Svilen-Stefanov: Yeah, I think the entire file seems like it could be in main.py. It kind of doesn't make too much sense to have main.py be 3 lines just importing another module.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+        return f"file:{file_path}", False
+    return f"region:{region_id}", True
+
+
+def _compare_modified_method_versions(
+    file_path: str,
+    old_body: str | None,
+    new_body: str | None,
+) -> tuple[bool, bool]:
+    """Return ``(semantically_unchanged, signature_changed)`` for a modified method."""
```
ivanmilevtues: I would split into two methods is_semantically_unchanged and is_signature_changed.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+                    old_body = strip_comments_from_source(fp, old_body)
+                semantically_unchanged, signature_changed = _compare_modified_method_versions(fp, old_body, body)
+                if semantically_unchanged:
+                    continue
+                if (
+                    file_delta.file_status == ChangeStatus.MODIFIED
+                    and not file_delta.added_methods
+                    and not file_delta.deleted_methods
+                    and not signature_changed
+                    and not up
```
ivanmilevtues: i don't get why not up but down is not used. Generally i think we can put the ChangeStatus.Modified body in a new method, with a quick comment explaining what are the conditions for something to happen.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+    base_ref: str,
+    parsed_diff: ParsedGitDiff | None = None,
+) -> TracePlan:
+    """Build grouped trace regions plus deterministic fast-path impact decisions."""
+    groups: dict[str, ChangeGroup] = {}
+    graph_metadata = _build_graph_region_metadata(upstream_index, downstream_index)
+
+    cosmetic_skipped = 0
+    extension_skipped = 0
+    fast_path_impacted_methods: set[str] = set()
```
ivanmilevtues: What is fast_path (it seems like it is only if they are affected upwards.)

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return None
+
+    lines = result.stdout.splitlines(keepends=True)
+    if start_line < 1 or end_line > len(lines):
+        return None
+    return "".join(lines[start_line - 1 : end_line])
+
+
+def _build_graph_region_metadata(
```
ivanmilevtues: I don't ge thow the regions are interlinked with the existing analysis.

codeboarding_cli/commands/full_analysis.py
Svilen-Stefanov, line 1: I think the naming here is suboptimal - we have `codeboarding_cli/commands/full.py`. It's probably better to have the follwoing structure:
- codeboarding_cli
   - parser.py
   - bootstrap.py
   - operations/
        - full_analysis.py
        - incremental_anaylsis.py

I think this gives us the possibility to extend this further in the future with other types of perations (run warning checks only, check architecture etc)

codeboarding_cli/commands/incremental.py
```diff
@@ -0,0 +1,118 @@
...
+    parser.add_argument("--base-ref", type=str, help="Base git ref/tree. Defaults to the last successful run.")
+    parser.add_argument(
+        "--target-ref",
+        type=str,
+        help=(
+            "Target git ref/tree. Must match the current checkout with a clean worktree; "
+            "defaults to the current working tree."
+        ),
+    )
+    parser.add_argument("--json", action="store_true", help="Emit structured JSON on stdout")
```
Svilen-Stefanov: Do we need this? Do we support any other format? Let's remove this if not needed.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+            _append_method_to_group(groups, region_key, graph_backed, fp, diff_text, ctx, up, down)
+
+        for mc in file_delta.deleted_methods:
+            ctx = ChangedMethodContext(
+                qualified_name=mc.qualified_name,
+                file_path=fp,
+                change_type=ChangeStatus.DELETED,
+                new_body=None,
+            )
+            up, down = _get_neighbors(upstream_index, downstream_index, mc.qualified_name)
```
ivanmilevtues: it could be me being stupid, but I felt like upstream and downstream indexes are outgoing and incoming edges.
Now the part that I don't get this index is at the latest snapshot (the moment for which we are iterating), in that latest snapshot delted methods will not be in the graph (they are deleted?), what is the purpose of this search?

codeboarding_cli/commands/incremental.py
```diff
@@ -0,0 +1,118 @@
...
+from diagram_analysis import RunContext
+from diagram_analysis.incremental_models import IncrementalSummary, IncrementalSummaryKind
+from utils import monitoring_enabled
+
+logger = logging.getLogger(__name__)
+
+
+def add_arguments(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument("--incremental", action="store_true", help="Use incremental analysis")
+    parser.add_argument("--base-ref", type=str, help="Base git ref/tree. Defaults to the last successful run.")
```
Svilen-Stefanov: Default to the last commit? Or what does "last successful run" mean here? I should provide a commit hash I suppose? Can we make this clearer?

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+        if trace_plan.fast_path_impacted_methods:
+            return TraceResult(
+                all_impacted_methods=trace_plan.fast_path_impacted_methods,
+                stop_reason=TraceStopReason.CLOSURE_REACHED,
+            )
+        logger.info("No traceable changed methods; skipping trace")
+        return TraceResult()
+
+    if trace_plan.parallel_safe:
+        max_workers = min(len(trace_plan.groups), cfg.max_parallel_regions)
```
ivanmilevtues: do we need the trace_plan.parallel_safe, it feels like we can just do:
min(len(trace_plan.groups), cfg.max_parallel_regions)

as the prallel_safe is trace_plan.groups>1, so it will end up teh same with no if/else and no head scratching what makes it safe (from my understanding it is not about safety but about if we can do it?)

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+
+def _trace_single_group(
+    group: ChangeGroup,
+    static_analysis: StaticAnalysisResults,
+    repo_dir: Path,
+    parsing_llm: BaseChatModel,
+    config: TraceConfig,
+) -> TraceResult:
+    """Run the semantic tracing loop for a single change region."""
+    from trustcall import create_extractor
```
ivanmilevtues: Global import

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+
+    resolver = MethodResolver(static_analysis, repo_dir)
+    extractor = create_extractor(parsing_llm, tools=[TraceResponse], tool_choice=TraceResponse.__name__)
+
+    prompt = _build_initial_prompt([group], config.max_neighbor_preview)
+    caching_supported = _supports_anthropic_prompt_caching(parsing_llm)
```
ivanmilevtues: why anthropic?, I feel like we should know this at LLM config level

codeboarding_cli/parser.py
```diff
@@ -0,0 +1,42 @@
...
+    parser = build_parser()
```
Svilen-Stefanov: The current design is a flat namespace with no real separation.
Both full.add_arguments() and incremental.add_arguments() dump all their flags into the same ArgumentParser. There are no subcommands. That creates several issues:
1. Every flag is always present on args, regardless of mode.

When --incremental is passed, args still has repositories, upload, partial_component_id, full, etc. When it's not passed, args still has base_ref, target_ref, json. 

2. Each command must manually reject the other's flags.

3. The validation is duplicated and incomplete.

Both commands independently re-derive output_dir (full.py:84, incremental.py:65), project_name (full.py:102, incremental.py:75), and call bootstrap_environment (full.py:90, incremental.py:69). The shared args like --local, --output-dir, --binary-location, --depth-level are defined in full.add_arguments but silently relied upon by incremental.run_from_args.

**What we should probably do:**
Use argparse subcommands - argparse.add_subparsers(). Each subcommand gets its own argument namespace, so invalid combinations are structurally impossible:

```
def build_parser():
    parser = argparse.ArgumentParser(...)
    # Shared args on the parent
    parser.add_argument("--local", type=Path, ...)
    parser.add_argument("--output-dir", type=Path, ...)
    parser.add_argument("--binary-location", type=Path, ...)
    parser.add_argument("--depth-level", type=int, default=1, ...)
    parser.add_argument("--enable-monitoring", action="store_true", ...)
    subs = parser.add_subparsers(dest="command")
    full_parser = subs.add_parser("full", ...)
    full_parser.add_argument("repositories", nargs="*", ...)
    full_parser.add_argument("--upload", ...)
    full_parser.add_argument("--partial-component-id", ...)
    full_parser.add_argument("--force", ...)
    incr_parser = subs.add_parser("incremental", ...)
    incr_parser.add_argument("--base-ref", ...)
    incr_parser.add_argument("--target-ref", ...)
    incr_parser.add_argument("--json", ...)
```

codeboarding_workflows/artifact_copy.py
Svilen-Stefanov, line 1: Can we improve the naming here as well? I want to make sure `codeboarding_workflows/files.py` makes sense. Right now the name is too generic / not descriptive enough for me.

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+        "Update impacted_methods (cumulative) and either request more methods or stop.\n"
+        "Only populate semantic_impact_summary if you conclude there is material semantic impact and closure has been reached.\n"
+    )
+    return "\n".join(parts)
+
+
+# ---------------------------------------------------------------------------
+# Core tracing loop
+# ---------------------------------------------------------------------------
+def _invoke_extractor_with_retry(
```
ivanmilevtues, line 163: we already have this type of method in the `agent.py`

diagram_analysis/incremental/tracer.py
```diff
@@ -0,0 +1,925 @@
...
+    if baseline_cfgs is not None:
+        index_args.append(baseline_cfgs)
+    upstream_idx, downstream_idx = _build_neighbor_indexes(*index_args)
+    return _build_trace_plan(delta, upstream_idx, downstream_idx, repo_dir, base_ref, parsed_diff)
+
+
+# ---------------------------------------------------------------------------
+# Prompt construction
+# ---------------------------------------------------------------------------
+_TRACE_SYSTEM = """\
```
ivanmilevtues: this should be in the prompts

codeboarding_workflows/files.py
```diff
@@ -0,0 +1,22 @@
+import logging
+import shutil
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def copy_files(temp_folder: Path, output_dir: Path) -> None:
```
Svilen-Stefanov: Can this be a utils function that accepts files as one arg and target_dir? Seems like something we might need elsewhere too.

diagram_analysis/analysis_patcher.py
```diff
@@ -0,0 +1,216 @@
...
+                        continue
+                    comp["file_methods"][fm_key] = ease_decode(fm, ["methods"])
+            encoded["components"][key] = ease_decode(comp, ["key_entities", "file_methods"])
+    return ease_decode(encoded, ["components", "components_relations"])
+
+
+# ---------------------------------------------------------------------------
+# Prompt construction
+# ---------------------------------------------------------------------------
+_PATCH_SYSTEM = """\
```
ivanmilevtues: prompts should be within the prompt factory

codeboarding_workflows/full.py
Svilen-Stefanov, line 1: Similar to the CLI comment - maybe rename the file name to make more sense individually (importing `full` seems dangerous in general)

Svilen-Stefanov (commented):
Leaving my comments for now. I'll continue later.

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+
+        impact_targets = [ic for ic in trace_result.impacted_components if ic.component_id in sub_analyses]
+        if not impact_targets:
+            return patched, failed
+
+        max_workers = min(len(impact_targets), 4)
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = {
+                executor.submit(
+                    patch_sub_analysis,
```
ivanmilevtues, line 450: I don't think this will generate a new high-level component, as it patcehs only sub analysis

codeboarding_cli/commands/full.py
brovatten: done

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
...
+            )
+
+        classify_scope(
+            trace_result,
+            root_analysis.file_to_component(),
+            self.static_analysis,
+            self.repo_location,
+        )
+
+        patched, failed = self._run_component_patches(sub_analyses, trace_result, parsing_llm)
```
ivanmilevtues: i feel like instead of returning failed we should just raise excpetion and catch it. there are lots of if failed around the code of this change and it makes it hard to read also it is weird all methods to return tuple[]

codeboarding_cli/commands/full_analysis.py
brovatten: makes sense. i took it one step further and did : 

      local_full.py      # current full.py lines 98–123                                                                                                         
      remote.py          # current full.py lines 125–177 + tqdm loop

diagram_analysis/diagram_generator.py
```diff
@@ -394,3 +412,242 @@ def generate_analysis(self) -> Path:
     def generate_analysis_smart(self) -> Path:
```
ivanmilevtues: remove this method

codeboarding_cli/commands/incremental.py
```diff
@@ -0,0 +1,118 @@
...
+            payload = incremental_error_payload(str(exc))
+        finally:
+            run_context.finalize()
+
+    _emit_payload(payload, args.json)
+
+
+def _emit_payload(payload: dict, emit_json: bool) -> None:
+    if emit_json:
+        print(json.dumps(payload, default=str, indent=2, sort_keys=True))
```
ivanmilevtues: We don't use prints. I suppose the idea is to send this to the wrapper as a message. We generally do not do this, I will have to look at the wrapper to see why.

Because in my head the API for incremental and full analysis is the same -> give me snapshot -> get a fresh analysis.json.

ivanmilevtues (approved):
I am done with this pass of the PR. Generally I like it conveptually, I am not super happy about implementation it is a complicated thing I agree, but it is extremely difficult to follow there are 1M branches off (feels like a lot of patchwork has happened).

Thinking of timing and so on, I am okay with most of it as long as it is functional but this will come around bite us in the ass if we do not clean up. We already suffered with the last incrmenetal analysis (My bad :'( ).

Excited to play with this, we can improve structure by adding diagram_analysis/incremental :) 


Few unsued things from an llm scan:
Symbol	Location	Used?	Action
classify_new_file_ranges	parsed_diff.py:275	No callers	Remove
EscalationLevel enum	incremental_models.py:117	Only as type annotation, never set	Remove or implement
COSMETIC_ONLY	incremental_models.py:126	Never produced	Remove or implement
RENAME_ONLY	incremental_models.py:127	Never produced	Remove or implement
FULL_FALLBACK	incremental_models.py:132	Never produced	Remove or implement
generate_analysis_smart	diagram_generator.py:412	Only caller is github_action.py	Inline caller, remove
use_source_as_diff_base param	run_metadata.py:55	Never passed externally	Make private

codeboarding_cli/commands/incremental.py
brovatten: Yup removed it. i think the idea was to return something to the user, but not a good way

codeboarding_cli/commands/incremental.py
brovatten: Yup unified this!

agents/analysis_patcher.py
```diff
@@ -0,0 +1,216 @@
+"""EASE-encoded JSON Patch flow for incremental sub-analysis updates.
+
+Given impacted components from the tracer, extracts the parent sub-analysis,
+EASE-encodes it, asks the LLM for RFC 6902 patches, applies them, validates
+the result, and merges back into the full analysis.
+"""
+
+import json
```
ivanmilevtues, line 13: this whole file should go to agents imo + we need to probably explain that it is trustcall inspired so we have a better idea of what is going on.

diagram_analysis/diagram_generator.py
brovatten: leaving as-is. The locals get reused as kwargs three times
below (MetaAgent, DetailsAgent, AbstractionAgent), so dropping them
saves 2 lines at the assignment but adds 6 self. tokens at the call
sites — and it'd diverge from the same local-alias pattern already used at diagram_generator.py:537-539

repo_utils/parsed_diff.py
brovatten: target_ref is intentionally empty in the worktree-diff path — incremental_pipeline.py:200 normalizes None to "" and passes it through, and the same     
  function relies on the falsy case at parsed_diff.py:240 to append untracked files. Dropping the guard would pass "" as an argv to git diff, which breaks
   the worktree mode — the "HEAD" default in the signature is just a default, not an invariant

repo_utils/parsed_diff.py
brovatten: acknowledge the duplication is real and the consolidation pattern already exists (change_detector.detect_changes_from_parsed_diff as precedent), but flag that -U0→-U3 is not header-equivalent — we need a body-walking classifier to preserve exact boundaries, which means resurrecting the function I just deleted under a different name or on the dataclass itself

codeboarding_cli/bootstrap.py
```diff
@@ -0,0 +1,40 @@
...
+from install import ensure_tools
+from logging_config import setup_logging
+from tool_registry import needs_install
+from user_config import ensure_config_template, load_user_config
+from vscode_constants import update_config
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_local_run_paths(args: argparse.Namespace) -> tuple[Path, Path, str]:
```
Svilen-Stefanov: Instead of `tuple[Path, Path, str]`, should we create a data class for this?

codeboarding_cli/commands/full_analysis.py
Svilen-Stefanov: I still see this and the below as `full.py` and `incremental.py`, instead of `full_analysis.py` and `incremental_anaylsis.py`.

codeboarding_cli/parser.py
Svilen-Stefanov: This is still a separate file. Can you remove it and move the full thing in `main.py`?

codeboarding_cli/commands/remote.py
```diff
@@ -0,0 +1,84 @@
...
+from monitoring.paths import get_monitoring_run_dir
+from repo_utils import get_repo_name, store_token
+from repo_utils.ignore import initialize_codeboardingignore
+from utils import monitoring_enabled
+
+logger = logging.getLogger(__name__)
+
+
+def run_from_args(args: argparse.Namespace, parser: argparse.ArgumentParser) -> None:
+    output_dir = Path.cwd() / ".codeboarding"
```
Svilen-Stefanov: .codeboarding should be a constant

Svilen-Stefanov (commented):
Leaving a couple things I saw. I'll continue tomorrow.

PR #322, opened 2026-04-28
Improve incremental analysis

Svilen-Stefanov:
---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/322" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/diagram_analysis/incremental/delta.py
+++ b/diagram_analysis/incremental/delta.py
@@ -36,6 +36,8 @@ class FileDelta:
     modified_methods: list[MethodChange] = field(default_factory=list)
     deleted_methods: list[MethodChange] = field(default_factory=list)
     renamed_qualified_names: dict[str, str] = field(default_factory=dict)
+    # Why: produced by the wrapper on revert so the IDE can drop stale overlays.
+    reset_methods: list[MethodChange] = field(default_factory=list)
 
     def to_dict(self) -> dict[str, Any]:
         return {
@@ -47,6 +49,7 @@ class FileDelta:
             "modified_methods": [m.to_dict() for m in self.modified_methods],
             "deleted_methods": [m.to_dict() for m in self.deleted_methods],
             "renamed_qualified_names": self.renamed_qualified_names,
+            "reset_methods": [m.to_dict() for m in self.reset_methods],
         }
 
 
@@ -67,6 +70,11 @@ class IncrementalDelta:
             fd.file_status != ChangeStatus.DELETED and not fd.modified_methods and not fd.deleted_methods
             for fd in self.file_deltas
         )
+
+    @property
+    def needs_semantic_trace(self) -> bool:
+        """True when modifications or additions remain."""
+        return any(fd.modified_methods or fd.added_methods for fd in self.file_deltas)
 
     def to_dict(self) -> dict[str, Any]:
         return {
--- a/diagram_analysis/incremental/pipeline.py
+++ b/diagram_analysis/incremental/pipeline.py
@@ -30,7 +30,7 @@ from diagram_analysis.run_metadata import METADATA_FILENAME, write_incremental_r
 from repo_utils.diff_parser import detect_changes
 from repo_utils.ignore import initialize_codeboardingignore
 from repo_utils import get_git_commit_hash, get_repo_state_hash
-from repo_utils.git_ops import resolve_ref, worktree_has_changes
+from repo_utils.git_ops import git_object_type, resolve_ref, worktree_has_changes
 from static_analyzer.analysis_result import StaticAnalysisResults
 from utils import ANALYSIS_FILENAME, CODEBOARDING_DIR_NAME, to_relative_path
 
@@ -94,7 +94,8 @@ def _resolve_source_identity(repo_dir: Path, ref: str | None) -> str:
 
     Why: a dirty worktree has no commit hash, so we fall back to a content
     hash (``get_repo_state_hash``) so later incremental runs can still match.
-    When *ref* is given, we resolve it to its commit SHA.
+    When *ref* is given, we resolve it to its commit SHA — or pass a tree-ish
+    ref through unchanged, since a tree SHA is already a stable identity.
     """
     if not ref:
         return (
@@ -102,6 +103,9 @@ def _resolve_source_identity(repo_dir: Path, ref: str | None) -> str:
             if worktree_has_changes(repo_dir, exclude_patterns=(CODEBOARDING_DIR_NAME,))
             else get_git_commit_hash(repo_dir)
         )
+
+    if git_object_type(repo_dir, ref) == "tree":
+        return resolve_ref(repo_dir, ref) or ref
 
     resolved = resolve_ref(repo_dir, ref)
     if resolved is None:
@@ -131,8 +135,16 @@ def _target_ref_matches_checkout(repo_dir: Path, target_ref: str) -> bool:
 
 
 def _validate_target_ref(repo_path: Path, resolved_target_ref: str) -> str | None:
-    """Return an error message if *resolved_target_ref* is incompatible with the current checkout."""
+    """Return an error message if *resolved_target_ref* is incompatible with the current checkout.
+
+    A tree-ish ref is content-addressed and independent of the worktree's
+    checkout state, so the HEAD-match and dirty-worktree checks are skipped
+    for trees. Wrapper-driven snapshot runs rely on this — they pass a
+    pre-staged ``git write-tree`` SHA as the target.
+    """
     if not resolved_target_ref:
+        return None
+    if git_object_type(repo_path, resolved_target_ref) == "tree":
         return None
     if not _target_ref_matches_checkout(repo_path, resolved_target_ref):
         return (
--- a/diagram_analysis/incremental/updater.py
+++ b/diagram_analysis/incremental/updater.py
@@ -12,7 +12,13 @@ import time
 from pathlib import PurePosixPath
 from typing import Callable
 
-from agents.agent_responses import AnalysisInsights, Component, FileEntry, FileMethodGroup, MethodEntry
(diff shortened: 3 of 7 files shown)


diagram_analysis/incremental/delta.py
```diff
@@ -36,6 +36,8 @@ class FileDelta:
     modified_methods: list[MethodChange] = field(default_factory=list)
     deleted_methods: list[MethodChange] = field(default_factory=list)
     renamed_qualified_names: dict[str, str] = field(default_factory=dict)
+    # Why: produced by the wrapper on revert so the IDE can drop stale overlays.
+    reset_methods: list[MethodChange] = field(default_factory=list)
```
ivanmilevtues, line 40: Dead code:
```Confirmed — your suspicion is correct. The reset_methods payload is effectively dead because of a producer/consumer mismatch:

Producer side ([analysis_controller.py:354-371](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-wrapper/codeboarding_pro/analysis_controller.py#L354-L371)):
Builds a FileDelta with reset_methods=[…]
Never sets any is_reset field (no such field exists on the dataclass — see [delta.py:29-53](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding/diagram_analysis/incremental/delta.py#L29-L53))

Consumer side ([MessageHandler.ts:77-86](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-vscode/backend/services/MessageHandler.ts#L77-L86)):
if (fileDelta.is_reset && fileDelta.reset_methods) {  // is_reset is always undefined
    target.reset_files.push({ ... });
    continue;
}
The IDE wraps reset_methods consumption behind fileDelta.is_reset, but that flag is never emitted by the wrapper or core. to_dict() in [delta.py:42-53](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding/diagram_analysis/incremental/delta.py#L42-L53) does not serialize is_reset (it doesn't exist), so the guard is always falsy and the reset branch is never taken.

Test gap: [test_incremental_reset.py:241-243](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-wrapper/tests/unit/test_incremental_reset.py#L241-L243) only asserts the reset_methods array is present in the wire payload — it does not assert the is_reset flag, which is why the regression slipped through. The Python side and the TS side were written against different contracts.

Why it might "work on main": if main was emitting a top‑level resetFiles field (the test even asserts "resetFiles" not in params, suggesting that's the older shape), the IDE may have been consuming that path instead. The current branch migrated the producer to per‑file reset_methods but the consumer still expects an is_reset discriminator that nobody sets.

Fundamental concern you're raising is also valid: a FileDelta with file_status=UNCHANGED and a populated reset_methods list is semantically ambiguous — reset is being overloaded onto the same struct used for diffs. Two cleaner options:
Add an explicit is_reset: bool field to FileDelta (and serialize it) — minimal fix that aligns producer/consumer.
Drop reset_methods entirely; on revert, emit a FileDelta with deleted_methods for the previously-tracked overlay methods plus a normal status update — the IDE already knows how to drop those.

Want me to apply fix (1) as the minimal-change repair, or refactor toward (2)?```

File: repo_utils/diff_parser.py [hunk 2]
```diff
@@ -177,33 +177,79 @@ def _finalize_file_diff(file_diff: FileChange, patch_lines: list[str]) -> FileCh
     return file_diff
 
 
+_DIFF_HEADER_RE = re.compile(r'^diff --git "?a/(.+?)"? "?b/(.+?)"?$')
```
ivanmilevtues, line 180: another good one.

diagram_analysis/incremental/updater.py
```diff
@@ -363,3 +367,89 @@ def apply_method_delta(
...
+        rel
+        for rel in analysis.components_relations
+        if rel.src_id not in removed_ids
+        and rel.dst_id not in removed_ids
+        and rel.src_name not in removed_names
+        and rel.dst_name not in removed_names
+    ]
+
+
+def prune_empty_components(
```
ivanmilevtues, line 394: There should be validation step after this step.

(So we should use our own validators after we toouch analysis.json).

I am saying this because it might leave disconnected components example:
A->B->C
You delete B: A [empty space] C

So we need to rerun your logic for wiring things. The incremental updater should be treated as of an agent so it should follow the same steps:
analyze -> validate -> rewire.

So in my head from orchestration standpoint it should fit in the same "template method".

For now I suggest just run the rewire method that adds the "calls" edges.

diagram_analysis/incremental/updater.py
```diff
@@ -363,3 +367,89 @@ def apply_method_delta(
...
+    for sub in sub_analyses.values():
+        all_components.extend(sub.components)
+
+    candidate_ids = {c.component_id for c in all_components if c.component_id and _component_is_empty(c)}
+    if not candidate_ids:
+        return set()
+
+    non_empty_ids = {c.component_id for c in all_components if c.component_id and not _component_is_empty(c)}
+
+    # Safety: never prune a component that has a non-empty descendant. The
```
ivanmilevtues: This shouldn't be the case, descendent is a child if I understand this corerctly.
If scoping works well (which it should), delete everything from the component down. 🔪

diagram_analysis/incremental/updater.py
```diff
@@ -363,3 +367,89 @@ def apply_method_delta(
...
+        _drop_relations(sub, removed_ids, removed_names)
+
+    for cid in list(sub_analyses.keys()):
+        if cid in removed_ids:
+            del sub_analyses[cid]
+
+    return removed_ids
+
+
+def drop_deltas_for_pruned_components(delta: IncrementalDelta, removed_ids: set[str]) -> None:
```
ivanmilevtues, line 450: This is patchwork, shouldn't exist imo. If you rerun incremental analysis, the file is deleted and the analysis.json doesn't have it so we won't ever get file monitor notifications for this thing.

ivanmilevtues (approved):
Well done some clean up comments :)

diagram_analysis/incremental/updater.py
Svilen-Stefanov: @brovatten take a look at this for the improvements you're working on. Given the urgency for today, let's do this in the next iteration.

diagram_analysis/incremental/delta.py
Svilen-Stefanov: It is used by the callers of the library to track modifications.

diagram_analysis/incremental/delta.py
```diff
@@ -68,6 +71,11 @@ def is_purely_additive(self) -> bool:
             for fd in self.file_deltas
         )
 
+    @property
+    def needs_semantic_trace(self) -> bool:
+        """True when modifications or additions remain."""
```
brovatten, line 76: I am not entirely sure here. 

Example:

[a, b]  - "sends to" - > [c]     

but the only communication between the two components are b and c. I think edge labels and the edges themselves can need an LLM patch

diagram_analysis/incremental/pipeline.py
```diff
@@ -94,7 +94,8 @@ def _resolve_source_identity(repo_dir: Path, ref: str | None) -> str:
 
     Why: a dirty worktree has no commit hash, so we fall back to a content
     hash (``get_repo_state_hash``) so later incremental runs can still match.
-    When *ref* is given, we resolve it to its commit SHA.
+    When *ref* is given, we resolve it to its commit SHA — or pass a tree-ish
+    ref through unchanged, since a tree SHA is already a stable identity.
```
brovatten, line 98: I dont get this comment, pls elaborate IRL :D

brovatten: like commit SHA or a tree, both are accepted?

brovatten: it was probably me who did this in the worktree implementation, but shouldn't it be either commit or tree and not both?

diagram_analysis/incremental/updater.py
```diff
@@ -363,3 +367,89 @@ def apply_method_delta(
...
         for component in sub.components:
             _sync_component_methods(component, files, deltas_by_path, parent_ids)
+
+
+def _component_is_empty(component: Component) -> bool:
+    """A component is empty when it owns zero methods across all its files."""
+    return not any(group.methods for group in component.file_methods)
+
+
+def _drop_relations(analysis: AnalysisInsights, removed_ids: set[str], removed_names: set[str]) -> None:
```
brovatten, line 383: docstring here pl0x

diagram_analysis/incremental/updater.py
```diff
@@ -363,3 +367,89 @@ def apply_method_delta(
...
+
+    removed_names = {c.name for c in all_components if c.component_id in removed_ids}
+
+    root.components = [c for c in root.components if c.component_id not in removed_ids]
+    _drop_relations(root, removed_ids, removed_names)
+
+    for sub in sub_analyses.values():
+        sub.components = [c for c in sub.components if c.component_id not in removed_ids]
+        _drop_relations(sub, removed_ids, removed_names)
+
```
brovatten, line 437: did u check devin comment above?

repo_utils/git_ops.py
```diff
@@ -210,6 +210,26 @@ def resolve_ref(repo_dir: Path, ref: str) -> str | None:
     return result.stdout.strip() or None
 
 
+def git_object_type(repo_dir: Path, ref: str) -> str | None:
```
brovatten, line 213: seems like a very strange function, but if it works it works

brovatten (approved):

Svilen-Stefanov:
We discussed that this is just patching the current incremental analysis to have something actually incremental. The next iteration should properly do the incremental updates.

PR #323, opened 2026-04-29
Iterative LLM-based incremental analysis

brovatten:
## Summary

Replaces the heuristic-only incremental path with an LLM-driven semantic tracer + structured patcher. The tracer walks call-graph neighbors with a hop budget to identify methods whose architecture descriptions are impacted by code changes; the patcher applies bounded-scope replacements via a structured extractor contract.

- Adds `diagram_analysis/incremental_tracer.py` (semantic tracing loop with per-region grouping + a deterministic cosmetic-only fast path), `analysis_patcher.py` (bounded-scope patch contract), `scope_planner.py` (pure helpers: ownership index, scope derivation, patch routing), `incremental_models.py` (trace-loop pydantic models).
- Adds `codeboarding_workflows/incremental.py` to orchestrate incremental runs and fall back to a full analysis when no baseline exists. `main.py` and `github_action.py` route through it directly (drops the function-local circular import via `DiagramGenerator.generate_analysis_smart`).
- Splits the 567-line `static_analyzer/semantic_diff.py` into `tree_sitter_parsers.py` (parser cache + `_LANG_CONFIGS`), `cosmetic_diff.py` (`is_file_cosmetic`, `check_syntax_errors`, `strip_comments_from_source`), `method_fingerprint.py` (whitespace-normalized fingerprints).
- Adds `static_analyzer/graph_query.py` (SCC/region helpers, neighbor indexes) and fixes `repo_utils/method_diff.py` to classify added vs changed methods inside mixed hunks (the existing test's `pytest.mark.skip` is removed).
- Trace and patch LLM call sites accept a `callbacks` list so usage flows into the existing `MONITORING_CALLBACK` contract; `apply_delta` returns new analyses instead of mutating in place.

Also includes the existing planning docs (`plan.md`, `workflow.md`) that were on this branch when the PR was originally opened.

## Summary by CodeRabbit

* **New Features**
  * Commit-based incremental semantic analysis that updates only impacted scopes
  * Cosmetic-vs-semantic change detection to skip non-functional edits
  * Tree-sitter powered multi-language parsing for broader language support

* **Improvements**
  * Deterministic, trace-driven scope derivation and safer, bounded patch application
  * Better handling of disconnected/added files and conservative widening on ambiguity
  * Faster, more reliable analysis cycles and improved test coverage


workflow.md
```diff
@@ -0,0 +1,108 @@
+# Incremental Workflow
+
+This document defines the intended normal incremental workflow for semantic incremental analysis.
+
+## Scope
+
+This is the workflow once a baseline `analysis.json` already exists and includes a valid `metadata.commit_hash`.
```
ivanmilevtues: analysis.json already has a commit_hash and snapshot_hash
How are those 3 different commit_hashes different? Makes sure it is clear.

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+Bootstrap is outside this contract:
+- No baseline yet: run baseline creation first.
+
+## Invariants
+
+- Incremental uses commit-based git diffing against the saved baseline.
+- Structural updates happen deterministically before semantic patching.
+- Tree-sitter cosmetic filtering happens before any semantic tracer LLM call.
+- Uncertainty widens patch scope.
+- A normal incremental run never ends in `requires_full_analysis`, `scoped_reanalysis`, `reexpand`, or an equivalent fallback.
```
ivanmilevtues: This names should be changed.
What is scoped_reanalysis, reexpand?
Reexpand I get, but how isn't scopred_reanalysis the same? we should be precise in the spec as I think scoped_reanalysis and reexpand mean the same thing so idk why they are different

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Workflow
+
+1. Baseline check
+- If `analysis.json` and `metadata.commit_hash` exist, continue.
+- If no baseline exists, exit the normal incremental path and create the baseline first.
```
ivanmilevtues: ?
If there is no baseline, we do a full analysis no? What does it mean craete baseline first?

Baseline analysis I suppose is what we mean here

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+
+2. Git diff
+- Compare the saved baseline commit against the current target.
+- If the diff is empty, write updated incremental metadata and exit with `no_changes`.
+- If the diff is non-empty, continue.
```
ivanmilevtues: Do we care about this 2 cases, I think the logic should be such that they by default would result in nothing happening.

No need to have early stop braniching.

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+5. Trace seeding
+- Seed the semantic tracer from the remaining changed methods.
+- If an added file has no graph-connected methods, do not abort. Carry it forward for synthetic subtree scope inference.
+
+6. Semantic tracing
+- Explore upstream callers and downstream callees through CFG neighbors.
+- Continue until closure, frontier exhaustion, or budget exhaustion.
+- Track both:
+  - `visited_methods`: methods seen during trace; this defines semantic patch scope.
+  - `impacted_methods`: methods judged to materially affect documentation.
```
ivanmilevtues: - `impacted_methods`: methods judged to materially affect documentation.

Here we don't give to the LLM the documentation so it cannot know or am I confusing something?

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+- Advance the baseline commit and incremental metadata.
+
+## Terminal States
+
+Normal incremental should end in exactly one of these states:
+
+- `no_changes`
+- `cosmetic_only`
+- `patched_from_trace`
+- `patched_from_conservative_widening`
```
ivanmilevtues: These two are never mentioned before, and I don't exactly get them.

workflow.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Added files without CFG edges
+
+- These files may produce no semantic trace hits.
+- They still need semantic patching.
+- Their patch scope is inferred from the subtree that the nearest analyzed file belongs to.
+
+### Ambiguous nearest ownership
+
+- If equally near files disagree on ownership, widen to the lowest common ancestor component.
```
ivanmilevtues: I thought we are not doing ownership, as per your point of biasing towards a direction.

plan.md
```diff
@@ -0,0 +1,626 @@
...
+- `diagram_analysis/incremental/tracer.py`
+- the diagram-specific parts of `diagram_analysis/incremental/trace_planner.py`
+
+into `diagram_analysis/incremental_tracer.py`.
+
+The planner should be an implementation detail of the tracer, not a separate top-level subsystem.
+
+### 7. Move patching into `diagram_analysis/analysis_patcher.py`
+
+Move `agents/analysis_patcher.py` to `diagram_analysis/analysis_patcher.py`.
```
ivanmilevtues: ? this is strictly incremental no?

plan.md
```diff
@@ -0,0 +1,626 @@
...
+
+This code patches `AnalysisInsights`, so it belongs to the diagram domain.
+
+As part of the move:
+
+- remove the EASE + RFC 6902 JSON Patch design
+- replace it with a simpler structured replacement model
```
ivanmilevtues: it shoulnd't be replacement model, but rather general extractor model/patching model.

plan.md
```diff
@@ -0,0 +1,626 @@
...
+If all changed methods/files are filtered out as cosmetic:
+
+- the run still succeeds
+- structural state and baseline metadata are updated
+- the semantic patch set is empty
+- no fallback or escalation result is returned
+
+### 8. Build a deterministic ownership index
+
+Before patch scoping, build a method-aware ownership index over the loaded analysis.
```
ivanmilevtues: why this?

workflow.md
brovatten: baseline_commit is the only one saved into analysis.json , the others are runtime

workflow.md
brovatten: Agreed, we should not escalate to full here

workflow.md
brovatten: This reexpand should actually not happen, good point

plan.md
brovatten: To only give it a part of the analysis that's relevant, and not the whole thang

workflow.md
brovatten: Let's do both, per ur suggestion

plan.md
```diff
@@ -0,0 +1,626 @@
+# Incremental Semantic Tracer Plan
+
+## Goal
+
+Build the semantic incremental analysis flow on top of the current `main` / `iterative-llm-based` direction, using the rebased `new-iterative` work as source material, while reshaping it into a flatter architecture that matches the rest of the repository.
+
+The target behavior is:
+
+- Incremental analysis compares git commits using the existing commit-based model.
+- Structural changes are computed deterministically first.
```
Svilen-Stefanov: What is this referring to - @ivanmilevtues says this should be CFG,

PR #326, opened 2026-04-30
Read .codeboardingignore from repo root, drop initialize helper

brovatten:
## Release Notes

* **Refactor**
  * Consolidated ignore configuration file location to the repository root (`.codeboardingignore`), eliminating the need for a subdirectory structure and simplifying configuration management.
  * Automatic ignore initialization is no longer performed; configuration is now read directly from the repository root when present.


Svilen-Stefanov (changes requested):
I'm confused about these changes, why are we removing `initialize_codeboardingignore` here? This means that the `.codeboardingignore` will never be created and considered by an analysis e.g when a user runs the core with PYPI?

Also the tests seem to be "passing" because we now create the file in the tests and verify that it was created?

ivanmilevtues:
inactive.

PR #333, opened 2026-05-03
Simplify incremental analysis

Svilen-Stefanov:
## Summary by CodeRabbit

* **New Features**
  * Cluster-driven incremental analysis: preserve existing component IDs, track per-cluster members, and target re-detailing to changed components to reduce LLM calls.

* **Refactor**
  * Incremental workflow rewritten to use a cluster-delta baseline approach; CLI output simplified and baseline refs deprecated.

* **Chores**
  * Removed legacy trace/patch incremental systems and associated prompt templates.

* **Tests**
  * Added comprehensive cluster-delta and incremental-agent tests; removed older trace/patch test suites.

* **Documentation**
  * Added interactive summary of the new incremental architecture.


diagram_analysis/diagram_generator.py
```diff
@@ -187,13 +197,21 @@ def pre_analysis(self):
...
         def get_static_with_injected_analyzer() -> StaticAnalysisResults:
             cache_dir = None if self.force_full_analysis else get_cache_dir(self.repo_location)
-            return self._get_static_from_injected_analyzer(cache_dir, skip_cache=True)
+            # SHA-gated load is the new safety net (was: blanket skip_cache=True).
+            # ``CODEBOARDING_DISABLE_CACHE_REUSE=1`` is the post-deploy kill
+            # switch that reverts to the old always-recompute behaviour without
+            # a code change; useful if telemetry surfaces a regression.
+            disable_reuse = os.getenv("CODEBOARDING_DISABLE_CACHE_REUSE", "").lower() in ("1", "true", "yes")
+            skip_cache = self.force_full_analysis or disable_reuse
+            if disable_reuse:
```
Svilen-Stefanov: This looks strange, we only log something in this branch, we say `skipping static analysis cache` but we don't check `skip_cache` above

agents/prompts/incremental_grouping.py
```diff
@@ -0,0 +1,77 @@
...
+
+Why this is provider-agnostic: the incremental grouping prompt is a thinner
+restatement of ``CLUSTER_GROUPING_MESSAGE`` plus a context block listing the
+existing components. It's a routing decision, not a creative-writing task, so
+the variation across LLM families is small enough that a single template
+suffices. If a provider eventually needs a tuned variant, fold it back into
+``prompt_factory`` like the other prompts.
+"""
+
+INCREMENTAL_GROUPING_MESSAGE = """\
```
ivanmilevtues: I noticed that the prompt doesn't contain the <tags> for some of the models so it seems as it is generated for gemini? I asked Gemini itself and it said the following:

```This looks like a prompt originally built for GPT-4 (using standard LangChain/Pydantic templating) that was later run through Claude. The developer likely noticed Claude getting stuck in infinite loops or over-reading files, and patched the prompt by adding the explicit "Tool usage policy" to rein in Claude's over-exploration habits.```

You can ask CC to make the prompts for each model we already have comments with what is somewhat specific for each.

File: agents/agent.py [hunk 3]
```diff
@@ -405,3 +427,36 @@ def _try_parse(self, message_content, parser):
                 except:
                     pass
         raise ValueError(f"Couldn't parse {message_content}")
+
+    @staticmethod
+    def _direct_pydantic_parse(response: str, return_type: type[BaseModel]):
```
ivanmilevtues: The function above this one tries to make what you are doing here, it might be bad, but now I think we have the same thing twice, so either extend the current one, or just remove this one. The usage of this method is at wrong location I wrote a comment.

agents/agent.py
```diff
@@ -322,6 +336,14 @@ def _parse_response(self, prompt, response, return_type, max_retries=5, attempt=
         if response is None or response.strip() == "":
             logger.error(f"Empty response for prompt: {prompt}")
 
+        # Fast-path: try a no-LLM parse first. The agent often emits valid JSON
+        # for ``return_type`` directly (especially for tightly-shaped outputs
+        # like ClusterAnalysis); when it does, the trustcall extractor's
+        # second LLM round-trip is pure overhead.
+        direct = self._direct_pydantic_parse(response, return_type)
```
ivanmilevtues: I am pretty sure this is pointless, as we don't ask the LLMs for structured output and that is why we have the extractor logic here.

ivanmilevtues (commented):
First comments.

agents/agent_responses.py
```diff
@@ -264,19 +317,36 @@ def file_to_component(self) -> dict[str, str]:
...
+    if only_new:
+        used_indices: set[int] = set()
+        for component in analysis.components:
+            if not component.component_id:
+                continue
+            tail = component.component_id.split(".")[-1]
+            if tail.isdigit():
+                used_indices.add(int(tail))
+        next_idx = max(used_indices, default=0) + 1
+        for component in analysis.components:
```
ivanmilevtues, line 392: This is a consumer, I think there is no need to yield? can't we do a normal for loop, it's a minor comment but I am like wut the heck iterators in python ;d

agents/cluster_methods_mixin.py
```diff
@@ -703,7 +707,7 @@ def _log_node_coverage(self, analysis: AnalysisInsights, total_nodes: int) -> No
         pct = (assigned_nodes / total_nodes * 100) if total_nodes else 0
         logger.info(f"Node coverage: {assigned_nodes}/{total_nodes} ({pct:.1f}%) nodes assigned to components")
 
-    def _build_files_index(self, analysis: AnalysisInsights) -> dict[str, FileEntry]:
+    def build_files_index(self, analysis: AnalysisInsights) -> dict[str, FileEntry]:
```
ivanmilevtues, line 710: > why public :D?

agents/incremental_agent.py
```diff
@@ -0,0 +1,578 @@
...
+from diagram_analysis.io_utils import normalize_repo_path
+from monitoring import trace
+from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.constants import Language
+from static_analyzer.graph import ClusterResult
+
+logger = logging.getLogger(__name__)
+
+
+class IncrementalAgent(ClusterMethodsMixin, CodeBoardingAgent):
```
ivanmilevtues, line 50: Please follow the existing pattern of having a `run` method. Now I have no idea where to look :D.

agents/incremental_agent.py
```diff
@@ -0,0 +1,578 @@
...
+        self.meta_context = meta_context
+        self.prompts = {
+            "group_delta": PromptTemplate(
+                template=get_incremental_grouping_message(),
+                input_variables=["project_name", "meta_context", "project_type", "existing_components", "cfg_clusters"],
+            ),
+        }
+
+    @trace
+    def step_group_delta(
```
ivanmilevtues: aah I see this is the run method :D

agents/agent.py
```diff
@@ -39,7 +40,15 @@ def __init__(
         system_message: str,
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
+        tool_names: list[str] | None = None,
```
ivanmilevtues: This I don't like. Because now it is random flying string here and there.

A better solution is to override the self.agent. Instantiating agents is not heavy at all, so we just override it in the agents which need special tooling imo.

The abstraction of passing llms all the way up just doesn't scale anymore so we can override the agent in the children.

utils.py
```diff
@@ -32,9 +32,26 @@ def remove_temp_repo_folder(temp_path: str):
...
+    :func:`get_artifact_dir` instead.
+    """
     return repo_dir / CODEBOARDING_DIR_NAME / "cache"
 
 
+def get_artifact_dir(repo_dir: Path) -> Path:
+    """Sibling-of-``analysis.json`` directory for run-artifact files.
+
+    Run artifacts are the outputs of one successful ``analyze()`` invocation
+    (e.g. ``analysis.json``, ``static_analysis.pkl``). They survive across
```
ivanmilevtues, line 48: Both this and the method above mention static analysis.pkl. Now I am unsure if it is in cache or just in .codeboarding. Can we double check actually?

ivanmilevtues: From what I see it is here not in the cache. But to me this is the static analysis cached right? So probably it should be in the /cache directory.

codeboarding_cli/commands/incremental_analysis.py
```diff
@@ -99,39 +64,47 @@ def run_from_args(args: argparse.Namespace, parser: argparse.ArgumentParser) ->
...
+    _emit(
+        {
+            "mode": RunMode.INCREMENTAL,
+            "error": message,
+            "requiresFullAnalysis": True,
+        }
+    )
+
+
+def _emit(payload: dict[str, Any]) -> None:
```
ivanmilevtues, line 136: This should probably become something more general and not just for incremental and this file. 
I feel like we should put it in utils or somewhere. It is okay to stay here just mentioning.

codeboarding_workflows/analysis.py
```diff
@@ -128,13 +137,25 @@ def run_incremental(
     project_name: str,
```
ivanmilevtues, line 165: It is a bit weird that run_incremental is here and not incremental but it is okay :D

codeboarding_cli/commands/incremental_analysis.py
```diff
@@ -19,42 +18,22 @@ def add_arguments(subparsers: argparse._SubParsersAction, parents: list[argparse
...
     parser.add_argument(
         "--base-ref",
         type=str,
-        help=(
-            "Git ref (commit SHA, branch, or tag) to diff against when computing the incremental update. "
-            "Defaults to the commit that produced the last successful analysis "
-            "(recorded in <output-dir>/incremental_run_metadata.json); "
-            "if no prior run is recorded, incremental mode falls back to a full analysis."
-        ),
+        default=None,
```
ivanmilevtues, line 28: Can you run an incremental w/o base-ref I think not right?

codeboarding_workflows/analysis.py
```diff
@@ -128,13 +137,25 @@ def run_incremental(
...
     run_id: str,
     log_path: str,
-    base_ref: str,
-    target_ref: str,
     depth_level: int = 1,
     monitoring_enabled: bool = False,
     static_analyzer=None,
-) -> IncrementalRunPayload:
-    """Incremental scope — diff against *base_ref* and propagate only the semantic deltas."""
+    base_ref: str | None = None,
```
ivanmilevtues: I would make it non-nullable and set the default to `last_successful_commit(output_dir).`

(Sorry I hate nullable params while agents love em)

codeboarding_workflows/analysis.py
```diff
@@ -128,13 +137,25 @@ def run_incremental(
...
+    target_ref: str | None = None,
```
ivanmilevtues: Set default to empty str: ""

codeboarding_workflows/analysis.py
```diff
@@ -145,4 +166,18 @@ def run_incremental(
         monitoring_enabled=monitoring_enabled,
         static_analyzer=static_analyzer,
     )
-    return run_incremental_pipeline(generator, base_ref=base_ref, target_ref=target_ref)
+    generator.source_sha = source_sha
+
+    effective_base = base_ref if base_ref is not None else last_successful_commit(output_dir)
+    if effective_base is None:
```
ivanmilevtues: I feel like this gate should be done on entry on args, we try to get last_succ if note done

codeboarding_workflows/analysis.py
```diff
@@ -145,4 +166,18 @@ def run_incremental(
...
+        logger.info("No baseline ref available; running unscoped incremental.")
+        generator.changes = None
+    else:
+        changes = detect_changes(repo_path, effective_base, target_ref or "")
+        if changes.error:
+            logger.warning("detect_changes failed (%s); running unscoped incremental.", changes.error)
+            generator.changes = None
```
ivanmilevtues: Yea this because I am sure it is used as gate keeper down the line makes no sense here (imo :D)
Also we create the generator in this class why not have it as a contructor var which defaults to none?

codeboarding_workflows/incremental.py
```diff
@@ -1,61 +1,38 @@
...
+    1. If no prior ``analysis.json`` is present, run full analysis.
+    2. Otherwise hand the loaded baseline to ``generate_analysis_incremental``,
+       which itself falls back to a full run when the cluster snapshot is
+       missing or the cluster delta produces nothing actionable.
+    """
     output_dir = generator.output_dir
     existing = load_full_analysis(output_dir)
-    metadata = load_analysis_metadata(generator.output_dir)
+    metadata = load_analysis_metadata(output_dir)
     if existing is None or metadata is None:
```
ivanmilevtues: from api standpoint, to me we should just fail with a message pelase run full analysis.

In our own UI we should disable buttons and so on. But in my head codeboarding_pro --incremental (should do an incremental to existing one, otherwise it is bad usage and you should run full which has it's own parameter)

And this will simplify the propagation of Nones down to here just so we can return full analysis.

codeboarding_workflows/incremental.py
```diff
@@ -1,61 +1,38 @@
...
-        logger.info("Incremental workflow fell back to full analysis.")
-        return generator.generate_analysis()
-    if isinstance(payload, NoChangesPayload):
-        return payload.analysis_path.resolve()
-    if isinstance(payload, IncrementalCompletedPayload) and payload.result.analysis_path is not None:
-        return payload.result.analysis_path.resolve()
 
-    logger.info("Incremental workflow produced no analysis path; running full analysis.")
-    return generator.generate_analysis()
+__all__ = ["run_incremental_workflow", "ANALYSIS_FILENAME"]
```
ivanmilevtues: this i don't think we know?

codeboarding_workflows/incremental.py
```diff
@@ -1,61 +1,38 @@
...
 
-    base_ref = metadata.get("commit_hash", "")
-    if not base_ref:
-        logger.info("Baseline analysis is missing commit_hash; running full analysis.")
-        return generator.generate_analysis()
+    root_analysis, sub_analyses = existing
 
-    target_ref = get_git_commit_hash(generator.repo_location)
-    changes = detect_changes(generator.repo_location, base_ref, target_ref, raise_on_error=True)
+    if not root_analysis.components:
```
ivanmilevtues: Welp this is a weird edge case but fine :D

Is it incremental in that case idk, I still think it should fail and say analysis json is empty and you should regen full. (from UI perspective we should gatekeep you from click)

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +479,152 @@ def generate_analysis_incremental(
...
         sub_analyses: dict[str, AnalysisInsights],
-        base_ref: str,
-        changes,
     ) -> Path:
-        if self.static_analysis is None:
+        """Cluster-driven incremental update of an existing ``analysis.json``.
+
+        Deterministic cluster delta, one LLM call to route delta clusters,
+        then ``_generate_subcomponents`` seeded with the changed components.
+        Falls back to a full run when no baseline cluster info exists.
```
ivanmilevtues, line 547: the details agent is not used in this method.

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +479,152 @@ def generate_analysis_incremental(
...
             self.pre_analysis()
         assert self.static_analysis is not None
+        assert self.abstraction_agent is not None
 
-        ownership_index = build_ownership_index(root_analysis, sub_analyses)
-        added_files, modified_files, deleted_files, rename_map = normalize_changes_for_delta(changes)
-        methods_by_file = self._collect_method_entries_from_static_analysis()
+        monitor = self.stats_writer if self.stats_writer else nullcontext()
+        with monitor:
+            # Scrub before cluster math: orphan-routed files never appear in
```
ivanmilevtues, line 556: this deserves its own method

diagram_analysis/diagram_generator.py
```diff
@@ -178,14 +213,20 @@ def pre_analysis(self):
...
+            # change; useful if telemetry surfaces a warm-start regression.
+            disable_reuse = os.getenv("CODEBOARDING_DISABLE_CACHE_REUSE", "").lower() in ("1", "true", "yes")
+            skip_cache = self.force_full_analysis or disable_reuse
+            if disable_reuse:
+                logger.info("CODEBOARDING_DISABLE_CACHE_REUSE set; skipping static analysis cache")
+            return self._get_static_from_injected_analyzer(skip_cache=skip_cache, source_sha=self.source_sha)
 
         def get_static_with_new_analyzer() -> StaticAnalysisResults:
             skip_cache = self.force_full_analysis
             if skip_cache:
```
ivanmilevtues, line 279: The two methods above I would put outside of this method, because it is getting quite lengthy with a self liek the self._get_static

diagram_analysis/diagram_generator.py
```diff
@@ -178,14 +213,20 @@ def pre_analysis(self):
         self._monitoring_agents["MetaAgent"] = self.meta_agent
 
         def get_static_with_injected_analyzer() -> StaticAnalysisResults:
-            cache_dir = None if self.force_full_analysis else get_cache_dir(self.repo_location)
-            return self._get_static_from_injected_analyzer(cache_dir, skip_cache=True)
+            # ``CODEBOARDING_DISABLE_CACHE_REUSE=1`` is the post-deploy kill
+            # switch that reverts to "always re-LSP everything" without a code
```
ivanmilevtues, line 267: I've never seen this sys variable, i think it is dead code.

agents/incremental_agent.py
```diff
@@ -0,0 +1,577 @@
...
+        seen.add(method.qualified_name)
+        out.append(method)
+    return out
+
+
+# ---------------------------------------------------------------------------
+# Prune (deterministic, no LLM)
+# ---------------------------------------------------------------------------
+def scrub_deleted_files(
+    root_analysis: AnalysisInsights,
```
ivanmilevtues, line 714:
```suggestion
def remove_deleted_files(
```

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +478,141 @@ def generate_analysis_incremental(
...
+            # any cluster, so deletes wouldn't surface via the delta alone.
+            live_files: set[str] = set()
+            for language in self.static_analysis.get_languages():
+                try:
+                    cfg = self.static_analysis.get_cfg(language)
+                except (ValueError, KeyError):
+                    continue
+                for node in cfg.nodes.values():
+                    if node.file_path:
+                        live_files.add(normalize_repo_path(node.file_path, self.repo_location))
```
ivanmilevtues, line 566:
```suggestion
            remove_deleted_files(root_analysis, sub_analyses, live_files)
```

diagram_analysis/diagram_generator.py
```diff
@@ -10,39 +10,42 @@
...
 from agents.abstraction_agent import AbstractionAgent
-from agents.agent_responses import AnalysisInsights, Component, MethodEntry
+from agents.agent_responses import AnalysisInsights, Component, MetaAnalysisInsights, MethodEntry
 from agents.details_agent import DetailsAgent
-from agents.llm_config import MONITORING_CALLBACK, initialize_llms
+from agents.incremental_agent import (
+    IncrementalAgent,
+    prune_empty_components,
+    repopulate_touched_scopes,
+    scrub_deleted_files,
```
ivanmilevtues:
```suggestion
    remove_deleted_files,
```

diagram_analysis/diagram_generator.py
```diff
@@ -71,13 +74,26 @@ def __init__(
         self.log_path = log_path
         self.monitoring_enabled = monitoring_enabled
         self.force_full_analysis = False  # Set to True to skip incremental updates
+        # Optional source-tree changeset for the iterative path. When set,
+        # the cluster delta drops drift qnames whose file is outside the
+        # diff AND outside the prior analysis (see ``compute_cluster_delta``).
+        # Set externally before calling ``generate_analysis_incremental``;
+        # the workflow function ``run_incremental_workflow`` doesn't need
+        # to grow a parameter for it.
```
ivanmilevtues: This is super hidden populate I didn't get it for a minute or more. As I mentioned I feel like it should be set via cntr, so it is obvious where it is set.

diagram_analysis/cluster_delta.py
```diff
@@ -0,0 +1,452 @@
...
+    """Compute per-language cluster deltas via seeded Leiden.
+
+    When ``changes`` is provided, qnames whose file is outside both the diff
+    and the prior analysis are dropped as drift. Qnames in the prior analysis
+    that vanish without appearing in the diff are kept but logged as
+    inconsistent. ``repo_dir`` normalizes CFG-absolute paths to repo-relative
+    posix so they match the diff. ``changes=None`` disables scoping.
+    """
+    delta = ClusterDelta()
+    diff_files = _changeset_to_path_set(changes) if changes is not None else None
```
ivanmilevtues, line 84: I hate nullables 💀
Can we make so that the ChangeSet has just empty files on creation instead of being None.

Conceptually makes so much more sense to me as well. Empty ChangeSet means no changes, None ChangeSet means unset changeSet to me.

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+) -> str:
+    """Resolve which file in *files_for_cluster* a particular qname lives in.
+
+    Prefers a dotted-path substring match (longest wins for specificity).
+    Falls back to any file the qname is associated with, then to the first
+    file in the cluster.
+    """
+    dotted = lambda fp: fp.replace("/", ".").rsplit(".", 1)[0]
+    matches = [fp for fp in files_for_cluster if dotted(fp) and dotted(fp) in qname]
+    if matches:
```
ivanmilevtues, line 691: This is a class method, put it in Component.

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+        return max(matches, key=lambda fp: len(dotted(fp)))
+    other = qname_to_files.get(qname, set())
+    if other:
```
ivanmilevtues, line 694: This is also a class method, add it in AnalysisInsights

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +479,141 @@ def generate_analysis_incremental(
...
+
+            delta = compute_cluster_delta(
+                old_snapshot,
+                self.static_analysis,
+                changes=self.changes,
+                repo_dir=self.repo_location,
+            )
+            if not delta.has_changes:
+                logger.info("Cluster delta is empty; rewriting current analysis without re-detailing.")
+                prune_empty_components(root_analysis, sub_analyses)
```
ivanmilevtues, line 591: Teh prune_empty_compenents will remove also edges so in the case of:
A -> B -> C
     ^
    remove this one

we end up with A C

But it might be A -> C. And we can rewire it based on the CFG, similar to the logic we have which add the 'calls' label

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+        }
+
+    @trace
+    def run(
+        self,
+        delta: ClusterDelta,
+        root_analysis: AnalysisInsights,
+        sub_analyses: dict[str, AnalysisInsights],
+    ) -> ClusterAnalysis:
+        """LLM call: route delta clusters to existing or new components."""
```
ivanmilevtues: I would rather keep it empty the meta_context_str.

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+        return result
+
+
+def _format_existing_components(
+    root_analysis: AnalysisInsights,
+    sub_analyses: dict[str, AnalysisInsights],
+    affected_files: set[str] | None = None,
+) -> str:
+    """Render existing components for the incremental grouping prompt.
+
```
ivanmilevtues, line 250: This should never happen so we can assert instead?

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+        (affected if owns_relevant else other).append(component)
+
+    sections: list[str] = []
+    if affected:
+        sections.append("### Affected components (full info)")
+        sections.extend(_format_component_line(c, include_description=True) for c in affected)
+    if other:
+        sections.append("\n### Other components (routing targets, names only)")
+        sections.extend(_format_component_line(c, include_description=False) for c in other)
+    return "\n".join(sections)
```
ivanmilevtues, line 275: Generally speaking LLMs are not too good with ids, we can leave them I think it is fine but it is something to keep in mind

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+    desc = (component.description or "").strip().replace("\n", " ")
+    return f'- {cid} "{component.name}" -- {desc}'
+
+
+# ---------------------------------------------------------------------------
+# Stitching (deterministic, no LLM)
+# ---------------------------------------------------------------------------
+class Verdict(StrEnum):
+    """Derived stitching verdict — not part of the LLM schema.
+
```
ivanmilevtues, line 291: decision #4?

agents/agent_responses.py
```diff
@@ -114,6 +114,39 @@ class ClustersComponent(LLMBaseModel):
     description: str = Field(
         description="Explanation of what this component does, its main flow, WHY these clusters are grouped together, how it interacts with other cluster groups, and the most important classes/methods (by their exact qualified names from the clusters)"
     )
+    existing_component_id: str | None = Field(
```
ivanmilevtues, line 192: So at the moment I feel like we have two things for no good reason.

I just now understand the thing, so I am thinking shouldn't we make the LLM guess the id always?
I.e. it seems to me like a good way to make the llm autoassign the level of change with a single simple id. 

I.e. if the component id is 2.4 that means parent is 2 <- and then if the .4 exists it is existing compoennt if it doesn't it is obv a new component.

I think it is just confusing to have the two ids if one can do the job. Or maybe it is better idk just my take on this, as I am usnure if parent_id is well explained for the LLM.

agents/incremental_agent.py
```diff
@@ -0,0 +1,656 @@
...
+        return root_analysis
+    return sub_analyses.setdefault(
+        parent_id,
+        AnalysisInsights(description="", components=[], components_relations=[]),
+    )
+
+
+def _parent_id_for_scope(
+    analysis: AnalysisInsights,
+    root_analysis: AnalysisInsights,
```
ivanmilevtues, line 541: probably just path is enoguh

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +481,147 @@ def generate_analysis_incremental(
...
+            delta_cluster_analysis = incremental_agent.run(delta, root_analysis, sub_analyses)
+
+            redetail_ids = stitch_delta(root_analysis, sub_analyses, delta_cluster_analysis, delta)
+
+            # Refresh first (per-component, siblings untouched), then prune —
+            # we only know a component is empty after rebuilding from live CFG.
+            repopulate_touched_scopes(
+                redetail_ids,
+                root_analysis,
+                sub_analyses,
```
ivanmilevtues, line 625: This I am not sure we want to run? This seems like generating extra layers or not enough layers?

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +481,147 @@ def generate_analysis_incremental(
...
-            parsing_llm,
-            parsed_diff=getattr(changes, "parsed_diff", None),
-            callbacks=callbacks,
-        )
+            removed_ids = prune_empty_components(root_analysis, sub_analyses)
+
+            if removed_ids:
+                redetail_ids -= removed_ids
+
+            redetail_components = _collect_components_by_id(redetail_ids, root_analysis, sub_analyses)
```
ivanmilevtues, line 634: It feels like these should be steps for the incremental_agent rather than steps in the diagram generator.

ivanmilevtues: It is also a cluster mixin agent so it has those methods

diagram_analysis/diagram_generator.py
```diff
@@ -448,70 +481,147 @@ def generate_analysis_incremental(
...
+                sub.files = self.abstraction_agent.build_files_index(sub)
+            unified_files = self.abstraction_agent.build_files_index(root_analysis)
+            for sub in sub_analyses.values():
+                for fp, entry in sub.files.items():
+                    unified_files.setdefault(fp, entry)
+            root_analysis.files = unified_files
+
+            commit_hash = get_git_commit_hash(self.repo_location)
+            analysis_path = save_analysis(
+                analysis=root_analysis,
```
ivanmilevtues, line 665: This method in general probably, like the other analysis methods needs to be shorten (later I suppose)

agents/agent.py
```diff
@@ -319,69 +331,46 @@ def _validation_invoke(
...
+        ignored — the local deterministic attempts can't change on retry, and
+        the LLM repair manages its own retry policy.
+        """
+        del max_retries, attempt  # signature kept for backwards-compat
+
         if response is None or response.strip() == "":
             logger.error(f"Empty response for prompt: {prompt}")
+            raise ValueError(f"Empty response for {return_type.__name__}")
 
-        def call_once():
```
ivanmilevtues, line 325: revert please

agents/incremental_agent.py
ivanmilevtues: this is done

PR #334, opened 2026-05-04
Fix typescript for multiconfigs

brovatten:
On a cold-started analyzer (no prior pkl), is either empty or populated by an in-process call to cfg.cluster() during pre_analysis (the cohesion health check). Either way it has no historical meaning, so cluster_delta compares the live partition against itself and trivially short-circuits the user's edits get silently absorbed without any re-detailing.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -142,6 +142,13 @@ class StaticAnalyzer:
         self.collected_diagnostics: dict[Language, FileDiagnosticsMap] = {}
         self._clients_started: bool = False
         self._cached_results: StaticAnalysisResults | None = None
+        # True iff the most recent ``analyze()`` loaded a prior pkl. Read by
+        # ``DiagramGenerator.generate_analysis_incremental`` to decide whether
+        # ``CallGraph._cluster_cache`` is meaningful as an "old snapshot": on
+        # a cold start the cache only holds clusters freshly computed from
+        # the live graph (e.g. by the cohesion health check), so cluster_delta
+        # would compare new-vs-new and trivially short-circuit.
+        self.warm_started: bool = False
 
     def __enter__(self) -> "StaticAnalyzer":
         self.start_clients()
@@ -461,11 +468,13 @@ class StaticAnalyzer:
         if skip_cache:
             logger.info("static_analysis_cache: outcome=bypass (skip_cache=True)")
             results = self._run_full_lsp_pass()
+            self.warm_started = False
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
                 logger.info("static_analysis_cache: outcome=miss_absent")
                 results = self._run_full_lsp_pass()
+                self.warm_started = False
             else:
                 cached_results, cached_sha = warm_start
                 logger.info(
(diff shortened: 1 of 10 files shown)


tests/integration/fixtures/real_projects/excalidraw_typescript.json
```diff
@@ -3,47 +3,36 @@
...
+      "Windows": 13061
     },
-    "packages_count": 56,
-    "call_graph_nodes": 4891,
-    "call_graph_edges": 12309,
-    "source_files_count": 2035,
+    "packages_count": 58,
+    "call_graph_nodes": 4902,
+    "call_graph_edges": 12407,
+    "source_files_count": 414,
```
Svilen-Stefanov, line 18: Seems like everything is in the expected ranges except the files we got cut 5x approximately.

static_analyzer/__init__.py
```diff
@@ -32,13 +32,22 @@ def _create_engine_configs(
     programming_languages: list[ProgrammingLanguage],
     repository_path: Path,
     ignore_manager: RepoIgnoreManager,
-) -> list[tuple[LanguageAdapter, Path]]:
+) -> tuple[list[tuple[LanguageAdapter, Path]], dict[Path, list[Path]]]:
```
Svilen-Stefanov: The return type here is obnoxious. :D

static_analyzer/typescript_config_scanner.py
```diff
@@ -1,54 +1,314 @@
...
+    filesystem fallback if tsc is unavailable). The engine should send exactly
+    these to the LSP client rooted at ``root`` — anything broader would re-claim
+    files that belong to a sibling project.
     """
 
+    root: Path
+    files: list[Path] = field(default_factory=list)
+
+
+_TS_EXTENSIONS = (".ts", ".tsx", ".mts", ".cts")
```
Svilen-Stefanov: We already have these in `LANGUAGE_EXTENSIONS`, we should reuse that instead of defining them again.

static_analyzer/typescript_config_scanner.py
```diff
@@ -1,54 +1,314 @@
...
+    if system_tsc is not None:
+        return [system_tsc, "--showConfig"]
+    return None
+
+
+def _resolve_system_tsc() -> str | None:
+    """Find a usable ``tsc`` binary on the system PATH.
+
+    Why ``shutil.which`` rather than hand-rolling the loop: on Windows the
+    binary may be ``tsc.cmd`` (npm-global), ``tsc.exe`` (some chocolatey/scoop
```
Svilen-Stefanov: This has grown a bit, let me see if we need it all or if we can somehow simplify this.

Svilen-Stefanov (commented):
I realised these things are not triggering the integration tests even though they should. Let's fix that

tests/integration/fixtures/real_projects/excalidraw_typescript.json
Svilen-Stefanov: I confirmed the new approximation makes much more sense, we actually count 404 not 414 files and that is pretty much the number of files I can see after the ignored patterns too.

PR #335, opened 2026-05-04
Add name-stability arbiter for incremental cluster naming

brovatten:


brovatten:
This was meant to keep the stability when incremental analysis was more llm-driven than it is now. Closing

PR #336, opened 2026-05-12
Update parsing model to gemini-3.1-flash-lite

Svilen-Stefanov:
## Summary by CodeRabbit

* **Chores**
  * Updated default model identifiers for Google, GLM, and OpenRouter providers.

* **Tests**
  * Added integration tests that verify configured provider models against public model catalogs (with caching and graceful skipping when catalogs are unreachable) and a guard test to ensure catalogs are reachable.
  * Updated live-model smoke-test datasets to use newer Google model entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CodeBoarding/CodeBoarding/pull/336?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -122,7 +122,7 @@ LLM_PROVIDERS = {
     "vercel": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="VERCEL_API_KEY",
-        agent_model="google/gemini-3-flash-preview",
+        agent_model="google/gemini-3-flash",
         parsing_model="openai/gpt-5-mini",
         llm_type=LLMType.GEMINI_FLASH,
         alt_env_vars=["VERCEL_BASE_URL"],
@@ -210,8 +210,8 @@ LLM_PROVIDERS = {
     "glm": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="GLM_API_KEY",
-        agent_model="glm-4-flash",
-        parsing_model="glm-4-flash",
+        agent_model="glm-4.7-flash",
+        parsing_model="glm-4.7-flash",
         llm_type=LLMType.GLM,
         alt_env_vars=["GLM_BASE_URL"],
         extra_args={
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -122,7 +122,7 @@ LLM_PROVIDERS = {
     "vercel": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="VERCEL_API_KEY",
-        agent_model="google/gemini-3-flash",
+        agent_model="google/gemini-3-flash-preview",
         parsing_model="openai/gpt-5-mini",
         llm_type=LLMType.GEMINI_FLASH,
         alt_env_vars=["VERCEL_BASE_URL"],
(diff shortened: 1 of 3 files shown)


ivanmilevtues (approved):
I think it looks good, we should just make sure that vercel and open router also switch the names as well. It could be the case that they switch names internally but not on the api level.

agents/llm_config.py
```diff
@@ -238,8 +238,8 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
     "openrouter": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="OPENROUTER_API_KEY",
-        agent_model="google/gemini-3-flash",
-        parsing_model="google/gemini-3.1-flash-lite-preview",
+        agent_model="google/gemini-3-flash-preview",
```
brovatten, line 241: upgrade this to flash 3.5 - i think it's released for us now

agents/llm_config.py
```diff
@@ -210,8 +210,8 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
     "glm": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="GLM_API_KEY",
-        agent_model="glm-4-flash",
-        parsing_model="glm-4-flash",
+        agent_model="glm-4.7-flash",
+        parsing_model="glm-4.7-flash",
```
brovatten, line 214: if you use glm as a provider, 5.1 is better i think (check this). they are all quite inexpensive

agents/llm_config.py
```diff
@@ -238,8 +238,8 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
     "openrouter": LLMConfig(
         chat_class=ChatOpenAI,
         api_key_env="OPENROUTER_API_KEY",
-        agent_model="google/gemini-3-flash",
-        parsing_model="google/gemini-3.1-flash-lite-preview",
+        agent_model="google/gemini-3-flash-preview",
+        parsing_model="google/gemini-3.1-flash-lite",
```
brovatten, line 242: u mentioned 3.1 was being phased out, for what? mby we should change this in advance if they already released replacement. or just know when

tests/integration/test_provider_models_live.py
```diff
@@ -0,0 +1,113 @@
...
+_UA = {"User-Agent": "codeboarding-tests"}  # models.dev 403s without a User-Agent.
+
+# LLM_PROVIDERS provider name -> models.dev provider key(s) whose model sets are unioned.
+_MODELSDEV_KEYS = {
+    "openai": ["openai"],
+    "anthropic": ["anthropic"],
+    "google": ["google"],
+    "cerebras": ["cerebras"],
+    "deepseek": ["deepseek"],
+    "glm": ["zhipuai"],
```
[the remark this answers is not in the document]
brovatten: did u check this comment?

brovatten (commented):
Thanks for doing this. Check my comment here

agents/llm_config.py
Svilen-Stefanov: Not the lite, the lite-preview is being phased out.

tests/integration/test_provider_models_live.py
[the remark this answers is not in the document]
Svilen-Stefanov: I had forgotten to push my latest change. It's addressed now.

agents/llm_config.py
Svilen-Stefanov: We just chatted, the reason we want to keep 3-flash is latency and cost.

PR #345, opened 2026-05-29
feat(config): support openai_base_url for self-hosted / OpenAI-compatible providers

scubamount:
Adds an `openai_base_url` field to the `[provider]` section of `config.toml`, so users who point the OpenAI client at a self-hosted or OpenAI-compatible endpoint (local inference server, gateway, or proxy) can configure it in one place instead of relying on an exported `OPENAI_BASE_URL` shell variable.

## Why

Today the OpenAI base URL can only be set through the environment. Projects that standardize on `config.toml` for provider/model settings have to special-case this one value via shell wrappers or exported env vars, which is easy to get wrong and inconsistent with how other provider settings are handled.

## What

- `user_config.py`: new `openai_base_url` field on `ProviderUserConfig`, loader wiring, and an entry in `_PROVIDER_KEY_TO_ENV` mapping it to the canonical `OPENAI_BASE_URL` env var.
- Commented template line in the generated `config.toml`.
- This mirrors the existing `ollama_base_url` handling exactly.

## Precedence

`apply_to_env()` does not overwrite an already-set environment variable, so an explicit `OPENAI_BASE_URL` in the shell still wins over the config value. Behavior is unchanged when the field is unset.

## Tests

- `apply_to_env` injects the configured base URL when no env var is set.
- An existing shell `OPENAI_BASE_URL` takes precedence.
- Loading from `config.toml` populates the field; absent -> `None`.

## Verification

Manually verified end-to-end against a local OpenAI-compatible server: with no `OPENAI_BASE_URL` set in the environment, requests routed to the base URL configured in `config.toml`.

---
This PR is independent of and can be merged before or after the companion timeouts PR; the only overlap is one adjacent commented template line in `config.toml`.

## Summary by CodeRabbit

* **New Features**
  * Users can set a custom OpenAI base URL in settings; environment variables (including shell) take precedence.
  * If a base URL is used without an API key, the app treats it as a keyless local endpoint and emits a warning (multiple-provider key conflicts still raise an error).

* **Tests**
  * Added comprehensive tests covering base-URL handling, precedence rules, and keyless-endpoint warning behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CodeBoarding/CodeBoarding/pull/345?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -85,9 +85,25 @@ class LLMConfig:
     parsing_temperature: float = LLMDefaults.DEFAULT_PARSING_TEMPERATURE
     extra_args: dict[str, Any] = field(default_factory=dict)
     alt_env_vars: list[str] = field(default_factory=list)
+    keyless_capable: bool = False
+    """Whether this provider can run without a real API key.
+
+    True for self-hosted / OpenAI-compatible endpoints where ``api_key_env`` (or
+    an ``alt_env_vars`` entry) is really a base-URL existence check rather than a
+    secret. When such a provider is the sole active one and no real key is set,
+    key validation warns instead of failing, and the client uses a placeholder.
+    """
 
     def get_api_key(self) -> str | None:
         return os.getenv(self.api_key_env)
+
+    def has_real_api_key(self) -> bool:
+        """True if the provider's primary API-key env var holds a value.
+
+        Distinct from ``is_active()``: a keyless-capable provider can be active
+        via a base-URL ``alt_env_vars`` entry while having no real key here.
+        """
+        return bool(os.getenv(self.api_key_env))
 
     def is_active(self) -> bool:
         """Check if any of the environment variables (primary or alternate) are set."""
@@ -113,6 +129,7 @@ LLM_PROVIDERS = {
         parsing_model="gpt-4o-mini",
         llm_type=LLMType.GPT4,
         alt_env_vars=["OPENAI_BASE_URL"],
+        keyless_capable=True,
         extra_args={
             "base_url": lambda: os.getenv("OPENAI_BASE_URL"),
             "max_tokens": None,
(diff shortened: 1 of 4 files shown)


brovatten (approved):
LGTM! Thanks for noticing this miss

ivanmilevtues (changes requested):
Thank you for the submission @scubamount. It is almost done the one thing left is to modifty `agents/llm_config.py`. At the moment if you are running a local model without API key it will fail because of no API key provided, to fix this we need to skip open_AI_key validation, we can just log a wanrning in that case.
```python
diff --git a/agents/llm_config.py b/agents/llm_config.py
index 83c5b17..c0829cf 100644
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -314,6 +314,8 @@ class LLMConfigError(ValueError):
 
 def validate_api_key_provided() -> None:
     """Raise LLMConfigError if zero or more than one LLM provider key is configured."""
+    if os.getenv("OPENAI_BASE_URL"):
+        if not os.getenv("OPENAI_API_KEY"):
+            logger.warning("OPENAI_BASE_URL is set without OPENAI_API_KEY — using 'no-key-required' as fallback")
+        return
     active = [name for name, config in LLM_PROVIDERS.items() if config.is_active()]
     if not active:
         required = sorted({config.api_key_env for config in LLM_PROVIDERS.values()})
```

scubamount:
thanks @ivanmilevtues! pushed a version based on your diff. it works and it's ready. i did try a slightly different shape though. let me know which you want, and i'll revert to your exact diff if you'd rather keep it minimal.

what i changed:

1. made "keyless" a flag on the provider (`keyless_capable: bool`) instead of an `OPENAI_*` check, and set it on openai since it already activates on `OPENAI_BASE_URL` via `alt_env_vars`. the idea was that any openai-compatible backend could reuse it. openai's the only one today though, so if that feels like over-engineering, your inline version is leaner and i'm fine switching back.

2. kept the multi-provider ambiguity check. this is the one real difference. your early `return` treats base_url as "this is my provider, done", which is reasonable. i went the other way: warn only when the single active provider is keyless with no key, so base_url plus a stray second key still raises instead of silently picking openai. depends what you want base_url to mean. fine to go your way here too.

3. small thing either way: the warning text "using 'no-key-required' as fallback". that fallback actually happens downstream at client construction (`api_key or "no-key-required"`), not in the validator, so i reworded it to just say we're treating it as a keyless local endpoint. matches where the fallback lives now.

added tests for the keyless paths and verified end to end against a local openai-compatible server with no key set. if you want the minimal version, i'll force-push your diff with just the warning-text fix.

agents/llm_config.py
```diff
@@ -113,6 +129,7 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
         parsing_model="gpt-4o-mini",
         llm_type=LLMType.GPT4,
         alt_env_vars=["OPENAI_BASE_URL"],
+        keyless_capable=True,
```
ivanmilevtues, line 132: Nice.

ivanmilevtues (approved):
I like your solution better. Looks good to me we can merge it!

PR #346, opened 2026-05-29
feat(llm): make agent and parsing timeouts configurable via config.toml

scubamount:
Adds two optional `[llm]` settings so users running slower or self-hosted backends can raise the timeouts without editing source.

## Why

The agent invoke timeout and the parsing model client timeout are currently hard-coded. On slower or local inference backends, large prompts (e.g. the cluster-grouping step) can exceed the built-in limit and fail, with no supported way to extend it short of patching the installed package.

While wiring this up I also noticed the parsing client was being constructed with **no timeout at all**, so a hung upstream could block indefinitely. This PR makes that an explicit, configurable value.

## What

- `agent_timeout_s` (optional): thread-join timeout applied to each `agent.invoke()` attempt.
- `parsing_timeout_s` (optional): HTTP client timeout for the parsing model.

Both thread through `configure_models()` into module-level settings read by `get_agent_timeout()` / `get_parsing_timeout()`. `agents/agent.py` and the CLI bootstrap use the accessors.

## Backward compatibility

Leaving both unset preserves prior behavior: the agent keeps its existing default, and the parsing client keeps its previous (no-timeout) construction — the timeout is applied only when explicitly configured.

## Files

- `user_config.py`: new fields + loader wiring + template comments.
- `agents/llm_config.py`: module globals, `configure_models` params, accessors, parsing-client timeout injection.
- `agents/agent.py`: use `get_agent_timeout(attempt)`.
- `codeboarding_cli/bootstrap.py`: pass configured timeouts through.
- tests: load/default coverage, accessor behavior, and parsing-client injection.

## Verification

Manually verified end-to-end against a slow local backend: `agent.invoke()` logs the configured timeout and the parsing client is constructed with it. Existing behavior confirmed unchanged when the settings are absent.

---
Independent of the companion `openai_base_url` PR; the only overlap is one adjacent commented template line in `config.toml`.

## Summary by CodeRabbit

* **New Features**
  * Configuration file now supports timeout settings for agent and parsing operations, allowing users to customize how long these operations can run before timing out.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CodeBoarding/CodeBoarding/pull/346?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -30,12 +30,22 @@ logger = logging.getLogger(__name__)
 # ---------------------------------------------------------------------------
 _agent_model_override: str | None = None
 _parsing_model_override: str | None = None
+_agent_timeout_override: int | None = None
+_parsing_timeout_override: int | None = None
+
+# Default thread-join timeouts for the agent invoke loop (seconds).
+# First attempt is shorter; retries get the longer budget. Overridden as a
+# single flat value when agent_timeout_s is configured (see get_agent_timeout).
+_DEFAULT_AGENT_TIMEOUT_FIRST = 300
+_DEFAULT_AGENT_TIMEOUT_RETRY = 600
 
 
 def configure_models(
     agent_model: str | None = None,
     parsing_model: str | None = None,
     api_keys: dict[str, str] | None = None,
+    agent_timeout_s: int | None = None,
+    parsing_timeout_s: int | None = None,
 ) -> None:
     """Set process-wide model and provider overrides.  Call this once at startup.
 
@@ -46,6 +56,12 @@ def configure_models(
     Keys already present in the shell environment are never overwritten, so
     CI/CD pipelines that export keys directly retain full control.
 
+    ``agent_timeout_s`` overrides the agent invoke thread-join timeout (a single
+    flat value for every attempt). ``parsing_timeout_s`` overrides the parsing
+    model's HTTP client timeout. Both are optional; when unset the built-in
+    defaults apply. Useful for slow self-hosted / local backends whose
+    long-context completions exceed the default budgets.
+
     Priority (highest to lowest):
       1. Shell environment variables (set before the process starts)
       2. ``api_keys`` passed here  /  values from ~/.codeboarding/config.toml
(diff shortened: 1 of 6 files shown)


agents/llm_config.py
```diff
@@ -30,12 +30,22 @@
 # ---------------------------------------------------------------------------
 _agent_model_override: str | None = None
 _parsing_model_override: str | None = None
+_agent_timeout_override: int | None = None
+_parsing_timeout_override: int | None = None
+
+# Default thread-join timeouts for the agent invoke loop (seconds).
+# First attempt is shorter; retries get the longer budget. Overridden as a
+# single flat value when agent_timeout_s is configured (see get_agent_timeout).
+_DEFAULT_AGENT_TIMEOUT_FIRST = 300
```
Svilen-Stefanov, line 39: Let's move these to `agents/constants.py`.

agents/llm_config.py
```diff
@@ -46,21 +56,46 @@ def configure_models(
...
+
+    When ``agent_timeout_s`` was configured, return it flat for every attempt
+    (B1a). Otherwise fall back to the default first/retry split.
+    """
+    if _agent_timeout_override is not None:
+        return _agent_timeout_override
+    return _DEFAULT_AGENT_TIMEOUT_FIRST if attempt == 0 else _DEFAULT_AGENT_TIMEOUT_RETRY
+
+
+def get_parsing_timeout() -> int | None:
```
Svilen-Stefanov, line 94: I think this is the main blocker for me at the moment. When the user doesn't set an overwrite, no timeout kwarg is passed and the parsing agent could hang forever.

Instead of preserving the bug and making it more explicit, it would be great to fix it here directly and handle the parsing timeouts in a similar way we handle the agent ones.

agents/llm_config.py
```diff
@@ -30,12 +30,22 @@
 # ---------------------------------------------------------------------------
 _agent_model_override: str | None = None
 _parsing_model_override: str | None = None
+_agent_timeout_override: int | None = None
```
Svilen-Stefanov, line 33: It would be great to update the README with these values as well to help folks understand this is configurable.

Svilen-Stefanov (changes requested):
Thank you @scubamount for the great suggestion. I made 3 minor suggestions, the most relevant being the parsing model timeout handling. After they are addreessed, I think we can merge this.

PR #350, opened 2026-05-31
fix(installer): SHA-verify existing native binaries against the registry pin

brovatten:
## Summary

`install_native_tools` had a `if binary_path.exists(): skip` shortcut that treated any existing file at `binary_path` as healthy, so a user who installed CodeBoarding before a tools-repo re-cut kept their stale binary indefinitely — pin bumps in `tool_registry/registry.py` were silently ignored.

I noticed this empirically: my own \`~/.codeboarding/servers/bin/macos/tokei\` is sha256 \`1a43d9a3…\` and 3.1 MB, but the currently-pinned binary at the same release URL is sha256 \`90ae8a2e…\` and 4.1 MB. Both are tokei 12.1.2, but they're different builds — the `tools-2026.04.05` release got re-cut after my install, and the installer never noticed.

## What changes

- Extract the hash-lookup precedence (asset-name → platform-suffix → None) into `_resolve_native_expected_hash` so it can be called twice: once before the skip, once inside the download path.
- Before the skip-if-installed shortcut, compute the on-disk SHA256 and compare to the pin. On mismatch, delete and re-fetch.
- On read failure (permissions, FS error), reinstall defensively rather than report a broken file as healthy.
- Compressed assets (rust-analyzer) keep the legacy skip — their pin is over the archive, not the decompressed binary.

## What this doesn't touch

The same drift bug exists in `install_archive_tool` (JDTLS, clangd) and `install_package_manager_tools` (csharp-ls), but they need different verification strategies (marker-file or `--version` checks). Deferred to follow-up issues so this PR stays focused.

## Test plan

- [x] New `TestInstallNativeToolsPinDrift` class — 4 tests:
  - matching SHA skips download
  - drifted SHA triggers reinstall with new bytes on disk
  - no-pin case preserves legacy skip behavior
  - unreadable existing binary triggers defensive reinstall
- [x] Full `tests/test_tool_registry.py` + `tests/test_registry_coverage.py` sweep — 120/120 passing, no regressions.

## Summary by CodeRabbit

* **New Features**
  * Installer now verifies native tool binaries using SHA256 pins (preferring per-asset, then per-platform), auto-replacing binaries when pins drift or files are unreadable.
  * Compressed-asset handling: archives are reinstalled when appropriate; existing compressed files are preserved if a refetch fails; unpinned binaries retain legacy behavior.

* **Tests**
  * Added tests for matching pins, pin drift replacements, unreadable binaries, legacy unpinned behavior, and compressed-asset reinstall/preserve scenarios.

### DIFF
--- a/tool_registry/installers.py
+++ b/tool_registry/installers.py
@@ -163,70 +163,45 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
 # -- Native binary installer --------------------------------------------------
 
 
-def _resolve_native_expected_hash(
+def _resolve_native_download_hash(
     source: GitHubToolSource,
     asset_name: str,
     compressed: bool,
     platform_suffix: str,
 ) -> str | None:
-    """SHA256 to expect for a NATIVE asset, or None when no pin applies.
-
-    Precedence:
-      1. ``sha256[asset_name]`` — exact-asset pin (works for any asset
-         including compressed; one entry per arch override).
-      2. ``sha256[platform_suffix]`` — per-platform pin used by tools we
-         republish ourselves (tokei, gopls); only consulted for
-         pre-extracted assets so a stale platform-keyed hash doesn't get
-         applied to a freshly compressed binary.
-      3. ``None`` — no pin, callers must skip verification.
-    """
+    """Resolve the SHA256 pin for the downloaded release asset."""
     if asset_name in source.sha256:
         return source.sha256[asset_name]
     if not compressed:
-        return source.sha256.get(platform_suffix or "")
+        if not source.sha256:
+            return None
+        platform_hash = source.sha256.get(platform_suffix or "")
+        if platform_hash is None:
+            raise ValueError(f"{asset_name}: missing SHA256 pin for platform {platform_suffix!r}")
+        return platform_hash
     return None
 
 
 def _existing_binary_is_usable(
     binary_path: Path,
-    expected_hash: str | None,
-    compressed: bool,
+    expected_hash: str,
     binary_name: str,
 ) -> bool:
-    """Return True when the on-disk binary can be trusted (skip reinstall).
-
-    Side effect: deletes ``binary_path`` on hash mismatch or read failure,
-    leaving the slot empty for the caller's subsequent download.
-
-    Trusts the existing file (returns True) when:
-
-    - The asset is compressed — the pin is over the archive, not the
-      extracted binary, so on-disk verification is structurally impossible.
-    - No SHA is pinned for this platform.
-    - The on-disk SHA matches the pin.
-
-    Returns False (and deletes the file) when the on-disk hash differs
-    from the pin, or the file exists but cannot be read.
-    """
-    if compressed or expected_hash is None:
-        logger.info("  %s: already installed (no on-disk pin to verify)", binary_name)
-        return True
+    """Return whether the on-disk binary can be reused."""
     try:
         actual = hashlib.sha256(binary_path.read_bytes()).hexdigest()
     except OSError as exc:
-        logger.warning("  %s: existing binary unreadable (%s); reinstalling", binary_name, exc)
-        binary_path.unlink(missing_ok=True)
+        logger.warning("  %s: existing binary unreadable (%s)", binary_name, exc)
         return False
     if actual == expected_hash:
         logger.info("  %s: already installed (sha256 verified)", binary_name)
         return True
     logger.info(
-        "  %s: pin drift (have %s..., want %s...); reinstalling",
+        "  %s: pin drift (have %s..., want %s...)",
         binary_name,
         actual[:12],
         expected_hash[:12],
     )
-    binary_path.unlink(missing_ok=True)
     return False
 
 
@@ -241,10 +216,7 @@ def install_native_tools(
     ``gopls``) and compressed-binary assets (``rust-analyzer``); the
     archive format is inferred from the asset filename suffix.
 
-    On-disk binaries are SHA256-verified against the pin before the
-    skip-if-installed shortcut. A drifted binary (pin moved upstream
-    after the user's first install) is removed and re-fetched so users
-    pick up corrected/updated artifacts automatically.
+    Pre-extracted on-disk binaries are SHA256-verified before reuse.
     """
     system = platform.system()
     suffix = PLATFORM_SUFFIX.get(system)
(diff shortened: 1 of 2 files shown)


tool_registry/installers.py
```diff
@@ -163,6 +163,73 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+    """SHA256 to expect for a NATIVE asset, or None when no pin applies.
+
+    Precedence:
+      1. ``sha256[asset_name]`` — exact-asset pin (works for any asset
+         including compressed; one entry per arch override).
+      2. ``sha256[platform_suffix]`` — per-platform pin used by tools we
+         republish ourselves (tokei, gopls); only consulted for
+         pre-extracted assets so a stale platform-keyed hash doesn't get
+         applied to a freshly compressed binary.
+      3. ``None`` — no pin, callers must skip verification.
```
ivanmilevtues: What does this mean? Is this happenning I think if it is we should probably fail rather than download on repeat.

Maybe we can just raise a ValueError

ivanmilevtues: Also the comment is longer than the function :D

tool_registry/installers.py
```diff
@@ -163,6 +163,73 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+
+def _existing_binary_is_usable(
+    binary_path: Path,
+    expected_hash: str | None,
+    compressed: bool,
+    binary_name: str,
+) -> bool:
+    """Return True when the on-disk binary can be trusted (skip reinstall).
+
+    Side effect: deletes ``binary_path`` on hash mismatch or read failure,
```
ivanmilevtues: not good?

tool_registry/installers.py
```diff
@@ -163,6 +163,73 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+    - The asset is compressed — the pin is over the archive, not the
+      extracted binary, so on-disk verification is structurally impossible.
+    - No SHA is pinned for this platform.
+    - The on-disk SHA matches the pin.
+
+    Returns False (and deletes the file) when the on-disk hash differs
+    from the pin, or the file exists but cannot be read.
+    """
+    if compressed or expected_hash is None:
+        logger.info("  %s: already installed (no on-disk pin to verify)", binary_name)
```
ivanmilevtues: This also sounds sus, we should probably raise and error as this shouldn't happen or?

tool_registry/installers.py
```diff
@@ -163,6 +163,73 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+        return True
+    try:
+        actual = hashlib.sha256(binary_path.read_bytes()).hexdigest()
+    except OSError as exc:
+        logger.warning("  %s: existing binary unreadable (%s); reinstalling", binary_name, exc)
+        binary_path.unlink(missing_ok=True)
```
ivanmilevtues: Yes I think if this func returns false the person interested if the binary is usable or not decides whether to delete it or what to do with it.

I would never expect a check to delete something.

tool_registry/installers.py
```diff
@@ -163,6 +163,73 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+        logger.info("  %s: already installed (sha256 verified)", binary_name)
+        return True
+    logger.info(
+        "  %s: pin drift (have %s..., want %s...); reinstalling",
+        binary_name,
+        actual[:12],
+        expected_hash[:12],
+    )
+    binary_path.unlink(missing_ok=True)
+    return False
```
ivanmilevtues, line 209: As the comment above I think we shouldn't delete here, so this can be just
```
logger.info(f"The file sha is {'alligned' if actual == expected_hash else 'misaligned.'}")
return actual == expected_hash

File: tool_registry/installers.py [hunk 6]
```diff
@@ -217,20 +286,11 @@ def install_native_tools(
...
-        #   3. ``None`` — no pin, ``download_asset`` skips verification.
-        if asset_name in source.sha256:
-            expected_hash: str | None = source.sha256[asset_name]
-        elif not compressed:
-            expected_hash = source.sha256.get(suffix or "")
-        else:
-            expected_hash = None
+        expected_hash = _resolve_native_expected_hash(source, asset_name, compressed, suffix or "")
+
+        if binary_path.exists() and _existing_binary_is_usable(binary_path, expected_hash, compressed, dep.binary_name):
```
ivanmilevtues: Yea here we can have a method which checks both:
if not is_existing_binary_usable():
     if exists(): delete
else:
   continue
or something like that idk. YOu get the idea.

ivanmilevtues (changes requested):
Cool change, I don't like the side-effect.
Unsure if we should merge this without @Svilen-Stefanov taking a look as he has done the install/release setup.

ivanmilevtues:
Also tests are failing:
`FAILED tests/test_windows_encoding.py::TestWindowsEncoding::test_no_cp1252_unencodable_characters_in_source - AssertionError: Lists differ: ["tests/test_tool_registry.py:1754:56  char '→' (U+2192)"] != []
`

tool_registry/installers.py
```diff
@@ -163,6 +163,48 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
 
 
+def _resolve_native_download_hash(
+    source: GitHubToolSource,
+    asset_name: str,
+    compressed: bool,
+    platform_suffix: str,
+) -> str | None:
+    """Resolve the SHA256 pin for the downloaded release asset."""
+    if asset_name in source.sha256:
```
Svilen-Stefanov, line 177: Here we fetch the asset name for both compressed and not compressed assets. I think that's fine because of the removed comment:
```
        # Hash lookup precedence:
        #   1. ``sha256[asset_name]`` — exact-asset pin (works for any asset
        #      including compressed; one entry per arch override).
        #   2. ``sha256[suffix]`` — per-platform pin used by tools we
        #      republish ourselves (tokei, gopls); only consulted for
        #      pre-extracted assets so a stale platform-keyed hash doesn't
        #      get applied to a freshly compressed binary.
        #   3. ``None`` — no pin, ``download_asset`` skips verification.
```

I think a comment like the above one is still helpful (and potentially even needed) to avoid confusion.

tool_registry/installers.py
```diff
@@ -163,6 +163,48 @@ def download_asset(url: str, destination: Path, expected_sha256: str | None = No
...
+        return source.sha256[asset_name]
+    if not compressed:
+        if not source.sha256:
+            return None
+        platform_hash = source.sha256.get(platform_suffix or "")
+        if platform_hash is None:
+            raise ValueError(f"{asset_name}: missing SHA256 pin for platform {platform_suffix!r}")
```
Svilen-Stefanov, line 184: I think this path is not tested in the tests. 
We should probably write a test with two deps where the first has a missing platform pin. Based on the current behavior, we will fail the entire thing if one install fails. That's probably fine for now but let's make it more explicit in the tests. 

One thing to point out here - we might actually want this to continue if the download is not required (e.g not everyone needs gopls). If we don't fail, that could complicate our lives down the line (partial downloads would become a pain). For now, failing hard here seems okay but let's discuss this more. 

cc @ivanmilevtues

tool_registry/installers.py
```diff
@@ -217,24 +258,23 @@ def install_native_tools(
...
-            expected_hash = source.sha256.get(suffix or "")
-        else:
-            expected_hash = None
+        download_hash = _resolve_native_download_hash(source, asset_name, compressed, suffix or "")
+        # Compressed pins apply to the archive, not the extracted binary.
+        existing_binary_hash = None if compressed else download_hash
+
+        if binary_path.exists():
+            if existing_binary_hash is None:
+                logger.info("  %s: already installed (no on-disk pin to verify)", dep.binary_name)
```
Svilen-Stefanov: This `None` branch is reached by two different cases - rust-analyzer (compressed, pin is on the archive so the extracted binary is intentionally unverifiable) and a genuinely unpinned native tool. 
For both of these, the logs are the same "no on-disk pin to verify" for both, which is wrong for rust-analyzer (it is pinned, just on the archive). 
I think this is a bit confusing as is right now, refer to [this](https://github.com/CodeBoarding/CodeBoarding/pull/350/changes#r3329808819) as well.

One way we could improve this is by branching on compressed explicitly so each skip says what's actually true.

Svilen-Stefanov (changes requested):
Added a couple of suggestions on how to make this easier to understand. I'd also want to confirm from both of you if we should stick with the hard failure path here.

ivanmilevtues:
no longer needed.

PR #354, opened 2026-06-04
Treat non-executable native LSP binaries as not-installed and self-heal

brovatten:
A native LSP binary (e.g. `gopls`) that exists on disk but lacks the execute bit caused a permanent `PermissionError [Errno 13]` at `Popen`, because the exec bit was only set on a fresh download while every "is it installed?" gate (`has_required_tools`, `resolve_config`, the install skip-on-exists branch) checked existence alone — so `needs_install()` stayed `False` and the documented restart→reinstall recovery was a no-op. This redefines "installed" for native binaries as exists-**and**-executable (re-arming `needs_install()` plus an in-place `chmod` repair so a binary that lost its `+x` self-heals on the next install instead of failing forever), and wraps the LSP `Popen` to raise an actionable error for the environmental causes (noexec mount / ownership) that code can't fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error messages provide actionable guidance when tool startup encounters permission issues
  * Native tool binaries are now validated to ensure they are executable and ready to run
  * Non-executable native tool binaries are automatically repaired on non-Windows systems, restoring proper permissions without requiring re-download

### DIFF
--- a/static_analyzer/engine/lsp_client.py
+++ b/static_analyzer/engine/lsp_client.py
@@ -95,14 +95,21 @@ class LSPClient:
         """Start the LSP server process and perform initialization handshake."""
         env = os.environ.copy()
         env.update(self._extra_env)
-        self._process = subprocess.Popen(
-            self._command,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.DEVNULL,
-            env=env,
-            cwd=str(self._project_root),
-        )
+        try:
+            self._process = subprocess.Popen(
+                self._command,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.DEVNULL,
+                env=env,
+                cwd=str(self._project_root),
+            )
+        except PermissionError as exc:
+            binary = self._command[0] if self._command else "<lsp>"
+            raise PermissionError(
+                f"LSP server binary is not executable: {binary}. Restart to reinstall, "
+                f"run 'chmod +x {binary}', or check that its directory is not mounted noexec."
+            ) from exc
 
         # Grab raw fd and close Python's BufferedReader immediately
         self._stdout_fd = os.dup(self._process.stdout.fileno())  # type: ignore[union-attr]
--- a/tool_registry/installers.py
+++ b/tool_registry/installers.py
@@ -206,6 +206,19 @@ def install_native_tools(
             continue
         binary_path = bin_dir / f"{dep.binary_name}{exe_suffix()}"
         if binary_path.exists():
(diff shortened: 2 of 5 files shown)


static_analyzer/engine/lsp_client.py
```diff
@@ -95,14 +95,21 @@ def start(self) -> dict | list | None:
...
+        try:
+            self._process = subprocess.Popen(
+                self._command,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.DEVNULL,
+                env=env,
+                cwd=str(self._project_root),
+            )
+        except PermissionError as exc:
```
ivanmilevtues, line 107: That is really cool with the from, this way you get a full stack trace.

ivanmilevtues (approved):
It fixed the issue for me and the code looks good.

Thanks for fixing.

tool_registry/installers.py
```diff
@@ -206,6 +206,19 @@ def install_native_tools(
             continue
         binary_path = bin_dir / f"{dep.binary_name}{exe_suffix()}"
         if binary_path.exists():
+            # Repair a binary that exists but lost its exec bit (a crash before
+            # the post-download chmod, or a mode-dropping copy/restore) — cheap
+            # vs a re-download, and breaks the "skip forever" EACCES loop.
+            if system != "Windows" and not os.access(binary_path, os.X_OK):
+                try:
+                    os.chmod(binary_path, 0o755)
```
Svilen-Stefanov, line 214: That's the actual fix, nice! Kind of crazy that we can do this, download a file and make it executable :D 
If I were a company, I'd be quite scared of this :D

static_analyzer/engine/lsp_client.py
```diff
@@ -95,14 +95,21 @@ def start(self) -> dict | list | None:
...
-        self._process = subprocess.Popen(
-            self._command,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.DEVNULL,
-            env=env,
-            cwd=str(self._project_root),
-        )
+        try:
+            self._process = subprocess.Popen(
```
Svilen-Stefanov, line 99: I guess this is just a secondary protection mechanism if the chmod didn't work (e.g for Windows or something)?

Svilen-Stefanov (approved):
Nice fix, thank you for adding it.

tool_registry/installers.py
brovatten: same :D

static_analyzer/engine/lsp_client.py
brovatten: It's to be able to log more clearly what went wrong

PR #355, opened 2026-06-05
Surface LSP server failures and right-size C# workspace-ready waits

brovatten:
Language servers' stderr was sent to `DEVNULL`, so failures like csharp-ls/Roslyn stalling on project load were completely invisible; this drains each server's stderr to the debug log instead. It also bounds the workspace-ready wait by the adapter's own `get_probe_timeout_minimum` (C#/Roslyn → 600s) instead of a hardcoded 300s, so large solutions aren't cut off early.

**Testing:** Touched-area unit tests pass (117), and an end-to-end C# analysis of the `csharp_edge_cases_project` fixture produced a correct 36-node / 34-edge call graph (Program.Main hub, virtual dispatch resolved, the intentional dead-code class isolated).

Companion (wrapper) PR: CodeBoarding/CodeBoarding-wrapper#56

## Summary by CodeRabbit

* **Bug Fixes**
  * Language server diagnostic and error messages are now captured and logged during startup (previously discarded).

* **Tests**
  * Updated test mocks for LSP startup behavior.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -251,7 +251,8 @@ class StaticAnalyzer:
                 # via ``wait_for_workspace_ready`` so the language-name
                 # check doesn't keep growing.
                 if adapter.wait_for_workspace_ready:
-                    engine_client.wait_for_server_ready()
+                    ready_timeout = max(adapter.get_probe_timeout_minimum(), 300)
+                    engine_client.wait_for_server_ready(timeout=ready_timeout)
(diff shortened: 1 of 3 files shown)


Svilen-Stefanov (approved):
The changes make sense to me but the described testing in the PR is not really testing them. 
Before merging, we need to verify two things:
1. Run a failure case with the changes before (from main) - verify the logs are not there, run changes after, verify the error logs show up
2. Verify the timeout fix is working by checking with a C# project that takes longer than 5min - verify it fails before and works after.

After you've tested both, we can merge it.

PR #357, opened 2026-06-05
Capture per-run LLM token usage on DiagramGenerator

brovatten:
Adds `StreamingStatsWriter.snapshot_usage()` and `DiagramGenerator.capture_token_usage()`, which record cumulative input/output/total tokens and the model name into `last_token_usage` after both full and incremental analyses. The value is also exposed for the component-expand path so the wrapper and extension can read per-run token usage in-band instead of scraping log lines.

Tested with `black`, `mypy`, and the monitoring/model-capabilities suites (114 tests pass), including two new `snapshot_usage` aggregation tests.

## Summary by CodeRabbit

* **New Features**
  * Final token-usage snapshots are captured and exposed after operations, including a readable "last usage" value when available.
  * Monitoring now preserves the finalized usage snapshot to ensure consistent reporting at completion.

* **Tests**
  * Added tests covering token-usage aggregation, empty-agent behavior, and finalization of usage snapshots.

### DIFF
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -19,7 +19,7 @@ from agents.incremental_agent import (
     remove_deleted_files,
     stitch_delta,
 )
-from agents.llm_config import initialize_llms
+from agents.llm_config import MONITORING_CALLBACK, initialize_llms
 from agents.meta_agent import MetaAgent
 from agents.planner_agent import get_expandable_components
 from diagram_analysis.analysis_json import (
@@ -117,6 +117,22 @@ class DiagramGenerator:
 
         self._monitoring_agents: dict[str, MonitoringMixin] = {}
         self.stats_writer: StreamingStatsWriter | None = None
+        self.last_token_usage: dict | None = None
+
+    def capture_token_usage(self) -> None:
+        """Snapshot cumulative token usage into ``last_token_usage``.
+
+        Reads live agent stats, so it is valid after a full or incremental run,
+        or a standalone ``process_component`` (expansion). No-op when monitoring
+        is disabled.
+        """
+        if not self.stats_writer:
+            self.last_token_usage = None
+            return
+        usage = self.stats_writer.snapshot_usage()
+        if usage["model_name"] == "unknown" and MONITORING_CALLBACK.model_name:
+            usage["model_name"] = MONITORING_CALLBACK.model_name
+        self.last_token_usage = usage
 
     def process_component(
         self, component: Component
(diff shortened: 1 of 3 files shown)


diagram_analysis/diagram_generator.py
```diff
@@ -117,6 +117,22 @@ def __init__(
 
         self._monitoring_agents: dict[str, MonitoringMixin] = {}
         self.stats_writer: StreamingStatsWriter | None = None
+        self.last_token_usage: dict | None = None
+
+    def capture_token_usage(self) -> None:
```
ivanmilevtues: I don't like this method here because the diagram_generator is already huge and we now have to call this method at every "LLM like" operation.

We already have such logic withing the monitors, and we can use their final result + we already store that in a json file. Effectively we already have this as a thing, so we can redesign it, but for sure we shouldn't have it twice.


I checked what I wrote with an agent, if useful or not you can decide.
```
StreamingStatsWriter (monitoring/writers.py) is already the thing that aggregates token usage across all agents via snapshot_usage() — it reads from each agent's MonitoringMixin, which has per-agent RunStats. The StreamingStatsWriter runs as a background thread during the analysis and already does a final flush in stop() (called via __exit__).
DiagramGenerator.capture_token_usage() (diagram_generator.py:122) just calls self.stats_writer.snapshot_usage() and stores the result on self.last_token_usage — it's a thin passthrough that exists purely so the wrapper can read it after the run.
The actual design oddity is: StreamingStatsWriter.__exit__ already calls stop() → _save_llm_usage() + _stream_token_usage() → snapshot_usage(). So the final aggregation is happening twice — once by the monitoring system's own lifecycle, and once by capture_token_usage() which is called manually right after in generate_analysis() (line 509) and generate_analysis_incremental() (line 697).
The cleaner design would be: StreamingStatsWriter already knows the final usage at stop() time. It could expose the finalized snapshot as an attribute (e.g. self.final_usage) set during stop(), and DiagramGenerator could just read self.stats_writer.final_usage instead of having its own capture_token_usage() / last_token_usage dance. Or more broadly, the wrapper could read usage directly from the monitoring output files (llm_usage.json) rather than reaching into the generator's internals.
So yes — the token usage snapshot logically belongs to the monitoring lifecycle (StreamingStatsWriter.stop()), not to DiagramGenerator. The current capture_token_usage is a leaky convenience that duplicates what monitoring already does.
```

ivanmilevtues (changes requested):
Thanks for working on this, I have a single comment around the design.

Our monitoring already collects the total usage so we should reuse that rather than introducing a new method which we need to call manually.

One more thing, this doesn't cover incremental analysis. And such misses would be covered if we collect directly in the collectors (montoring logic).

PR #358, opened 2026-06-05
feat: add LiteLLM proxy as a configurable LLM provider

ShukhratE:
## What
Adds a `litellm` LLM provider that targets an OpenAI-compatible [LiteLLM proxy server](https://docs.litellm.ai/docs/simple_proxy), configured via `LITELLM_API_KEY` and `LITELLM_BASE_URL`.

- `agents/llm_config.py` — new `litellm` entry in `LLM_PROVIDERS`, reusing `ChatOpenAI` pointed at the proxy's base URL.
- `user_config.py` — `litellm_api_key` / `litellm_base_url` added to the `[provider]` schema, env-var mapping, loader, and generated `config.toml` template.
- `README.md` — documented the new keys and added LiteLLM proxy to the supported providers list.
- Tests in `tests/agents/test_llm_config.py` and `tests/test_user_config.py`.

## Why
Users running a LiteLLM proxy (a single gateway in front of many providers) couldn't point CodeBoarding at it without abusing the generic OpenAI provider and manually overriding the base URL. A first-class `litellm` provider makes this a documented, supported path — consistent with the existing gateway providers (deepseek, glm, kimi, openrouter). Since a LiteLLM proxy is OpenAI-compatible, no new dependency is required.

## How (tested)
- `uv run pytest --ignore=tests/integration tests/agents/test_llm_config.py tests/test_user_config.py` — 102 passed
- `black --check` — clean
- `mypy` — clean on changed files

New tests cover: provider key validation, config→env injection of the proxy key/URL, and that `base_url`/`api_key`/`model` are wired through to the chat client.

## Usage
```toml
# ~/.codeboarding/config.toml
[provider]
litellm_api_key  = "sk-..."                 # LiteLLM server key
litellm_base_url = "http://localhost:4000"  # LiteLLM server URL

[llm]
agent_model   = "your-proxy-model"
parsing_model = "your-proxy-parsing-model"
```
Or via env vars: `LITELLM_API_KEY`, `LITELLM_BASE_URL`, `AGENT_MODEL`, `PARSING_MODEL` (shell env takes precedence over the config file). Model names are defined by your proxy, so set `agent_model`/`parsing_model` to match — the placeholder defaults (`gpt-4o`/`gpt-4o-mini`) will be rejected if your proxy doesn't expose them.

## Summary by CodeRabbit

* **New Features**
  * Added LiteLLM proxy as a supported LLM provider (keyless-capable; configurable via base URL and optional API key).
  * User config now supports LiteLLM keys and base URL and will inject them into the environment when applied.

* **Documentation**
  * Updated docs and sample config to include LiteLLM proxy settings.

* **Bug Fixes**
  * Normalized Ollama host/base-URL handling to accept host-only values.

* **Tests**
  * Added tests for LiteLLM initialization, key/base-url validation, and config env injection.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -67,6 +67,9 @@ class LLMConfig:
     Configuration for LLM providers.
 
     Attributes:
+        activation_envs: Env vars that select this provider — active iff any is set.
+        api_key_env: Env var holding the provider's secret, or None when the
+                     underlying SDK reads its credentials from the environment itself.
         agent_model: The "agent" model used for complex reasoning and agentic tasks.
         parsing_model: The "parsing" model used for fast, cost-effective extraction and parsing tasks.
         agent_temperature: Temperature for the agent model. Defaults to 0 for deterministic behavior
@@ -77,40 +80,37 @@ class LLMConfig:
     """
 
     chat_class: Type[BaseChatModel]
-    api_key_env: str
+    activation_envs: list[str]
     agent_model: str
     parsing_model: str
     llm_type: LLMType
     agent_temperature: float = LLMDefaults.DEFAULT_AGENT_TEMPERATURE
     parsing_temperature: float = LLMDefaults.DEFAULT_PARSING_TEMPERATURE
     extra_args: dict[str, Any] = field(default_factory=dict)
-    alt_env_vars: list[str] = field(default_factory=list)
-    required_env_vars: list[str] = field(default_factory=list)
+    api_key_env: str | None = None
     keyless_capable: bool = False
     """Whether this provider can run without a real API key.
 
-    True for self-hosted / OpenAI-compatible endpoints where ``api_key_env`` (or
-    an ``alt_env_vars`` entry) is really a base-URL existence check rather than a
-    secret. When such a provider is the sole active one and no real key is set,
-    key validation warns instead of failing, and the client uses a placeholder.
+    True for self-hosted / OpenAI-compatible endpoints that accept
+    unauthenticated requests. When such a provider is the sole active one and
+    no real key is set, key validation warns instead of failing, and the
+    client uses a placeholder.
     """
 
     def get_api_key(self) -> str | None:
-        return os.getenv(self.api_key_env)
+        return os.getenv(self.api_key_env) if self.api_key_env else None
 
     def has_real_api_key(self) -> bool:
-        """True if the provider's primary API-key env var holds a value.
+        """True if the provider's API-key env var holds a value.
 
         Distinct from ``is_active()``: a keyless-capable provider can be active
-        via a base-URL ``alt_env_vars`` entry while having no real key here.
+        via a base-URL activation var while having no real key here.
         """
-        return bool(os.getenv(self.api_key_env))
+        return bool(self.api_key_env and os.getenv(self.api_key_env))
 
     def is_active(self) -> bool:
-        """Check if any of the environment variables (primary or alternate) are set."""
-        if os.getenv(self.api_key_env):
-            return True
-        return any(os.getenv(var) for var in self.alt_env_vars)
+        """Check if any of the activation environment variables are set."""
+        return any(os.getenv(var) for var in self.activation_envs)
 
     def get_resolved_extra_args(self) -> dict[str, Any]:
         resolved = {}
@@ -125,11 +125,11 @@ class LLMConfig:
 LLM_PROVIDERS = {
     "openai": LLMConfig(
         chat_class=ChatOpenAI,
+        activation_envs=["OPENAI_API_KEY", "OPENAI_BASE_URL"],
         api_key_env="OPENAI_API_KEY",
         agent_model="gpt-4o",
         parsing_model="gpt-4o-mini",
         llm_type=LLMType.GPT4,
-        alt_env_vars=["OPENAI_BASE_URL"],
         keyless_capable=True,
         extra_args={
             "base_url": lambda: os.getenv("OPENAI_BASE_URL"),
@@ -140,11 +140,11 @@ LLM_PROVIDERS = {
     ),
     "vercel": LLMConfig(
         chat_class=ChatOpenAI,
+        activation_envs=["VERCEL_API_KEY", "VERCEL_BASE_URL"],
         api_key_env="VERCEL_API_KEY",
         agent_model="google/gemini-3-flash",
         parsing_model="openai/gpt-5-mini",
         llm_type=LLMType.GEMINI_FLASH,
-        alt_env_vars=["VERCEL_BASE_URL"],
         extra_args={
             "base_url": lambda: os.getenv("VERCEL_BASE_URL", f"https://ai-gateway.vercel.sh/v1"),
             "max_tokens": None,
@@ -154,6 +154,7 @@ LLM_PROVIDERS = {
     ),
     "anthropic": LLMConfig(
         chat_class=ChatAnthropic,
+        activation_envs=["ANTHROPIC_API_KEY"],
         api_key_env="ANTHROPIC_API_KEY",
         agent_model="claude-sonnet-4-6",
         parsing_model="claude-haiku-4-5",
@@ -166,6 +167,7 @@ LLM_PROVIDERS = {
     ),
     "google": LLMConfig(
         chat_class=ChatGoogleGenerativeAI,
+        activation_envs=["GOOGLE_API_KEY"],
         api_key_env="GOOGLE_API_KEY",
         agent_model="gemini-3-flash-preview",
         parsing_model="gemini-3.1-flash-lite",
@@ -178,7 +180,8 @@ LLM_PROVIDERS = {
     ),
     "aws": LLMConfig(
         chat_class=ChatBedrockConverse,
-        api_key_env="AWS_BEARER_TOKEN_BEDROCK",  # Used for existence check
+        # No api_key_env: botocore reads AWS_BEARER_TOKEN_BEDROCK from the environment itself.
+        activation_envs=["AWS_BEARER_TOKEN_BEDROCK"],
         agent_model="anthropic.claude-sonnet-4-6",
         parsing_model="claude-haiku-4-5",
         llm_type=LLMType.CLAUDE_SONNET,
@@ -190,6 +193,7 @@ LLM_PROVIDERS = {
     ),
     "cerebras": LLMConfig(
         chat_class=ChatCerebras,
+        activation_envs=["CEREBRAS_API_KEY"],
         api_key_env="CEREBRAS_API_KEY",
         agent_model="zai-glm-4.7",
         parsing_model="gpt-oss-120b",
@@ -202,7 +206,11 @@ LLM_PROVIDERS = {
     ),
     "ollama": LLMConfig(
         chat_class=ChatOllama,
-        api_key_env="OLLAMA_BASE_URL",  # Used for existence check
+        # OLLAMA_HOST is Ollama's canonical host var; the client falls back to it
+        # when no base_url is passed, and sends OLLAMA_API_KEY (Ollama cloud) itself.
+        activation_envs=["OLLAMA_BASE_URL", "OLLAMA_HOST"],
+        api_key_env="OLLAMA_API_KEY",
+        keyless_capable=True,
         agent_model="qwen3:30b",
         parsing_model="qwen2.5:7b",
         llm_type=LLMType.GEMINI_FLASH,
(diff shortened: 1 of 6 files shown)


agents/llm_config.py
```diff
@@ -266,6 +249,23 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
             "max_retries": 0,
         },
     ),
+    # LiteLLM proxy server: OpenAI-compatible gateway in front of any provider.
+    # Model names are defined by the proxy, so AGENT_MODEL / PARSING_MODEL (or the
+    # [llm] section of config.toml) should normally override the placeholder defaults.
+    "litellm": LLMConfig(
+        chat_class=ChatOpenAI,
+        api_key_env="LITELLM_API_KEY",
```
Svilen-Stefanov, line 284: I'm not an expert on litellm but isn't it more similar to ollama than other providers - the base url is mandatory and the key is optional?

If that's the case, shouldn't we make this:
```suggestion
        api_key_env="LITELLM_BASE_URL",
```
@brovatten @ivanmilevtues I feel like `api_key_env` here is very confusing as it on the one hand should point to a key, on the other points to a base url when that's the mandatory component. We should refactor this separately to something like `activation_env ` maybe?

agents/llm_config.py
```diff
@@ -266,6 +249,23 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
+        agent_model="gpt-4o",
+        parsing_model="gpt-4o-mini",
+        llm_type=LLMType.GPT4,
+        alt_env_vars=["LITELLM_BASE_URL"],
+        extra_args={
+            "base_url": lambda: os.getenv("LITELLM_BASE_URL"),
```
Svilen-Stefanov, line 290: Does it make sense to provide a default here? [Here](https://docs.litellm.ai/docs/proxy/docker_quick_start) I see http://0.0.0.0:4000/ as the default example.

agents/llm_config.py
```diff
@@ -249,6 +249,23 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
+            "max_tokens": None,
+            "timeout": None,
+            "max_retries": 0,
+        },
```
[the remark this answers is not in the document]
Svilen-Stefanov: This is essentially a different way to phrase what I suggested above: https://github.com/CodeBoarding/CodeBoarding/pull/358/changes#r3365671923

Svilen-Stefanov (commented):
Thank you, @ShukhratE! I added a couple of questions / comments that I'd like to hear your opinion on. 
There also seem to be 2 unresolved conflicts, could you pull from the latest main branch?

ShukhratE:
Thanks for the review. Pushed an update (rebased onto main to pull in the openai_base_url work from #345):

- LiteLLM is now activated by `LITELLM_BASE_URL` (mandatory, like the ollama provider) instead of the API key; `LITELLM_API_KEY` is optional and a keyless proxy falls back to a placeholder key.
- This closes the bug where only setting `LITELLM_API_KEY` would silently route requests to the default OpenAI endpoint — base_url is now always wired through.
- Docs/template updated to show base_url as required, key as optional.
- Added a keyless-proxy test; full suite + black + mypy green.

Left two items as follow-ups since they felt like separate calls:
- Renaming `api_key_env` → `activation_env` repo-wide — agree it's confusing, but better as its own refactor than bundled here.
- Default base URL (`http://0.0.0.0:4000`): left unset on purpose so a missing proxy fails loudly rather than silently targeting a default. Happy to add a default if you'd prefer.

agents/llm_config.py
ShukhratE: Done - switched activation to LITELLM_BASE_URL (mandatory) with the key now optional, matching the ollama pattern you mentioned

agents/llm_config.py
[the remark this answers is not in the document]
ShukhratE: Fixed. LITELLM_BASE_URL is now the required activation variable and is always passed to the client, so a key-only config can no longer fall through to the default OpenAI endpoint. A keyless proxy uses a placeholder key. Added a test covering the keyless base-URL-only case

agents/llm_config.py
ShukhratE: Left it unset intentionally so a missing proxy fails loudly during validation instead of silently hitting a default address. The docker default (http://0.0.0.0:4000) is  reasonable - let me know if to add it as the
default if you'd prefer that over requiring explicit config

brovatten: I agree with @ShukhratE . The default pattern we have is a template in the `config.toml` e.g.
`litellm_base_url = "http://localhost:4000"  # LiteLLM server URL`
like implemented, and not in the lifecycle

agents/llm_config.py
```diff
@@ -266,6 +266,26 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
     ),
+    # LiteLLM proxy server: OpenAI-compatible gateway in front of any provider.
+    # LITELLM_BASE_URL is the activation/mandatory variable (the proxy address,
+    # like Ollama); LITELLM_API_KEY is optional since some proxies are keyless.
+    # Model names are defined by the proxy, so AGENT_MODEL / PARSING_MODEL (or the
+    # [llm] section of config.toml) should normally override the placeholder defaults.
+    "litellm": LLMConfig(
+        chat_class=ChatOpenAI,
+        api_key_env="LITELLM_BASE_URL",
+        agent_model="gpt-4o",
```
brovatten, line 285: Just curious @ShukhratE , is the model you would use yourself?
I am not using the OpenAI API but I'm guessing this one is a bit outdated, or what do you think?
The idea is to not throw the heaviest model at this, but still a capable one - and slightly cheaper one for the parsing

No need to edit this.

agents/llm_config.py
```diff
@@ -266,6 +266,26 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
         },
     ),
+    # LiteLLM proxy server: OpenAI-compatible gateway in front of any provider.
+    # LITELLM_BASE_URL is the activation/mandatory variable (the proxy address,
+    # like Ollama); LITELLM_API_KEY is optional since some proxies are keyless.
+    # Model names are defined by the proxy, so AGENT_MODEL / PARSING_MODEL (or the
+    # [llm] section of config.toml) should normally override the placeholder defaults.
+    "litellm": LLMConfig(
+        chat_class=ChatOpenAI,
+        api_key_env="LITELLM_BASE_URL",
```
brovatten: Shouldn't this be `litellm_api_key`?

agents/llm_config.py
```diff
@@ -266,6 +266,26 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
+        agent_model="gpt-4o",
+        parsing_model="gpt-4o-mini",
+        llm_type=LLMType.GPT4,
+        keyless_capable=True,
+        extra_args={
+            "base_url": lambda: os.getenv("LITELLM_BASE_URL"),
+            "api_key": lambda: os.getenv("LITELLM_API_KEY") or "no-key-required",
```
brovatten: I think we should drop "no-key-required" and just let os.getenv("LITELLM_API_KEY") become None.
since we set this later [here](https://github.com/CodeBoarding/CodeBoarding/pull/358/changes#diff-a8f8eb94cd27ee2916ed2e57d633fc09ba6227ec1cc5878145a0ecf86606afc8L310)

agents/llm_config.py
```diff
@@ -266,6 +266,26 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
+    # like Ollama); LITELLM_API_KEY is optional since some proxies are keyless.
+    # Model names are defined by the proxy, so AGENT_MODEL / PARSING_MODEL (or the
+    # [llm] section of config.toml) should normally override the placeholder defaults.
+    "litellm": LLMConfig(
+        chat_class=ChatOpenAI,
+        api_key_env="LITELLM_BASE_URL",
+        agent_model="gpt-4o",
+        parsing_model="gpt-4o-mini",
+        llm_type=LLMType.GPT4,
+        keyless_capable=True,
```
brovatten, line 288: We should add this to all LLMs in the future to avoid warnings @Svilen-Stefanov . For now OK

agents/llm_config.py
brovatten: Actually OK it says we need this, otherwise it would not activate litellm. 
So maybe just add a end-of-the-line comment
`# Used for existence check`

ivanmilevtues (approved):
Seems good to me, I did try to test it with a proxied litellm it seemed to work.

Thanks for adding this @ShukhratE!

agents/llm_config.py
Svilen-Stefanov: @brovatten - look at this code here. This is essentially what's the confusing bit. Whatever env var you set to `api_key_env` is considered the primary env var (so until now, whatever env var you set to `api_key_env` was considered the only `required_env_var`).  
Given that you now introduce `required_env_vars`, I guess you need to change this as well.

If you modify it here though, this will have impact to the other configs as well - so you might want to update them too.
We're starting to have quite a few attributes here and I'm not sure we need all of them. We've accumulated four overlapping fields for activation/validation (api_key_env, alt_env_vars, required_env_vars, keyless_capable), and for litellm `LITELLM_BASE_URL` is listed as both an alternate and required — which reads as a contradiction.

agents/llm_config.py
```diff
@@ -85,6 +85,7 @@ class LLMConfig:
     parsing_temperature: float = LLMDefaults.DEFAULT_PARSING_TEMPERATURE
     extra_args: dict[str, Any] = field(default_factory=dict)
     alt_env_vars: list[str] = field(default_factory=list)
+    required_env_vars: list[str] = field(default_factory=list)
```
Svilen-Stefanov: I think I'd call this `activation_env` right now, set it as a string (as we currently don't have use cases with more than one activation env) and update all configs. 
The difference between activation and required is OR vs AND semantics. 
Activation = Any one of these present would work
Required = All need to be there for it to work

Not sure if we want the first or the second. If you think the second makes more sense, we might want to rename `is_active` method I pointed out below.

Svilen-Stefanov (commented):
Leaving fresh comments as you requested.

agents/llm_config.py
```diff
@@ -77,39 +80,37 @@ class LLMConfig:
...
     def has_real_api_key(self) -> bool:
-        """True if the provider's primary API-key env var holds a value.
+        """True if the provider's API-key env var holds a value.
 
         Distinct from ``is_active()``: a keyless-capable provider can be active
-        via a base-URL ``alt_env_vars`` entry while having no real key here.
+        via a base-URL activation var while having no real key here.
         """
-        return bool(os.getenv(self.api_key_env))
+        return bool(self.api_key_env and os.getenv(self.api_key_env))
```
Svilen-Stefanov: nit: can't this just be 
```suggestion
        return self.get_api_key() is not None
```
?

agents/llm_config.py
```diff
@@ -124,11 +125,11 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
 LLM_PROVIDERS = {
     "openai": LLMConfig(
         chat_class=ChatOpenAI,
+        activation_envs=["OPENAI_API_KEY", "OPENAI_BASE_URL"],
```
Svilen-Stefanov: Given that we do `activation_envs` now and that's an OR dependency, this means that I can provide `OPENAI_BASE_URL` without a key AND provide a key without a `OPENAI_BASE_URL`, right?

user_config.py
```diff
@@ -30,7 +30,10 @@
     "glm_api_key": "GLM_API_KEY",
     "kimi_api_key": "KIMI_API_KEY",
     "ollama_base_url": "OLLAMA_BASE_URL",
+    "ollama_api_key": "OLLAMA_API_KEY",
```
Svilen-Stefanov, line 35: One comment here - I think the original intent behind this is to only contain the `activation_envs` - aka the env vars that are needed (not optional) per provider. 
Do you think we should add all values that are present in `activation_envs` in the providers and remove the ones that are not? 
Also if that's what we decide to do, it would be great to add a test that validates this and throws a clear error so that any contributors understand the idea behind it as well, wdyt?

Svilen-Stefanov (approved):
I haven't tested this myself and given that we did some refactoring, it would be great if you can test it with 2-3 providers to make sure things are still working. 
Thank you for the refactor!

agents/llm_config.py
brovatten: Yup. As docstring states:
`A provider is "active" when any of its ``activation_envs`` is set.`

agents/llm_config.py
brovatten: Took the dedup, with one amendment: `is not None` flips the empty-string case. An env var set to `""` (blank `.env` line, `export VAR=` in CI) currently counts as "no key" — consistent with `is_active()`'s truthiness and the `api_key or "no-key-required"` fallback — but `"" is not None` is `True`, which would suppress the keyless warning while the placeholder key still gets sent. Went with `bool(self.get_api_key())` instead, plus a test pinning the empty-string semantics: c24317fd.

agents/llm_config.py
brovatten: Yes

agents/llm_config.py
brovatten: Yup good catch

user_config.py
brovatten: I think it's better to keep all of them together. The optional ones are specified in the init of the class when you say whether it's okay to do it without a key.

You are alluding correctly that the naming "activate" is misleading - i changed it to "selected" instead. ANd adding the test!

PR #363, opened 2026-06-06
Fix C# LSP startup on modern .NET

hassan-albalawi:
## What changed

- Detect `.slnx` files as C# solution roots so modern .NET repos are analyzed once at the solution level instead of as many separate `.csproj` roots.
- Add adapter-level C# startup handling so CodeBoarding no longer blocks on an unreliable startup workspace-ready signal before the later probe and diagnostics waits run.
- Bump the managed `csharp-ls` install to `0.24.0` and let `dotnet tool install` choose the package default target framework.

## Why

On a .NET 10-only host, the current `csharp-ls 0.20.0` install can succeed but the resulting binary still targets `Microsoft.NETCore.App 9.0.0` and exits with code 150 at launch. Modern `.slnx` repositories are also missed by the scanner, which makes CodeBoarding fall back to per-project C# analysis and multiplies startup waits.

## Verification

- `uv run pytest tests/static_analyzer/test_csharp_adapter.py tests/static_analyzer/test_csharp_config_scanner.py tests/test_tool_registry.py -q`
- Real `dotnet tool install csharp-ls --version 0.20.0 --framework net10.0 --tool-path ...` on a .NET 10-only host installs but the binary exits 150 without roll-forward.
- Real `dotnet tool install csharp-ls --version 0.24.0 --tool-path ...` installs and runs on the same host.
- Real `.slnx` repo check against `wc3-platform-main-clean`: current scanner returns one solution root; main-style `*.sln` scanning returns eight `.csproj` roots.

## Summary by CodeRabbit

* **New Features**
  * Detects .slnx solution files for C# projects.
  * Per-project .NET SDK resolution and automatic install when required.

* **Improvements**
  * Upgraded csharp-ls tooling to v0.24.0 and improved environment/restore handling for C# projects.
  * Package-manager installs are fingerprinted to avoid unnecessary reinstalls.
  * Adapter readiness and DOTNET env behavior made more robust.

* **Bug Fixes**
  * Prevents caching empty/invalid analysis results when language servers fail.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -252,7 +252,7 @@ class StaticAnalyzer:
                 # via ``wait_for_workspace_ready`` so the language-name
                 # check doesn't keep growing.
                 if adapter.wait_for_workspace_ready:
-                    engine_client.wait_for_server_ready(timeout=adapter.workspace_ready_timeout)
+                    engine_client.wait_for_server_ready()
                     logger.info(f"{adapter.language} workspace ready: {time.monotonic() - t_lsp_started:.1f}s")
 
                 started.append((engine_config, engine_client))
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -246,7 +246,7 @@ class StaticAnalyzer:
                 t_lsp_started = time.monotonic()
                 logger.info(f"{adapter.language} LSP start: {t_lsp_started - t_start:.1f}s")
 
-                # Some LSP servers (JDTLS, rust-analyzer, csharp-ls) load
+                # Some LSP servers (JDTLS, rust-analyzer) load
                 # workspace metadata asynchronously and only respond to
                 # cross-file queries once that's complete. Adapters opt in
                 # via ``wait_for_workspace_ready`` so the language-name
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -44,6 +44,10 @@ class EngineConfig:
     source_files: list[Path] = field(default_factory=list)
 
 
+class StaticAnalysisFatalError(RuntimeError):
+    """Raised when continuing would produce misleading cached analysis."""
+
+
 def _create_engine_configs(
     programming_languages: list[ProgrammingLanguage],
     repository_path: Path,
@@ -213,6 +217,7 @@ class StaticAnalyzer:
         started: list[tuple[EngineConfig, LSPClient]] = []
         attempted: list[str] = []
         failed_languages: list[str] = []
+        failed_details: list[str] = []
 
         for engine_config in self._engine_configs:
             adapter, project_path = engine_config.adapter, engine_config.project_path
@@ -226,7 +231,7 @@ class StaticAnalyzer:
                 adapter.prepare_project(project_path)
                 command = adapter.get_lsp_command(project_path)
                 init_options = adapter.get_lsp_init_options(self.ignore_manager)
-                extra_env = adapter.get_lsp_env()
+                extra_env = adapter.get_lsp_env(project_path)
                 # Node-based LSPs spawn child ``node`` processes by name; on
                 # a Node-less host the embedded runtime's dir must be on PATH.
                 ensure_node_on_path(command, extra_env)
@@ -257,12 +262,13 @@ class StaticAnalyzer:
 
                 started.append((engine_config, engine_client))
 
-            except Exception:
+            except Exception as exc:
                 logger.exception(
                     f"Failed to start engine LSP client for {adapter.language}; "
                     f"skipping this language and continuing"
                 )
                 failed_languages.append(adapter.language)
+                failed_details.append(f"{adapter.language}: {exc}")
                 if engine_client is not None:
                     try:
                         engine_client.shutdown()
@@ -273,13 +279,18 @@ class StaticAnalyzer:
 
         if not started:
             self._clients_started = False
-            raise RuntimeError(f"Failed to start any engine LSP client (attempted: {', '.join(attempted) or 'none'})")
+            details = f"; failures: {'; '.join(failed_details)}" if failed_details else ""
+            raise RuntimeError(
+                f"Failed to start any engine LSP client (attempted: {', '.join(attempted) or 'none'}){details}"
+            )
 
         if failed_languages:
+            details = f" Details: {'; '.join(failed_details)}." if failed_details else ""
             logger.warning(
                 f"Proceeding with partial LSP coverage. "
                 f"Failed: {', '.join(failed_languages)}. "
-                f"Started: {', '.join(s.adapter.language for s, _ in started)}"
+                f"Started: {', '.join(s.adapter.language for s, _ in started)}."
+                f"{details}"
             )
 
         self._engine_clients = started
@@ -530,10 +541,11 @@ class StaticAnalyzer:
                 )
                 results = self._update_cached_results(cached_results, cached_sha)
 
+        self._validate_analysis_results(results)
+        results.diagnostics = self.collected_diagnostics
         self._cached_results = results
         self._pending_source_sha = source_sha
         self._pending_cache_dir = cache_dir
-        results.diagnostics = self.collected_diagnostics
         return results
 
     def _run_full_lsp_pass(self) -> StaticAnalysisResults:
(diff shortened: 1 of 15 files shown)


static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -117,7 +117,20 @@ def get_workspace_settings(self) -> dict | None:
 
     @property
     def wait_for_workspace_ready(self) -> bool:
```
Svilen-Stefanov: pointer

static_analyzer/__init__.py
```diff
@@ -251,7 +251,7 @@ def start_clients(self) -> None:
                 # via ``wait_for_workspace_ready`` so the language-name
                 # check doesn't keep growing.
                 if adapter.wait_for_workspace_ready:
-                    engine_client.wait_for_server_ready()
+                    engine_client.wait_for_server_ready(timeout=adapter.workspace_ready_timeout)
```
Svilen-Stefanov: Does this work? I see that [here](https://github.com/CodeBoarding/CodeBoarding/pull/363/changes#r3369613557) we set the C# adapter to return `False` so the newly introduced timeout seems to not be actually working?

tool_registry/registry.py
```diff
@@ -213,26 +213,24 @@ def is_available_on_host(self) -> bool:
...
+    # roll-forward on .NET-10-only hosts, but may time out or return no symbols
+    # for large SDK-style solutions.
     ToolDependency(
         key="csharp",
         binary_name="csharp-ls",
         kind=ToolKind.PACKAGE_MANAGER,
         config_section=ConfigSection.LSP_SERVERS,
         source=PackageManagerToolSource(
-            tag="0.20.0",
+            tag="0.24.0",
```
Svilen-Stefanov, line 226: This PR raises the min requirement to .NET 10. .NET 9's end of life is indeed approaching but I'm uncertain if we want to remove support for it completely. @hassan-albalawi , thank you for making this work for .NET 10 though, it's for sure the most relevant one for the future.
@ivanmilevtues @brovatten what do you think?

Svilen-Stefanov (changes requested):
@hassan-albalawi thank you very much for the contribution and for making this work on .NET 10. 

There's one dead code comment I left that would be great to improve before merging this. I also saw that this PR essentially makes .NET 10 required, which I'll check with the team and get back to you.

tool_registry/registry.py
brovatten: If it's old enough that it's being phased out, I'd take the hit for now. I think keeping the latest one is easier than keeping it compatible for both - and if someone raises an issue we can look into it.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -196,22 +209,30 @@ def prepare_project(self, project_root: Path) -> None:
...
-        """Set DOTNET_ROOT when not already in the environment.
+        """Set DOTNET_ROOT and DOTNET_ROLL_FORWARD when not already in the environment.
 
         csharp-ls requires the .NET runtime to be discoverable. On systems
         where the SDK is installed via a package manager (e.g. Homebrew on
         macOS), the ``DOTNET_ROOT`` variable may not be set, causing
         csharp-ls to fail at startup.  This resolves the runtime location
         from the ``dotnet`` binary on PATH.
+
+        DOTNET_ROLL_FORWARD=Major: csharp-ls 0.20.0 on NuGet only ships a
```
brovatten: This shouldn't be mentioned since we now target 0.24.0 right?

tool_registry/registry.py
ivanmilevtues: Let me test it out, I had some issues with .NET 10 on linux hense we put it back to .NET 9

static_analyzer/engine/adapters/csharp_adapter.py
hassan-albalawi: Good catch. That wording was left over from the earlier approach where we kept csharp-ls at 0.20.0 and used roll-forward for .NET-10-only hosts. Since this PR now targets 0.24.0, I’ll generalize this comment and remove the stale version-specific explanation

static_analyzer/__init__.py
hassan-albalawi: Correct. The timeout path is only used by adapters that opt into wait_for_workspace_ready. C# intentionally opts out because csharp-ls does not reliably emit that readiness signal, so the C# override was dead code. I’ll remove the unused override and keep the C#-specific probe/diagnostic waits as the actual readiness handling.

Svilen-Stefanov: Thank you

brovatten (approved):
LGTM, thanks @hassan-albalawi !

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -117,7 +117,15 @@ def get_workspace_settings(self) -> dict | None:
...
-        return True
+        """Skip the startup workspace-ready gate for csharp-ls.
+
+        csharp-ls does not consistently emit a solution/workspace-loaded signal
+        when CodeBoarding launches it per project directory. Waiting here makes
+        otherwise usable projects fail before the later probe/diagnostics waits
+        can run. Keep those later C#-specific waits instead; they use real LSP
+        requests/diagnostic quiescence and have larger timeouts.
+        """
+        return False
```
Svilen-Stefanov: I think this still doesn't quite make sense. If we want to skip the wait time here, we can just remove this as `language_adapter.py` already sets this to `False` - we don't need to overwrite it with the same value.

static_analyzer/engine/language_adapter.py
```diff
@@ -168,6 +168,19 @@ def wait_for_workspace_ready(self) -> bool:
         """
         return False
 
+    @property
+    def workspace_ready_timeout(self) -> int:
```
Svilen-Stefanov: I think both this and the addition `timeout=adapter.workspace_ready_timeout` don't make a difference right now. The default was set to 300 before as well and we don't change it here either. Is there a reason why you wanted to add this?

static_analyzer/__init__.py
Svilen-Stefanov: As mentioned [here](https://github.com/CodeBoarding/CodeBoarding/pull/363/changes#r3372000859), I think this isn't really needed.

Svilen-Stefanov (changes requested):
I left a couple more code quality asks, thank you @hassan-albalawi for the fast iterations here!

static_analyzer/dotnet_sdk.py
```diff
@@ -0,0 +1,327 @@
...
+    if private_probe and private_probe.ok and _has_sdk_major(private_dotnet, project_root, private_env, TOOL_SDK_MAJOR):
+        return DotnetSdkResolution(
+            dotnet_path=str(private_dotnet),
+            env=private_env,
+            source="private",
+            global_json=global_json,
+            requested_version=requested_version,
+        )
+
+    system_dotnet = shutil.which("dotnet")
```
ivanmilevtues, line 107: Is this windows compatibile?

static_analyzer/dotnet_sdk.py
```diff
@@ -0,0 +1,327 @@
...
+
+def _install_channel(channel: str, install_dir: Path) -> None:
+    logger.info("Installing .NET SDK channel %s into %s", channel, install_dir)
+    _run_install_script(["--channel", channel], install_dir)
+
+
+def _run_install_script(args: list[str], install_dir: Path) -> None:
+    install_dir.mkdir(parents=True, exist_ok=True)
+    script = _download_install_script()
+    if platform.system() == "Windows":
```
ivanmilevtues, line 258: Guessing from this, I think the shutil.which works cross platfrom. (For my prev comment)

ivanmilevtues (approved):
I say fix the merge conflict, no real comment I think it needs a bit more testing. PLease run the C# tests cross platfrom for this PR (the integration tests) if they pass we are ready to merge.

Thoughs @Svilen-Stefanov/ @brovatten

brovatten:
Thanks a lot @hassan-albalawi for the work!
If you're interested, what added: The C# LSP is strict to adhere to the **same** .NET that the project configurations have. So we had to make it dynamic to the project .NET version.

PR #368, opened 2026-06-07
Add anonymous PostHog telemetry with opt-out

ivanmilevtues:
## Summary

Adds anonymous, opt-out usage telemetry to the OSS package, reporting to PostHog (US cloud). On by default; disable with `CODEBOARDING_TELEMETRY=false` or `DO_NOT_TRACK=1`.

## What's collected
- `analysis_started` / `analysis_completed` events emitted from the core kernels (`run_full` / `run_partial` / `run_incremental`), so token/duration/outcome are captured whether invoked via the OSS CLI or the VSCode extension.
- Properties: `command`, `version`, `depth_level`, `status`, `duration_ms`, `model_name`, token counts, plus `source` and an anonymous `distinct_id`.
- **Never** collected: source code, file/repo names, paths, prompts, model outputs, API keys, or any PII. GeoIP disabled.

## Identity
`telemetry/device_id.py` reproduces the VSCode extension's device fingerprint (SHA-256 of system UUID / disk serial / CPU model) so a user's OSS and VSCode usage correlate anonymously. Parity is enforced by `tests/test_device_id_parity.py` (runs the TS and Python implementations and asserts equality).

## New files
- `telemetry/` package (`service.py`, `device_id.py`, `events.py`)
- `TELEMETRY.md` transparency doc + README section
- `tests/test_device_id_parity.py`

## Dependency
- Adds `posthog>=3.7`.

## Verification
- black, mypy, and the workflow/CLI/main/parity tests pass.
- PostHog ingest smoke-tested (HTTP 200).

The companion VSCode extension change (transport swap + passing the shared id/source via env) is in a separate PR on CodeBoarding-vscode.

## Summary by CodeRabbit

* **New Features**
  * Added optional anonymous, aggregate telemetry enabled by default: captures analysis lifecycle events, command outcomes, durations, token usage, and one-time repo scan summaries; opt out via CODEBOARDING_TELEMETRY=false or DO_NOT_TRACK=1.
  * Runtime now sets a default source tag ("oss") at startup.

* **Documentation**
  * Added detailed telemetry docs and README section describing collected fields, privacy protections, opt-out, and where events are sent.


telemetry/events.py
```diff
@@ -0,0 +1,127 @@
...
+            ],
+            "stack": ",".join(sorted(pl.language for pl in languages)),
+        },
+    )
+
+
+def _token_usage() -> dict:
+    """Snapshot of the process-global token counters (best effort)."""
+    try:
+        from agents.llm_config import MONITORING_CALLBACK
```
ivanmilevtues: no local imports

telemetry/service.py
```diff
@@ -0,0 +1,96 @@
+"""Minimal anonymous telemetry, modeled on browser-use's ProductTelemetry.
+
+Identity rules (so VSCode and standalone OSS runs correlate in one project):
```
ivanmilevtues: This comment I don't like

telemetry/device_id.py
```diff
@@ -0,0 +1,97 @@
...
+import subprocess
+
+
+def _sha256(text: str) -> str:
+    return hashlib.sha256(text.encode()).hexdigest()
+
+
+def _run(cmd: str) -> str:
+    """Mirror Node's ``execSync(cmd).toString()`` (shell, stdout as text)."""
+    return subprocess.check_output(cmd, shell=True, text=True, stderr=subprocess.DEVNULL)
```
ivanmilevtues: We shouldn't run subprocesses here like wut?

codeboarding_workflows/analysis.py
```diff
@@ -57,6 +58,7 @@ def build_generator(
     )
 
 
+@track_analysis
```
brovatten: We dont track for the extension or github action rn. The wrapper and the github action calls `run_incremental_workflow(generator)` directly, so those are not tracked right now. I'd add it there aswell.

codeboarding_workflows/analysis.py
```diff
@@ -112,13 +115,14 @@ def run_partial(
     if metadata is None:
         raise BaselineUnavailableError(f"No baseline analysis.json found in '{output_dir}'. Run a full analysis first.")
 
+    depth_level = int(metadata.get("depth_level", 1))
```
brovatten, line 116: Never fallback on depth. leads to misleading analytics or unexpected downstream errors

brovatten: atleast for partial, there it should already exists a depth level

telemetry/device_id.py
```diff
@@ -0,0 +1,111 @@
...
+from pathlib import Path
+
+
+def _sha256(text: str) -> str:
+    return hashlib.sha256(text.encode()).hexdigest()
+
+
+def _shell(cmd: str) -> str:
+    """Run an OS tool and return stdout (macOS/Windows only — see module docs)."""
+    return subprocess.check_output(cmd, shell=True, text=True, stderr=subprocess.DEVNULL)
```
brovatten, line 16: We use timeouts elsewhere, should we do it here too?

telemetry/events.py
```diff
@@ -0,0 +1,146 @@
...
+                "duration_ms": round((time.monotonic() - started) * 1000),
+                "model_name": after.get("model_name"),
+                "total_tokens": after.get("total_tokens", 0) - before.get("total_tokens", 0),
+                "input_tokens": after.get("input_tokens", 0) - before.get("input_tokens", 0),
+                "output_tokens": after.get("output_tokens", 0) - before.get("output_tokens", 0),
+            }
+            if error_type is not None:
+                props["error_type"] = error_type
+            telemetry.capture("analysis_completed", props)
+            telemetry.flush()
```
brovatten, line 145: what if it's unavailable for somereason. should we do max retries = 1?

telemetry/service.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Public PostHog project key (safe to ship; it is write-only ingest).
+POSTHOG_PROJECT_API_KEY = os.getenv("CODEBOARDING_POSTHOG_KEY", "phc_BQWpoXuPYQhW7mPWQcRv4yzSfuoAmh48EmXuUpeXPUB2")
+POSTHOG_HOST = os.getenv("CODEBOARDING_POSTHOG_HOST", "https://us.i.posthog.com")
+
+
+def _telemetry_disabled() -> bool:
+    if os.getenv("DO_NOT_TRACK", "").strip().lower() in ("1", "true", "yes"):
+        return True
+    return os.getenv("CODEBOARDING_TELEMETRY", "true").strip().lower() == "false"
```
brovatten, line 18: Symmetry? one checks ("1", "true", "yes") one does "false"

codeboarding_workflows/analysis.py
brovatten: I can adress this in my PR

brovatten (approved):
Looks great, thanks! posted a few nits

codeboarding_workflows/analysis.py
Svilen-Stefanov: Hm, yeah, I don't remember why this has a default here, generally we should deduct the depth for the partial analysis. IIrc, we only pass it here and ignore it / deduct it futher down the line, either way, this needs a refactor. I imagine this is not a straightforward change here because of the expectations down the call stack.

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+        return "unknown"
+
+
+# Repos already reported this process, so a scan that runs twice per analysis
+# (StaticAnalyzer init + diagram generation) emits a single event. Paths stay
+# in-memory only and are never sent.
+_scanned_repos: set[str] = set()
+
+
+def track_tech_stack(repo_path, total_loc: int, languages) -> None:
```
Svilen-Stefanov: nit: typing is missing for some of the attributes

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+        "error_message": _truncate(str(exc), _MAX_ERROR_MESSAGE_CHARS),
+        "error_stacktrace": _truncate(_format_stacktrace(exc), _MAX_STACKTRACE_CHARS),
+    }
+    run_id = _current_run_id.get()
+    if run_id is not None:
+        props["run_id"] = run_id
+    if extra:
+        props.update(extra)
+    telemetry.capture("error", props)
+    telemetry.flush()
```
Svilen-Stefanov, line 164: I see that PostHog already has a method called [capture_exception](https://posthog.com/docs/libraries/python#capturing-exceptions). Why are we not using that and trying to define our own version of it?

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+                "output_tokens": after.get("output_tokens", 0) - before.get("output_tokens", 0),
+            }
+            if error_type is not None:
+                props["error_type"] = error_type
+            if error_message:
+                props["error_message"] = error_message
+            if error_stacktrace:
+                props["error_stacktrace"] = error_stacktrace
+            telemetry.capture("analysis_completed", props)
+            telemetry.flush()
```
Svilen-Stefanov, line 145: I think you can again use the PostHog in-build exception capturing mechanism here as mentioned in my other comment:
https://github.com/CodeBoarding/CodeBoarding/pull/368/changes#r3384482602

Not an expert but this manual exception handling seems wrong.

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+            "model_name": stats.get("model_name"),
+            "total_tokens": usage.get("total_tokens", 0),
+            "input_tokens": usage.get("input_tokens", 0),
+            "output_tokens": usage.get("output_tokens", 0),
+        }
+    except Exception:
+        return {}
+
+
+def _bound_arg(func, args, kwargs, name: str):
```
Svilen-Stefanov: This function seems like an overkill for reading 2 arguments? 

We call inspect.signature().bind() to dig run_id and depth_level out of the call? 
if anything goes wrong, the bare except just returns None and the property is just not there. 

Since `track_analysis` only ever wraps three known functions that all take run_id as a keyword, can't kwargs.get("run_id") do the same job?

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+    run so a multi-repo loop in one process attributes usage to the right run.
+
+    ``run_id`` (the same id ``monitoring`` uses) is stamped on both events so
+    ``analysis_started`` / ``analysis_completed`` can be paired even when runs
+    overlap in one process.
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        base = {"command": func.__name__.removeprefix("run_"), "version": _app_version()}
```
Svilen-Stefanov: Can't we make our life a bit nicer with Pydantic here. I'm thinking creating a separate module with all events we want to emit and using them something like:
```
from pydantic import BaseModel

class AnalysisCompleted(BaseModel):
    command: str
    version: str
    status: str
    duration_ms: int
    model_name: str | None = None
    total_tokens: int = 0
    input_tokens: int = 0
    output_tokens: int = 0
    run_id: str | None = None
    error_type: str | None = None
    error_message: str | None = None

# at the call site:
event = AnalysisCompleted(command=cmd, version=_app_version(), status=status, ...)
telemetry.capture("analysis_completed", event.model_dump(exclude_none=True))
```

Maybe the command and even other things could be defined / constant, idk.

telemetry/service.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _init(self) -> None:
+        self._client = None
+        self._user_id: str | None = None
+
+        if _telemetry_disabled() or not POSTHOG_PROJECT_API_KEY:
+            return
+
+        try:
+            from posthog import Posthog  # type: ignore[import-not-found]
```
Svilen-Stefanov: Does this need to be a local import?

telemetry/service.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def flush(self) -> None:
+        if self._client is not None:
+            try:
+                self._client.flush()
+            except Exception:
+                pass
+
+
+telemetry = ProductTelemetry()
```
Svilen-Stefanov, line 97: Interesting. Is this where we want this to live?

telemetry/events.py
```diff
@@ -0,0 +1,199 @@
...
+        props["run_id"] = run_id
+    telemetry.capture("repo_scanned", props)
+
+
+def _token_usage() -> dict:
+    """Snapshot of the process-global token counters (best effort)."""
+    try:
+        stats = MONITORING_CALLBACK.stats.to_dict()
+        usage = stats.get("token_usage", {})
+        return {
```
Svilen-Stefanov: The pydantic comment from below applies to all of these.

Svilen-Stefanov (commented):
The code quality here could be better but it seems like it's working. 
If you don't have time to work on my suggestions, we can implement them later, adding what I found so far here.

PR #370, opened 2026-06-09
Cover generator analysis telemetry runs

brovatten:
This PR moves analysis lifecycle telemetry to the DiagramGenerator chokepoints so full and incremental runs are captured consistently across OSS, GitHub Action, VS Code, and other callers. It also keeps partial analysis explicitly instrumented, preserves requested command attribution for incremental fallbacks, and documents the GitHub Action telemetry source.


ivanmilevtues:
Thanks for doing this PR, I ended up incorporating it as feedback for my PR which is already merged.

PR #371, opened 2026-06-09
Add telemetry source attribution

brovatten:
Propagate an explicit CodeBoarding source across core, wrapper, VS Code, and the GitHub Action so telemetry can attribute user-facing entry points. The wrapper keeps source attribution per session and injects it into spawned core processes without relying on mutable global environment state.

## Summary by CodeRabbit

* **Chores**
  * Updated environment context handling in GitHub action workflow
  * Cleaned up internal code comments


ivanmilevtues (approved):
Thanks for adding that. I will merge it directly.

PR #372, opened 2026-06-10
Skip LSP startup when the static-analysis pkl is fresh

brovatten:
Every analyze path paid full LSP startup (workspace-ready waits up to 300s per Java/Rust/C# config, dotnet restore, diagnostics waits) before the pkl was consulted, because `analyze()` requires running clients and only then reads the cache.

This adds `StaticAnalyzer.load_cache_if_fresh` — an LSP-free freshness probe (tag SHA → git diff per engine config, filtered to adapter extensions → unpickle) — and wires it in so a fresh pkl with no relevant changes is served without spawning any LSP:

- `get_static_analysis()` probes before entering the client lifecycle.
- `analyze()` tries the fast path before raising when clients aren't started; behavior with running clients is unchanged.

Guards: removed-language buckets, concurrent re-saves, and probe errors all fall back to the normal LSP path; `CODEBOARDING_DISABLE_CACHE_REUSE` disables the fast path.

## Summary by CodeRabbit

* **New Features**
  * Fast cache path: returns cached analysis results when sources are unchanged, honors the cache-reuse disable switch, and respects skip-cache behavior. CLI now probes the fresh cache first and returns immediately on a hit.
* **Tests**
  * Added comprehensive tests covering cache hits/misses, git probing failures, skip-cache behavior, in-memory memoization, pending-state handling, and CLI fast-path behavior.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -374,18 +374,11 @@ class StaticAnalyzer:
         return cached_results
 
     def load_cache_if_fresh(self, cache_dir: Path) -> StaticAnalysisResults | None:
-        """Serve the SHA-tagged pkl without LSP clients when git reports no relevant changes.
+        """Serve the SHA-tagged pkl without LSP clients, or None when not safely fresh.
 
-        LSP-free freshness probe: load the pkl + tag SHA, ask git which files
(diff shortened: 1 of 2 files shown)


brovatten:
Problematic PR

PR #374, opened 2026-06-10
Gate --binary-location native commands on runnability, self-healing the exec bit

brovatten:
## Problem

#354 fixed the `PermissionError [Errno 13]` on a non-executable gopls for the default servers-dir flow, but the `--binary-location` flow bypasses all of it: `bootstrap_environment` calls `update_config(binary_location)` *instead of* `ensure_tools()` (so no install and no chmod self-heal), and `update_command_paths` rewrote native-tool commands (gopls, csharp-ls, rust-analyzer, tokei) to absolute paths under `<binary_location>/bin/<platform>/` with no exists/X_OK check. A binary left at 0644 by a mode-dropping unpack then reached `Popen` as an absolute path and failed with EACCES — and because the path was absolute, `build_config`'s PATH fallback was disabled, so retries failed deterministically. Reproducible on main by pointing `--binary-location` at `~/.codeboarding/servers`, which yields the exact traceback from the original report. The same bypass is reachable via `health_main` and the Pro wrapper's `_apply_binary_location`/serve path.

## Fix

`update_command_paths` now only commits the absolute path when the binary is actually runnable, via a new `_runnable_native()` gate that:

- requires a regular file (`isfile`, not `exists` — a traversable *directory* named like the tool passes an exists+X_OK check and recreates the same EACCES-with-dead-fallback failure)
- restores a lost exec bit in place (`chmod 0o755`), mirroring the `tool_registry.installers` repair, but refuses to chmod through a symlink since `binary_location` is caller-supplied
- on Windows accepts the `.exe` sibling of the suffix-less path that `CreateProcess` resolves
- keeps the bare command name on failure (logged), so `build_config`'s PATH fallback stays available — failure parity with the default servers-dir path

The candidate is re-derived from the bare tool name on every call, so re-invocation re-validates against the current bin dir instead of trusting a stale absolute path from an earlier call.

The helper lives in `vscode_constants.py` (stdlib-only) rather than `tool_registry.paths` because `tool_registry.manifest` imports this module — importing back would be circular.

## Tests

New `TestUpdateCommandPathsNativeGate` covers: executable binary wired absolute (gopls + tokei, i.e. both config sections), exec-bit repair (with log assertion), missing binary keeps the bare name (POSIX and Windows), unrepairable binary keeps the bare name (chmod failure, with warning assertion), directory and symlink impostors rejected, re-invocation after the old location disappears, and the Windows `.exe`-sibling contract. Exec-bit tests are skipped on Windows hosts where chmod can't drop executability.

Full unit suite passes (1612 passed, 28 skipped); black and mypy clean.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for native command path handling, covering executable binaries, permission repair, missing or non-executable binaries, directory/name collisions, symlink edge cases, reinvocation behavior, and Windows-specific resolution.

* **Improvements**
  * Smarter handling of native command paths: prefer absolute paths only for runnable binaries, fall back to bare command names otherwise, with permission-repair attempts and better Windows compatibility.

### DIFF
--- a/vscode_constants.py
+++ b/vscode_constants.py
@@ -1,5 +1,8 @@
+import logging
 import os
 import platform
+
+logger = logging.getLogger(__name__)
 
 
 def get_bin_path(bin_dir):
@@ -10,6 +13,30 @@ def get_bin_path(bin_dir):
             f"Unsupported platform: {system}. The extension currently supports Windows, macOS, and Linux."
         )
     return os.path.join(bin_dir, "bin", subdirs[system])
+
+
+def _runnable_native(path: str) -> bool:
+    """True when a native binary exists and is executable, restoring a lost exec bit when possible.
+
+    Why: a mode-dropping unpack/copy leaves the binary at 0644; wiring its absolute
+    path into the command guarantees EACCES at Popen and disables the PATH fallback
+    in tool_registry.build_config. Mirrors the repair in tool_registry.installers
+    (not imported from there: tool_registry.manifest imports this module).
+    """
+    if platform.system().lower() == "windows":
+        # No exec bit on Windows; Popen resolves the suffix-less path to .exe itself.
+        return os.path.exists(path) or os.path.exists(path + ".exe")
+    if not os.path.exists(path):
+        return False
+    if os.access(path, os.X_OK):
+        return True
+    try:
+        os.chmod(path, 0o755)
+        logger.info("Restored exec bit on %s", path)
+        return True
+    except OSError:
+        logger.warning("%s exists but is not executable and chmod failed; check ownership/mount", path)
+        return False
 
 
 def update_command_paths(bin_dir):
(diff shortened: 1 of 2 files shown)


vscode_constants.py
```diff
@@ -12,6 +15,32 @@ def get_bin_path(bin_dir):
     return os.path.join(bin_dir, "bin", subdirs[system])
 
 
+def _runnable_native(path: str) -> bool:
+    """True when a native binary is a runnable file, restoring a lost exec bit when possible.
+
+    Why: a 0644 binary wired as an absolute path fails EACCES at Popen with the PATH
+    fallback disabled. Mirrors the tool_registry.installers repair (not imported from
```
Svilen-Stefanov, line 22: The comment here suggest that we're trying to avoid some circular dependency. If that's the case, we should just extract that into a helper that both of the files are calling.

Svilen-Stefanov (commented):
It would be great if we could reproduce this first and then make sure that the changes here indeed fix it.

PR #377, opened 2026-06-11
test: label pytest telemetry source

ivanmilevtues:
## Summary
- Configure pytest runs to set `CODEBOARDING_SOURCE=tests` so telemetry from tests is not classified as core usage.

## Test plan
- [x] `uv run pytest tests/test_telemetry_events.py -q`

## Summary by CodeRabbit

* **Tests**
  * Updated test configuration setup to ensure consistent test environment behavior.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,11 @@
+import os
+
+import pytest
+
+
+os.environ["CODEBOARDING_SOURCE"] = "tests"
+
+
+@pytest.fixture(autouse=True)
+def label_test_telemetry(monkeypatch):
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+import pytest
+
+
+os.environ["CODEBOARDING_SOURCE"] = "tests"
```
Svilen-Stefanov, line 6: Ah, nice, thx!

Svilen-Stefanov (approved):
Thx, good catch.

PR #381, opened 2026-06-15
feat: per-method and per-file content_hash in analysis.json

brovatten:
Records a content fingerprint (hash) of every method body and every whole file into analysis.json, computed while analysis runs and preserved across saves. This lets the diff later detect when a method's code or a file's contents change. Older files without fingerprints fall back to current behavior.

## Release Notes

**New Features**
- Added deterministic content hashing for methods and whole files to support reliable incremental updates.
- Included `content_hash` in analysis outputs and JSON persistence, with restoration during rehydration.
- Incremental scope repopulation now populates method entries with source-derived hash values when available.

**Tests**
- Added/expanded test coverage for hash generation, defaulting behavior, JSON propagation/restoration, and edge cases.

**Chores**
- Regenerated the health report JSON with updated metrics.

### DIFF
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -250,6 +250,10 @@ class MethodEntry(BaseModel):
     start_line: int = Field(description="Starting line number in the file.")
     end_line: int = Field(description="Ending line number in the file.")
     node_type: str = Field(description="Node type name matching NodeType enum (e.g. METHOD, FUNCTION, CLASS).")
+    content_hash: str = Field(
+        default="",
+        description="Truncated SHA-256 of the method's source lines; '' when source was unavailable.",
+    )
 
     def __hash__(self) -> int:
         return hash(self.qualified_name)
@@ -286,6 +290,10 @@ class FileEntry(BaseModel):
     methods: list[MethodEntry] = Field(
         default_factory=list,
         description="Methods and functions in this file, sorted by start line.",
+    )
+    content_hash: str = Field(
+        default="",
+        description="Truncated SHA-256 of the entire file's bytes; '' when source was unavailable.",
     )
 
 
--- a/agents/cluster_methods_mixin.py
+++ b/agents/cluster_methods_mixin.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 import os
 from collections import defaultdict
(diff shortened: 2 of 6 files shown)


agents/agent_responses.py
```diff
@@ -250,6 +250,10 @@ class MethodEntry(BaseModel):
     start_line: int = Field(description="Starting line number in the file.")
     end_line: int = Field(description="Ending line number in the file.")
     node_type: str = Field(description="Node type name matching NodeType enum (e.g. METHOD, FUNCTION, CLASS).")
+    content_hash: str = Field(
```
ivanmilevtues, line 253: This shouldn't got to the LLM and at the moment it will go. We are flooding the  the llm with stiff, I think I added a param, cannot recall at the moment which will filter them for the llm prompt creation

agents/agent_responses.py
```diff
@@ -287,6 +291,10 @@ class FileEntry(BaseModel):
         default_factory=list,
         description="Methods and functions in this file, sorted by start line.",
     )
+    content_hash: str = Field(
+        default="",
+        description="Truncated SHA-256 of the entire file's bytes; '' when source was unavailable.",
```
ivanmilevtues, line 296: This shouldn't got to the LLM and at the moment it will go. We are flooding the the llm with stiff, I think I added a param, cannot recall at the moment which will filter them for the llm prompt creation

agents/cluster_methods_mixin.py
```diff
@@ -40,6 +41,32 @@
...
 
 
+def _read_source_lines(repo_dir: Path, rel_path: str, cache: dict[str, list[str] | None]) -> list[str] | None:
+    """Read and cache a file's lines (repo-relative path). None if unreadable."""
+    if rel_path not in cache:
+        try:
+            # splitlines() normalizes line endings, so trailing-newline-only changes aren't detected.
+            cache[rel_path] = (repo_dir / rel_path).read_text(encoding="utf-8", errors="replace").splitlines()
+        except OSError:
+            cache[rel_path] = None
```
ivanmilevtues, line 51: maybe log something here, so we know it doesn't work - logs or analytics even idk if it is that special for analytics

ivanmilevtues: Or just break the whole thing thing and let it throw the OSS error

agents/cluster_methods_mixin.py
```diff
@@ -40,6 +41,32 @@
...
+    return cache[rel_path]
+
+
+def _hash_method_body(lines: list[str] | None, start_line: int, end_line: int) -> str:
+    """Truncated SHA-256 of source lines [start_line-1:end_line]. '' when unavailable."""
+    if lines is None or start_line < 1 or end_line < start_line:
+        return ""
+    body = "\n".join(lines[start_line - 1 : end_line])
+    return hashlib.sha256(body.encode("utf-8")).hexdigest()[:16]
```
ivanmilevtues, line 60: Why [:16]? Is there a chance for this to be not unique? (I feel like there is and I am not sure what do we win by :16_

diagram_analysis/analysis_json.py
```diff
@@ -94,6 +94,10 @@ class MethodIndexEntry(BaseModel):
     start_line: int = Field(description="Starting line number in the file.")
     end_line: int = Field(description="Ending line number in the file.")
     type: str = Field(description="Node type name (METHOD, FUNCTION, CLASS, ...).")
+    content_hash: str = Field(
```
ivanmilevtues, line 97: Now here these are fine, maybe we can remove them whatsoever from the analysis_results.py and just keep them here.

ivanmilevtues (changes requested):
Awesome, I really dislike sending data to the LLMs which is irrelevant. If that is fixed it is an approve from my side.

Svilen-Stefanov (changes requested):
I agree with Ivan's comments. The changes I imagine should probably happen after the analysis is done, right before we save it.

agents/cluster_methods_mixin.py
brovatten: Highly unlikely, 5 × 10⁻²⁰ to be exact.
Why use many bytes when few bytes do trick - that's the idea

agents/cluster_methods_mixin.py
brovatten: Yup good catch

Svilen-Stefanov:
Superseded by #401, which carries the full git-free content-versioning story (this PR's `content_hash` plus `source_tree_hash`, the incremental bridge, and review fixes). Closing to keep one live PR per repo. New stack: Core #401 → Wrapper CodeBoarding-wrapper#76 → VSCode CodeBoarding-vscode#209, all based on `support-cross-component-relationships`.

PR #383, opened 2026-06-17
refactor: separate LLM-schema models from runtime data models

brovatten:
## Why

Our component / relation / analysis models were doing two jobs at once: they were both the JSON schema we send to the LLM **and** the in-memory object we fill with internal data (component IDs, file/method groups, content hashes, CFG edge counts). Because of that, internal fields kept leaking into the LLM prompt schema — which we'd been masking with per-field hidden/exclude flags and a schema-pruning pass.

## What

Split each model into a small **LLM-facing base** (only the fields the LLM actually produces) and a **runtime subclass** that adds the internal fields. The agents parse the base model and "promote" it to the runtime model right after parsing, so internal data simply can't reach the LLM anymore — no hiding flags, no pruning needed.

Also includes the review fixes for #381: an out-of-bounds guard in the hash helper, a debug log when a source file can't be read, a documented reason for the 16-char (64-bit) hash, an explicit type annotation, and module-scope test imports.

## Notes

- Stacked on top of `feat/content-hash-modification-diff` (base of this PR), not `main`.
- `content_hash` stays on the runtime `MethodEntry`/`FileEntry` (it's load-bearing for the incremental carry-through) — it's now structurally off the LLM path rather than hidden from it.
- Verified: full unit suite green (1661 passed), `mypy` clean, `black` clean.


agents/agent_responses.py
```diff
@@ -162,21 +162,26 @@ def __str__(self):
...
-    src_id: str = Field(default="", description="Component ID of the source.", exclude=True)
-    dst_id: str = Field(default="", description="Component ID of the destination.", exclude=True)
-    edge_count: int = Field(default=0, description="Number of CFG edges backing this relation.", exclude=True)
-    is_static: bool = Field(default=False, description="True if derived from static CFG analysis.", exclude=True)
 
     def llm_str(self):
         return f"({self.src_name}, {self.relation}, {self.dst_name})"
 
 
+class Relation(RelationLLM):
```
Svilen-Stefanov: This seems a bit strange - that a more generic thing `Relation` inherits from a more specific one `RelationLLM`.
If we're splitting them in two, I wonder if the other attributes are even needed here or we just kept them all together because we needed different attributes in different places.

Applies to all other similar classes in this PR.

static_analyzer/cluster_relations.py
```diff
@@ -84,7 +85,7 @@ def build_component_relations(
 
 
 def merge_relations(
-    llm_relations: list[Relation],
+    llm_relations: Sequence[RelationLLM],
```
Svilen-Stefanov: Interesting that we're changing the type here

agents/agent_responses.py
```diff
@@ -313,54 +318,44 @@ class Component(LLMBaseModel):
...
     )
-
     component_id: str = Field(
         default="",
         description="Deterministic unique identifier for this component.",
-        exclude=True,
-        json_schema_extra={"hidden": True},
     )
 
-    def llm_str(self):
```
ivanmilevtues, line 337: Is this unused now?

I think it is used indirectly from the AnalysisInsights.llm_str() so probably it should be implemented.

agents/incremental_agent.py
```diff
@@ -594,7 +595,7 @@ def repopulate_touched_scopes(
 
 
 def _build_node_lookup(
-    static_analysis, cluster_results: dict[str, ClusterResult], repo_dir: Path
+    static_analysis: StaticAnalysisResults, cluster_results: dict[str, ClusterResult], repo_dir: Path
```
Svilen-Stefanov, line 594: Thanks

agents/cluster_methods_mixin.py
```diff
@@ -41,30 +41,41 @@
 logger = logging.getLogger(__name__)
 
 
+# 16 hex chars = 64 bits of SHA-256. The hash only flags whether a method body
+# or file changed between two analyses — not a security or uniqueness guarantee
+# — so 64 bits is ample (collisions are astronomically unlikely across a repo's
+# ~10^4-10^5 entries) while keeping analysis.json compact.
+_HASH_LEN = 16
```
ivanmilevtues: This should be in a constants.py file somewhere.

In the core we keep all the constants in a constatnts.py files

agents/agent_responses.py
ivanmilevtues: This and the other classes which are not AgentResponses, we can put into a different file as this one grew quite a lot, seems like a good moment to shave-off a bit from it.

agents/agent_responses.py
```diff
@@ -570,7 +583,7 @@ def llm_str(self):
 class ScopeRelations(LLMBaseModel):
```
ivanmilevtues, line 275: Probably it is good to rename all LLMTypes to have LLM at the end so that it is consistent. As now I feel like we have AnlaysisInsight and AnalysisInsightLLM and I know that one is the LLM response the other is enhanced with extra data.

With ScopeRelations this isn't clear.

Svilen-Stefanov (commented):
Leaving my comments here, haven't approved yet for 2 reasons:
- I'd like to see if the 2 objects can be independent or if there's another way to do the inheritance
- it seems like you haven't tried running an analysis e2e (or at least it's not documented in the PR description), making sure we're not breaking anything

ivanmilevtues (approved):
Thanks for addressing this 2 real comments:
1. check the agent_reposne.Component.llm_str() comment I think this shoulnd't be delted
2. split them into 2 files

There are few other comments as well.

agents/agent_responses.py
brovatten: I think it only make sense to rename when it has a counterpart (sibling) that is not used by the LLM

brovatten: check other PR comment

static_analyzer/cluster_relations.py
brovatten: changed back

brovatten:
@Svilen-Stefanov I ran it for markitdown and it succeeded

ivanmilevtues:
became stale and it will be fixed in a future pr

PR #384, opened 2026-06-17
Update CodeBoarding workflow configuration

Svilen-Stefanov:
This PR updates your existing [CodeBoarding](https://codeboarding.org) workflow
configuration (`codeboarding.yml` and `codeboarding-sync.yml`) to the current recommended shape. It regenerates the file(s)
so they:

- grant `id-token: write`, so the **free hosted tier** works with no secret, and
- wire the `OPENROUTER_API_KEY` and `CODEBOARDING_LICENSE` repository secrets, so
  adding either one is picked up automatically (the action only reads these from its
  inputs — without the wiring a configured secret would silently go unused).

Only CodeBoarding’s own workflow files are touched. If you’d customized one of these,
review the diff before merging.

— opened for you by CodeBoarding

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflows to enhance CodeBoarding integration with additional authentication credentials.
  * Improved GitHub Actions security configuration by removing workflow-level permissions and implementing job-scoped least-privilege access controls.

### DIFF
--- a/.github/workflows/codeboarding.yml
+++ b/.github/workflows/codeboarding.yml
@@ -2,46 +2,37 @@ name: CodeBoarding review
 
 on:
   pull_request:
-    # Generate once when the PR becomes reviewable; 'closed' only cancels an
-    # in-flight review (see concurrency below), it doesn't start one.
     types: [opened, reopened, ready_for_review, closed]
   issue_comment:
-    # Enables refreshing on demand by commenting /codeboarding.
     types: [created]
 
-permissions:
-  # write lets the action commit the generated .codeboarding/analysis.json back to
-  # the PR branch so the comment links to the webview diff (same-repo PRs). Drop to
-  # `read` to keep the comment without that link.
-  contents: write
-  pull-requests: write
-  issues: write
+# No workflow-level permissions: each job requests only what it needs (least
+# privilege), so the default token starts with none.
+permissions: {}
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/codeboarding.yml
```diff
@@ -2,46 +2,37 @@ name: CodeBoarding review
...
-permissions:
-  # write lets the action commit the generated .codeboarding/analysis.json back to
-  # the PR branch so the comment links to the webview diff (same-repo PRs). Drop to
-  # `read` to keep the comment without that link.
-  contents: write
-  pull-requests: write
-  issues: write
+# No workflow-level permissions: each job requests only what it needs (least
+# privilege), so the default token starts with none.
+permissions: {}
```
ivanmilevtues, line 11: If this is not needed then do we need it in the yml at all?

ivanmilevtues (commented):
My comment here is do we need an empty yml property?

This seems very unintuitive and ultimately I think it is fine to skip it if we are defining it in other scopers.

ivanmilevtues (approved):

.github/workflows/codeboarding.yml
Svilen-Stefanov: There's a semantic difference between omitting it and adding it:
- omitting = uses the repository's default permission set
- explicitly mentioning it = explicitly grants the token no scopes at all at the workflow level, regardless of the repo default

I think both are fine, I don't have a strong preference.

PR #385, opened 2026-06-18
fix: getFileStructure dir default + omit temperature for sampling-free Anthropic models

brovatten:
## Description

Fixes two latent bugs in the LLM path: `getFileStructure` required a positional `dir` even though its args schema defaults to `"."`, so the agent's empty-args tool calls raised `TypeError` and burned the full retry budget before recovering. And `_initialize_llm` always passed `temperature`, which returns a 400 on Anthropic models that removed sampling params (Opus 4.7+/4.8 and the Claude 5 family) — it now omits `temperature` for those models while still sending `temperature=0` to models that accept it (sonnet-4-6/haiku-4-5, OpenAI/Google/etc.).

## How it was tested

Added offline regression tests (`TestFileStructureToolDefaultDir`, `TestTemperatureGating`) and ran the affected suites — 124 passed, 3 skipped, with `mypy` and `black` clean. Verified the temperature behavior against the live Anthropic API with minimal calls: `claude-opus-4-7`/`4-8` return a 400 when `temperature` is sent and succeed without it, while `claude-haiku-4-5`/`sonnet-4-6` accept it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with certain Anthropic Claude models by automatically omitting unsupported sampling parameters (including temperature) when required.

* **New Features**
  * File structure tool now defaults to the project root directory when no directory is specified.

* **Tests**
  * Added coverage to verify correct sampling-parameter handling for Claude model variants.
  * Added coverage to confirm the file structure tool’s default directory behavior.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -25,6 +25,23 @@ logger = logging.getLogger(__name__)
 
 _OPENROUTER_FALLBACK_CONTEXT_WINDOW = ContextWindow(1_048_576, 65_536, is_fallback=True)
 
+# Model families that reject sampling params (temperature/top_p/top_k) with HTTP 400.
+# Why: Anthropic removed them starting with Opus 4.7; sending temperature (even 0) 400s.
+# Substrings match bare IDs and Bedrock-prefixed forms (e.g. us.anthropic.claude-opus-4-8-v1:0).
+_SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS = (
+    "claude-opus-4-7",
+    "claude-opus-4-8",
+    "claude-fable-5",
+    "claude-mythos-5",
+)
+
+
+def _model_accepts_temperature(model_name: str) -> bool:
+    """False for models that reject sampling params; temperature must be omitted for those."""
+    lowered = model_name.lower()
+    return not any(s in lowered for s in _SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS)
+
+
 # ---------------------------------------------------------------------------
 # Module-level model overrides – set once by the orchestrator (main.py) and
 # consumed by initialize_llms() without needing to thread the values through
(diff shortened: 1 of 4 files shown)


agents/llm_config.py
```diff
@@ -25,6 +25,23 @@
 
 _OPENROUTER_FALLBACK_CONTEXT_WINDOW = ContextWindow(1_048_576, 65_536, is_fallback=True)
 
+# Model families that reject sampling params (temperature/top_p/top_k) with HTTP 400.
+# Why: Anthropic removed them starting with Opus 4.7; sending temperature (even 0) 400s.
+# Substrings match bare IDs and Bedrock-prefixed forms (e.g. us.anthropic.claude-opus-4-8-v1:0).
+_SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS = (
+    "claude-opus-4-7",
```
ivanmilevtues, line 32: I am unsure, but I think I observerd this failure with Haiku? (Very possible for me to have messed it up)

agents/llm_config.py
```diff
@@ -25,6 +25,23 @@
...
+    "claude-opus-4-8",
+    "claude-fable-5",
+    "claude-mythos-5",
+)
+
+
+def _model_accepts_temperature(model_name: str) -> bool:
+    """False for models that reject sampling params; temperature must be omitted for those."""
+    lowered = model_name.lower()
+    return not any(s in lowered for s in _SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS)
```
ivanmilevtues, line 42: isnt' there a smarter way to figure this out w/o having to hardcode it. I.e. newer langchain version which somehow supports it under the hood?

Do langchain just expect you to not pass a temperature if you are using one of these models?

It seems like they might: https://github.com/langchain-ai/langchain-aws/issues/1094

ivanmilevtues (changes requested):
Thanks for taking a look, can you just double check 2 things that I asked in the comments and reping me. I just don't want to maintain this hardcoded list of model names because as soon as something new comes out we won't remember to maintain it :D.

Also the pre-commit fails, I noticed after submitting

agents/llm_config.py
brovatten: for me haiku worked

ivanmilevtues (approved):
okay good

PR #392, opened 2026-06-22
fix(static-analysis): exclude machine-generated bindings (fixes tinygrad full-analysis timeout)

brovatten:
## What & why

`codeboarding full` timed out on tinygrad because static-analysis Phase 2 issues one LSP `references` query per symbol position, and tinygrad's machine-generated ctypes bindings (`tinygrad/runtime/autogen/`) inflate that to ~109k positions / ~256k symbols — roughly 3 hours of work that blows the worker's wall-clock deadline. This excludes machine-generated code from analysis via ignore-template path patterns (`**/autogen/**`, `*_pb2.py`, …) plus a content heuristic in `discover_source_files` for large ctypes/clang2py binding modules, cutting tinygrad to ~6.8k positions / ~56k symbols.

## How we tested it

Reran the exact failing command (`codeboarding full --local <tinygrad> --depth-level 2`): it now runs to completion and writes a valid `analysis.json` (6 components) — Phase-2 reference-probe positions dropped **109,435 → 6,862 (16×)** with no degradation (previously it died at 24% after 50 min) — plus new unit tests cover both exclusion layers, including a no-false-positive case for hand-written ctypes code.

## Summary by CodeRabbit

* **New Features**
  * Enhanced detection and exclusion of machine-generated code, including autogen directories and protobuf/gRPC artifacts, for cleaner analysis results.
  * Added filtering of large ctypes binding files during source discovery to prevent analyzing generated bindings.

* **Tests**
  * Added comprehensive test coverage for generated code exclusion patterns and detection logic.


static_analyzer/engine/language_adapter.py
```diff
@@ -18,6 +19,30 @@
...
+    """
+    try:
+        size = path.stat().st_size
+        with path.open("rb") as fh:
+            head = fh.read(4096)
+    except OSError:
+        return False
+    if _GENERATED_MARKER_RE.search(head):
+        return True
+    return b"import ctypes" in head and size > 200_000
```
Svilen-Stefanov: This constant is a bit random. Maybe we should pick a constant with some reasoning behind it and explain that? 
I was thinking about other approaches here but can't directly come up with something amazing - probably needs some more research - one of the alternatives is to see the ratio of ctypes against the rest and try to decide based on that (just makes another type of assumption though - I guess the assumption we make should depend on the target user groups).

static_analyzer/engine/language_adapter.py
```diff
@@ -18,6 +19,30 @@
...
+
+    Why: such files carry enormous symbol counts but no architectural signal, and they make the
+    LSP references phase (one query per symbol position) explode — e.g. tinygrad's
+    ``runtime/autogen`` ctypes bindings are ~82% of its trackable positions. Detected by an
+    explicit generated-header marker, or a large ctypes-binding module (clang2py/ctypesgen style).
+    """
+    try:
+        size = path.stat().st_size
+        with path.open("rb") as fh:
+            head = fh.read(4096)
```
Svilen-Stefanov: Why do we look at the first 4KB only?

static_analyzer/engine/language_adapter.py
```diff
@@ -18,6 +19,30 @@
 
 logger = logging.getLogger(__name__)
 
+_GENERATED_MARKER_RE = re.compile(
```
Svilen-Stefanov: I think that's fine but one thing we should probably decide and try to follow is what to put in `_ALWAYS_IGNORED_DIRS` and what to handle separately like here. E.g not sure if `autogen` and `_pb2` should go into `_ALWAYS_IGNORED_DIRS` actually.

repo_utils/ignore.py
```diff
@@ -110,6 +110,13 @@
 # Generated code
 *.pb.go
 **/generated_parser*
+# Machine-generated bindings (ctypes/clang2py, protobuf, gRPC): enormous symbol
+# counts, no architectural signal, and they explode the LSP references phase.
+**/autogen/**
+**/_autogen/**
+*_pb2.py
+*_pb2.pyi
+*_pb2_grpc.py
```
Svilen-Stefanov, line 119: Do we need `*_pb2_grpc.pyi` here as well?

Svilen-Stefanov (approved):
Left some comments that I think should be addressed before merging. 

One thing I noticed is that the integration tests are failing, which might indicate that we're ignoring more things that we should. Looking at the Java one, it could also be a flakiness we haven't experienced before - but we should definitely look closely into that before merging (make sure there are no regressions). If there are, we could potentially remove the second part of the PR (detecting auto-generated files), only keep the exact ignored dirs and leave the second part for a separate PR. 

Thanks @brovatten!

ivanmilevtues:
inactivity

PR #396, opened 2026-06-25
feat(cli): print webview view-instructions after local analysis

Svilen-Stefanov:
## What

After a local **full**, **partial**, or **incremental** analysis finishes, the CLI now prints clear instructions for viewing the generated `analysis.json` in the hosted webview — open `app.codeboarding.org` and use **"Load a file"** to pick the file.

No upload, no local server, works in every browser.

## Why

We explored auto-opening the result (a short-lived loopback server serving the file to `app.codeboarding.org/?src=...`), but dropped it: the JSON is too large for a URL, the hosted webview that renders `?src=` isn't deployed yet, and current Chrome/Edge **Local Network Access** restrictions make a public-site → `127.0.0.1` fetch unreliable. Manual "Load a file" is the route that works today, everywhere, with zero infrastructure.

## Details

- New `codeboarding_cli/view_instructions.py` with a single `print_view_instructions(analysis_path)` (logs the webview URL + the file path; warns and no-ops if the file is missing; never raises).
- Wired into the end of `full`, `partial`, and `incremental`.
- For `incremental`, the instructions log to **stderr** (via the logger), so the **stdout JSON wire contract** consumed by the VS Code wrapper stays clean — verified.

## Testing

- `uv run pytest --ignore=tests/integration` → **1656 passed, 28 skipped**.
- `uv run mypy .` clean on changed files; `black` clean (pre-commit passing).
- **Manually tested** — this is what's printed at the end of a local analysis run:

```
Analysis ready. View your architecture diagram:

  1. Open the webview:   https://app.codeboarding.org
  2. Click 'Load a file' and pick:
       /Users/path...
```

## Summary by CodeRabbit

* **New Features**
  * Added user-facing instructions after analysis runs to help you open and view generated results in the web interface.
  * Local analysis output now includes a clear path and viewing guidance.

* **Bug Fixes**
  * Improved handling when the generated analysis file is missing, with a helpful warning instead of failing silently.
  * Kept result output available while adding the new viewing message.

### DIFF
--- a/codeboarding_cli/commands/incremental_analysis.py
+++ b/codeboarding_cli/commands/incremental_analysis.py
@@ -6,6 +6,7 @@ from typing import Any
 
 from agents.llm_config import LLMConfigError
 from codeboarding_cli.bootstrap import bootstrap_environment, resolve_local_run_paths
+from codeboarding_cli.view_instructions import print_view_instructions
 from codeboarding_workflows.analysis import BaselineUnavailableError, run_incremental
 from diagram_analysis import RunContext
 from diagram_analysis.io_utils import load_analysis_commit_hash
@@ -119,6 +120,8 @@ def run_from_args(args: argparse.Namespace, parser: argparse.ArgumentParser) -> 
                 "analysis_path": str(analysis_path),
             }
         )
+        # Human-facing hint (logs to stderr, so the stdout JSON contract stays clean).
+        print_view_instructions(analysis_path)
     finally:
         run_context.finalize()
 
--- a/codeboarding_cli/commands/full_analysis.py
+++ b/codeboarding_cli/commands/full_analysis.py
@@ -6,6 +6,7 @@ from tqdm import tqdm
 
 from agents.llm_config import LLMConfigError
 from codeboarding_cli.bootstrap import bootstrap_environment, resolve_local_run_paths
+from codeboarding_cli.view_instructions import print_view_instructions
 from codeboarding_workflows.analysis import run_full
 from codeboarding_workflows.orchestration import run_analysis_pipeline
 from codeboarding_workflows.rendering import render_docs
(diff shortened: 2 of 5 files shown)


codeboarding_cli/commands/incremental_analysis.py
```diff
@@ -119,6 +120,8 @@ def run_from_args(args: argparse.Namespace, parser: argparse.ArgumentParser) ->
                 "analysis_path": str(analysis_path),
             }
         )
+        # Human-facing hint (logs to stderr, so the stdout JSON contract stays clean).
+        print_view_instructions(analysis_path)
```
ivanmilevtues, line 124: It seems better if we print this in the main function of the main file:
```
def main(argv: list[str] | None = None) -> None:
    os.environ.setdefault("CODEBOARDING_SOURCE", "oss")
    if argv is None:
        argv = sys.argv[1:]
    argv = _inject_default_subcommand(list(argv))
    parser = build_parser()
    args = parser.parse_args(argv)
    if args.command == "incremental":
        incremental_analysis.run_from_args(args, parser)
        return
    if args.command == "partial":
        partial_analysis.run_from_args(args, parser)
        return
    full_analysis.run_from_args(args, parser)
    print_view_instructions(args.local_path) # or w.e. the things is.
```

ivanmilevtues (approved):
Generally looks good, my only comment is to put the print in the main function in main.py.

codeboarding_cli/commands/incremental_analysis.py
Svilen-Stefanov: It's slightly more complicated than this, because we also support remote runs (with a link). Here we don't have the full path to the analysis.json yet, we could pass it back but the code changes are slightly larger, which is why I decided for the current design. 
I'm more inclined to merge it as is as we'll probably need to update this later either way but I can show you how it would look like otherwise as well.

PR #399, opened 2026-07-02
First iteration on edges implementation

ivanmilevtues:
## Summary by CodeRabbit

* **New Features**
  * Added a directed “component bridge edges” lookup tool for component-group pairs.
  * Expanded the analysis pipeline with dedicated API-surface and relation-analysis stages.
  * Relations now carry concrete evidence with key/all edge details and call-site metadata, including unified JSON round-tripping.
* **Bug Fixes**
  * Relation merging/validation now uses full per-edge static evidence more consistently, with partial scoring for partially supported results.
* **Tests**
  * Added/updated coverage for bridge edges, relation evidence/serialization, unified JSON import/export, and the revised relation validation behavior.

## Benchmark: depth 2 self-analysis

Command used for each run:

```bash
rm -rf .codeboarding
codeboarding --local "$(pwd)" --depth-level 2 --force
```

Target project: `CodeBoarding` repo analyzing itself. `.codeboarding` was removed before every run.

| Branch | Commit | Run 1 | Run 2 | Average |
| --- | --- | ---: | ---: | ---: |
| `bridge-edge-relations` | `63a1f9f` | 209.829s | 218.222s | 214.026s |
| `main` | `26004f2` | 129.326s | 118.099s | 123.713s |

Result: this branch is about 1.73x as slow as `main`, roughly 73% slower in this run set.


diagram_analysis/analysis_json.py
```diff
@@ -146,6 +169,85 @@ def _method_key(file_path: str, qualified_name: str) -> str:
...
+
+    sites = source_inspector.find_call_sites(file_path)
+    if not sites:
+        return [RelationCallSiteJson()]
+
+    start_line = reference.reference_start_line or 0
+    end_line = reference.reference_end_line or 0
+    matching_sites = [site for site in sites if start_line and end_line and start_line <= site.line <= end_line]
+    selected_sites = matching_sites or sites
+    return [RelationCallSiteJson(line=site.line, column=site.column) for site in selected_sites]
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in c0feb2b by removing the source-inspector lookup/fallback from analysis.json serialization. Relation edges now serialize only their explicit call_sites; if none were produced upstream, the JSON emits an empty array instead of inferring every call site in the file.

agents/prompts/abstract_prompt_factory.py
```diff
@@ -69,3 +69,56 @@ def get_planning_message(self) -> str:
...
+{static_call_evidence}
+
+Create the component relations. Do not limit relations to static calls. First reason from component APIs, runtime dispatch, plugin hooks, REST/queues/files/config, reflection/imports, and data flow. Use static calls only as one evidence source.
+
+For each relation:
+- src_name and dst_name must exactly match component names
+- relation should be a short architectural phrase
+- evidence should concisely explain the communication mechanism
+- key_edges should contain 1-3 important source-to-target code references when possible, similar to key_entities
+- avoid generic implementation-only calls and avoid adding relations solely because a static edge exists"""
```
ivanmilevtues, line 118: These here should be per model prompts, not general like they are at the moment.

agents/cluster_methods_mixin.py
ivanmilevtues, line 891: Please add a comment as a response here @ampagent which shows how does this string look exactly.

agents/relation_edges.py
```diff
@@ -0,0 +1,106 @@
+from agents.agent_responses import Relation, RelationEdge
+
+RelationEdgeIdentity = tuple[
+    str, str, str, str, int | None, int | None, int | None, int | None, tuple[tuple[int, int], ...]
+]
+
+
+def relation_edge_identity(edge: RelationEdge) -> RelationEdgeIdentity:
```
ivanmilevtues: @ampagent this should be a method within the RelationEdge.

agents/relation_edges.py
```diff
@@ -0,0 +1,106 @@
...
+
+def merge_relation_edges(
+    key_edges: list[RelationEdge], all_edges: list[RelationEdge]
+) -> tuple[list[RelationEdge], list[RelationEdge]]:
+    merged_key_edges = unique_relation_edges(key_edges)
+    merged_all_edges = unique_relation_edges([*all_edges, *merged_key_edges])
+    return merged_key_edges, merged_all_edges
+
+
+def relation_pair_key(relation: Relation, fallback_to_names: bool = False) -> tuple[str, str]:
```
ivanmilevtues: this should be a method within the Relation class. @ampagent

agents/validation.py
```diff
@@ -53,6 +56,7 @@ class ValidationResult:
 
     is_valid: bool
     feedback_messages: list[str] = field(default_factory=list)
+    score: float | None = None
```
ivanmilevtues: @ampagent Why is this nullable? Can it be not?

agents/validation.py
```diff
@@ -1,12 +1,13 @@
 """Validation utilities for LLM agent outputs."""
```
ivanmilevtues, line 1: The validation.py has become too big, we need to split it in categoties, maybe per step of validation or have helpers and keep only the public methods here, the ones which are in fact part of the abstraction_agent.py and the details_agent.py and the incremental_agent.py, and add a subpacakge for the validation where we can put all the validation stuff. @ampagent

agents/validation.py
```diff
@@ -613,6 +863,34 @@ def _build_cluster_edge_lookup(
     return cluster_edge_lookup
 
 
+def _component_cluster_ids(components: list[Component], cluster_analysis: ClusterAnalysis) -> dict[str, list[int]]:
```
ivanmilevtues, line 778: cluster related and/or edge related methods should be part of places where that is related to those.
@ampagent

diagram_analysis/analysis_json.py
```diff
@@ -234,24 +302,31 @@ def _hydrate_component_methods_from_refs(
         logger.warning("Missing method index entry for %d ref(s): %s", len(missing), missing)
 
 
-def _relation_to_json(r: Relation) -> RelationJson:
+def _relation_to_json(r: Relation, repo_dir: Path | None = None) -> RelationJson:
```
ivanmilevtues: Is this nullable? Can we do it non-nullable?

@ampagent

diagram_analysis/analysis_json.py
```diff
@@ -234,24 +302,31 @@ def _hydrate_component_methods_from_refs(
...
+        key_edges=[_relation_edge_to_json(edge, repo_dir) for edge in r.key_edges],
         src_id=r.src_id,
         dst_id=r.dst_id,
-        edge_count=r.edge_count,
         is_static=r.is_static,
+        all_edges=[_relation_edge_to_json(edge, repo_dir) for edge in r.all_edges],
     )
 
 
+def _collapse_component_relations(relations: list[Relation]) -> list[Relation]:
```
ivanmilevtues: This method is bad, @ampagent (also please update our AGENTS.MD to be explicit that we shouldn't have single line methods liek this one) + also add it in the review.md

diagram_analysis/analysis_json.py
```diff
@@ -292,13 +369,17 @@ def from_analysis_to_json(
     analysis: AnalysisInsights,
     expandable_components: list[Component],
     sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
+    repo_dir: Path | None = None,
```
ivanmilevtues: @ampagent nullable repo_dir.

diagram_analysis/analysis_json.py
```diff
@@ -365,14 +446,16 @@ def build_unified_analysis_json(
     file_coverage_summary: FileCoverageSummary | None = None,
     commit_hash: str = "",
     snapshot_commit: str | None = None,
+    repo_dir: Path | None = None,
```
ivanmilevtues: @ampagent (why is this nullable).

Can you please check if this is in fact nullable or we should do it in some other form.

static_analyzer/cluster_relations.py
```diff
@@ -21,8 +28,67 @@ class ClusterRelation:
 
     src_cluster_id: str  # component's component_id, e.g. "1.2"
     dst_cluster_id: str  # e.g. "3"
-    edge_count: int = 0  # number of CFG edges crossing this boundary
-    sample_edges: list[tuple[str, str]] = field(default_factory=list)  # representative (src, dst) node pairs
+    all_edges: list[RelationEdge] = field(default_factory=list)
+
+
+def _call_site_int(value: Hashable | None) -> int:
```
ivanmilevtues: can we have this as non-nullable @ampagent

static_analyzer/graph.py
```diff
@@ -109,7 +109,10 @@ def add_call_site(self, call_site: Mapping[str, Hashable]) -> None:
 
     @staticmethod
     def _normalize_call_site(call_site: Mapping[str, Hashable]) -> dict[str, Hashable]:
-        return dict(call_site)
+        normalized = dict(call_site)
+        if "file" not in normalized and "file_path" in normalized:
```
ivanmilevtues, line 113: This seems like backward compatibility we should have onlfy file_path now. Can we check why do we have this check. Maybe it hsould stay unsure @ampagent

static_analyzer/reference_resolve_mixin.py
ivanmilevtues: split into to 2 funcs fix_key_entities_refs and fix_edge_refs @ampagent

agents/prompts/claude_prompts.py
```diff
@@ -334,14 +334,14 @@
...
-5. Define relationships between sub-components (max 2 per component pair; avoid paired sends/returns where ComponentA sends a message to ComponentB and ComponentB returns the result)
+5. Do not define relationships yet; relationships are discovered in a later API-surface step
 6. Provide a one-paragraph description of the subsystem's main flow and purpose
 
 Guidelines for {project_type} projects:
 - Aim for 3-8 final sub-components
 - Merge related cluster groups that serve a common purpose
 - Each sub-component should have clear boundaries
-- Include only architecturally significant relationships
+- Focus on component boundaries; relationships are discovered after components are finalized
```
Svilen-Stefanov, line 344: nit: does it make sense to mention this twice (also part of step 5)

agents/prompts/abstract_prompt_factory.py
```diff
@@ -69,3 +69,56 @@ def get_planning_message(self) -> str:
...
+    def get_api_surfaces_message(self) -> str:
+        return API_SURFACES_MESSAGE
+
+    def get_relation_analysis_message(self) -> str:
+        return RELATION_ANALYSIS_MESSAGE
+
+
+API_SURFACES_MESSAGE = """Analyze component API surfaces for `{project_name}`.
+
+Project Context:
```
Svilen-Stefanov: Don't we have this in any of our prompts? If that's the case, should we at least extract this somewhere? I also wonder if there's a way to share this between prompts so that we don't waste tokens by sending this on almost any prompt.

agents/prompts/gemini_flash_prompts.py
```diff
@@ -309,15 +309,15 @@
...
 - Each sub-component should have clear boundaries
-- Include only architecturally significant relationships
+- Focus on component boundaries; relationships are discovered after components are finalized
 
-Each sub-component must have a clear name, a description of what it does, and reference the exact named cluster groups it encompasses (use exact group names). Include 2-5 key entities per sub-component — the most important classes/methods — mentioning their qualified names and source files. Describe the subsystem's main flow and purpose in one paragraph. Define relationships between sub-components, but limit to at most 2 per component pair and avoid paired sends/returns (e.g. ComponentA sends a message to ComponentB and ComponentB returns the result).
+Each sub-component must have a clear name, a description of what it does, and reference the exact named cluster groups it encompasses (use exact group names). Include 2-5 key entities per sub-component — the most important classes/methods — mentioning their qualified names and source files. Describe the subsystem's main flow and purpose in one paragraph. Do not define relationships yet.
```
Svilen-Stefanov, line 320: Interesting, that this diverges from the claude and deepseek prompts.

AGENTS.md
```diff
@@ -16,7 +16,7 @@
...
 - **Respect project structure**: Code is organized by functional domain (e.g., `agents/`, `static_analyzer/`, `output_generators/`, `monitoring/`). Place new code in the appropriate directory and follow existing module patterns.
 - **Add imports at the top of the file**: avoid function or class level imports. Do not use `if TYPE_CHECKING` import blocks; keep runtime imports explicit and resolve cycles by moving shared types or decoupling modules instead.
-- **Avoid nested/anonymous functions that grow their enclosing function**: don't define inner `def`s or non-trivial `lambda`s that push a function past a screenful or capture enclosing state to do real work. Extract them to module-level (or a method on the relevant type) so the parent stays readable and the helper is testable and reusable. A one-line `lambda` as a `key=`/`sorted()` argument is fine; a multi-branch closure that another block depends on is not.
+- **Avoid trivial pass-through methods**: don't add one-line functions or methods that only call another helper with fixed arguments. Inline the call, or give the helper a real domain responsibility that hides meaningful complexity.
```
Svilen-Stefanov, line 19: Thanks for adding this. I wonder if we should actually have our AGENTS.md and REVIEW.md in a centralized place to share among all of our repos?

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
     src_name: str = Field(description="Source component name")
     dst_name: str = Field(description="Target component name")
+    evidence: str = Field(
+        default="",
+        description=(
+            "Optional concrete evidence for relations that are not direct static calls, "
+            "such as REST endpoints, queues, plugin registration, subprocesses, reflection, or config-driven wiring."
+        ),
+    )
+    key_edges: list[RelationEdge] = Field(
```
Svilen-Stefanov, line 275: src and dst seem a bit redundant here - seems like a roll up to help us identify the component relationships faster but I'm wondering what's a better way to do this. I'll think about this some more and we can discuss it.

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
+        exclude=True,
+        json_schema_extra={"hidden": True},
+    )
 
-    def llm_str(self):
+    def llm_str(self) -> str:
         return f"({self.src_name}, {self.relation}, {self.dst_name})"
 
+    def pair_key(
+        self, fallback_to_names: bool = False, include_relation: bool = False
```
Svilen-Stefanov: just checking the fallback is intended (or if that's an agent decision?)

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
+        src = self.src_id or self.src_name if fallback_to_names else self.src_id
+        dst = self.dst_id or self.dst_name if fallback_to_names else self.dst_id
+        if include_relation:
+            return (src, dst, self.relation)
+        return (src, dst)
+
+    def with_merged_edges(
+        self,
+        *,
+        src_id: str | None = None,
```
Svilen-Stefanov: Should all of these be nullable?

agents/tools/base.py
```diff
@@ -17,6 +19,9 @@ class RepoContext(BaseModel):
     ignore_manager: RepoIgnoreManager
     static_analysis: Optional[StaticAnalysisResults] = None
     changes: ChangeSet | None = None
+    cluster_analysis: ClusterAnalysis = Field(default_factory=lambda: ClusterAnalysis(cluster_components=[]))
```
Svilen-Stefanov, line 22: Hm, I wonder if this should ever have such a default - it seems like it's probably a fallback somewhere, where an error occurred and we don't raise / propagate it up?

agents/tools/base.py
```diff
@@ -17,6 +19,9 @@ class RepoContext(BaseModel):
...
+    cluster_results: dict[str, ClusterResult] = Field(default_factory=dict)
```
Svilen-Stefanov, line 23: nit: I know this is unrelated, just still don't quite like that we use `str` here instead of a proper type. If we have a proper type, we could always validate the `str` even if it's still a str wrapper.

Svilen-Stefanov (commented):
Leaving some initial comments.

diagram_analysis/analysis_json.py
```diff
@@ -142,6 +169,51 @@ def _method_key(file_path: str, qualified_name: str) -> str:
...
+        description=edge.description,
+    )
+
+
+def _relation_edge_from_json(edge: dict, methods_index: dict[str, MethodIndexEntry]) -> RelationEdge:
+    source_key = edge.get("source")
+    target_key = edge.get("target")
+    if not isinstance(source_key, str) or not isinstance(target_key, str):
+        raise ValueError("Relation edge endpoints must be method-index keys")
+    source = methods_index[source_key]
```
Svilen-Stefanov: blocker-ish: we do `methods_index[source_key]` directly here. Won't this KeyError the moment an edge endpoint isn't in the index (external targets like `openai.OpenAI`, or a resolved method not owned by any component)? Feels like it'd crash the whole `parse_unified_analysis`. Can we do `.get()` and skip + log instead, like we already do in `_reconstruct_files_index` lower down? This also bites us in the wrapper since it only catches `ValueError` there.

diagram_analysis/analysis_json.py
```diff
@@ -260,14 +335,21 @@ def from_component_to_json_component(
...
             for c in sub_analysis.components
         ]
+        nested_relations = [
+            _relation_to_json(r, repo_dir)
+            for r in merge_relations_by_pair(
+                sub_analysis.components_relations, fallback_to_names=True, include_relation=True
+            )
+        ]
 
     return ComponentJson(
```
Svilen-Stefanov, line 313: blocker-ish: I might be misreading, but it looks like we compute `nested_relations` here and then never pass it into `ComponentJson(...)`? So the nested `components_relations` always come out empty. Either we're silently dropping data or the whole block is dead. Given we re-project the flat set per level anyway, maybe we just delete the computation?

static_analyzer/cluster_relations.py
```diff
@@ -304,54 +293,67 @@ def merge_relations(
...
-            merged.append(
-                Relation(
-                    relation=llm_rel.relation,
-                    src_name=llm_rel.src_name,
-                    dst_name=llm_rel.dst_name,
-                    src_id=src_id,
-                    dst_id=dst_id,
-                    edge_count=0,
-                    is_static=False,
+            if llm_rel.key_edges:
```
Svilen-Stefanov: blocker-ish: this seems to drop LLM-only relations that have `evidence` but no `key_edges`. Isn't that exactly the case `evidence` was added for? The validator explicitly accepts evidence-only relations, so the two disagree here. Intended?

agents/relation_edges.py
```diff
@@ -0,0 +1,73 @@
+from agents.agent_responses import Relation, RelationEdge
+
+RelationEdgeIdentity = tuple[
+    str, str, str, str, int | None, int | None, int | None, int | None, tuple[tuple[int, int], ...]
+]
+
+
+def relation_edge_identity(edge: RelationEdge) -> RelationEdgeIdentity:
```
Svilen-Stefanov: This whole module feels like thin wrappers forwarding to methods that already live on `Relation` / `RelationEdge` (`relation_edge_identity` -> `edge.identity()`, `unique_relation_edges` -> `Relation._unique_edges`, etc). Can't the callers just use the methods directly? It's also kind of the exact pattern our own REVIEW.md tells us not to do :D

agents/relation_edges.py
```diff
@@ -0,0 +1,73 @@
...
+    return Relation._unique_edges(edges)
+
+
+def merge_relation_edges(
+    key_edges: list[RelationEdge], all_edges: list[RelationEdge]
+) -> tuple[list[RelationEdge], list[RelationEdge]]:
+    return Relation._merge_edges(key_edges, all_edges)
+
+
+def relation_pair_key(relation: Relation, fallback_to_names: bool = False) -> tuple[str, str] | tuple[str, str, str]:
```
Svilen-Stefanov: `relation_pair_key` (and `relation_with_merged_edges` below) don't seem to be called from anywhere outside this module. Is it dead?

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+    )
+
+
+def validate_relation_evidence(result: AnalysisInsights, context: ValidationContext) -> ValidationResult:
+    """Validate LLM relations against directed CFG edges or explicit runtime evidence."""
+    if not context.llm_cluster_analysis or not context.cluster_results or not context.cfg_graphs:
+        logger.warning("[Validation] Missing static context for relation evidence validation")
+        return ValidationResult(is_valid=True)
+
+    components = result.components if hasattr(result, "components") else context.components
```
Svilen-Stefanov: these are typed for `AnalysisInsights` but we actually call them with `ComponentRelations`, which is why we need the `hasattr(result, 'components')` check. Can we fix the annotation to a shared type/protocol instead of duck-typing?

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+def _has_external_unresolved_endpoint(edge, source_node, target_node, context: ValidationContext) -> bool:
+    """Return true when one endpoint is repo code and the other looks external."""
+    if source_node is not None and target_node is None:
+        return not _looks_internal_reference(edge.target, context)
+    if source_node is None and target_node is not None:
+        return not _looks_internal_reference(edge.source, context)
+    return False
+
+
+def _looks_internal_reference(reference, context: ValidationContext) -> bool:
```
Svilen-Stefanov: this "internal reference" heuristic looks copy-pasted verbatim between here and `_looks_internal_reference` in `reference_resolve_mixin.py`. Should we pull it into one shared helper?

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+        for node in static_analysis.iter_reference_nodes(lang):
+            tokens.update(_reference_tokens(node.fully_qualified_name))
+    return tokens
+
+
+def _reference_tokens(qualified_name: str) -> list[str]:
+    return [token.lower() for token in re.split(r"[.:/\\]+", qualified_name) if token]
+
+
+def _same_endpoint_key_edges(relation, context: ValidationContext) -> list[str]:
```
Svilen-Stefanov: `_same_endpoint_key_edges` (and `_source_reference_identity` below) don't seem to be called anywhere. Can we drop them?

static_analyzer/cluster_relations.py
```diff
@@ -22,8 +28,67 @@ class ClusterRelation:
...
+        evidence=evidence,
+        key_edges=[],
+        src_id=src_id,
+        dst_id=dst_id,
+        is_static=is_static,
+        all_edges=unique_relation_edges(edges),
+    )
+
+
+def _append_or_merge_relation(relations: list[Relation], relation: Relation) -> None:
```
Svilen-Stefanov: `_append_or_merge_relation` is a one-line pass-through to the imported `append_or_merge_relation`, and it's used inconsistently. Same wrapper pattern as `relation_edges.py`, could we just call the import directly?

static_analyzer/reference_resolve_mixin.py
```diff
@@ -182,10 +265,143 @@ def _remove_unresolved_references(self, analysis: AnalysisInsights):
...
+    def _attach_static_call_sites(self, edge: RelationEdge) -> None:
+        """Copy call-site metadata from the exact static CFG edge when present."""
+        static_edge = self._find_static_edge(edge)
+        if static_edge is None:
+            return
+        edge.call_sites = [
+            {"line": int(site.get("line", 0)), "column": int(site.get("column", 0))} for site in static_edge.call_sites
+        ]
+
+    def _has_static_edge(self, edge: RelationEdge) -> bool:
```
Svilen-Stefanov: `_has_static_edge` looks like it has no callers. Leftover?

agents/tools/component_bridge_edges.py
```diff
@@ -0,0 +1,84 @@
...
+    source_group_names: list[str] = Field(
+        description="source_group_names for the source component whose outgoing calls should be checked"
+    )
+    destination_group_names: list[str] = Field(
+        description="source_group_names for the destination component whose incoming calls should be checked"
+    )
+
+
+class ComponentBridgeEdgesTool(BaseRepoTool):
+    name: str = "getComponentBridgeEdges"
```
Svilen-Stefanov, line 21: nit: this one tool has like five different names (file `read_component_edges`, class `ComponentBridgeEdgesTool`, input `ComponentBridgeEdgesInput`, tool name `getComponentBridgeEdges`, property `read_component_edges_tool`).  Should we pick more similar names?

agents/abstraction_agent.py
```diff
@@ -130,25 +157,90 @@ def step_final_analysis(
...
+        api_surfaces: ComponentApiSurfaces,
+        cluster_analysis: ClusterAnalysis,
+        cluster_results: dict[str, ClusterResult],
+    ) -> None:
+        logger.info(f"[AbstractionAgent] Discovering component relations for: {self.project_name}")
+        meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
+        project_type = self.meta_context.project_type if self.meta_context else "unknown"
+        static_call_evidence = self.build_scope_cfg_string(analysis)
+        cfg_graphs = self.static_analysis.available_cfgs()
+        self.toolkit.context.cluster_analysis = cluster_analysis
```
Svilen-Stefanov, line 216: we're mutating the shared `RepoContext` in place here just to get data into `read_component_edges`. Feels a bit fragile, is there a way to pass it explicitly instead?

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
         return f"`{self.qualified_name}`:{self.reference_start_line}-{self.reference_end_line}"
 
 
+class RelationEdge(LLMBaseModel):
+    """A source-to-target code reference that supports a component relation."""
+
+    source: SourceCodeReference = Field(description="Source method/class/config reference for this interaction.")
+    target: SourceCodeReference = Field(description="Target method/class/config reference for this interaction.")
+    description: str = Field(default="", description="Short explanation of how source reaches or configures target.")
+    call_sites: list[dict[str, int]] = Field(
```
Svilen-Stefanov: nit: `call_sites` is `list[dict[str,int]]` here but the source CFG edge is `list[dict[str,Hashable]]`, so we end up needing some helpers. Should we align the types?

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
         return f"({self.src_name}, {self.relation}, {self.dst_name})"
 
+    def pair_key(
+        self, fallback_to_names: bool = False, include_relation: bool = False
+    ) -> tuple[str, str] | tuple[str, str, str]:
+        src = self.src_id or self.src_name if fallback_to_names else self.src_id
+        dst = self.dst_id or self.dst_name if fallback_to_names else self.dst_id
+        if include_relation:
+            return (src, dst, self.relation)
+        return (src, dst)
```
Svilen-Stefanov, line 324: Not sure if we shouldn't have a uniform return type here instead of two separate possible ones

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
+        exclude=True,
+        json_schema_extra={"hidden": True},
+    )
+
+    def llm_str(self) -> str:
+        return f"{self.source} -> {self.target}: {self.description}"
+
+    def identity(
+        self,
+    ) -> tuple[str, str, str, str, int | None, int | None, int | None, int | None, tuple[tuple[int, int], ...]]:
```
Svilen-Stefanov: Lol, this return type is a killer :D

agents/agent_responses.py
```diff
@@ -166,20 +167,133 @@ def __str__(self):
...
+        return (src, dst)
+
+    def with_merged_edges(
+        self,
+        *,
+        src_id: str | None = None,
+        dst_id: str | None = None,
+        src_name: str | None = None,
+        dst_name: str | None = None,
+    ) -> "Relation":
```
Svilen-Stefanov: We won't need this if we import `from __future__ import annotations`

agents/validation.py
```diff
@@ -15,14 +16,15 @@
...
-# Coverage validators (cluster + group name) are the most critical — they
-# determine whether every cluster is accounted for and properly assigned.
-# Key-entity validation is secondary (auto-correctable, less structural).
+# Coverage and relation evidence validators are the most critical structural
+# checks. Key-entity validation is secondary (auto-correctable, less structural).
 VALIDATOR_WEIGHTS: dict[str, float] = {
     "validate_cluster_coverage": 20.0,
     "validate_group_name_coverage": 20.0,
     "validate_key_entities": 5.0,
+    "validate_relations": 30.0,
```
Svilen-Stefanov, line 34: Is this just a ballpark number? I don't think we need to benchmark this as long as it works, just wondering if there's anything else to add above.

agents/tools/component_bridge_edges.py
```diff
@@ -0,0 +1,84 @@
...
+    description: str = (
+        "Returns directed static CFG method/function calls between two candidate components. "
+        "Use this before adding a static relationship: pass each component's source_group_names. "
+        "If it returns no bridge edges, only add the relation when you found concrete runtime evidence."
+    )
+    args_schema: ArgsSchema = ComponentBridgeEdgesInput
+
+    def _run(self, source_group_names: list[str], destination_group_names: list[str]) -> str:
```
Svilen-Stefanov, line 29: I see strings everywhere :D

agents/abstraction_agent.py
```diff
@@ -130,25 +157,90 @@ def step_final_analysis(
...
                 validate_group_name_coverage,
                 validate_key_entities,
             ],
             context=context,
             max_validation_attempts=3,
         )
+        return AnalysisInsights(
+            description=architecture.description,
+            components=architecture.components,
+            components_relations=[],
```
Svilen-Stefanov, line 179: I wonder if we populate them later, what's the best design to do this, this should be fine for now, just mentioning.

agents/abstraction_agent.py
```diff
@@ -130,25 +157,90 @@ def step_final_analysis(
...
+            project_name=self.project_name,
+            component_summaries=analysis.llm_str(),
+            static_call_evidence=static_call_evidence,
+            meta_context=meta_context_str,
+            project_type=project_type,
+        )
+        return self._validation_invoke(
+            prompt,
+            ComponentApiSurfaces,
+            validators=[],
```
Svilen-Stefanov, line 198: It's a bit strange we don't have validators but we have a validation context.

agents/abstraction_agent.py
```diff
@@ -166,12 +258,15 @@ def run(self):
...
+        api_surfaces = self.step_api_surfaces(analysis)
+
+        # Step 7: Discover relations from API surfaces and attach deterministic all_edges
+        self.step_relation_analysis(analysis, api_surfaces, cluster_analysis, cluster_results)
 
-        # Step 7: Fix source code reference lines (resolves reference_file paths for key_entities)
+        # Step 8: Fix source code reference lines (resolves reference_file paths for key_entities and key_edges)
         analysis = self.fix_source_code_reference_lines(analysis)
-        # Step 8: Ensure unique key entities across components
+        # Step 9: Ensure unique key entities across components
```
Svilen-Stefanov, line 269: These steps keep piling up, I wonder how the runtime has been increasing with them. Not a real concern, just a thought.

agents/agent.py
```diff
@@ -82,6 +82,10 @@ def read_cfg_tool(self):
     def read_method_invocations_tool(self):
         return self.toolkit.read_method_invocations
 
+    @property
+    def read_component_edges_tool(self):
```
Svilen-Stefanov: Seems a bit scatchy.

agents/agent_responses.py
```diff
@@ -383,6 +497,81 @@ def file_to_component(self) -> dict[str, str]:
...
+    component_name: str = Field(description="Exact component name this API surface describes.")
+    provided_interfaces: list[SourceCodeReference] = Field(
+        default_factory=list,
+        description="Methods/classes/config symbols this component exposes or uses as entrypoints.",
+    )
+    consumed_interfaces: list[SourceCodeReference] = Field(
+        default_factory=list,
+        description="Methods/classes/config symbols this component calls, configures, imports, or expects from others.",
+    )
+    incoming_mechanisms: list[str] = Field(
```
Svilen-Stefanov: I find this phrasing a bit strange - wouldn't communication / apis etc be more understandable?

agents/agent_responses.py
```diff
@@ -383,6 +497,81 @@ def file_to_component(self) -> dict[str, str]:
...
+
+    api_surfaces: list[ComponentApiSurface] = Field(description="API surface for each component in this scope.")
+
+    def llm_str(self):
+        if not self.api_surfaces:
+            return "No component API surfaces found."
+        return "\n".join(surface.llm_str() for surface in self.api_surfaces)
+
+
+class ComponentRelations(LLMBaseModel):
```
Svilen-Stefanov, line 639: nit: looking at this and the above, I wonder if there's some sort of a generic way to wrap such multiple objects so that we don't need a separate class for each one. Not sure if it's gonna simplify things but just thought of it given that this and the above class are almost identical.

agents/details_agent.py
```diff
@@ -176,24 +210,92 @@ def step_final_analysis(
...
+        )
         self._analysis_cache.store(
             cache_key,
             result,
             run_id=self.run_id,
         )
         return result
 
+    @trace
+    def step_api_surfaces(self, analysis: AnalysisInsights) -> ComponentApiSurfaces:
```
Svilen-Stefanov, line 236: I think there's some redundancy between the agents here.

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+        return False
+    if tokens[0] in _internal_reference_roots(context):
+        return True
+    return any(token.startswith("_") and token in _internal_reference_tokens(context) for token in tokens)
+
+
+def _internal_reference_roots(context: ValidationContext) -> set[str]:
+    roots: set[str] = set()
+    for token in _internal_reference_tokens(context):
+        if token not in {"packages", "src", "lib"}:
```
Svilen-Stefanov: What are we doing here? Should these refer to the ignored dirs or?

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+            roots.add(token)
+    return roots
+
+
+def _internal_reference_tokens(context: ValidationContext) -> set[str]:
```
Svilen-Stefanov: Generally not sure if these methods belong here.

agents/validation.py
```diff
@@ -551,7 +563,246 @@ def validate_relation_component_names(result: AnalysisInsights, _context: Valida
...
+                cfg = context.static_analysis.get_cfg(lang)
+            except ValueError:
+                continue
+            for edge in cfg.edges:
+                if edge.get_source() == source_qname and edge.get_destination() == target_qname:
+                    return True
+    return False
+
+
+def _node_identity(node) -> str:
```
Svilen-Stefanov: Don't we have this defined somewhere else?

agents/validation.py
Svilen-Stefanov: Yeah, I think a lot of the methods here might be duplicates, misplaced etc.

diagram_analysis/analysis_json.py
```diff
@@ -142,6 +169,51 @@ def _method_key(file_path: str, qualified_name: str) -> str:
     return f"{file_path}|{qualified_name}"
 
 
+def _source_reference_method_key(reference: SourceCodeReference, repo_dir: Path | None) -> str:
+    file_path = normalize_repo_path(reference.reference_file or "", repo_dir)
+    return _method_key(file_path, reference.qualified_name)
+
+
+def _call_site_from_dict(call_site: dict[str, int]) -> RelationCallSiteJson:
```
Svilen-Stefanov: Should this be a different constructor or is this the right place for this?

diagram_analysis/analysis_json.py
```diff
@@ -142,6 +169,51 @@ def _method_key(file_path: str, qualified_name: str) -> str:
...
+    return RelationCallSiteJson(line=int(call_site.get("line", 0)), column=int(call_site.get("column", 0)))
+
+
+def _relation_edge_to_json(edge: RelationEdge, repo_dir: Path | None) -> RelationEdgeJson:
```
Svilen-Stefanov: Can the path be not-nullable?

diagram_analysis/analysis_json.py
```diff
@@ -142,6 +169,51 @@ def _method_key(file_path: str, qualified_name: str) -> str:
...
+    call_sites = [_call_site_from_dict(site) for site in edge.call_sites]
+    return RelationEdgeJson(
+        source=_source_reference_method_key(edge.source, repo_dir),
+        target=_source_reference_method_key(edge.target, repo_dir),
+        call_sites=call_sites,
+        description=edge.description,
+    )
+
+
+def _relation_edge_from_json(edge: dict, methods_index: dict[str, MethodIndexEntry]) -> RelationEdge:
```
Svilen-Stefanov: nit: better typing for edge?

diagram_analysis/analysis_json.py
```diff
@@ -248,6 +322,7 @@ def from_component_to_json_component(
     expandable_components: list[Component],
     sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
     processed_ids: set[str] | None = None,
+    repo_dir: Path | None = None,
```
Svilen-Stefanov: same here - not nullable ideally, is this something your agent ignores or is it not part of the AGENTS.md?

diagram_analysis/analysis_json.py
```diff
@@ -524,10 +617,12 @@ def _extract_analysis_recursive(
                 relation=r["relation"],
                 src_name=r["src_name"],
                 dst_name=r["dst_name"],
+                evidence=r.get("evidence", ""),
+                key_edges=[_relation_edge_from_json(edge, methods_index or {}) for edge in r.get("key_edges", [])],
```
Svilen-Stefanov: Some silent empty dicts and lists here, checking if it's intended.

diagram_analysis/io_utils.py
```diff
@@ -60,6 +60,13 @@ def __init__(self, output_dir: Path) -> None:
         # routinely steal multi-second cycles from contended writers.
         self._lock = FileLock(output_dir / f"{ANALYSIS_FILENAME}.lock", timeout=30)
 
+    def _repo_dir_for_source_lookup(self) -> Path:
+        if self._output_dir.name == "run-output" and self._output_dir.parent.name == ".codeboarding":
```
Svilen-Stefanov: I think we have a constant for `.codeboarding`, not sure about run-output, what is it actually used for?

static_analyzer/analysis_result.py
```diff
@@ -210,6 +210,14 @@ def get_cfg(self, language: Language) -> CallGraph:
             return bucket.cfg.graph
         raise ValueError(f"Control flow graph for language '{language}' not found in results.")
 
+    def available_cfgs(self) -> dict[str, CallGraph]:
```
Svilen-Stefanov, line 213: Don't we have this elsewhere?

static_analyzer/cluster_relations.py
```diff
@@ -22,8 +28,67 @@ class ClusterRelation:
...
-    sample_edges: list[tuple[str, str]] = field(default_factory=list)  # representative (src, dst) node pairs
+    all_edges: list[RelationEdge] = field(default_factory=list)
+
+
+def _call_site_int(value: Hashable) -> int:
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str):
+        return int(value)
+    return 0
```
Svilen-Stefanov: Checking if this is intended?

static_analyzer/cluster_relations.py
```diff
@@ -22,8 +28,67 @@ class ClusterRelation:
...
+
+
+def _call_sites_from_cfg_edge(edge: Edge) -> list[dict[str, int]]:
+    return [
+        {"line": _call_site_int(call_site.get("line", 0)), "column": _call_site_int(call_site.get("column", 0))}
```
Svilen-Stefanov: Same here, some defaults that I want to double-check.

static_analyzer/cluster_relations.py
```diff
@@ -22,8 +28,67 @@ class ClusterRelation:
...
+            qualified_name=edge.dst_node.fully_qualified_name,
+            reference_file=edge.dst_node.file_path,
+            reference_start_line=edge.dst_node.line_start,
+            reference_end_line=edge.dst_node.line_end,
+        ),
+        call_sites=_call_sites_from_cfg_edge(edge),
+    )
+
+
+def _relation_with_edges(
```
Svilen-Stefanov: Same here - could these be constructors or something?

static_analyzer/graph.py
Svilen-Stefanov: I wonder if `file_path` is present somewhere else and it's confusing agents. Maybe it's worth checking once more?

static_analyzer/reference_resolve_mixin.py
```diff
@@ -182,10 +265,143 @@ def _remove_unresolved_references(self, analysis: AnalysisInsights):
...
+        """Return true when an unresolved name appears to refer to repo code."""
+        tokens = self._symbol_tokens(qualified_name)
+        if not tokens:
+            return False
+        if tokens[0] in self._internal_reference_roots():
+            return True
+        internal_tokens = self._internal_reference_tokens()
+        return any(token.startswith("_") and token in internal_tokens for token in tokens)
+
+    def _internal_reference_roots(self) -> set[str]:
```
Svilen-Stefanov: Some of these look like duplicates to me as well. I think I mentioned above for one of them.

Svilen-Stefanov (approved):
I added loads of comments and marked the ones I think might be blocking with a `blocker-ish` prefix. Approving as this could be merged after the comments are addressed. 

One thing that scares me a bit - there seem to be some very nasty things that could creep in (shadowing methods, multiple definitions and copies) so we should keep those in mind if bugs start to appear and we should probably add something to CodeBoarding itself to catch these.

agents/prompts/claude_prompts.py
ivanmilevtues: hmm I will can skip this comment for now, this said, we need to setup and use evals to be able to say such things and optimize the prompts.

agents/prompts/gemini_flash_prompts.py
ivanmilevtues: It still says that we don't need to do the relationships, they have differet styles I am unsure why it is the case for these to make better sense and be optimized we will need to define an eval.

Further probably that eval should be on a step basis so taht we can optimize steps not the e2e workflow.

AGENTS.md
ivanmilevtues: I was thinking about that but in many cases I work on one repo at a time and then it won't pick up a shared location, we can have rules or w.e. setup claude/codex/amp/opencode can ingest in their system promtps. That has a drawback of always being applied though.

agents/tools/base.py
ivanmilevtues: This is so that the type is not nullable as we instantiate it and set it later on in the agent pipelines, we can make it nullable and fail if it stays like that which I think will never happen by the way we set it in the agents.

static_analyzer/cluster_relations.py
ivanmilevtues: The llm cannot populate all_edges (that is reserved for population from static analysis + key_edges[as they might be not in the stat analysis]), the llm populates key_edges only and the evidence is for those.

To get to this block the logic is:
1. we do not have static edges (edges from stat analysis)
2. we do have in fact key_edges -> add those then.

I will simplify so it is easier to udnerstand.

agents/relation_edges.py
ivanmilevtues: yea good point idk :D, will be fixed maybe as I prompted to put the method in the class beforehand and it did put em but didn't remove the method idk

agents/prompts/abstract_prompt_factory.py
ivanmilevtues: Fixed in ed71ff2: removed this Project Context block from the shared abstract prompt factory. Model-specific prompt wording should stay in the model-specific prompt files, and project context is already part of the system prompt so we should not spend tokens repeating it here.

agents/agent_responses.py
ivanmilevtues: Fixed in ed71ff2: the fallback was just backwards compatibility at rest. `Relation.pair_key()` is now ID-only; no source/destination-name fallback.

agents/agent_responses.py
ivanmilevtues: Fixed in ed71ff2: removed these nullable override parameters entirely. The relation already has knowable `src_id`/`dst_id`/names, so `with_merged_edges()` now uses the values on `Relation` directly.

agents/tools/base.py
ivanmilevtues: I checked this one while addressing the batch. This `str` is not a cluster id; it is the language key for `cluster_results` (`language -> ClusterResult`). So the cluster-id wrapper types are not the right replacement here. I left it unchanged in ed71ff2 rather than swapping in the wrong type.

diagram_analysis/analysis_json.py
ivanmilevtues: Fixed in ed71ff2. `_relation_edge_from_json()` now uses `.get()`, raises a `ValueError` for a missing endpoint, and `_extract_analysis_recursive()` catches that per edge, logs it, and skips the bad edge instead of crashing the whole parse.

diagram_analysis/analysis_json.py
ivanmilevtues: You were right. We computed `nested_relations` and then dropped them. Fixed in ed71ff2 by passing `components_relations=nested_relations` into `ComponentJson`.

static_analyzer/cluster_relations.py
ivanmilevtues: Simplified in ed71ff2. The branch now computes `static_edges = static_rel.all_edges if static_rel else []`, skips only when there are neither static edges nor LLM key_edges, and then uses one append/merge path with `is_static=bool(static_edges)`. Evidence-only relations are still not enough here because this merge path only materializes relations with concrete edge data; validation is the separate place where evidence-only LLM relations are accepted before this projection.

agents/relation_edges.py
ivanmilevtues: Fixed in ed71ff2. Removed the thin wrappers (`relation_edge_identity`, `unique_relation_edges`) and updated callers to use `edge.identity()` / `Relation._unique_edges()` directly.

agents/relation_edges.py
ivanmilevtues: Fixed in ed71ff2. `relation_pair_key` and `relation_with_merged_edges` were dead/pass-through helpers, so I removed them.

agents/relation_edges.py
```diff
@@ -0,0 +1,73 @@
...
+    return relation.pair_key(fallback_to_names=fallback_to_names)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed by deleting this wrapper in ed71ff2 rather than extending it. Callers now use the `Relation` method directly, and the compatibility/fallback plumbing is gone.

agents/validation.py
ivanmilevtues: Fixed in ed71ff2. Added a small `RelationValidationTarget` protocol with `components_relations`, and `validate_relation_evidence()` now uses that instead of pretending everything is `AnalysisInsights` plus `hasattr` duck typing.

agents/validation.py
ivanmilevtues: Fixed in ed71ff2. Pulled the duplicated internal-reference heuristic into `static_analyzer/internal_references.py` and reused it from both validation and `ReferenceResolverMixin`.

agents/validation.py
ivanmilevtues: Fixed in ed71ff2. Both `_same_endpoint_key_edges` and `_source_reference_identity` were unused and have been removed.

static_analyzer/cluster_relations.py
ivanmilevtues: Fixed in ed71ff2. Removed `_append_or_merge_relation` and now call the imported `append_or_merge_relation` directly.

static_analyzer/reference_resolve_mixin.py
ivanmilevtues: Fixed in 8b5091e. `_has_static_edge()` had no callers and was just a leftover wrapper around `_find_static_edge()`, so I removed it.

diagram_analysis/analysis_json.py
ivanmilevtues: Addressed in 8b5091e for the concrete one-line helper pattern here: removed `_to_method_qualified_name()` and inlined `method.qualified_name` at the call sites. REVIEW.md already has the no trivial pass-through methods rule, so I did not add another duplicate rule there.

agents/validation.py
ivanmilevtues: Made a small cleanup in 8b5091e while scanning this area: removed a validation pass-through cluster-edge helper and kept one shared edge-set helper instead. I did not do the broader validation package split in this pass because that is larger than dead-code/leftover cleanup.

agents/tools/component_bridge_edges.py
ivanmilevtues: True this probably is the result of few iteartions.

ivanmilevtues: Fixed in a433cd5. Aligned the internal naming around `component_bridge_edges`: renamed the module to `agents/tools/component_bridge_edges.py`, renamed the toolkit property/cache key to `component_bridge_edges`, and renamed the agent accessor to `component_bridge_edges_tool`. I left the external LangChain tool name `getComponentBridgeEdges` unchanged since that is the model-facing tool name.

agents/abstraction_agent.py
ivanmilevtues: that's a valid question :D.

agents/abstraction_agent.py
ivanmilevtues: Yea we instantiate the tools when we create the agents. So I am unsure how can we do this in runtime. It is quite implicit I agree.

agents/agent_responses.py
ivanmilevtues: Fixed in 7f8986a. Added a named `RelationEdgeIdentity` type alias and changed `RelationEdge.identity()` to return that alias instead of the full tuple inline. I also kept the widened call-site value type and made the coordinate conversion explicit so full-repo mypy stays green.

agents/validation.py
ivanmilevtues: Yes, it is just to draw attention that it is more important than other things.

agents/abstraction_agent.py
ivanmilevtues: true I am dropping the context.

ivanmilevtues: Also setting the attempts to 1.

agents/abstraction_agent.py
ivanmilevtues: I will leave it unresolved so I can add the numbers after finishing with the review.

agents/agent.py
ivanmilevtues: the idea is for the LLM to be able to explore the edges between component without us having to blow-up again the context with the full CFG. So it can potentially explore what is interesting to it.

agents/agent_responses.py
ivanmilevtues: agrred fixed.

agents/agent_responses.py
ivanmilevtues: I think it is hard since the strings are important as they are passed to the LLMs. what we can do is split them into multiple files as we have made this file quite big, but proly another task.

We can add it to our weekly clean-up tasks.

agents/details_agent.py
ivanmilevtues: yes, there is duplicated structure, but generally details agent is a copy which is operating on a scoped subanalysis.
I would avoid flattening it into a root-level/shared workflow that obscures the detail scope, because tmrw we will change one thing in one of the agents and it will affect the other.

Also I think we might want to rework both of them into analysis agent and always have scoping in abstraction have full scope as I find myself optimizing more the global agent since I see it immediately after the generaiton.

agents/validation.py
ivanmilevtues: shame on me, reworked now hopefully it works better now.

Here we are trying to udnerstand if something is referring to an interntal project or to an external dependency and as I was using lots of markitdown the agent overfit to that.

static_analyzer/analysis_result.py
ivanmilevtues: I don't think so, also asked an agent and it said now usnure, hope it is not duplications:d

static_analyzer/cluster_relations.py
ivanmilevtues: Should never happen now it will raise a value error.

ivanmilevtues:
Benchmark results for depth 2 self-analysis are in.

Command used for each run:

```bash
rm -rf .codeboarding
codeboarding --local "$(pwd)" --depth-level 2 --force
```

Target project: the CodeBoarding repo analyzing itself. The .codeboarding directory was removed before every run.

| Branch | Commit | Run 1 | Run 2 | Average |
| --- | --- | ---: | ---: | ---: |
| bridge-edge-relations | 63a1f9f | 209.829s | 218.222s | 214.026s |
| main | 26004f2 | 129.326s | 118.099s | 123.713s |

Result: this branch is about 1.73x as slow as main, roughly 73% slower in this run set.

Also pushed the circular dependency fix in 63a1f9f. The issue was the import cycle through utils -> repo_utils.__init__ -> repo_utils.ignore -> utils. CODEBOARDING_DIR_NAME now lives in constants.py and both utils.py and repo_utils/ignore.py import it from there.

static_analyzer/cluster_relations.py
```diff
@@ -294,64 +221,62 @@ def merge_relations(
...
-                    relation=llm_rel.relation,
-                    src_name=llm_rel.src_name,
-                    dst_name=llm_rel.dst_name,
-                    src_id=src_id,
-                    dst_id=dst_id,
-                    edge_count=0,
-                    is_static=False,
-                )
-            )
+        if not static_edges and not llm_rel.key_edges:
```
Svilen-Stefanov: blocker-ish: won't this silently drop any LLM relation that only has textual evidence (REST/queue/plugin/config wiring) and no matching static edge? `if not static_edges and not llm_rel.key_edges: continue` never looks at `llm_rel.evidence`, and the tail loop only re-adds unmatched static edges. So a `publishes to` relation with evidence but no CFG edge passes `validate_relation_evidence` and then disappears here before it ever reaches analysis.json. That feels like exactly the case the whole evidence field was added for? Could we keep it when `llm_rel.evidence` is set too?

static_analyzer/reference_resolver.py
```diff
@@ -0,0 +1,437 @@
...
+            reference.reference_file = None
+        return False
+
+    def _try_qualified_name_as_path(
+        self, reference: SourceCodeReference, qname: str, lang: Language, file_candidates: list[str] | None = None
+    ) -> bool:
+        file_path = qname.replace(".", os.sep)
+        full_path = os.path.join(self.repo_dir, file_path)
+        file_ref = ".".join(full_path.rsplit(os.sep, 1))
+        paths = [
```
Svilen-Stefanov: nit: this extension list duplicates `LANGUAGE_EXTENSIONS` from constants.py (which we already import from here), and it's already drifted - it's missing .rs / .cs / the C++ ones / .mts/.cts, so a Rust/C#/C++ qname just won't resolve by path. Can't we derive the list from `LANGUAGE_EXTENSIONS` so they can't drift?

diagram_analysis/analysis_json.py
```diff
@@ -142,8 +177,27 @@ def _method_key(file_path: str, qualified_name: str) -> str:
...
 
 
-def _to_method_qualified_name(method: MethodEntry) -> str:
-    return method.qualified_name
+def _source_reference_method_key(reference: SourceCodeReference, repo_dir: Path) -> str:
+    file_path = normalize_repo_path(reference.reference_file or "", repo_dir)
+    return _method_key(file_path, reference.qualified_name)
+
+
+def _call_site_int(value: Hashable) -> int:
```
Svilen-Stefanov: nit: this `_call_site_int` is byte-for-byte the same as `_call_site_coordinate` in agent_responses.py, and there's a third slightly different copy inlined in reference_resolver.py that already disagrees (no str guard). This keeps getting replicated - could we pull it into one shared helper? I'll make a ticket so we don't forget.

agents/agent_responses.py
```diff
@@ -166,20 +170,223 @@ def __str__(self):
...
+    return 0
+
+
+class RelationEdge(LLMBaseModel):
+    """A source-to-target code reference that supports a component relation."""
+
+    source: SourceCodeReference = Field(description="Source method/class/config reference for this interaction.")
+    target: SourceCodeReference = Field(description="Target method/class/config reference for this interaction.")
+    description: str = Field(default="", description="Short explanation of how source reaches or configures target.")
+    call_sites: list[dict[str, Hashable]] = Field(
```
Svilen-Stefanov: nit: `str` (well, `list[dict[str, Hashable]]`) here again where a proper type would validate. We already have `RelationCallSiteJson` for exactly this {line, column} shape. A small typed call-site model would drop the ad-hoc `_call_site_coordinate` coercion sprinkled through from_dict/from_edge/identity. Not for this PR necessarily, but worth a ticket?

static_analyzer/cluster_relations.py
```diff
@@ -40,229 +43,156 @@ def build_node_to_component_map(analysis: AnalysisInsights) -> dict[str, str]:
...
+    candidates = [
+        rel
+        for rel in llm_relations
+        if rel.src_id
+        and rel.dst_id
+        and is_self_or_descendant(src_id, rel.src_id)
+        and is_self_or_descendant(dst_id, rel.dst_id)
+    ]
+    if not candidates:
+        return "calls"
```
Svilen-Stefanov: nit: magic constant `"calls"` here - it's also hardcoded again down in the static-only branch of merge_relations. Could we make it a named constant (something like `DEFAULT_STATIC_RELATION_LABEL`) so the two producers can't drift and the consumers have one thing to match on?

agents/validation.py
```diff
@@ -613,6 +775,42 @@ def _build_cluster_edge_lookup(
...
+    component_to_clusters: dict[str, list[int]] = {}
+    for component in components:
+        cluster_ids: list[int] = []
+        for group_name in component.source_group_names:
+            cluster_ids.extend(group_to_cluster_ids.get(group_name, []))
+        component_to_clusters[component.name] = cluster_ids
+    return component_to_clusters
+
+
+def _check_directed_edge_between_cluster_sets(
```
Svilen-Stefanov: nit: this looks like a thin pass-through around `_cluster_sets_have_edge` with the same args and no extra logic, single caller. Leftover from the refactor? Could the caller just call `_cluster_sets_have_edge` directly and we drop this?

agents/agent_responses.py
```diff
@@ -166,20 +170,223 @@ def __str__(self):
...
+    str, str, str, str, int | None, int | None, int | None, int | None, tuple[tuple[int, int], ...]
+]
+
+
+def _call_site_coordinate(value: Hashable) -> int:
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str):
+        return int(value)
+    return 0
```
Svilen-Stefanov: just checking, is returning `0` for anything that isn't int/str what we want here? These coordinates feed `identity()` (the dedup key) and get serialized, so a silently-zeroed coord could collapse two distinct call sites. I wonder if this should raise on an unexpected type instead of quietly defaulting to 0?

agents/agent_responses.py
Svilen-Stefanov: Pointer to this in case it got lost: 

> src and dst seem a bit redundant here - seems like a roll up to help us identify the component relationships faster but I'm wondering what's a better way to do this. I'll think about this some more and we can discuss it.

agents/abstraction_agent.py
Svilen-Stefanov: I did not expect that increase, seems a bit much but not an issue rn.

Svilen-Stefanov (approved):
Skimmed through it again, leaving some more comments if you want to take a look before merging.

static_analyzer/cluster_relations.py
ivanmilevtues: Good catch, yes this was being dropped in merge_relations even though validation allows relation-level evidence. I updated the condition so evidence-only LLM relations are preserved, and added a warning when we keep a relation that has neither static edges nor key_edges.
I’m still a bit unsure how comfortable we should be with relations where we cannot point to code locations. I.e. REST, we should point to both the endpoint and the client call/config. But I can see why preserving this is useful for runtime/config/plugin-driven wiring, and we can improve the evidence quality over time.

diagram_analysis/analysis_json.py
ivanmilevtues: I promoted the existing coordinate parser to a shared call_site_coordinate() helper and reused it from analysis_json.py, agent_responses.py, and reference_resolver.py. This also makes the reference resolver handle string call-site coordinates consistently with the other two paths.

static_analyzer/cluster_relations.py
ivanmilevtues: done added it to constants

agents/agent_responses.py
ivanmilevtues: yea done, I think i addressed this somewhere else as well 🤔 , it is raising an error, this shouldn't happen.

agents/validation.py
ivanmilevtues: Yep, that was leftover from the refactor. I inlined the single caller to use _cluster_sets_have_edge(...) directly and removed _check_directed_edge_between_cluster_sets

static_analyzer/reference_resolver.py
ivanmilevtues: yepp, fixed.

PR #400, opened 2026-07-04
fix: cancel zombie agent invoke threads on timeout

### LINKED ISSUE #216: Zombie requests
We have an issue in the [agent.py](https://github.com/CodeBoarding/CodeBoarding/blob/2db15d9ebd6530ffbd679aba81f072ec33df8a8e/agents/agent.py#L196), that the `thread.join()` does not actually close the thread. You would think the `raise` does, but it doesn't. Consequently, during high load, we may created multiple zombie threads that queries the model providers. Throttling our key, and creating unnesessary token usage.

Here's an example for a level 3 markitdown generation, that created one of these logs
`2026-02-24 00:09:06 ERROR    [agents.agent:200] Agent invoke thread still running after 300s timeout`
Imagine now for multiple users on the same API key, this could become a problem.
Logs:
https://gist.github.com/brovatten/748cbf0ac7f979c5830f29a5a7ab2c3f

mvanhorn:
## Summary
Timed-out agent invocations no longer leave zombie worker threads querying the provider behind retries. `_invoke` now signals a cancellation event that callback hooks observe at each LLM/tool call boundary, and the retry path waits for the previous worker to actually stop before starting a new attempt, discarding the stale worker's eventual result.

## Why this matters
Issue #216: when an agent invocation times out, the worker thread keeps running the in-flight provider call while `_invoke` retries, so two (or more) threads can hammer the provider simultaneously, burning API quota and occasionally interleaving results from an abandoned attempt. The fix is cooperative: a `threading.Event` per invocation, a callback that raises at `on_llm_start`/`on_chat_model_start`/`on_tool_start` once cancellation is set, and a bounded join on the old worker before any retry proceeds, so a worker blocked inside a provider call can never overlap a retry.

## Testing
Extended `tests/agents/test_agent.py` with cases for cancellation at each callback boundary, the timeout-while-blocked-in-provider-call path (retry held until the worker stops), and discarding a stale worker's late result. `black --check` passes on the touched files.

Fixes #216

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for agent requests, with clearer behavior when work continues past the limit.
  * Ensured late/continued executions no longer trigger extra retries or interfere with subsequent attempts.
  * Updated agent timeout logic to reliably re-raise original errors when failures occur before the limit.
  * Added a consistent LLM request timeout across supported providers to prevent requests from hanging indefinitely.

### DIFF
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Any
 
 from google.api_core.exceptions import ResourceExhausted
-from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.exceptions import OutputParserException
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import SystemMessage, HumanMessage, AIMessage, ToolMessage
@@ -33,41 +32,19 @@ logger = logging.getLogger(__name__)
 
 _AGENT_INVOKE_INITIAL_TIMEOUT_SECONDS = 300.0
 _AGENT_INVOKE_RETRY_TIMEOUT_SECONDS = 600.0
-_TIMED_OUT_THREAD_JOIN_TIMEOUT_SECONDS = 5.0
+_AGENT_RECURSION_LIMIT = 40
 
 
 class EmptyExtractorMessageError(ValueError):
     """Raised when extractor returns an empty message payload."""
 
 
-class _AgentInvocationCancelled(TimeoutError):
-    """Raised inside an abandoned agent thread after caller timeout."""
+class _AgentInvocationControlFlowError(Exception):
+    """Base for invocation control flow that must bypass retries."""
 
 
-class _AgentInvocationStillRunning(TimeoutError):
+class _AgentInvocationStillRunning(_AgentInvocationControlFlowError):
     """Raised when a timed-out invocation worker is still in flight."""
-
-
-class _CancellationCallback(BaseCallbackHandler):
-    """Stops LangGraph at the next LLM or tool boundary after timeout."""
-
-    raise_error = True
-
-    def __init__(self, cancellation_event: threading.Event):
-        self.cancellation_event = cancellation_event
-
-    def _raise_if_cancelled(self) -> None:
-        if self.cancellation_event.is_set():
-            raise _AgentInvocationCancelled("Agent invocation cancelled after timeout")
-
-    def on_llm_start(self, serialized: dict[str, Any], prompts: list[str], **_kwargs: Any) -> None:
-        self._raise_if_cancelled()
-
-    def on_chat_model_start(self, serialized: dict[str, Any], messages: list[list[Any]], **_kwargs: Any) -> None:
-        self._raise_if_cancelled()
-
-    def on_tool_start(self, serialized: dict[str, Any], input_str: str, **_kwargs: Any) -> None:
-        self._raise_if_cancelled()
 
 
 class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
@@ -137,7 +114,7 @@ class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
 
         Classification applied per exception:
         - ``TimeoutError``: backoff ``min(10·2^n, 120)``, raise on exhaustion.
-        - Timed-out workers still running after cancellation: raise immediately.
+        - Timed-out workers still running: raise immediately.
         - ``ResourceExhausted``: backoff ``min(30·2^n, 300)``, raise on exhaustion.
         - ``status_code == 404``: raise immediately (retired model ID, etc.).
         - Other exceptions: backoff ``min(10·2^n, 120)``, return fallback string
@@ -176,7 +153,7 @@ class CodeBoardingAgent(ReferenceResolverMixin, MonitoringMixin):
             if getattr(exc, "status_code", None) == 404:
                 logger.error(f"Permanent HTTP 404 — not retrying: {type(exc).__name__}: {exc}")
                 return RetryDecision(action=RetryAction.GIVE_UP)
-            if isinstance(exc, _AgentInvocationStillRunning):
+            if isinstance(exc, _AgentInvocationControlFlowError):
                 logger.error("Timed-out agent invoke worker is still running; not retrying while it is in flight")
                 return RetryDecision(action=RetryAction.GIVE_UP)
             if isinstance(exc, ResourceExhausted):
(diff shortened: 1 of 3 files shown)


ivanmilevtues:
Hi @mvanhorn thanks for the PR I have assigned myself to review it later today. Thanks for the contribution.

agents/agent.py
```diff
@@ -27,11 +31,45 @@
...
+        self.cancellation_event = cancellation_event
+
+    def _raise_if_cancelled(self) -> None:
+        if self.cancellation_event.is_set():
+            raise _AgentInvocationCancelled("Agent invocation cancelled after timeout")
+
+    def on_llm_start(self, serialized: dict[str, Any], prompts: list[str], **_kwargs: Any) -> None:
+        self._raise_if_cancelled()
+
+    def on_chat_model_start(self, serialized: dict[str, Any], messages: list[list[Any]], **_kwargs: Any) -> None:
```
Svilen-Stefanov: @ivanmilevtues  just so I understand the guarantee here: these hooks fire *before* each LLM/tool call, right? So if the worker is already blocked inside one long provider HTTP call when we set the event, this raise can't fire until that call returns on its own. Which means for the exact case in the issue (thread wedged in a single provider call), we don't actually interrupt it, we just stop the *next* step. I think that's fine, I just want to make sure we're all clear that this is cooperative, not a real kill. See my bigger comment on the join below.

agents/agent.py
```diff
@@ -27,11 +31,45 @@
...
+    raise_error = True
+
+    def __init__(self, cancellation_event: threading.Event):
+        self.cancellation_event = cancellation_event
+
+    def _raise_if_cancelled(self) -> None:
+        if self.cancellation_event.is_set():
+            raise _AgentInvocationCancelled("Agent invocation cancelled after timeout")
+
+    def on_llm_start(self, serialized: dict[str, Any], prompts: list[str], **_kwargs: Any) -> None:
```
Svilen-Stefanov: nit: for chat models langchain dispatches `on_chat_model_start`, not `on_llm_start`, so this one never fires for our agents afaik. Not harmful, just wondering if it's worth keeping?

agents/agent.py
```diff
@@ -27,11 +31,45 @@
...
 
 class EmptyExtractorMessageError(ValueError):
     """Raised when extractor returns an empty message payload."""
 
 
+class _AgentInvocationCancelled(TimeoutError):
+    """Raised inside an abandoned agent thread after caller timeout."""
+
+
+class _AgentInvocationStillRunning(TimeoutError):
```
Svilen-Stefanov: both of these subclass `TimeoutError`, and the whole thing only works because `classify` checks `_AgentInvocationStillRunning` *before* the generic TimeoutError tail. It's correct today but a bit fragile, if someone reorders those branches or adds an `except TimeoutError` upstream it silently changes. 
A potentially more reliable approach would be to give the control-flow ones their own base (not a TimeoutError) so they can't be caught by accident?

agents/agent.py
```diff
@@ -135,6 +176,9 @@ def classify(exc: Exception, attempt: int) -> RetryDecision:
             if getattr(exc, "status_code", None) == 404:
                 logger.error(f"Permanent HTTP 404 — not retrying: {type(exc).__name__}: {exc}")
                 return RetryDecision(action=RetryAction.GIVE_UP)
+            if isinstance(exc, _AgentInvocationStillRunning):
+                logger.error("Timed-out agent invoke worker is still running; not retrying while it is in flight")
+                return RetryDecision(action=RetryAction.GIVE_UP)
```
Svilen-Stefanov, line 158: just checking the behavior change here: before, a timed-out run would retry (up to 5x, which is the zombie-stacking we're fixing, agreed that's bad). Now if the worker won't stop we GIVE_UP and re-raise, and since nothing up in `analysis.py` catches TimeoutError, the whole analysis run dies. I think that's the right call (better to fail loud than stack zombies) but it does mean a slow-but-recoverable call that used to eventually succeed on retry now just fails. 

@ivanmilevtues wdyt?

agents/agent.py
```diff
@@ -186,6 +238,14 @@ def invoke_target():
         if thread.is_alive():
             # Thread is still running - timeout occurred
             logger.error(f"Agent invoke thread still running after {timeout_seconds}s timeout")
+            cancellation_event.set()
+            thread.join(timeout=_TIMED_OUT_THREAD_JOIN_TIMEOUT_SECONDS)
```
Svilen-Stefanov: This is really the main blocker for me. 

Iiuc, the reason we need this whole grace-join + give-up path is that a provider call can block forever, and the only reason it can block forever is that we set `timeout: None` on every provider in `llm_config.py`. If we were to set a real per-request timeout there (say ~120s, something under the 300/600 worker timeouts, `max_retries` is already 0 so the SDK won't silently re-drive it), wouldn't the blocking call itself return with a Timeout at the socket level? 
Then the worker would actually unwind, and this stuck-worker branch becomes a true last resort instead of the expected outcome for a merely slow call? 

@ivanmilevtues wdyt?

agents/agent.py
```diff
@@ -161,22 +205,30 @@ def on_exhausted(exc: Exception) -> str:
...
+        cancellation_event = threading.Event()
+        cancellation_callback = _CancellationCallback(cancellation_event)
+        invoke_callbacks = [*callback_list, cancellation_callback]
 
         def invoke_target():
             try:
                 response = self.agent.invoke(
                     {"messages": [self.system_message, HumanMessage(content=prompt)]},
-                    config={"callbacks": callback_list, "recursion_limit": 40},
+                    config={"callbacks": invoke_callbacks, "recursion_limit": 40},
```
Svilen-Stefanov: nit: `recursion_limit: 40` is still a magic constant, and it's sitting right in the path you just cleaned up with the timeout constants. Could we pull it out next to those as `_AGENT_RECURSION_LIMIT = 40` while we're here? (`max_attempts = 5` and the `404` too, but those are pre-existing, fine to leave.)

Svilen-Stefanov (changes requested):
Thanks for taking this on @mvanhorn and sorry for the delay on my side.
 
The zombie thread issue has been biting us. I read through it and I think the mechanism seems plausible. My main concern is that we are building a fairly involved cooperative-cancellation machine to work around something we could fix at the root. 
We currently pass `timeout: None` to every provider in `llm_config.py`, so a single provider call can block forever. A real per-request timeout there would deterministically unwedge the exact stuck-in-one-HTTP-call case that this whole callback dance can't actually interrupt. Left that as the main comment plus a few smaller ones. 

I'll let @ivanmilevtues chip in as well with his thoughts.

mvanhorn:
Good call - I agree the root fix is the cleaner path. Pushed ddc7838, which sets a real per-request timeout on every provider in `llm_config.py` (a shared `LLM_REQUEST_TIMEOUT_SECONDS = 120`, comfortably under the 300/600s worker timeouts, and `max_retries` is already 0 so a stuck call now surfaces deterministically). With that in place I tore out the cooperative-cancellation machinery it was working around - the `_CancellationCallback` grace-join/give-up dance is gone, and on timeout the worker just raises now.

I also folded in your smaller points: the control-flow exception has its own base (`_AgentInvocationControlFlowError`) so `classify` no longer relies on branch ordering, `recursion_limit` is now `_AGENT_RECURSION_LIMIT`, and the dead `on_llm_start` hook is gone (chat models only fire `on_chat_model_start`). Full suite passes locally. Happy to hear @ivanmilevtues's thoughts too.

PR #401, opened 2026-07-04
feat(core): git-free content versioning + per-method/file content hashes

Svilen-Stefanov:
## What this PR does

Git-free **content versioning** for Core, and a single internal-detection incremental path.

Every file and method in `analysis.json` carries a truncated-SHA-256 `content_hash`, plus a whole-repo `metadata.source_tree_hash`. A full analysis also writes a `fingerprint.json` sidecar (whole analyzable tree). Together these let a consumer answer "have I analyzed this exact source before?" and compute the changed-file set by diffing content-hash maps — no git.

### Incremental analysis

`run_incremental(repo_path, output_dir, project_name, run_id, log_path, …)` detects changes **internally**: it fingerprints the working tree and diffs it against the whole-tree `fingerprint.json` sidecar. No caller passes a changed-file set, git refs, or a source SHA — the CLI (`codeboarding incremental --local <path>`) and the wrapper both just point at a directory. Static-analysis warm-start re-LSPs only the changed files (an empty change set re-LSPs nothing, rather than falling back to a full pass).

**Fail fast, never silently full.** Incremental requires a trustworthy baseline: the whole-tree sidecar. When it is missing (a baseline that predates content versioning, or was never seeded), incremental raises `BaselineUnavailableError` and the wire contract reports `requiresFullAnalysis: true` — the surface layer prompts "run a full analysis first" instead of silently doing a full run or emitting an empty-but-successful update. The baseline can live purely locally (a prior run in the same output dir); committing `.codeboarding/` (`analysis.json` + `fingerprint.json`) lets it travel with a branch or to another machine.

### Content hashing

`agents/content_hash.py` is a dependency-free leaf module (`hash_method_body`, `hash_whole_file`, `read_source_lines`, `MethodRef`/`MethodSpan`/`SourceCache`). Decode/encode use `surrogateescape` so invalid-UTF-8 bytes stay distinct. Method spans and file-index keys are posix-normalized (`normalize_repo_path`) so path lookups match across OSes. `hash_repo_source_files` prunes ignored directories during `os.walk` (never descends into `.git`/`node_modules`).

Hashes are truncated SHA-256 to 16 hex (64 bits). Truncation samples 64 avalanche-mixed bits — it behaves as an ideal 64-bit hash, not a weakened one. The diff compares each method against its own prior hash by `(file, qualified_name)`, so the only failure mode is a method edited to a body colliding with its own old hash: `1/2^64 ≈ 5e-20` per edit. Collisions here are not adversarial, so 16 hex keeps `analysis.json` compact with no practical risk.

### Cleanups

- `ChangeSet` carries only `files` (a content-hash diff has no git refs); git detection stays available via `diff_parser.detect_changes`.
- The ref-only `readGitDiff` LLM tool is removed; the ref-free `listGitChanges` stays.
- `metadata.schema_version` is gone — `content_hash` presence is the real shape gate.

## Testing

`mypy` + `black` clean; `pytest -m "not integration"` passes. Validated end-to-end on a real checkout: full depth-2 → edit → incremental → edit → incremental, each staying `mode: incremental` with a scoped 1-file re-LSP; and a legacy/no-sidecar baseline correctly refusing with a "run a full analysis first" message.

**Incremental eval** (`google/gemini-3-flash-preview`, judge `gemini-3.1-pro`): both small-scope scenarios pass all isolation checks with a perfect structure verdict (1.00) and no fallback to full.

| scenario | full | incremental | speedup |
|---|---|---|---|
| codeboarding-small-scope | 101.7s / 186k tok | 20.6s / 16k tok | ~5x, ~12x fewer tokens |
| markitdown-small-scope | 84.4s / 173k tok | 10.4s / 9k tok | ~8x, ~20x fewer tokens |

Per-method content hashing verified accurate against the real git diff — the edited method's hash changed and untouched methods' did not (e.g. `ProjectScanner.scan` changed, `__init__` unchanged), so a two-state `methods_index` diff yields exactly the modified method set.

## Summary by CodeRabbit

* **New Features**
  * Incremental analysis no longer relies on git references; it detects changes using repository fingerprints from a persisted sidecar.
  * Analysis now stores stable content fingerprints for both files and methods, plus a repo-wide source-tree hash to improve repeatability.

* **Bug Fixes**
  * Incremental runs refresh file/method details from the current checkout to avoid stale spans.
  * Static analysis warm-start is now scoped to the changed files only.

* **Chores**
  * Updated CLI help/tool guidance and README to reflect the git-free incremental workflow and fail-fast baseline requirements.
  * Added/updated tests for fingerprinting, hashing, and incremental scoping.


agents/agent_responses.py
ivanmilevtues: Not from this change but this is in-fact not an agent reponse, we should probably split it and add it into a different file, as my first reaction here was why is this ignored from the LLM -> because it is not sent to the LLM.

agents/agent_responses.py
```diff
@@ -488,6 +492,10 @@ class FileEntry(BaseModel):
         default_factory=list,
         description="Methods and functions in this file, sorted by start line.",
     )
+    content_hash: str = Field(
```
ivanmilevtues: Same comment as for the MethodEntry: Not from this change but this is in-fact not an agent reponse, we should probably split it and add it into a different file, as my first reaction here was why is this ignored from the LLM -> because it is not sent to the LLM.

agents/cluster_methods_mixin.py
```diff
@@ -639,6 +646,11 @@ def _is_more_specific(candidate: str, current: str) -> bool:
                 start_line=node.line_start,
                 end_line=node.line_end,
                 node_type=node.type.name,
+                content_hash=hash_method_body(
```
ivanmilevtues, line 651: hmm really thinking if we are doing ourselves this, maybe we should hash the method body without comments because we either way want to avoid these things as they are irrelevant for the incremental analysis.

However if we add this, we won't show a method diff for the comments when there is such. That's why probably what you did for the extension would also make sense here - such changes are detected runtime in memory because treesitter is fast and cheap.

What do you think on this case when it is a bit incremental func vs diffing?

agents/content_hash.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class MethodSpan(NamedTuple):
+    """A method's 1-based inclusive line range in its file."""
+
+    start_line: int
+    end_line: int
+
+
+def read_source_lines(repo_dir: Path, rel_path: str, cache: SourceCache) -> list[str] | None:
```
ivanmilevtues: Somehow I feel like read_source_lines is not part of content_hash but more on the source_inspector side of things. This said source inspector is for method resolution not full file reading, so unsure. Still feels like it is a standalone thing not part of the hashing.

agents/content_hash.py
```diff
@@ -0,0 +1,65 @@
...
+        try:
+            cache[rel_path] = (
+                (repo_dir / rel_path).read_text(encoding=SOURCE_ENCODING, errors=SOURCE_DECODE_ERRORS).splitlines()
+            )
+        except OSError:
+            cache[rel_path] = None
+    return cache[rel_path]
+
+
+def hash_method_body(lines: list[str] | None, start_line: int, end_line: int) -> str:
```
ivanmilevtues: Give an empty list in instead of None here and make it non-nullable

agents/content_hash.py
```diff
@@ -0,0 +1,65 @@
...
+    Why '' on an out-of-range span: a stale line range must not hash to a stable
+    but meaningless value that compares equal across unrelated code.
+    """
+    if lines is None or start_line < 1 or end_line < start_line or end_line > len(lines):
+        return ""
+    body = "\n".join(lines[start_line - 1 : end_line])
+    return hashlib.sha256(body.encode(SOURCE_ENCODING, errors=SOURCE_DECODE_ERRORS)).hexdigest()[:16]
+
+
+def hash_whole_file(lines: list[str] | None) -> str:
```
ivanmilevtues: pass in an empty list instead of None.

agents/cluster_methods_mixin.py
```diff
@@ -791,25 +803,59 @@ def _log_node_coverage(self, analysis: AnalysisInsights, total_nodes: int) -> No
         logger.info(f"Node coverage: {assigned_nodes}/{total_nodes} ({pct:.1f}%) nodes assigned to components")
 
     def build_files_index(self, analysis: AnalysisInsights) -> dict[str, FileEntry]:
+        file_cache: SourceCache = {}
```
ivanmilevtues: It feels like we should build the file_cache already when we are running the static analysis. Now we are rebuilding it here and also in _build_file_methods_from_nodes.

I also think we should proly have some class for the Hashing and put the file_cache directly there and then we just ask it to give us the hash for method X?

agents/cluster_methods_mixin.py
```diff
@@ -791,25 +803,59 @@ def _log_node_coverage(self, analysis: AnalysisInsights, total_nodes: int) -> No
...
+                            copied.content_hash = ""
+                        methods_by_qname[method.qualified_name] = copied
 
                 entry.methods = sorted(
                     methods_by_qname.values(),
                     key=lambda m: (m.start_line, m.end_line, m.qualified_name),
                 )
         return files
 
+    def _live_cfg_method_spans(self) -> dict[MethodRef, MethodSpan]:
```
ivanmilevtues: kinda blocker: Hmm I don't get why do we need this method. There are 2 cases:
1. Normal analysis (full analysis): this is the initial analysis so the indexing is brand new and hense it should not need to do "updates"
2. Incremental analysis: iiuc the method aims to solve a case here, however I believe that we should start with updating the static analysis in a way which it is up-to-date to start with. If that is th case and the stat analysis is updated -> then we don't need the live_cfg_method.

I see that the point is for untouched (un-updated) components we will get wrong LoC allocations, but if that has changed then there has been a change in fact. So they should be modified. This seems to me as patchwork on a root cause problem - how do we decide if a component has been touched or not? I asked an agent to confirm my theory, and it did confirm imo.

I would either fix it here (which I might prefer :D),  or leave it as part of the incremental v2 (which we can also do there but it might get forgotten, which I think might happen). I think it has to go either way: fix, no fix let it be broken.

I will put it in here unsure if it is useful.
```The disconnect
The incremental flow loads AnalysisInsights from disk (analysis.py:228), where every MethodEntry was serialized with its baseline start_line/end_line/content_hash from methods_index (analysis_json.py:573-580). The CFG gets re-analyzed and IS current — but nobody rebuilds the MethodEntry objects from it for untouched components.
Why not just rebuild all MethodEntry from the up-to-date CFG?
You could — that's exactly what populate_file_methods does. But it requires cluster results to know which method goes to which component, and the incremental flow deliberately skips recomputing clusters for untouched components (that's the whole performance win of incremental). It carries forward the LLM's component→method assignment and only refreshes the metadata (spans, hashes).
So what is _live_cfg_method_spans really doing?
It is the "propagate the up-to-date CFG into the indexing" step you're describing. It's just done lazily inside build_files_index rather than as a separate pre-pass. The refresh has to happen somewhere, and right now this method is that somewhere.
If you wanted to eliminate the method, the architectural alternative would be: after loading the baseline, run an explicit pre-pass that refreshes all MethodEntry spans from the live CFG, then build_files_index could trust them. Same work, different location. But you can't skip the work entirely — the gap between "CFG is current" and "MethodEntry objects are current" is real, and something has to bridge it.```

File: agents/cluster_methods_mixin.py [hunk 8]
```diff
@@ -791,25 +803,59 @@ def _log_node_coverage(self, analysis: AnalysisInsights, total_nodes: int) -> No
         logger.info(f"Node coverage: {assigned_nodes}/{total_nodes} ({pct:.1f}%) nodes assigned to components")
 
     def build_files_index(self, analysis: AnalysisInsights) -> dict[str, FileEntry]:
+        file_cache: SourceCache = {}
+        live_spans = self._live_cfg_method_spans()
```
ivanmilevtues: nit: I would name this just method_spans? Or something like that unsure, live_spans didn't really click to me

agents/cluster_methods_mixin.py
```diff
@@ -791,25 +803,59 @@ def _log_node_coverage(self, analysis: AnalysisInsights, total_nodes: int) -> No
...
                 methods_by_qname = {m.qualified_name: m for m in entry.methods}
                 for method in fmg.methods:
                     if method.qualified_name not in methods_by_qname:
-                        methods_by_qname[method.qualified_name] = method.model_copy(deep=True)
+                        copied = method.model_copy(deep=True)
+                        # Rehash from the live CFG span, not the carried-forward one:
+                        # a stale span can point at unrelated code (edit above the
+                        # method) and a carried hash would mask a body-only edit.
+                        # Absent from the live CFG -> '' rather than hash a stale span.
+                        live_span = live_spans.get(MethodRef(fmg.file_path, method.qualified_name))
```
ivanmilevtues: This is patchwork see the comment on the self._live_cfg_method_spans() method.

codeboarding_workflows/analysis.py
```diff
@@ -174,24 +164,20 @@ def run_incremental(
     project_name: str,
```
ivanmilevtues: I would change these params to accept run_context and run_paths to reduce number of parameters passed around.

codeboarding_workflows/analysis.py
```diff
@@ -200,10 +186,11 @@ def run_incremental(
         raise BaselineUnavailableError(f"No baseline analysis.json found in '{output_dir}'. Run a full analysis first.")
     depth_level = int(metadata.get("depth_level", 1))
 
-    detected = detect_changes(repo_path, base_ref, target_ref)
-    if detected.error:
-        raise BaselineUnavailableError(f"Could not compute diff against baseline {base_ref!r}: {detected.error}")
-    changes = detected
+    changes = detect_changes_from_fingerprint(repo_path, output_dir)
```
ivanmilevtues: Here we can just pass run_paths so it is simpler.

codeboarding_workflows/analysis.py
```diff
@@ -174,24 +164,20 @@ def run_incremental(
...
-    the diff cannot be computed against the given baseline — callers should
-    surface a "run full analysis" prompt rather than silently degrading to an
-    unscoped run.
-    """
-    logger.info(
-        f"Running INCREMENTAL analysis workflow for project '{project_name}' "
-        f"(base={base_ref!r}, target={target_ref!r})."
-    )
+    Change detection is internal and git-free: fingerprint ``repo_path`` and diff
+    it against the baseline hashes in the existing ``analysis.json``. No caller
```
ivanmilevtues: Super simple and straight-forward I like it. iterative on what you have on disk - done.
no tracing up, no figuring out where to trace, vs base or vs target or w.e.
Clean and easy.

codeboarding_workflows/analysis.py
```diff
@@ -216,7 +203,8 @@ def run_incremental(
         static_analyzer=static_analyzer,
         changes=changes,
     )
-    generator.source_sha = source_sha
+    # source_sha (the warm-start tag) is computed by the generator from live
```
ivanmilevtues: Drop these comments?

diagram_analysis/analysis_json.py
```diff
@@ -113,6 +121,10 @@ class MethodIndexEntry(BaseModel):
     start_line: int = Field(description="Starting line number in the file.")
     end_line: int = Field(description="Ending line number in the file.")
     type: str = Field(description="Node type name (METHOD, FUNCTION, CLASS, ...).")
+    content_hash: str = Field(
```
ivanmilevtues, line 117: We should make a new task to unify/simplify and remove half of the data classes as they are the same but different. Can you please do that in Linear (that is a perfect task for an autonomous agent to execute on and for us to review)

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
     return {
         file_path: FileEntryJson(
             method_keys=[_method_key(file_path, m.qualified_name) for m in entry.methods],
+            content_hash=entry.content_hash,
         )
         for file_path, entry in files_index.items()
     }
 
 
+def tree_hash_from_file_hashes(file_hashes: dict[str, str]) -> str:
```
ivanmilevtues: This method should probably go in the hashing methods content_hash.py or if we follow my other comment to make it more like a service call thath HashingService and use it.

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
...
+    skipped so a partial read can't silently collide two different trees. Public
+    so the wrapper can reproduce ``source_tree_hash`` from a fingerprint map.
+    """
+    parts = [f"{path}:{digest}" for path, digest in sorted(file_hashes.items()) if digest]
+    if not parts:
+        return ""
+    return hashlib.sha256("\n".join(parts).encode("utf-8")).hexdigest()
+
+
+def hash_repo_source_files(repo_dir: Path) -> dict[str, str]:
```
ivanmilevtues: Same comment as the one above:
This method should probably go in the hashing methods content_hash.py or if we follow my other comment to make it more like a service call thath HashingService and use it.

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
...
+            content_hash=entry.content_hash,
         )
         for file_path, entry in files_index.items()
     }
 
 
+def tree_hash_from_file_hashes(file_hashes: dict[str, str]) -> str:
+    """SHA-256 over sorted ``path:hash`` lines. '' when no files carry a hash.
+
+    Why: the aggregate source-state key. Files with an unknown ('') hash are
```
ivanmilevtues: I think no need for why :d

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
...
+    parts = [f"{path}:{digest}" for path, digest in sorted(file_hashes.items()) if digest]
+    if not parts:
+        return ""
+    return hashlib.sha256("\n".join(parts).encode("utf-8")).hexdigest()
+
+
+def hash_repo_source_files(repo_dir: Path) -> dict[str, str]:
+    """Fingerprint every non-ignored file under *repo_dir* as ``{posix_path: sha16}``.
+
+    Why: the source-tree hash must cover the WHOLE analyzable tree (docs, configs,
```
ivanmilevtues: I think there is no need for why ;d

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
...
+    unclustered source), not just files that landed in a component. Otherwise a
+    consumer that fingerprints the working tree (the wrapper) can never reproduce
+    this hash. Uses the same ignore rules + whole-file hashing as clustering.
+
+    Ignored directories are pruned during the walk, so ``.git`` / ``node_modules``
+    are never descended into.
+    """
+    ignore = RepoIgnoreManager(repo_dir)
+    result: dict[str, str] = {}
+    for dirpath, dirnames, filenames in os.walk(repo_dir):
```
ivanmilevtues: Hmm I don't like this way of doing this. We should reuse the analyzed files collection we have at the end of the analysis. This way we don't need to rerun the ignore manager again and run only for the analyzed files. Probably this step should also happen around the analysis not in the serialization.

diagram_analysis/analysis_json.py
```diff
@@ -224,11 +238,60 @@ def _build_file_entry_json_from_files(files_index: dict[str, FileEntry]) -> dict
...
+                continue
+            digest = hash_whole_file(lines)
+            if digest:
+                result[rel.as_posix()] = digest
+    return result
+
+
+def compute_source_tree_hash(repo_dir: Path) -> str:
+    """The canonical source-tree version key: whole-repo walk, hashed and aggregated."""
+    return tree_hash_from_file_hashes(hash_repo_source_files(repo_dir))
```
ivanmilevtues: Same comment as above, probably has to go in the hashing service thingy.
Also it is a single line calling I think it is fine to have two methods chained rather than making another one.

diagram_analysis/analysis_json.py
```diff
@@ -400,22 +464,41 @@ def build_unified_analysis_json(
     analysis: AnalysisInsights,
     expandable_components: list[Component],
     repo_name: str,
-    repo_dir: Path,
+    source_root: Path,
     sub_analyses: dict[str, tuple[AnalysisInsights, list[Component]]] | None = None,
     file_coverage_summary: FileCoverageSummary | None = None,
     commit_hash: str = "",
-    snapshot_commit: str | None = None,
+    repo_dir: Path | None = None,
```
ivanmilevtues: Don't really see when repo_dir is different than source_root, this is the only place where we have such separation. Can you revisit this? Is this intentional?

Also repo_dir is nullable it used to be not, so pretty sure we can keep it non-nullable

diagram_analysis/io_utils.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, output_dir: Path) -> None:
         output_dir.mkdir(parents=True, exist_ok=True)
         self._output_dir = output_dir
         self._analysis_path = output_dir / ANALYSIS_FILENAME
+        # Last repo_dir written, reused by write_sub so sub-analysis saves keep
+        # the same whole-tree source_tree_hash the root write computed.
+        self._repo_dir: Path | None = None
```
ivanmilevtues: This shouldn't be nullable we for sure have it, esp if we have output_dir.

We can pass in run_paths rather than dirs here.

diagram_analysis/io_utils.py
```diff
@@ -116,7 +119,7 @@ def write(
         repo_name: str = "",
         file_coverage_summary: FileCoverageSummary | None = None,
         commit_hash: str = "",
-        snapshot_commit: str | None = None,
+        repo_dir: Path | None = None,
```
ivanmilevtues: non-nullable

diagram_analysis/io_utils.py
```diff
@@ -183,9 +197,19 @@ def _write_with_lock_held(
...
     ) -> Path:
         """Write ``analysis.json`` — caller must already hold ``self._lock``."""
+        if repo_dir is not None:
+            self._repo_dir = repo_dir
+        else:
+            repo_dir = self._repo_dir
+        # If no repo_dir is available (e.g. a bare sub-analysis write), preserve the
```
ivanmilevtues: If it is a bare sub-analysis write we still have repo-dir no?

This seems some sort of a backward/interface compat, rather clean up and not have if/else checks like that.

diagram_analysis/io_utils.py
```diff
@@ -238,19 +252,17 @@ def _write_with_lock_held(
...
             expandable_components=expandable,
             repo_name=repo_name,
-            repo_dir=self._repo_dir_for_source_lookup(),
+            source_root=self._repo_dir_for_source_lookup(),
             sub_analyses=sub_analyses_tuples,
             file_coverage_summary=file_coverage_summary,
             commit_hash=commit_hash,
-            snapshot_commit=snapshot_commit,
+            repo_dir=repo_dir,
+            source_tree_hash_override=source_tree_hash_override,
```
ivanmilevtues: I don't see the point of this override, we should always have some sort of hash if we are analyzing something. If it is empty empty I think it is an error - "you are trying to analyze an empty dir" or something like that. The failure will happen much earlier than serialization of the analysis result (I think)

diagram_analysis/io_utils.py
```diff
@@ -166,6 +169,17 @@ def write_sub(
 
             return self._write_with_lock_held(root_analysis, all_expandable_ids, sub_analyses, repo_name)
 
+    def _read_existing_source_tree_hash(self) -> str:
```
ivanmilevtues: I think this will end-up redundant following my comment from line: 207 (https://github.com/CodeBoarding/CodeBoarding/pull/401#discussion_r3558639930)

At this stage (serialization to json) I think we don't need to read/understand the prev hashes, this should care mostly (probably ony) about putting the info into json without corruping the files.

diagram_analysis/diagram_generator.py
```diff
@@ -232,6 +234,21 @@ def _write_file_coverage(self) -> None:
             f.write(report.model_dump_json(indent=2, exclude_none=True))
         logger.info(f"File coverage report written to {coverage_path}")
 
+    def _changed_files_for_static_analysis(self) -> set[Path] | None:
```
ivanmilevtues, line 234: It feels like this should be part of the StaticAnalyzer rather than the diagram generator itself, I see that it is using the self.changes here. I am a bit lazy to figure out how, but it should be internal method for the analyzer, maybe pass the changes in the ctor for the StaticAnalyzer idk.

diagram_analysis/diagram_generator.py
```diff
@@ -448,6 +473,7 @@ def submit_component(comp: Component, lvl: int):
                                 sub_analyses=sub_analyses,
                                 repo_name=self.repo_name,
                                 commit_hash=commit_hash,
+                                repo_dir=self.repo_location,
```
ivanmilevtues, line 476: here again I would pass the Run_paths instead of two dirs, otput and repo. (I wrote a comment somewhere about this API)

diagram_analysis/diagram_generator.py
```diff
@@ -562,12 +588,17 @@ def finalize_and_save(
...
+            repo_dir=self.repo_location,
         ).resolve()
         if seed_delta is not None:
             self._seed_incremental_cluster_cache(seed_delta)
         if persist_side_artifacts:
             self._write_file_coverage()
             self._persist_static_analysis_artifact()
+            # Whole-tree fingerprint sidecar: the next incremental diffs against
+            # the same file set source_tree_hash covers (not the component-only
+            # files block). Same hasher, so the digest matches metadata.source_tree_hash.
```
ivanmilevtues: I'd drop the comment.

diagram_analysis/diagram_generator.py
```diff
@@ -729,8 +760,10 @@ def generate_analysis_incremental(
             )
             if not delta.has_changes:
                 logger.info("Cluster delta is empty; rewriting current analysis without re-detailing.")
-                # No structural change: the loaded baseline's relations already
-                # are the global set, so finalize's rebuild is a no-op here.
+                # No structural change, but a body-only edit still moves content
+                # hashes — refresh the files index from live source so they don't
+                # go stale (relations are already the global set here).
+                self._refresh_files_index(root_analysis, sub_analyses)
```
ivanmilevtues, line 772: if there are no changes why do we need to refresh the files index? This is again relevant to my longer commet from https://github.com/CodeBoarding/CodeBoarding/pull/401#discussion_r3558417730

diagram_analysis/diagram_generator.py
```diff
@@ -815,6 +837,30 @@ def generate_analysis_incremental(
             )
             return analysis_path
 
+    def _refresh_files_index(
```
ivanmilevtues, line 846: Yes this makes sense and is relevant ot what I mentioned here: https://github.com/CodeBoarding/CodeBoarding/pull/401#discussion_r3558417730

I got a but puzzled here why do we need to refresh things now twice. Uhmm I think the other thing has to go :D

diagram_analysis/io_utils.py
```diff
@@ -350,29 +362,32 @@ def load_analysis_commit_hash(output_dir: Path) -> str | None:
     return sha if isinstance(sha, str) and sha else None
 
 
-def load_snapshot_commit(output_dir: Path) -> str | None:
-    """Return the top-level ``snapshotCommit`` from live ``analysis.json``.
+# Whole-tree fingerprint sidecar. analysis.json's ``files`` block covers only
+# component-assigned files; the sidecar covers the whole analyzable tree, so the
+# incremental diff also sees changes to docs/configs/unclustered source.
+FINGERPRINT_FILENAME = "fingerprint.json"
```
ivanmilevtues: This constant has to go with ANALYSIS_FILENAME somewhere else, atm they are in `utils.py` there was a campaign to put them in a constants files, maybe we should maybe just put it with the others both ar fine by me.

repo_utils/diff_parser.py
```diff
@@ -43,12 +43,12 @@ def detect_changes(
     try:
```
ivanmilevtues: I think this method now is dead, and when removed will introduce more dead code so please double check.

ivanmilevtues: I suspect the whole file to go after this type of change.

static_analyzer/__init__.py
```diff
@@ -496,17 +496,25 @@ def analyze(
         skip_cache: bool = False,
         source_sha: str | None = None,
         persist_cache: bool = True,
+        changed_files: set[Path] | None = None,
```
ivanmilevtues: I think this should either come in consturctor style or be computed internally and not passed into the analyze. (also leave it be empty and not none)

static_analyzer/__init__.py
```diff
@@ -628,14 +646,19 @@ def _update_cached_results(self, cached_results: StaticAnalysisResults, cached_s
...
-                    "falling back to full re-LSP for this language"
-                )
-                changed_files = None
+            if supplied_changed_files is not None:
+                # Scope the repo-wide set to this language's project root so a
+                # multi-language repo doesn't re-LSP every changed file per engine.
+                changed_files = {f for f in supplied_changed_files if f.is_relative_to(project_path)}
+            else:
+                try:
+                    changed_files = set(get_changed_files_since(project_path, cached_sha))
```
ivanmilevtues: This I think should be part of the way we get the changed files, which imo should be via ctor or something else as I commented around somewhere else as well. So that the update_cached_results just uses the changed_files and we don't have the if/elses and then the whole supply that goes 10 methods deep from the diagram_generator.py to here.

tests/codeboarding_cli/test_incremental_analysis.py
```diff
@@ -11,136 +11,42 @@
...
+    assert kwargs["repo_path"] == tmp_path
+    assert kwargs["output_dir"] == tmp_path / ".codeboarding"
+    # No change-detection inputs cross the boundary — Core detects internally.
+    assert "base_ref" not in kwargs
+    assert "target_ref" not in kwargs
+    assert "changes" not in kwargs
+    assert "source_sha" not in kwargs
+
+
+def test_incremental_no_git_ref_flags_in_parser() -> None:
```
ivanmilevtues: pointless test I think?

tests/codeboarding_cli/test_incremental_analysis.py
```diff
@@ -11,136 +11,42 @@
...
-
-
-def test_incremental_flags_default_to_none_in_parser() -> None:
-    args = build_parser().parse_args(["incremental", "--local", "/tmp/repo"])
-    assert args.base_ref is None
-    assert args.target_ref is None
+    assert kwargs["repo_path"] == tmp_path
+    assert kwargs["output_dir"] == tmp_path / ".codeboarding"
+    # No change-detection inputs cross the boundary — Core detects internally.
+    assert "base_ref" not in kwargs
```
ivanmilevtues: pointless asserts imo

AGENTS.md
```diff
@@ -26,12 +26,17 @@
...
 - **Language Server Protocol (LSP) integration**: Static analysis runs via LSP servers configured in `static_analysis_config.yml`. Supported languages include Python, TypeScript, Go, PHP, and Java. LSP servers are installed by `setup.py`.
 
-### 5. Git Workflow and Branch Management
+### 5. Fail Fast — Never Silently Degrade on Unexpected State
+- **Raise, don't silently fall back**: when an operation cannot do what the caller asked — and doing something *else* would have large, surprising impact — raise a specific exception instead of quietly substituting different behavior. The caller decides how to recover.
+- **Incremental never silently becomes full**: if incremental analysis cannot proceed (no baseline, a baseline that predates content versioning, an unindexable/unparseable file, an empty or untrustworthy change set), raise (`BaselineUnavailableError` / `IncrementalCacheMissingError`) so the surface layer can prompt "run a full analysis first". Do NOT run a full analysis, return an empty-but-successful update, or mark every method deleted. A silent full run discards the existing depth/component IDs and misleads the user into thinking the incremental succeeded.
```
ivanmilevtues, line 32: That is nice

AGENTS.md
```diff
@@ -26,12 +26,17 @@
...
+- **Prefer a loud, specific error over a plausible-but-wrong result**: a missed change that the user never learns about is worse than a clear failure they can act on. Reserve graceful degradation for genuinely non-critical, clearly-logged paths.
```
ivanmilevtues, line 33: That is also good

README.md
```diff
@@ -137,6 +137,14 @@ python main.py partial --local ./my-project --component-id "1.2"
 python main.py full https://github.com/pytorch/pytorch
 ```
 
+> **Incremental needs a baseline.** `incremental` diffs the working tree against the previous
```
ivanmilevtues, line 140: Yep good addition.

ivanmilevtues (changes requested):
Great work here.

Main comments are stylistic and few on patchwork.

Main concern is around re-patching things around incremental analysis. There is some dead code left.

Anyway great change and simplification on how we do analysis (based on what you have on disk, dead simple, done).

diagram_analysis/analysis_json.py
Svilen-Stefanov: Dropped the why :D

agents/content_hash.py
Svilen-Stefanov: Done, read_source_lines returns [] now and the hashers take a non-nullable list[str]. SourceCache is dict[str, list[str]] too.

agents/content_hash.py
Svilen-Stefanov: Yep done, same as above, empty list instead of None.

agents/content_hash.py
Svilen-Stefanov: I kept it in content_hash.py for now. It is the reader behind both hash_method_body and hash_whole_file and shares the SourceCache type, so it felt more at home with the hashers than in source_inspector (which is method resolution, not file reading). Can move it later if it bugs you.

agents/cluster_methods_mixin.py
Svilen-Stefanov: I left the full-body hash for now. The treesitter runtime-diff you mention is the nicer long term answer imo though, so I'd rather do it there than fork the hashing here. Let's keep it for now and revisit in incremental v2, wdyt?

diagram_analysis/analysis_json.py
Svilen-Stefanov: Moved tree_hash_from_file_hashes, hash_repo_source_files and compute_source_tree_hash into content_hash.py. Bonus, fingerprint_diff now imports from the leaf module instead of pulling in analysis_json.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Moved this one too, same place.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Moved. And agree it is a single-line chain, so I kept compute_source_tree_hash as the two-method chain you suggested rather than inlining.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Dropped ;d

diagram_analysis/analysis_json.py
Svilen-Stefanov: Half done, and good question on the wrapper. First half: the walk now happens once in pre_analysis (around the analysis, not in serialization) and the fingerprint is reused for source_sha, the sidecar and every save, so no more re-walking per save.

On reusing the analyzed-files set: the reason it can't be that isn't really the wrapper doing its own thing, the wrapper actually calls Core's hash_repo_source_files directly (it delegates so the maps stay byte-identical). The real reason is timing: the wrapper fingerprints the frozen source BEFORE running any analysis, precisely to answer have I analyzed this exact source before without paying for the analysis. At that point there is no analyzed-files list yet, it only exists after the LSP pass. So even if we updated the wrapper, it has nothing to reuse there. Plus the sidecar is meant to cover the whole tree (docs, configs, unclustered source), and the analyzed set is only the LSP'd source, so the two would stop matching. So I kept the walk but made it run once. Ok with that, or do you see a way around the timing?

Side note, I moved these hashers to agents/content_hash.py in this PR and the wrapper imported them from analysis_json, so I fixed that import on the wrapper branch too.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Good catch, they were always the same repo root. Collapsed to one non-nullable repo_dir. repo_dir now only relativizes paths, and the source_tree_hash comes in precomputed, so the override machinery is gone.

diagram_analysis/io_utils.py
Svilen-Stefanov: Made it non-nullable and dropped the _repo_dir cache. write_sub recovers the root from the output-dir layout, so we always have it.

diagram_analysis/io_utils.py
Svilen-Stefanov: Done, non-nullable now.

diagram_analysis/io_utils.py
Svilen-Stefanov: Right, we do have it. Removed the if/else, write_sub carries the repo_dir forward from the output-dir layout so there is no bare-write branch anymore.

diagram_analysis/io_utils.py
Svilen-Stefanov: Agreed, dropped the override. It is a single precomputed source_tree_hash passed in now, no empty fallback. write_sub carries the prior hash forward from the on-disk metadata it already reads, so no extra I/O.

diagram_analysis/io_utils.py
Svilen-Stefanov: Yep this went away with the override. write_sub now just reuses the source_tree_hash already in the on-disk metadata, no re-reading logic needed.

diagram_analysis/io_utils.py
Svilen-Stefanov: Moved FINGERPRINT_FILENAME to utils.py next to ANALYSIS_FILENAME.

diagram_analysis/analysis_json.py
Svilen-Stefanov: Good idea. I split MethodEntry/FileEntry/FileMethodGroup out of agent_responses into agents/file_index_models.py in this PR (they are internal, never sent to the LLM, which is what confused you below). The deeper unify with MethodIndexEntry/FileEntryJson I left as a ticket, will create it in Linear.

agents/agent_responses.py
Svilen-Stefanov: Agree, these aren't agent responses. Moved MethodEntry, FileEntry and FileMethodGroup into agents/file_index_models.py so the LLM-response models and the internal index models don't live in the same file anymore.

agents/agent_responses.py
Svilen-Stefanov: Same, moved FileEntry out with MethodEntry into agents/file_index_models.py.

agents/cluster_methods_mixin.py
Svilen-Stefanov: You were right, this was patchwork. Fixed at the root. build_files_index is a pure assembler now, it hashes each method at the span it carries, no CFG lookup. The carry-forward span refresh is one explicit pre-pass, refresh_method_spans_from_cfg, run once before the index build. _live_cfg_method_spans and the two-layer refresh are gone.

agents/cluster_methods_mixin.py
Svilen-Stefanov: Fixed, see the reply on _live_cfg_method_spans. The rehash-from-live-CFG logic moved out of build_files_index into the explicit refresh pre-pass.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: The refresh is still needed even with no structural change because a body-only edit moves content hashes while relations stay the global set. But it is one explicit pre-pass now (refresh_method_spans_from_cfg then build_files_index), not the hidden double-refresh from before.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Yep, the other thing went :D There is a single refresh path now, refresh_method_spans_from_cfg feeds build_files_index in one place.

agents/cluster_methods_mixin.py
Svilen-Stefanov: Renamed. The method is _cfg_method_spans now and it feeds refresh_method_spans_from_cfg, so live_spans as a name is gone.

agents/cluster_methods_mixin.py
Svilen-Stefanov: Good point, this was reading every file twice. There is one shared SourceCache now threaded through _build_file_methods_from_nodes and build_files_index, so each file is read once per populate pass. I didn't go as far as building it during static analysis, that felt like a bigger change, but the double-read is gone. A proper HashingService we can do as a follow-up if you like.

static_analyzer/__init__.py
Svilen-Stefanov: Moved it to the StaticAnalyzer constructor and left it as an empty-vs-None set. analyze() no longer takes changed_files.

static_analyzer/__init__.py
Svilen-Stefanov: Done. The supplied-vs-git branch is now a small helper, _changed_files_for_language, and analyze() reads self.changed_files, so no more threading it 10 methods deep from the generator.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Moved the ownership into StaticAnalyzer (changed_files is a ctor arg now). I kept _changed_files_for_static_analysis on the generator though, since it is the translation from the generator's ChangeSet to abs paths, that way StaticAnalyzer stays decoupled from the ChangeSet type.

codeboarding_workflows/analysis.py
Svilen-Stefanov: Done, run_incremental takes run_paths and run_context now. Turned out RunContext already existed and LocalRunPaths was basically RunPaths, so I moved RunPaths next to RunContext and folded LocalRunPaths into it.

codeboarding_workflows/analysis.py
Svilen-Stefanov: Yep, it reads run_paths.repo_path / run_paths.output_dir now.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Done here too, save happens via the generator which now holds run_paths, so no more passing the two dirs around at the call sites.

codeboarding_workflows/analysis.py
Svilen-Stefanov: Dropped.

diagram_analysis/diagram_generator.py
Svilen-Stefanov: Dropped it, well, compressed it to one line since the whole-tree vs component-only distinction is a bit non-obvious. Can drop entirely if you'd rather.

repo_utils/diff_parser.py
Svilen-Stefanov: Confirmed dead, only tests called it. Removed the whole file, and followed the cascade, the git_ops diff primitives it fed (run_raw_diff, fetch_all, list_untracked_files) are gone too with their tests and the stale doc refs.

tests/codeboarding_cli/test_incremental_analysis.py
Svilen-Stefanov: Dropped the not-in-kwargs asserts, agree they weren't proving much.

tests/codeboarding_cli/test_incremental_analysis.py
Svilen-Stefanov: Removed that test.

diagram_analysis/diagram_generator.py
ivanmilevtues: nice thanks.

agents/cluster_methods_mixin.py
ivanmilevtues: love this phrase: `pure assembler now `

agents/cluster_methods_mixin.py
ivanmilevtues: Yep we should not forget in v2 thou :D, so we should add some context in the task.

diagram_analysis/analysis_json.py
ivanmilevtues: I am fine with merging like that. 

On the timing issue, are we not analyzing on the frozen set of files, as otherwise I think there is not too much to freeze about. If we are not analyzing the frozen set we are also not analyzing the set of files for whcih we did the sha diff?

ivanmilevtues (approved):
Thanks for addressing, I have answered my thoughts (shallow) on the problem around timing issues pre/post freezing for sha checkup.

PR #402, opened 2026-07-04
Add call-site metadata to static analysis edges

ivanmilevtues:
Ivan:
This seems like a big change but in fact I am adding/saving infromation which we weren't. That is the call site infromation when collecting edges:
https://github.com/CodeBoarding/CodeBoarding/pull/402#discussion_r3523367633
and
https://github.com/CodeBoarding/CodeBoarding/pull/402#discussion_r3523367714

I have done a TDD aproach for the integration tests and rust setup came out buggy, so now we have a better handling for that. You can see it in the `install.py` and the language adapter, effectively it wasn't starting properly but we weren't failing (that behavior was on my mac) becuase we could get a lot of nodes, but 0 edges.

## Summary
- Add call-site metadata to static analysis edge models/builders/converters while keeping the edge call_sites contract non-null.
- Preserve repeated-call occurrence details through graph/language-result/incremental conversion paths.
- Update edge-case fixtures and tests across supported languages to cover repeated call sites.

## Validation
- `uv run pytest tests/static_analyzer/test_call_site_locations.py tests/static_analyzer/test_graph.py tests/static_analyzer/test_edge_builder.py tests/static_analyzer/test_call_graph_builder.py tests/static_analyzer/test_result_converter.py tests/static_analyzer/test_cluster_relations.py -q` -> 135 passed
- `uv run mypy static_analyzer/engine/models.py static_analyzer/engine/edge_builder.py static_analyzer/engine/call_graph_builder.py static_analyzer/engine/result_converter.py static_analyzer/graph.py static_analyzer/language_results.py static_analyzer/incremental_orchestrator.py` -> success
- `uv run pytest tests/integration/test_lsp_analysis_for_edge_cases.py::TestEdgeCases::test_call_site_occurrences -m integration -q` -> 7 passed, 1 skipped

## Known limitation
- Full edge-case integration currently has only Rust failures due to zero CFG edges from unhealthy rust-analyzer/Cargo: 2 failed, 53 passed, 17 skipped. This remains a Rust implementation/toolchain gap.

## Summary by CodeRabbit

* **New Features**
  * Call graphs now carry per-edge call-site metadata (file/line/column) end-to-end, including export and graph comparisons.
  * Edge-case integration fixtures were extended to include expected call-site locations.
  * Rust analysis adds workspace readiness and Cargo toolchain health validation before extraction.
* **Bug Fixes**
  * Duplicate edges are consolidated while preserving all recorded call-site locations.
  * Graph operations/conversions now retain call-site metadata consistently (or warn/error when absent).
* **Tests**
  * Expanded call-site occurrence consistency checks and added Rust/LSP health-handling test coverage.
  * Updated fixtures and unit tests to the new call-site-aware edge structure.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -465,7 +465,7 @@ class StaticAnalyzer:
             if not call_sites:
                 return []
 
-            queries = [(file_path, line, char) for line, char in call_sites]
+            queries = [(file_path, site.line - 1, site.column - 1) for site in call_sites]
             results, _ = client.send_definition_batch(queries)
 
             resolved = file_path.resolve()
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -178,6 +178,7 @@ class StaticAnalyzer:
         self.collected_diagnostics: dict[Language, FileDiagnosticsMap] = {}
         self._clients_started: bool = False
         self._cached_results: StaticAnalysisResults | None = None
+        self._persist_cache: bool = True
         # ``stop_clients`` writes the pkl using ``_pending_source_sha`` as the
         # tag value (a diff-base for the next warm-start, NOT a cache gate).
         # ``analyze()`` updates it on every call so the latest run's SHA
@@ -307,7 +308,8 @@ class StaticAnalyzer:
         """
         if not self._clients_started:
             return
-        self.flush_cache()
+        if self._persist_cache:
+            self.flush_cache()
         for engine_config, client in self._engine_clients:
             try:
                 client.shutdown()
@@ -493,6 +495,7 @@ class StaticAnalyzer:
         cache_dir: Path,
         skip_cache: bool = False,
         source_sha: str | None = None,
+        persist_cache: bool = True,
     ) -> StaticAnalysisResults:
         """Analyze the repository, warm-starting from the SHA-tagged pkl when present.
 
@@ -516,6 +519,7 @@ class StaticAnalyzer:
                 "LSP clients are not running. Call start_clients() or use StaticAnalyzer as a context manager "
                 "('with StaticAnalyzer(...) as sa:') before calling analyze()."
             )
+        self._persist_cache = persist_cache
 
         if not skip_cache and self._cached_results is not None:
             logger.info("static_analysis_cache: outcome=memhit")
@@ -778,6 +782,7 @@ def get_static_analysis(
     cache_dir: Path,
     skip_cache: bool = False,
     source_sha: str | None = None,
+    persist_cache: bool = True,
 ) -> StaticAnalysisResults:
     """CLI orchestrator: get static analysis results with full LSP lifecycle management.
 
@@ -793,12 +798,18 @@ def get_static_analysis(
         source_sha: Canonical source-state identifier (typically a git tree SHA)
             stamped onto the freshly-saved pkl as a diff base for the next
             warm-start.
+        persist_cache: If False, skip writing the run artifact on teardown.
 
     Returns:
         StaticAnalysisResults reflecting the live source state.
     """
     analyzer = StaticAnalyzer(repo_path)
     with analyzer:
-        results = analyzer.analyze(skip_cache=skip_cache, source_sha=source_sha, cache_dir=cache_dir)
+        results = analyzer.analyze(
+            cache_dir=cache_dir,
+            skip_cache=skip_cache,
+            source_sha=source_sha,
+            persist_cache=persist_cache,
+        )
     results.diagnostics = analyzer.collected_diagnostics
     return results
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -465,7 +465,7 @@ class StaticAnalyzer:
             if not call_sites:
                 return []
 
-            queries = [(file_path, site.line - 1, site.column - 1) for site in call_sites]
+            queries = [(file_path, site.lsp_line, site.lsp_column) for site in call_sites]
             results, _ = client.send_definition_batch(queries)
 
             resolved = file_path.resolve()
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -592,7 +592,21 @@ class StaticAnalyzer:
                     analysis={},
                     diagnostics={},
                 )
-        logger.info(f"Static analysis complete: {results}")
+        summaries = []
+        for language in results.get_languages():
+            try:
+                cfg = results.get_cfg(language)
+                node_count = len(cfg.nodes)
+                edge_count = len(cfg.edges)
+            except ValueError:
+                node_count = 0
+                edge_count = 0
+            summaries.append(
+                f"{language.value}: {len(results.get_source_files(language))} files, "
+                f"{sum(1 for _ in results.iter_reference_nodes(language))} references, "
+                f"{node_count} nodes, {edge_count} edges"
+            )
+        logger.info("Static analysis complete: %s", "; ".join(summaries) or "no languages")
         return results
 
     def _update_cached_results(self, cached_results: StaticAnalysisResults, cached_sha: str) -> StaticAnalysisResults:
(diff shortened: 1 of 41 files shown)


static_analyzer/engine/edge_builder.py
```diff
@@ -348,7 +353,9 @@ def _resolve_definitions(
                             if parent_qname in st.symbols:
                                 parent_sym = st.symbols[parent_qname]
                                 if _is_valid_edge(caller, parent_sym):
-                                    edge_set.add((caller.qualified_name, parent_qname))
+                                    _add_edge_site(
```
ivanmilevtues: the actuall addition of the call_site we already had the information didn't save and use it.

static_analyzer/engine/edge_builder.py
```diff
@@ -335,7 +337,10 @@ def _resolve_definitions(
                     if not _is_valid_edge(caller, target):
                         continue
 
-                    edge_set.add((caller.qualified_name, target.qualified_name))
+                    call_site = _call_site(file_path, site_line, site_col)
+                    _add_edge_site(
```
ivanmilevtues: the actuall addition of the call_site we already had the information didn't save and use it.

static_analyzer/__init__.py
```diff
@@ -464,7 +465,7 @@ def discover_file_dependencies(self, file_path: Path) -> list[str]:
             if not call_sites:
                 return []
 
-            queries = [(file_path, line, char) for line, char in call_sites]
+            queries = [(file_path, site.line - 1, site.column - 1) for site in call_sites]
```
Svilen-Stefanov: I really don't like the fact with these +/-1s we have with the LSPs. I think we should probably have a class with 2 methods - human readable lines and LSP-readable ones and use these if possible.

static_analyzer/engine/source_inspector.py
```diff
@@ -1,19 +1,90 @@
...
+import tree_sitter_python
+import tree_sitter_rust
+import tree_sitter_typescript
+
+logger = logging.getLogger(__name__)
+
+
+LanguageFactory = Callable[[], object]
+
+_LANGUAGE_BY_SUFFIX: dict[str, LanguageFactory] = {
```
Svilen-Stefanov, line 40: It seems like there's a discrepandy between the suffixes here and the ones from `constants.py`, some missing ones being - .mjs/.cjs/.mts/.cts etc.
We should probably have a consistent mapping between those.

static_analyzer/graph.py
```diff
@@ -270,6 +297,12 @@ def _prune_method_cluster_paths(self, surviving_nodes: dict[str, Node]) -> Metho
     def visit_paths(self, fn: Callable[[str], str]) -> None:
         for node in self.nodes.values():
             node.file_path = fn(node.file_path)
+        for edge in self.edges:
+            for site in edge.call_sites:
+                if "file" in site:
+                    site["file"] = fn(site["file"])
+                elif "file_path" in site:
```
Svilen-Stefanov: Do we need both file and file_path? Do they have different semanitcs?

static_analyzer/graph.py
```diff
@@ -173,11 +195,16 @@ def add_edge(self, src_name: str, dst_name: str) -> None:
...
+                self._edge_by_key[edge_key].add_call_site(dict(call_site))
             return
 
-        edge = Edge(self.nodes[src_name], self.nodes[dst_name])
+        edge = Edge(self.nodes[src_name], self.nodes[dst_name], [])
+        for call_site in call_sites:
+            edge.add_call_site(dict(call_site))
         self.edges.append(edge)
         self._edge_set.add(edge_key)
+        self._edge_by_key[edge_key] = edge
```
Svilen-Stefanov, line 212: Do we need both an edge set and the dict?

Svilen-Stefanov (commented):
Leaving my comments for now.

static_analyzer/engine/source_inspector.py
ivanmilevtues: I agree, good catch reusing those constants now.

static_analyzer/graph.py
ivanmilevtues: nope, dict is enough.

static_analyzer/__init__.py
ivanmilevtues: yup changed to have a static_analyzer/engine/models.py CallSite to have method which do those operations with +/- with methods:
```python
@dataclass
class CallSite:
    """A physical source location where an edge is invoked."""

    file: str
    line: int
    column: int

    @classmethod
    def from_lsp_position(cls, file: str, line: int, column: int) -> "CallSite":
        return cls(file=file, line=line + 1, column=column + 1)

    @property
    def human_line(self) -> int:
        return self.line

    @property
    def human_column(self) -> int:
        return self.column

    @property
    def lsp_line(self) -> int:
        return self.line - 1

    @property
    def lsp_column(self) -> int:
        return self.column - 1
```

static_analyzer/graph.py
ivanmilevtues: only file, backwards...

tests/integration/fixtures/edge_cases/java_edge_cases.json
```diff
@@ -120,137 +120,1008 @@
...
-    ["src.main.java.core.UserProfile.displayName()", "src.main.java.core.UserProfile.email"],
-    ["src.main.java.core.UserProfile.displayName()", "src.main.java.core.UserProfile.name"]
+    {
+      "source": "src.main.java.app.Main.main(String[])",
+      "destination": "src.main.java.core.Animal.getName()",
+      "call_sites": [
+        {
+          "file": "src/main/java/app/Main.java",
+          "line": 138,
+          "column": 72
```
Svilen-Stefanov, line 167: Do we need columns here? Does this help for cases where you have multiple calls on the same line?

static_analyzer/engine/result_converter.py
```diff
@@ -65,7 +65,21 @@ def convert_to_codeboarding_format(
         dst = edge.destination
         if call_graph.has_node(src) and call_graph.has_node(dst):
             try:
-                call_graph.add_edge(src, dst)
+                if edge.call_sites:
+                    for site in edge.call_sites:
+                        call_graph.add_edge(
+                            src,
+                            dst,
+                            call_sites=[{"file": site.file, "line": site.line, "column": site.column}],
```
Svilen-Stefanov: This seems like a list of sites - idk if we need the dictionary part added here.

static_analyzer/graph.py
```diff
@@ -89,17 +101,32 @@ def get_destination(self) -> str:
     def __repr__(self) -> str:
         return f"Edge({self.src_node.fully_qualified_name} -> {self.dst_node.fully_qualified_name})"
 
+    def add_call_site(self, call_site: dict) -> None:
+        call_site = self._normalize_call_site(call_site)
+        if call_site not in self.call_sites:
```
Svilen-Stefanov: nit: this is currently a list, if it's a set, this will be faster. Do we need to preserve order here?

static_analyzer/graph.py
```diff
@@ -89,17 +101,32 @@ def get_destination(self) -> str:
...
+            self.call_sites.append(call_site)
+
+    @staticmethod
+    def _normalize_call_site(call_site: dict) -> dict:
+        normalized = dict(call_site)
+        if "file" not in normalized and "file_path" in normalized:
```
Svilen-Stefanov: Is this a left-over - didn't you mean we only need one of them?

Svilen-Stefanov (commented):
Leaving current comments - will continue later.

static_analyzer/engine/result_converter.py
ivanmilevtues: true!

tests/integration/fixtures/edge_cases/java_edge_cases.json
ivanmilevtues: We do if you have mutlpe calls on the same line, yes it does help. I tested it :D

static_analyzer/graph.py
ivanmilevtues: oh my god I removed this and it came back :@

PR #405, opened 2026-07-10
fix: use baseline lineage for recursive scope snapshots

ivanmilevtues:
## Summary
- Fix recursive incremental scope diffs so the old scoped snapshot is built from `static_analysis.incremental_base_results` when available.
- Keep target static analysis for the new scoped cluster results, preserving the intended old-vs-new split.
- Add a regression test where unchanged methods have stable baseline lineage but shuffled target lineage, verifying the old snapshot uses the baseline cluster IDs.

## Test plan
- [x] `uv run pytest tests/diagram_analysis/test_diagram_analysis.py -k "recursive_scope_old_snapshot_uses_incremental_base_lineage or incremental_refresh_updates_existing_parent_scope or incremental_refresh_skips_child_scope_when_local_diff_is_empty"`
- [x] `uv run pytest tests/diagram_analysis/test_diagram_analysis.py`

## Notes
- The working tree had pre-existing `.codeboarding/` changes before this branch was created. They were not staged or included in this PR.

### DIFF
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -883,7 +883,10 @@ def _build_scope_incremental_inputs(
     changes: ChangeSet | None,
     repo_dir: Path,
 ) -> tuple[dict[str, ClusterResult], StructuralClusterDiff]:
-    old_snapshot = scoped_snapshot_for_component(component, scope_id, incremental_agent)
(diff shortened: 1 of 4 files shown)


ivanmilevtues:
an attempt at a demo that I was trying to create.

PR #406, opened 2026-07-11
fix(github-action): use Core's git-free internal change detection

Svilen-Stefanov:
## What this PR does

`github_action.py` (the entry point Core's own `.github/workflows/docs.yml` calls) drove incremental analysis through the low-level `run_incremental_workflow` kernel with a generator built **without** a change set. So `generator.changes` was `None` and the cluster delta ran **unscoped** — the action never participated in the git-free "Core detects changes internally" contract that the CLI and wrapper adopted in the content-versioning stack.

### Changes

- **Route `generate_analysis` through `run_incremental(RunPaths, RunContext)`** — Core fingerprints the tree and diffs it against the seeded `fingerprint.json` sidecar itself, and reuses the baseline's `depth_level` instead of the env var.
- **Seed `fingerprint.json`** alongside `analysis.json` in `_seed_existing_analysis` — it's the whole-tree baseline the git-free detection needs. Without it, incremental would `BaselineUnavailableError` every time.
- **Fall back to a full run on `BaselineUnavailableError`** — a CI docs build can't prompt the user, so this preserves the old resilience (a legacy/no-sidecar baseline degrades to full instead of failing the build).
- **Tests** — patch `_run_analysis`; add coverage for sidecar seeding and both the incremental and full-fallback branches.

## Depends on

The content-versioning stack (#401) — this uses the new git-free `run_incremental(RunPaths, RunContext)` signature. Based off `feat/content-versioning-stateless`.

## Note — separate follow-up

The published **CodeBoarding-action** repo has the *same* class of break at a larger scale: its `engine_adapter.py` calls `run_incremental`/`run_full` with the old `base_ref`/`target_ref`/`source_sha` signatures and will `TypeError` when #401 merges. That migration (action goes git-free, seed `fingerprint.json`) is tracked separately.

## Testing

`mypy` + `black` clean; `pytest tests/test_github_action.py` — 15 passed.


ivanmilevtues:
WE will redo this one

PR #407, opened 2026-07-11
Preserve static analysis SHA after health reads

ivanmilevtues:
## Summary
- treat static-analysis artifacts loaded for health/status checks as read-only
- prevent Wrapper LSP shutdown from rewriting the pickle without a source SHA and deleting its sidecar
- add a regression test covering the health-load then flush lifecycle

## Root cause
The live Wrapper analyzer loaded `static_analysis.pkl` for health reporting, which populated its in-memory cache without a pending source SHA. On shutdown, Core persisted that read-only cache with `source_sha=None`, removing `static_analysis.sha` and making the next incremental run fail.

## Verification
- `pytest -q tests/static_analyzer/test_static_analyzer_start_clients.py` (11 passed)

## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed the disk-cache loading API to `load_cached_analysis`.
  * Removed the `persist_cache` option from `analyze()` and from the `get_static_analysis()` invocation.
* **Bug Fixes**
  * Ensured cache artifacts loaded from disk are treated as read-only and won’t be rewritten during flush/shutdown.
  * Cache saving now occurs only after fresh analysis results are produced, preventing unnecessary/incorrect persistence behavior.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -178,7 +178,13 @@ class StaticAnalyzer:
         self.collected_diagnostics: dict[Language, FileDiagnosticsMap] = {}
         self._clients_started: bool = False
         self._cached_results: StaticAnalysisResults | None = None
-        self._persist_cache: bool = True
+        # True once ``analyze()`` has produced fresh results this session that
+        # the on-disk pkl doesn't have yet. Only ``analyze()`` sets it;
+        # ``flush_cache``/``stop_clients`` write the pkl iff it is set. A
+        # read-only ``load_cached_analysis`` never flips it, so rehydrating the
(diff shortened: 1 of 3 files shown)


Svilen-Stefanov (changes requested):
How did you test this, I don't quite understand the solution. Can you maybe just clarify in the PR or double check the solution?

Svilen-Stefanov (approved):
I checked this, made some minor clarification renames but this looks right. Take a look if the renames make more sense to you and lmk what you think.

ivanmilevtues:
Yes thanks for the renames and also the comments I think it makes it easier to understand as it is quite puzzling ;d

PR #408, opened 2026-07-11
Align incremental analysis with the extended relation pipeline

ivanmilevtues:
## Summary
- align incremental relation generation with the API-surface and evidence-validation stages used by full and detail analysis
- preserve evidence, key/all edges, and call-site metadata through global relation projection and JSON round trips
- repair missing planner component IDs only when prior cluster ownership or an exact existing component name identifies one unique target
- canonicalize unique hyphen/underscore qualified-name aliases and let deterministic materialization select missing optional key entities
- preserve warm-start CFG edges by validating cached inbound and outbound cross-boundary references against the live LSP
- resolve outbound definitions to the most-specific symbol instead of every enclosing class
- fail explicitly on ambiguous plans or required scope regeneration; incremental analysis never silently becomes full analysis

## Incremental pipeline evaluation

**Source:** local editable core at `../CodeBoarding-core/CodeBoarding`  
**Judge model:** Gemini 3.1 Pro

### LLM scorecard — depth 2 vs depth 3

| Metric | D2 | D3 |
|---|---:|---:|
| Components judged | 5 | 6 |
| Mean component score | 0.48 | 0.63 |
| Correct | 0 | 1 |
| Partial | 5 | 5 |
| Incorrect | 0 | 0 |
| LLM structural score | Not completed | 1.00 — justified |
| Deterministic structural checks | PASS | PASS |
| Edge-integrity check | PASS — 98.1% | PASS — 94.9% |

### Interpretation

- D2 is weak: every changed component was judged partial. The main issues are false method removals and absorbing MCP/OCR changes into unrelated existing components.
- D3 is better: one component was judged fully correct, with the remaining five partial. It correctly added the OCR component and received a fully justified structural score.
- No incorrect verdicts were recorded at either depth.
- D2’s structural LLM call hung after the five component judgments, so its structural score is unavailable; its deterministic structural check passed.

## Repository validation
- `uv run black .` passed
- `mypy .` passed across 312 source files
- repository suite: 1,775 passed, 28 skipped
- coverage gate passed at 80.05% (required 80%)
- focused incremental/diagram/static suites passed, including regressions for the supplied planner and edge failures

## Local test-environment caveat
- the repository-wide command has one unrelated failure in `tests/test_pyproject_packages.py`: user-populated `temp/repos/*` clones are discovered as undeclared CodeBoarding packages

## Summary by CodeRabbit

* **New Features**
  * Incremental updates now use staged relation discovery with deterministic relation IDs and improved preservation/repair of component key-entity metadata.
  * Relation endpoint evidence is indexed from source-level references, and diagram JSON/file indexing now supports nested sub-analyses.
* **Bug Fixes**
  * Improved handling for missing/invalid relation endpoints and clearer failures for invalid incremental plans or missing relation context.
  * Warm-start cross-boundary edge restoration now requires live reference validation.
* **Tests**
  * Expanded coverage for incremental planning/repair, relation generation, diagram refresh/indexing, warm-start edge restoration, and reference/key-entity resolution.

## Incremental Pipeline Eval — Run 3

After `005e7ef` — **Harden incremental analysis pipeline**

**Engine:** CodeBoarding Core commit `005e7ef` (`3d287cc` Restructure of method calling + reference resolver, plus `005e7ef` Harden incremental analysis pipeline)  
**Agent:** `google/gemini-3-flash-preview`  
**Judge:** `google/gemini-3.1-pro-preview`  
**Date:** 2026-07-12

### Summary scorecard — Run 2 vs Run 3

| Scenario | Depth | Run 2 eval | Run 3 eval | Run 2 judge | Run 3 judge |
|---|---:|---:|---:|---:|---:|
| `markitdown-small-scope` | 2 | 11/11 (100%) | 11/11 (100%) | — (0 changes) | — (0 changes) |
| `markitdown` (full) | 2 | 8/9 (89%) | 8/9 (89%) | 0.68 | **0.78** |
| `markitdown-small-scope-l3` | 3 | 11/11 (100%) | 11/11 (100%) | — (0 changes) | — (0 changes) |
| `markitdown-l3` (full) | 3 | 8/9 (89%) | 8/9 (89%) | 0.68 | 0.63 |

### Small-scope results — perfect stability at both depths

| Scenario | Depth | Baseline | Incremental | Token savings |
|---|---:|---:|---:|---:|
| `markitdown-small-scope` | 2 | 530s / 1.68M tok / 26 comps | 15.3s / 24.4K tok / 26 comps | **69x** |
| `markitdown-small-scope-l3` | 3 | 980s / 7.19M tok / 77 comps | 21.1s / 24.6K tok / 77 comps | **292x** |

Zero component churn, zero relation churn, and zero edge churn at both depths: **11/11 PASS**.

#### Eval checks

| Check | Result | Detail |
|---|---|---|
| `depth_level_unchanged` | PASS | expected 2; baseline=2, incremental=2 |
| `baseline_has_components` | PASS | 15 components |
| `incremental_has_components` | PASS | 23 components |
| `baseline_artifact_present` | PASS | artifact written |
| `incremental_artifact_present` | PASS | artifact written |
| `l1_component_count_stable` | PASS | 4→6, delta +2, max ±2 |
| `total_component_churn_bounded` | **FAIL** | 15→23, +53%, max ±50% (3pp over threshold) |
| `relation_pair_churn_bounded` | PASS | 42 pairs, 19 added, 1 removed, 48% churn |
| `edge_endpoint_churn_bounded` | PASS | 0 churn |

#### Performance

| Phase | Time | Tokens | Components |
|---|---:|---:|---:|
| Baseline | 433s | 1.22M | 15 (4 L1 + 11 L2) |
| Incremental | 481s | 806K | 23 (6 L1 + 17 L2) |
| Savings | — | **1.5x fewer tokens** | +8 new comps (OCR + MCP) |

#### LLM judge — depth-2 full

Mean: **0.78**, up from 0.68.

| Component | Status | Verdict | Score |
|---|---|---|---:|
| OCR-Enhanced Converters (5) | added | correct | **1.00** |
| MCP Server Interface (6) | added | correct | **1.00** |
| Binary & AI-Powered Converters (3) | modified | correct | **1.00** |
| Textual & Structured Converters (2) | modified | partial | 0.50 |
| Conversion Orchestrator (1) | modified | partial | 0.40 |

**Improvement vs Run 2:** Binary & AI-Powered Converters moved from partial (0.60) to correct (1.00). The judge noted the `accepts` baseline-correction artifact but scored it correct because “the core changes are accurately and comprehensively captured.”

**Structure verdict:** justified (1.00). The +2 L1 and +6 L2 components fully map to the new `markitdown-ocr` and `markitdown-mcp` packages.

#### Eval checks

| Check | Result | Detail |
|---|---|---|
| `depth_level_unchanged` | PASS | expected 3; baseline=3, incremental=3 |
| `baseline_has_components` | PASS | 86 components |
| `incremental_has_components` | PASS | 112 components |
| `baseline_artifact_present` | PASS | artifact written |
| `incremental_artifact_present` | PASS | artifact written |
| `l1_component_count_stable` | PASS | 6→8, delta +2, max ±2 |
| `total_component_churn_bounded` | PASS | 86→112, +30%, max ±50% |
| `relation_pair_churn_bounded` | PASS | 223 pairs, 54 added, 3 removed, 26% churn |
| `edge_endpoint_churn_bounded` | **FAIL** | 3 base edges, 3 churned (+100%); tiny-base artifact on non-static edges |

#### Performance

| Phase | Time | Tokens | Components |
|---|---:|---:|---:|
| Baseline | 961s | 7.25M | 86 (6 L1 + 20 L2 + 60 L3) |
| Incremental | 801s | 2.28M | 112 (8 L1 + 26 L2 + 78 L3) |
| Savings | **1.2x faster** | **3.2x fewer tokens** | +26 new comps |

#### LLM judge — depth-3 full

Mean: **0.63**, with 7 changes judged.

| Component | Status | Verdict | Score | Core issue |
|---|---|---|---:|---|
| OCR-Enhanced Conversion Suite (7) | added | correct | **1.00** | None |
| MCP Server Interface (8) | added | correct | **1.00** | None |
| Component 4 | modified | partial | 0.70 | New files/methods correctly caught; accepts baseline-recovery noise |
| Component 5 | modified | partial | 0.60 | File-level additions correct; method extraction gaps on `DocumentIntelligenceConverter` |
| Orchestration & Registry (2) | modified | partial | 0.50 | Correctly found new converters; false `accepts` removals |
| Entry Point & Input Handling (1) | modified | partial | 0.40 | `_cu_converter.py` mis-assigned to this component |
| Conversion Framework (3) | modified | incorrect | 0.20 | Flagged as modified with hallucinated reasons; missed the real `_markitdown._convert` change |

**Structure verdict:** justified (1.00). The +2 L1, +6 L2, and +18 L3 components all map to new packages and modules.

#### What improved from Run 2 to Run 3

- Depth-2 judge score improved from 0.68 to **0.78** (+10pp), with one component moving from partial to fully correct.
- Edge churn now passes at depth 2; the previous tiny-base edge volatility is resolved.
- Depth-3 total churn improved from 35% to **30%**, with more components correctly carried forward from the baseline.

#### What regressed

- Depth-3 judge score decreased from 0.68 to **0.63** (-5pp). A new component split created 7 judged changes instead of 6, and Conversion Framework scored 0.20 because the pipeline flagged it as modified with hallucinated reasons while missing the real `MarkItDown._convert` change.
- Depth-2 `total_component_churn_bounded` now fails at +53%, compared with +40% previously. The baseline produced fewer components (15 instead of 20), so the same +8 absolute growth is a higher percentage. This is baseline non-determinism rather than an incremental regression.

#### Persistent issue across all runs

The `accepts` method hallucination remains: baseline extraction misses `accepts` methods on several converters, so the incremental pass “discovers” them and reports them as added or removed. This affects 4 of 5 modified components at depth 2 and 5 of 7 at depth 3. The judge identifies this as a baseline-completeness gap rather than an incremental-logic error.

### DIFF
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -22,6 +22,7 @@ from agents.prompts import (
     get_api_surfaces_message,
     get_relation_analysis_message,
     get_system_message,
+    format_project_system_message,
 )
 from agents.relation_edges import index_relation_endpoints
 from agents.repair import ComponentRepairContext, repair_component_group_names, repair_key_entities
@@ -53,13 +54,7 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
     ):
-        meta_context_str = meta_context.llm_str() if meta_context else "No project context available."
-        project_type = meta_context.project_type if meta_context else "unknown"
-        system_message = get_system_message().format(
-            project_name=project_name,
-            project_type=project_type,
-            meta_context=meta_context_str,
-        )
+        system_message = format_project_system_message(get_system_message(), project_name, meta_context)
         super().__init__(repo_dir, static_analysis, system_message, agent_llm, parsing_llm)
 
         self.project_name = project_name
@@ -68,11 +63,11 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         self.prompts = {
             "group_clusters": PromptTemplate(
                 template=get_cluster_grouping_message(),
-                input_variables=["project_name", "cfg_clusters", "meta_context", "project_type"],
+                input_variables=["cfg_clusters"],
             ),
             "final_analysis": PromptTemplate(
                 template=get_final_analysis_message(),
-                input_variables=["project_name", "cluster_analysis", "meta_context", "project_type"],
+                input_variables=["cluster_analysis"],
             ),
             "api_surfaces": PromptTemplate(
                 template=get_api_surfaces_message(),
@@ -95,9 +90,6 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
     def step_clusters_grouping(self, cluster_results: dict[str, ClusterResult]) -> ClusterAnalysis:
         logger.info(f"[AbstractionAgent] Grouping CFG clusters for: {self.project_name}")
 
-        meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
-        project_type = self.meta_context.project_type if self.meta_context else "unknown"
-
         programming_langs = self.static_analysis.get_languages()
 
         # Measure everything that wraps cfg_clusters (system message + rendered
@@ -105,10 +97,7 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         # the input window before budgeting the cluster string.
         overhead_chars = len(str(self.system_message.content)) + len(
             self.prompts["group_clusters"].format(
-                project_name=self.project_name,
                 cfg_clusters="",
-                meta_context=meta_context_str,
-                project_type=project_type,
             )
         )
         cluster_str = self._build_cluster_string(
@@ -116,18 +105,13 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         )
 
         prompt = self.prompts["group_clusters"].format(
-            project_name=self.project_name,
             cfg_clusters=cluster_str,
-            meta_context=meta_context_str,
-            project_type=project_type,
         )
 
-        cluster_analysis = self._invoke_repair_validate(
+        cluster_analysis = self._invoke_validate(
             prompt,
             ClusterAnalysis,
-            repairs=[],
             validators=[validate_cluster_coverage],
-            repair_context=None,
             validation_context=ValidationContext(
                 cluster_results=cluster_results,
                 expected_cluster_ids=get_all_cluster_ids(cluster_results),
@@ -142,18 +126,12 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
     ) -> AnalysisInsights:
         logger.info(f"[AbstractionAgent] Generating final component analysis for: {self.project_name}")
 
-        meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
-        project_type = self.meta_context.project_type if self.meta_context else "unknown"
-
         cluster_str = llm_cluster_analysis.llm_str() if llm_cluster_analysis else "No cluster analysis available."
 
         group_names = [cc.name for cc in llm_cluster_analysis.cluster_components] if llm_cluster_analysis else []
 
         prompt = self.prompts["final_analysis"].format(
-            project_name=self.project_name,
             cluster_analysis=cluster_str,
-            meta_context=meta_context_str,
-            project_type=project_type,
         )
 
         if group_names:
@@ -198,15 +176,7 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             component_summaries=analysis.llm_str(),
             static_call_evidence=static_call_evidence,
         )
-        return self._invoke_repair_validate(
-            prompt,
-            ComponentApiSurfaces,
-            repairs=[],
-            validators=[],
-            repair_context=None,
-            validation_context=None,
-            max_validation_attempts=1,
-        )
+        return self._parse_invoke(prompt, ComponentApiSurfaces)
 
     @trace
     def step_relation_analysis(
@@ -227,12 +197,10 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             api_surfaces=api_surfaces.llm_str(),
             static_call_evidence=static_call_evidence,
         )
-        relation_result = self._invoke_repair_validate(
+        relation_result = self._invoke_validate(
             prompt,
             ComponentRelations,
-            repairs=[],
             validators=[validate_relations],
-            repair_context=None,
             validation_context=ValidationContext(
                 cluster_results=cluster_results,
                 cfg_graphs=cfg_graphs,
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -23,6 +23,8 @@ from agents.prompts import (
     get_relation_analysis_message,
     get_system_message,
 )
+from agents.relation_edges import index_relation_endpoints
+from agents.repair import ComponentRepairContext, repair_component_group_names, repair_key_entities
 from agents.validation import (
     ValidationContext,
     validate_cluster_coverage,
@@ -51,7 +53,14 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
     ):
-        super().__init__(repo_dir, static_analysis, get_system_message(), agent_llm, parsing_llm)
+        meta_context_str = meta_context.llm_str() if meta_context else "No project context available."
+        project_type = meta_context.project_type if meta_context else "unknown"
+        system_message = get_system_message().format(
+            project_name=project_name,
+            project_type=project_type,
+            meta_context=meta_context_str,
+        )
+        super().__init__(repo_dir, static_analysis, system_message, agent_llm, parsing_llm)
 
         self.project_name = project_name
         self.meta_context = meta_context
@@ -68,22 +77,16 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             "api_surfaces": PromptTemplate(
                 template=get_api_surfaces_message(),
                 input_variables=[
-                    "project_name",
                     "component_summaries",
                     "static_call_evidence",
-                    "meta_context",
-                    "project_type",
                 ],
             ),
             "relation_analysis": PromptTemplate(
                 template=get_relation_analysis_message(),
                 input_variables=[
-                    "project_name",
                     "component_summaries",
                     "api_surfaces",
                     "static_call_evidence",
-                    "meta_context",
-                    "project_type",
                 ],
             ),
         }
@@ -119,11 +122,13 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             project_type=project_type,
         )
 
-        cluster_analysis = self._validation_invoke(
+        cluster_analysis = self._invoke_repair_validate(
             prompt,
             ClusterAnalysis,
+            repairs=[],
             validators=[validate_cluster_coverage],
-            context=ValidationContext(
+            repair_context=None,
+            validation_context=ValidationContext(
                 cluster_results=cluster_results,
                 expected_cluster_ids=get_all_cluster_ids(cluster_results),
             ),
@@ -163,14 +168,20 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             llm_cluster_analysis=llm_cluster_analysis,
         )
 
-        architecture = self._validation_invoke(
+        architecture = self._invoke_repair_validate(
             prompt,
             ComponentArchitecture,
+            repairs=[repair_component_group_names, repair_key_entities],
             validators=[
                 validate_group_name_coverage,
                 validate_key_entities,
             ],
-            context=context,
+            repair_context=ComponentRepairContext(
+                reference_resolver=self.reference_resolver,
+                cluster_results=cluster_results,
+                llm_cluster_analysis=llm_cluster_analysis,
+            ),
+            validation_context=context,
             max_validation_attempts=3,
         )
         return AnalysisInsights(
@@ -182,21 +193,18 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
     @trace
     def step_api_surfaces(self, analysis: AnalysisInsights) -> ComponentApiSurfaces:
         logger.info(f"[AbstractionAgent] Analyzing component API surfaces for: {self.project_name}")
-        meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
-        project_type = self.meta_context.project_type if self.meta_context else "unknown"
         static_call_evidence = self.build_scope_cfg_string(analysis)
         prompt = self.prompts["api_surfaces"].format(
-            project_name=self.project_name,
             component_summaries=analysis.llm_str(),
             static_call_evidence=static_call_evidence,
-            meta_context=meta_context_str,
-            project_type=project_type,
         )
-        return self._validation_invoke(
+        return self._invoke_repair_validate(
             prompt,
             ComponentApiSurfaces,
+            repairs=[],
             validators=[],
-            context=None,
+            repair_context=None,
+            validation_context=None,
             max_validation_attempts=1,
         )
 
@@ -209,26 +217,23 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         cluster_results: dict[str, ClusterResult],
     ) -> None:
         logger.info(f"[AbstractionAgent] Discovering component relations for: {self.project_name}")
-        meta_context_str = self.meta_context.llm_str() if self.meta_context else "No project context available."
-        project_type = self.meta_context.project_type if self.meta_context else "unknown"
         static_call_evidence = self.build_scope_cfg_string(analysis)
         cfg_graphs = self.static_analysis.available_cfgs()
         self.toolkit.context.cluster_analysis = cluster_analysis
         self.toolkit.context.cluster_results = cluster_results
         self.toolkit.context.cfg_graphs = cfg_graphs
         prompt = self.prompts["relation_analysis"].format(
-            project_name=self.project_name,
             component_summaries=analysis.llm_str(),
             api_surfaces=api_surfaces.llm_str(),
             static_call_evidence=static_call_evidence,
-            meta_context=meta_context_str,
-            project_type=project_type,
         )
-        relation_result = self._validation_invoke(
+        relation_result = self._invoke_repair_validate(
             prompt,
             ComponentRelations,
+            repairs=[],
             validators=[validate_relations],
-            context=ValidationContext(
+            repair_context=None,
+            validation_context=ValidationContext(
                 cluster_results=cluster_results,
                 cfg_graphs=cfg_graphs,
                 repo_dir=str(self.repo_dir),
@@ -266,7 +271,9 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
 
         # Step 8: Fix source code reference lines (resolves reference_file paths for key_entities and key_edges)
         analysis = self.reference_resolver.fix_source_code_reference_lines(analysis)
-        # Step 9: Ensure unique key entities across components
+        # Step 9: Index relation endpoints after reference resolution
+        index_relation_endpoints(analysis, self.repo_dir)
+        # Step 10: Ensure unique key entities across components
         self._ensure_unique_key_entities(analysis)
 
         return analysis, cluster_results
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -227,7 +227,13 @@ class CodeBoardingAgent(MonitoringMixin):
         assert isinstance(response, str), f"Expected a string as response type got {response}"
         return self._parse_response(prompt, response, type, include_hidden=include_hidden)
 
-    def _score_result(self, result, validators: list, context) -> tuple[float, list[tuple[float, str]]]:
+    def _repair_result(self, result, repairs: list, repair_context) -> None:
+        """Apply deterministic repairs to one parsed candidate."""
+        for repair in repairs:
+            logger.info("[Repair] Applying %s", repair.__name__)
+            repair(result, repair_context)
+
+    def _score_result(self, result, validators: list, validation_context) -> tuple[float, list[tuple[float, str]]]:
         """Run all validators on a result and return (score, prioritized_feedback).
 
         The score is computed using weighted validators where coverage-related
@@ -241,7 +247,7 @@ class CodeBoardingAgent(MonitoringMixin):
         validator_results: list[tuple] = []
         weighted_feedback: list[tuple[float, str]] = []
         for validator in validators:
-            validator_result: ValidationResult = validator(result, context)
+            validator_result: ValidationResult = validator(result, validation_context)
             validator_results.append((validator, validator_result))
             if not validator_result.is_valid:
                 weight = VALIDATOR_WEIGHTS.get(validator.__name__, DEFAULT_VALIDATOR_WEIGHT)
@@ -254,44 +260,25 @@ class CodeBoardingAgent(MonitoringMixin):
         score = score_validation_results(validator_results)
         return score, weighted_feedback
 
-    def _validation_invoke(
+    def _invoke_repair_validate(
         self,
         prompt: str,
         return_type: type,
+        repairs: list,
         validators: list,
-        context,
+        repair_context,
+        validation_context,
         max_validation_attempts: int = 1,
         include_hidden: bool = False,
     ):
-        """
-        Invoke LLM with validation, feedback loop, and best-of-N selection.
-
-        Each attempt (initial + retries) is scored using weighted validators.
-        Coverage validators (validate_cluster_coverage, validate_group_name_coverage)
-        are weighted ~2x higher than other validators, so the selection strongly
-        favours results with complete coverage.
-
-        If any attempt scores perfectly (all validators pass), it is returned
-        immediately. Otherwise the highest-scoring result across all attempts is
-        returned.
-
-        Args:
-            prompt: The original prompt
-            return_type: Pydantic type to parse into
-            validators: List of validation functions to run
-            context: ValidationContext with data needed for validation
-            max_validation_attempts: Maximum validation attempts (initial attempt included).
-                Retries occur only when this value is greater than 1. (default: 1)
-
-        Returns:
-            The highest-scoring result of return_type across all attempts
-        """
+        """Parse, repair, and validate each candidate before best-of-N selection."""
         # Compute the maximum possible score so we can detect a perfect result
         max_possible_score = sum(VALIDATOR_WEIGHTS.get(v.__name__, DEFAULT_VALIDATOR_WEIGHT) for v in validators)
 
         result = self._parse_invoke(prompt, return_type, include_hidden=include_hidden)
+        self._repair_result(result, repairs, repair_context)
         logger.info(
-            "[Validation] Parsed %s: %s",
+            "[Validation] Parsed and repaired %s: %s",
             return_type.__name__,
             result.llm_str()[:500],
         )
@@ -304,7 +291,7 @@ class CodeBoardingAgent(MonitoringMixin):
         critical_threshold = 10.0
 
         for attempt in range(1, max_validation_attempts + 1):
-            score, weighted_feedback = self._score_result(result, validators, context)
+            score, weighted_feedback = self._score_result(result, validators, validation_context)
 
             logger.info(
                 f"[Validation] Attempt {attempt}/{max_validation_attempts} "
@@ -348,6 +335,7 @@ class CodeBoardingAgent(MonitoringMixin):
                 f"with {len(weighted_feedback)} feedback items"
             )
             result = self._parse_invoke(feedback_prompt, return_type, include_hidden=include_hidden)
+            self._repair_result(result, repairs, repair_context)
 
         return best_result
 
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -297,14 +297,41 @@ class CodeBoardingAgent(MonitoringMixin):
         max_validation_attempts: int = 1,
         include_hidden: bool = False,
     ) -> ResultT:
-        """Parse, repair, and validate each candidate before best-of-N selection."""
+        """Bind deterministic repairs to every candidate before validation."""
+
+        def repair_candidate(result: ResultT) -> None:
+            self._repair_result(result, repairs, repair_context)
+
+        return self._invoke_validate(
+            prompt,
+            return_type,
+            validators,
+            validation_context,
+            max_validation_attempts=max_validation_attempts,
+            include_hidden=include_hidden,
+            candidate_repair=repair_candidate,
+        )
+
+    def _invoke_validate(
+        self,
+        prompt: str,
+        return_type: type[ResultT],
+        validators: list[Callable[[ResultT, ValidationContextT], ValidationResult]],
+        validation_context: ValidationContextT,
+        max_validation_attempts: int = 1,
+        include_hidden: bool = False,
+        candidate_repair: Callable[[ResultT], None] | None = None,
+    ) -> ResultT:
+        """Validate parsed candidates and return the best-scoring result."""
         # Compute the maximum possible score so we can detect a perfect result
         max_possible_score = sum(VALIDATOR_WEIGHTS.get(v.__name__, DEFAULT_VALIDATOR_WEIGHT) for v in validators)
 
         result = self._parse_invoke(prompt, return_type, include_hidden=include_hidden)
-        self._repair_result(result, repairs, repair_context)
+        if candidate_repair is not None:
+            candidate_repair(result)
         logger.info(
-            "[Validation] Parsed and repaired %s: %s",
+            "[Validation] Parsed%s %s: %s",
+            " and repaired" if candidate_repair is not None else "",
             return_type.__name__,
             result.llm_str()[:500],
         )
@@ -361,7 +388,8 @@ class CodeBoardingAgent(MonitoringMixin):
                 f"with {len(weighted_feedback)} feedback items"
             )
             result = self._parse_invoke(feedback_prompt, return_type, include_hidden=include_hidden)
-            self._repair_result(result, repairs, repair_context)
+            if candidate_repair is not None:
+                candidate_repair(result)
 
         return best_result
 
--- a/agents/agent.py
+++ b/agents/agent.py
@@ -1,6 +1,8 @@
 import json
 import logging
+from collections.abc import Callable
 from pathlib import Path
+from typing import Protocol, TypeVar
 
 from google.api_core.exceptions import ResourceExhausted
 from langchain_core.exceptions import OutputParserException
@@ -27,6 +29,15 @@ from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.reference_resolver import StaticReferenceResolver
 
 logger = logging.getLogger(__name__)
+
+ParseResultT = TypeVar("ParseResultT")
+ResultT = TypeVar("ResultT", bound="RepairValidationResult")
+RepairContextT = TypeVar("RepairContextT")
+ValidationContextT = TypeVar("ValidationContextT")
+
+
+class RepairValidationResult(Protocol):
+    def llm_str(self) -> str: ...
 
 
 class EmptyExtractorMessageError(ValueError):
@@ -222,18 +233,33 @@ class CodeBoardingAgent(MonitoringMixin):
         except Empty:
             raise RuntimeError("Agent invocation completed but no result was returned")
 
-    def _parse_invoke(self, prompt: str, type: type, include_hidden: bool = False):
+    def _parse_invoke(
+        self,
+        prompt: str,
+        return_type: type[ParseResultT],
+        include_hidden: bool = False,
+    ) -> ParseResultT:
         response = self._invoke(prompt)
         assert isinstance(response, str), f"Expected a string as response type got {response}"
-        return self._parse_response(prompt, response, type, include_hidden=include_hidden)
+        return self._parse_response(prompt, response, return_type, include_hidden=include_hidden)
 
-    def _repair_result(self, result, repairs: list, repair_context) -> None:
+    def _repair_result(
+        self,
+        result: ResultT,
+        repairs: list[Callable[[ResultT, RepairContextT], None]],
+        repair_context: RepairContextT,
+    ) -> None:
         """Apply deterministic repairs to one parsed candidate."""
         for repair in repairs:
             logger.info("[Repair] Applying %s", repair.__name__)
             repair(result, repair_context)
 
-    def _score_result(self, result, validators: list, validation_context) -> tuple[float, list[tuple[float, str]]]:
+    def _score_result(
+        self,
+        result: ResultT,
+        validators: list[Callable[[ResultT, ValidationContextT], ValidationResult]],
+        validation_context: ValidationContextT,
+    ) -> tuple[float, list[tuple[float, str]]]:
         """Run all validators on a result and return (score, prioritized_feedback).
 
         The score is computed using weighted validators where coverage-related
@@ -244,7 +270,7 @@ class CodeBoardingAgent(MonitoringMixin):
         weight descending, so that the LLM focuses on the most critical issues
         (cluster/group coverage) before lower-priority ones (key entities).
         """
-        validator_results: list[tuple] = []
+        validator_results: list[tuple[Callable[[ResultT, ValidationContextT], ValidationResult], ValidationResult]] = []
         weighted_feedback: list[tuple[float, str]] = []
         for validator in validators:
             validator_result: ValidationResult = validator(result, validation_context)
@@ -263,14 +289,14 @@ class CodeBoardingAgent(MonitoringMixin):
     def _invoke_repair_validate(
         self,
         prompt: str,
-        return_type: type,
-        repairs: list,
-        validators: list,
-        repair_context,
-        validation_context,
+        return_type: type[ResultT],
+        repairs: list[Callable[[ResultT, RepairContextT], None]],
+        validators: list[Callable[[ResultT, ValidationContextT], ValidationResult]],
+        repair_context: RepairContextT,
+        validation_context: ValidationContextT,
         max_validation_attempts: int = 1,
         include_hidden: bool = False,
-    ):
+    ) -> ResultT:
         """Parse, repair, and validate each candidate before best-of-N selection."""
         # Compute the maximum possible score so we can detect a perfect result
         max_possible_score = sum(VALIDATOR_WEIGHTS.get(v.__name__, DEFAULT_VALIDATOR_WEIGHT) for v in validators)
--- a/agents/cluster_ids.py
+++ b/agents/cluster_ids.py
@@ -1,3 +1,6 @@
+from agents.scope_ids import ROOT_SCOPE_ID
+
+
 type GraphClusterId = int
 type CodeBoardingClusterId = str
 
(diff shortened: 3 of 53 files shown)


agents/incremental_planning_agent.py
```diff
@@ -37,10 +39,24 @@
...
+"""
+
+
 @dataclass
 class ScopeOperationValidationContext:
     expected_cluster_refs: set[ClusterRef] = field(default_factory=set)
     existing_component_ids: set[str] = field(default_factory=set)
+    known_qnames: set[str] = field(default_factory=set)
+    component_ids_by_cluster_ref: dict[ClusterRef, str] = field(default_factory=dict)
+    component_ids_by_name: dict[str, str] = field(default_factory=dict)
```
ivanmilevtues: This should go into the prompts and it should be added into each prompt factory. And this message should be just part of the prompt to which we are appending it.

agents/incremental_planning_agent.py
```diff
@@ -111,10 +131,14 @@ def decide_scope_update(
         validation = validate_scope_update_decision(decision, context)
         if not validation.is_valid:
             logger.error(
-                "Incremental planning decision remained invalid after retries; will still pass it on. Issues: %s",
+                "Incremental planning decision remained invalid after retries: %s",
                 validation.feedback_messages,
             )
```
ivanmilevtues, line 114: The IncrementalPlannerAgent has a system prompt which has the project_metacontext, we shouldn't also have it in the concrete prompt here.

agents/incremental_planning_agent.py
```diff
@@ -14,6 +14,7 @@
     AnalysisInsights,
```
ivanmilevtues, line 13: This should go into the validation.py file and not here.

diagram_analysis/diagram_generator.py
```diff
@@ -668,10 +668,10 @@ def _apply_incremental_scope_recursively(
             removed_ids=set(apply_result.removed_ids),
             regenerate_scope=apply_result.regenerate_scope,
         )
-        if apply_result.refresh_ids:
+        if apply_result.refresh_ids or apply_result.removed_ids:
             result.touched_scopes.add(scope_id)
         if result.regenerate_scope:
-            return result
+            raise IncrementalScopeRegenerationRequiredError(scope_id)
```
ivanmilevtues: We should also log something + send an error event (it might be auto-picked) so that we can see how often this is happening.

diagram_analysis/analysis_json.py
```diff
@@ -227,6 +227,33 @@ def _build_methods_index_from_files(files_index: dict[str, FileEntry]) -> dict[s
     return methods_index
 
 
+def _add_relation_endpoints_to_methods_index(
```
ivanmilevtues: This seems as patchwork, this should happen as one of the steps/updates that the incremental agent does.

agents/incremental_agent.py
```diff
@@ -56,6 +68,30 @@ def __init__(
...
+                ],
+            ),
+            "relation_analysis": PromptTemplate(
+                template=get_relation_analysis_message(),
+                input_variables=[
+                    "project_name",
+                    "component_summaries",
+                    "api_surfaces",
+                    "static_call_evidence",
+                    "meta_context",
```
ivanmilevtues: Meta context should be part of the system prompt

agents/incremental_agent.py
```diff
@@ -56,6 +68,30 @@ def __init__(
...
         self.meta_context = meta_context
+        self._scope_relation_contexts: dict[str, _ScopeRelationContext] = {}
+        self.prompts = {
+            "api_surfaces": PromptTemplate(
+                template=get_api_surfaces_message(),
+                input_variables=[
+                    "project_name",
+                    "component_summaries",
+                    "static_call_evidence",
+                    "meta_context",
```
ivanmilevtues: meta context is part of the system prompt as well as few of the other input vars

static_analyzer/cluster_relations.py
```diff
@@ -122,13 +122,15 @@ def _collect_component_names(
 def _collect_llm_relations(
     root_analysis: AnalysisInsights, sub_analyses: dict[str, AnalysisInsights]
 ) -> list[Relation]:
-    relations = list(root_analysis.components_relations)
+    relations: list[Relation] = []
     for _, sub_analysis in sorted(sub_analyses.items()):
         relations.extend(sub_analysis.components_relations)
+    # Refreshed scope relations override stale global copies loaded on root.
```
Svilen-Stefanov: blocker-ish: I don't think this comment actually holds. Appending root last only matters if something downstream respects list order, but both consumers below re-sort. `_ancestor_relation` sorts by `(-depth, src_id, dst_id, relation)` and the second loop re-sorts by `(src_id, dst_id, relation)`, so for the same (src_id, dst_id) pair the tie-break is the alphabetically-smaller `relation` label, not whichever copy is fresher. So a stale root relation from the previous run (we write globalized relations back to `root_analysis.components_relations` in `rebuild_global_relations` and reload them next run) can win over the refreshed scope copy. Since relation label/evidence stability is basically the point of the PR, can we resolve same-pair collisions by freshness explicitly instead of relying on sort order?

static_analyzer/cluster_relations.py
```diff
@@ -138,11 +140,27 @@ def _ancestor_relation_label(src_id: str, dst_id: str, llm_relations: list[Relat
         and is_self_or_descendant(dst_id, rel.dst_id)
     ]
     if not candidates:
-        return DEFAULT_STATIC_RELATION_LABEL
+        return None
     candidates.sort(
         key=lambda rel: (-(rel.src_id.count(".") + rel.dst_id.count(".")), rel.src_id, rel.dst_id, rel.relation)
```
Svilen-Stefanov: This is the static-backed half of it: when two candidates share the same (src_id, dst_id) at the same depth but have different `relation` labels, `candidates[0]` after this sort is just the alphabetically-first label. And the winner supplies `.relation`, `.evidence` and the inherited `key_edges` below, so stale metadata rides along. Won't a stale `"calls"` beat a fresh `"publishes to"` here purely on alphabetical order?

static_analyzer/cluster_relations.py
```diff
@@ -172,17 +190,34 @@ def build_global_relations(
...
+                dst_id=dst_id,
+                is_static=True,
+                all_edges=all_edges,
+            )
+        append_or_merge_relation(
+            global_relations,
+            relation,
         )
 
     for llm_rel in sorted(llm_relations, key=lambda rel: (rel.src_id, rel.dst_id, rel.relation)):
```
Svilen-Stefanov: Same thing for the LLM-only (no static edge) pairs, just via a different route. This re-sort throws away the root-last ordering, and then `append_or_merge_relation` merges by `pair_key` which ignores the label, and `merge_edges_from` keeps the existing relation's label + only takes evidence when the current one is empty. So again the alphabetically-first of a stale-vs-fresh pair wins. Worth fixing together with the one above, and a test with differing labels on the two copies would lock it in (the current override test reuses the same label so it can't fail).

agents/repair.py
```diff
@@ -0,0 +1,255 @@
...
+    ScopeUpdateDecision,
+    ScopedClusterRef,
+)
+from diagram_analysis.cluster_delta import ClusterRef
+from static_analyzer.graph import ClusterResult
+from static_analyzer.reference_resolver import StaticReferenceResolver
+
+logger = logging.getLogger(__name__)
+
+_EXISTING_COMPONENT_ACTIONS = frozenset(
```
Svilen-Stefanov: nit: `_EXISTING_COMPONENT_ACTIONS` here is identical to the one in `validation.py`, and `_cluster_ref_from_scoped_ref` at the bottom of this file is too. Both operate on the same `ScopeOperationAction`/`ScopedClusterRef` types from agent_responses and gate the same existing-component contract. Could we define these once (agent_responses.py?) and import in both, so a future action doesn't get added to one and not the other?

agents/repair.py
```diff
@@ -0,0 +1,255 @@
...
+
+        component.key_entities = repaired_references
+
+    if canonicalized_count:
+        logger.info("Repaired %d key-entity qualified name(s)", canonicalized_count)
+    if dropped_qnames:
+        logger.info("Dropped invalid or out-of-scope key entities: %s", sorted(dropped_qnames))
+
+
+def _nodes_in_scope(cluster_results: dict[str, ClusterResult]) -> set[str]:
```
Svilen-Stefanov: nit: `_nodes_in_scope` is the same triple comprehension as `_scope_key_entity_qnames` in incremental_planning_agent.py, and `_normalize_component_name` (line 116) is copied verbatim there too. planning_agent already imports from this module, so it can just reuse one exported helper. This kind of duplication keeps growing across the incremental files, could we pull the shared bits into one place?

agents/incremental_agent.py
```diff
@@ -274,6 +360,58 @@ def generate_all_scope_relations(
         if all_llm_rels:
             _log_scope_relations_summary(all_llm_rels)
 
+    def _scope_contexts(self) -> dict[str, _ScopeRelationContext]:
+        contexts = getattr(self, "_scope_relation_contexts", None)
```
Svilen-Stefanov: nit: `__init__` always sets `self._scope_relation_contexts = {}`, so this `getattr(..., None)` plus the `if contexts is None` branch is dead, the method is just `return self._scope_relation_contexts`. Can we inline it at the two call sites and drop the accessor?

diagram_analysis/diagram_generator.py
```diff
@@ -848,25 +860,32 @@ def _refresh_files_index(
...
-            self.abstraction_agent.refresh_method_spans_from_cfg(analysis)
-        for sub in sub_analyses.values():
-            sub.files = self.abstraction_agent.build_files_index(sub)
-        unified_files = self.abstraction_agent.build_files_index(root_analysis)
+        """Rebuild live per-scope file indexes and union them into the root index."""
+        assert self.static_analysis is not None
+        analyses = (root_analysis, *sub_analyses.values())
+        for analysis in analyses:
+            refresh_method_spans_from_cfg(analysis, self.static_analysis, self.repo_location)
+            analysis.files = build_files_index(analysis, self.repo_location)
```
Svilen-Stefanov: nit: each analysis gets its own fresh source cache here since we don't pass a `source_cache` into `build_files_index`. Root and the sub-analyses usually share file paths, so the same file gets read + whole-file-hashed once per analysis. Could we thread one `SourceCache` dict through the loop to read each file once? (refresh_method_spans doesn't need it, it reads CFG spans.)

static_analyzer/incremental_orchestrator.py
```diff
@@ -264,7 +269,20 @@ def _containing_callable_nodes(call_graph: CallGraph, file_path: Path, line: int
 
 
 def _definition_nodes(call_graph: CallGraph, definition: dict) -> list[Node]:
-    return [node for node in call_graph.nodes.values() if _definition_points_to_node(definition, node)]
+    matches = [node for node in call_graph.nodes.values() if _definition_points_to_node(definition, node)]
+    if not matches:
+        return []
+    return [
```
Svilen-Stefanov: blocker-ish?: I want to double check this one since it goes to the "add the right edges" goal. Before this PR `_definition_nodes` returned every node whose span contains the definition position, now we keep only the innermost one. For a constructor call, doesn't the definition land inside both `Foo.__init__` and `class Foo` (the class span wraps its methods)? If so the old code emitted both `caller -> __init__` and `caller -> Foo`, and now we drop the `caller -> Foo` edge. The full build deliberately adds that class-parent edge (edge_builder.py, the `is_callable(target.kind) and target.parent_chain` block), and this is the only place incremental adds these. So won't incremental now be missing the caller->class edge a full run produces? 

I don't know what the right behavior should be, just noting the potential inconsistency.

static_analyzer/reference_resolver.py
```diff
@@ -37,20 +46,67 @@ def fix_source_code_reference_lines(self, analysis: AnalysisInsights) -> Analysi
...
+        seen_qnames: set[str] = set()
+        canonicalized_count = 0
+        unresolved_qnames: set[str] = set()
+        for reference in references:
+            original_qname = reference.qualified_name
+            if force_resolve:
+                reference.reference_file = None
+                reference.reference_start_line = None
+                reference.reference_end_line = None
+                resolved = self._resolve_symbol_reference(reference, original_qname.replace(os.sep, "."))
```
Svilen-Stefanov: I'm a bit worried this force-resolve can move a key entity out of its component. On main we regenerated touched key entities from the component's own `file_methods`, so they were pinned to the component's files. Here we null the file and re-resolve via exact -> loose -> token match, and `_try_symbol_token_match` pulls candidates repo-wide through `iter_reference_nodes`. So if exact+loose both miss (say an unrelated edit shifted things) and there's a unique same-leaf match in another file, won't `a.py::Handler.run` silently rebind to `b.py::Worker.run` and hop components? Nothing downstream seems to catch it (validate_key_entities only rejects empty, remove_unresolved only drops non-existent files). Could we scope the symbol/token match to the component's own file_paths first, and only fall back repo-wide if that fails?

tests/agents/test_incremental_planning_agent.py
```diff
@@ -217,6 +249,252 @@ def test_validate_scope_update_decision_allows_create_when_llm_chooses_new_compo
...
+        existing_component_ids={"1"},
+    )
+
+    result = validate_scope_update_decision(decision, context)
+
+    assert not result.is_valid
+    assert "noop operations must preserve" in result.feedback_messages[0]
+
+
+def test_repair_scope_update_decision_clears_name_used_to_route_noop() -> None:
```
Svilen-Stefanov, line 465: nit / test note: both `decide_scope_update` tests mock `_invoke_repair_validate`, so the real repair -> second validate -> raise path isn't exercised end to end. Combined with the same-label thing I flagged in cluster_relations, I think a couple of the stability guarantees here aren't actually pinned by a test yet. Not blocking, but worth a targeted test (e.g. a NOOP routed by cluster ownership that still carries a description, and a differing-label stale-vs-fresh relation).

agents/prompts/abstract_prompt_factory.py
```diff
@@ -77,9 +77,7 @@ def get_relation_analysis_message(self) -> str:
         return RELATION_ANALYSIS_MESSAGE
 
 
-API_SURFACES_MESSAGE = """Analyze component API surfaces for `{project_name}`.
-
-Project Type: {project_type}
```
Svilen-Stefanov, line 82: Makes sense that we don't need this everywhere. I feel like we probably need to remove it from other prompts as well.

agents/prompts/claude_prompts.py
```diff
@@ -423,9 +419,17 @@
 - For modified clusters, preserve the existing owning component shown by its clusters=[...] list; use update_component for that owner instead of moving the cluster to another component.
 - For new clusters, decide from the structural diff whether they extend an existing responsibility or introduce a new component; do not infer this from file/package layout alone.
 - For reshaped groups, follow overlap counts to keep old cluster ownership stable. Only assign a reshaped new cluster to a different component when the diff proves a real responsibility move.
-- Do not reparent existing components. If reparenting seems required, use regenerate_scope.
+- Do not reparent existing components. Preserve their current scope; reparenting is not a valid incremental operation.
```
Svilen-Stefanov: Seems a bit strong, probably not true long term but it should for sure be stricter.

agents/prompts/claude_prompts.py
```diff
@@ -423,9 +419,17 @@
...
 - Every modified/new/reshaped new-side cluster listed below must appear in exactly one operation's cluster_refs.
 
+<architecture_output_contract>
+- This step plans component boundaries only. Do not define component relations; API surfaces and relations are generated later.
+- Preserve an existing component's name, description, and key entities unless its architectural responsibility changed.
+- For create_component, provide a clear name and description. Select up to 5 key_entities only when their exact qualified names are available; otherwise leave them empty. Key entities are not synthesized later.
+- For update_component, include refreshed name, description, or key_entities only when the component's responsibility changed. An empty key_entities list preserves the current selection.
+- For update_component, delete_component, and noop, copy the exact component_id from the existing-components list.
```
Svilen-Stefanov: I was expecting 3 "switch case"-ish branches here but this one is a bit inclusinve of the second one, not sure what's better for the agent.

agents/abstraction_agent.py
```diff
@@ -119,11 +122,13 @@ def step_clusters_grouping(self, cluster_results: dict[str, ClusterResult]) -> C
             project_type=project_type,
         )
 
-        cluster_analysis = self._validation_invoke(
+        cluster_analysis = self._invoke_repair_validate(
             prompt,
             ClusterAnalysis,
+            repairs=[],
```
Svilen-Stefanov: A bit strange that we call repair with 0 repairs. Also seems redundant to have the repairs + repair context but I think we have a similar approach elsewhere. Still, seems a bit redundant - there are other occurrences of this.

agents/agent.py
```diff
@@ -227,7 +227,13 @@ def _parse_invoke(self, prompt: str, type: type, include_hidden: bool = False):
         assert isinstance(response, str), f"Expected a string as response type got {response}"
         return self._parse_response(prompt, response, type, include_hidden=include_hidden)
 
-    def _score_result(self, result, validators: list, context) -> tuple[float, list[tuple[float, str]]]:
+    def _repair_result(self, result, repairs: list, repair_context) -> None:
```
Svilen-Stefanov: nit: stronger typing?

agents/agent.py
```diff
@@ -227,7 +227,13 @@ def _parse_invoke(self, prompt: str, type: type, include_hidden: bool = False):
...
+        """Apply deterministic repairs to one parsed candidate."""
+        for repair in repairs:
+            logger.info("[Repair] Applying %s", repair.__name__)
+            repair(result, repair_context)
+
+    def _score_result(self, result, validators: list, validation_context) -> tuple[float, list[tuple[float, str]]]:
```
Svilen-Stefanov: same, stronger typing, maybe check elsewhere too

agents/cluster_ids.py
```diff
@@ -9,6 +12,10 @@ def sort(cls, cluster_ids: set[GraphClusterId]) -> list[GraphClusterId]:
 
 
 class CodeBoardingClusterIds:
+    @classmethod
+    def prefix_for_scope(cls, scope_id: str) -> CodeBoardingClusterId:
+        return "" if scope_id == ROOT_SCOPE_ID else scope_id
```
Svilen-Stefanov, line 17: Uh, this is a bit confusing with this type aliases and I'm wondering if this is the right place for this method given the import but seems fine for now.

agents/details_agent.py
```diff
@@ -55,7 +57,14 @@ def __init__(
         parsing_llm: BaseChatModel,
         run_id: str,
     ):
-        super().__init__(repo_dir, static_analysis, get_system_details_message(), agent_llm, parsing_llm)
+        meta_context_str = meta_context.llm_str() if meta_context else "No project context available."
```
Svilen-Stefanov: Isn't all of this shared between 4 different agents:
```
        meta_context_str = meta_context.llm_str() if meta_context else "No project context available."
        project_type = meta_context.project_type if meta_context else "unknown"
        system_message = get_system_message().format(
            project_name=project_name,
            project_type=project_type,
            meta_context=meta_context_str,
        )
```
Can't we reuse it? There's probably more potential for code reuse among the agents overall.

agents/details_agent.py
```diff
@@ -235,21 +250,18 @@ def step_final_analysis(
...
             component_summaries=analysis.llm_str(),
             static_call_evidence=static_call_evidence,
-            meta_context=meta_context_str,
-            project_type=project_type,
         )
-        return self._validation_invoke(
+        return self._invoke_repair_validate(
             prompt,
             ComponentApiSurfaces,
+            repairs=[],
```
Svilen-Stefanov: There're some more empty repairs than full ones it seems.

agents/relation_edges.py
```diff
@@ -1,4 +1,8 @@
-from agents.agent_responses import Relation, RelationEdge
+from pathlib import Path
+
+from agents.agent_responses import AnalysisInsights, Relation, RelationEdge
+from agents.file_index_models import FileEntry, MethodEntry
+from repo_utils.path_utils import normalize_repo_path
 
 
 def merge_relation_edges(
```
Svilen-Stefanov: nit: this is a one-line pass-through to `Relation._merge_edges` with the same args, so it's the kind of wrapper AGENTS.md asks us to inline. Both callers (cluster_relations.py) already import `Relation` and build a `Relation(...)` right after, so they could call `Relation._merge_edges` directly (or we promote it to public if the underscore is the reason for the wrapper). Not important, just noticed it while reading the edge merging.

agents/details_agent.py
```diff
@@ -357,7 +366,10 @@ def run(self, component: Component):
         # Step 9: Fix source code reference lines (resolves reference_file paths)
         analysis = self.reference_resolver.fix_source_code_reference_lines(analysis)
 
-        # Step 10: Ensure unique key entities across components
+        # Step 10: Index relation endpoints after reference resolution
+        index_relation_endpoints(analysis, self.repo_dir)
```
Svilen-Stefanov, line 338: Just to confirm, this does impact the full analysis as well, right? I expect the runtime here to increase again.

agents/relation_edges.py
```diff
@@ -29,3 +33,31 @@ def merge_relations_by_pair(relations: list[Relation], include_relation: bool =
     for relation in relations:
         append_or_merge_relation(merged, relation, include_relation=include_relation)
     return merged
+
+
+def index_relation_endpoints(analysis: AnalysisInsights, repo_dir: Path) -> None:
```
Svilen-Stefanov, line 33: Follow-up from reviewing the incremental/relation code: I did a dead-code + duplication sweep over the files this PR touches (and the modules around them). Most of it isn't on the changed lines so I couldn't leave inline comments (they'd land on unrelated code), so grouping here. None of this blocks the PR, but a lot of it sits right in the blast radius, so might be worth clearing while we're in here or in a quick follow-up.

**Duplication (the one I'd actually like to fix):**
- `static_analyzer/reference_resolver.py:424` `_symbol_tokens` is byte-for-byte identical to `reference_tokens` in `static_analyzer/internal_references.py:19` (same regex, same list comp), and reference_resolver already imports from that module. Could we drop the static method and reuse `reference_tokens` at the 4 call sites? It's on the resolution hot path so nice to have one copy.

**Dead pydantic response models in `agents/agent_responses.py`** (each has zero references anywhere, incl. tests and structured-output dispatch, so they look like leftovers from the incremental refactor):
- `CFGAnalysisInsights` + `CFGComponent` (:704 / :685)
- `ExpandComponent` (:719) - looks superseded by the deterministic `should_expand_component()` in planner_agent
- `ValidationInsights` (:729)
- `UpdateAnalysis` (:742) - orphaned after its caller was removed
- `FilePath` (:882)
- `Relation.analysis_dump()` (:380) - unwired serialization helper, real path is `_relation_to_json`
- `AnalysisInsights.file_to_component()` (:532)

**Dead validators in `agents/validation.py`** (not in any `validators=[...]` list or `VALIDATOR_WEIGHTS`, so `agent.py`'s name-based dispatch can never reach them):
- `validate_existing_component_ids` (:240) - its job is already covered by the live `validate_scope_update_decision`
- `validate_scope_relation_names` (:629) - and it's the only consumer of the `ScopeRelations` model (`agent_responses.py:808`), so that model dies with it (live scope relations parse into `ComponentRelations`)

**Test-only helpers (keep-or-drop call, not defects):**
- `validation.py:716` `_check_edge_between_cluster_sets` - only `test_validation.py` calls it (its delegates `_build_cluster_edge_lookup`/`_cluster_sets_have_edge` are used by the live `validate_relation_evidence`, so dropping the wrapper is safe)
- `diagram_analysis/cluster_snapshot.py:86` `snapshot_from_cluster_results` - only tests; its own docstring says "(test helper)"

**Minor dead pass-throughs (lower priority, probably a ticket):**
- `diagram_analysis/diagram_generator.py:645` `_collect_method_entries_from_static_analysis` - never called
- `diagram_analysis/io_utils.py` `load_root_analysis` (:274), `load_sub_analysis`/`save_sub_analysis` (:370/:375), `detect_expanded_components` (:180) - unreferenced one-line store wrappers
- `static_analyzer/cluster_helpers.py:496` `get_files_for_cluster_ids` - zero refs

**One cluster outside the PR area** (ticket, not this PR): the validation-prompt getters `get_validator_system_message` / `get_component_validation_component` / `get_relationships_validation` plus their constants are duplicated across all 6 provider prompt files and are only reachable through `get_all_prompts()`, whose sole caller is `test_prompts.py`. So `get_all_prompts()` + those three getters across every provider are effectively test-only and could go together.

You can decide what you address and what we create a ticket for

agents/incremental_agent.py
```diff
@@ -90,6 +125,13 @@ def update_scope(
                     result.removed_ids.add(operation.component_id)
                 continue
             if operation.action == ScopeOperationAction.NOOP:
+                component = components_by_id.get(operation.component_id or "")
```
Svilen-Stefanov, line 131: Do we need the or ""? Can't this have a default get if the component is not found?

agents/repair.py
```diff
@@ -0,0 +1,255 @@
...
+)
+_KEY_ENTITY_METADATA_ACTIONS = frozenset(
+    {
+        ScopeOperationAction.CREATE_COMPONENT,
+        ScopeOperationAction.UPDATE_COMPONENT,
+    }
+)
+
+
+class ComponentRepairTarget(Protocol):
```
Svilen-Stefanov, line 36: Exactly the same as `ComponentValidationTarget`, not sure if we need both. I generally wonder if there are more such cases and if the type aliasing made things worse but let's see.

diagram_analysis/diagram_generator.py
```diff
@@ -857,25 +869,32 @@ def _refresh_files_index(
...
+            index_relation_endpoints(analysis, self.repo_location)
+
+        unified_files = root_analysis.files
         for sub in sub_analyses.values():
             for fp, entry in sub.files.items():
-                unified_files.setdefault(fp, entry)
+                indexed_entry = unified_files.get(fp)
+                if indexed_entry is None:
```
Svilen-Stefanov: Interesting that this union does `= method` (aliases the sub-analysis's live MethodEntry into the root index), while the save-time twin in `analysis_json._build_files_index_from_analysis` does `= method.model_copy(deep=True)`. So here the root and the sub-analysis end up sharing the same MethodEntry objects, and anything that later mutates a method in the root index in place (index_relation_endpoints bumps start/end on REFERENCE entries, for example) would also mutate the sub-analysis copy. Is that intended? Feels like the two "same" merges have quietly drifted. This is really the symptom of there being no owner for the file-index merge rule (it lives in 4 places now); I think a small `FileEntry.merge_from(other)` / a `FileIndex` value object that owns build + merge + REFERENCE-precedence + sort would kill the drift and the aliasing question.

agents/relation_edges.py
```diff
@@ -29,3 +33,31 @@ def merge_relations_by_pair(relations: list[Relation], include_relation: bool =
...
+                    (method for method in entry.methods if method.qualified_name == reference.qualified_name),
+                    None,
+                )
+                if indexed is None:
+                    entry.methods.append(
+                        MethodEntry(
+                            qualified_name=reference.qualified_name,
+                            start_line=reference.reference_start_line or 0,
+                            end_line=reference.reference_end_line or 0,
+                            node_type="REFERENCE",
```
Svilen-Stefanov: `node_type="REFERENCE"` is a string that isn't a `NodeType` member (`NodeType.from_name("REFERENCE")` would raise), and the "a real METHOD supersedes a REFERENCE placeholder" rule is then re-spelled at ~4 sites (here, analysis_json.py:182, diagram_generator.py:883, analysis_json.py:537), two of which this PR adds. Since "is this a real analyzed method or just a relation pointer" is a real concept, could we make it one, e.g. an `is_reference: bool` (or a small kind enum) on MethodEntry plus one `prefer()`/`merge` helper that owns the precedence? Right now the sentinel and its precedence live in a free-form `str` and get hand-maintained everywhere.

agents/relation_edges.py
```diff
@@ -29,3 +33,31 @@ def merge_relations_by_pair(relations: list[Relation], include_relation: bool =
...
         append_or_merge_relation(merged, relation, include_relation=include_relation)
     return merged
+
+
+def index_relation_endpoints(analysis: AnalysisInsights, repo_dir: Path) -> None:
+    """Add relation endpoint references to the analysis file index."""
+    for relation in analysis.components_relations:
+        for edge in [*relation.key_edges, *relation.all_edges]:
+            for reference in (edge.source, edge.target):
+                file_path = normalize_repo_path(reference.reference_file or "", repo_dir)
```
Svilen-Stefanov: This uses `""` as the file key when an endpoint has no `reference_file` (external / REST / queue endpoints, which are legitimately fileless). Two things fall out: every fileless endpoint from unrelated relations collapses into one `analysis.files[""]` bucket, and while `_build_file_entry_json_from_files` drops the `""` entry (the `if file_path` guard), `_build_methods_index_from_files` does not, so orphaned `"|qname"` rows persist in the methods_index. Could we just skip indexing endpoints whose `reference_file` is None here (they're already represented by the RelationEdge itself), rather than routing them through a `""` key? I wouldn't make reference_file non-nullable though, external endpoints really don't have a file.

agents/incremental_agent.py
```diff
@@ -195,61 +238,104 @@ def _patch_scope_file_methods(
...
+        return self.step_relation_analysis(
+            scope,
+            scope_name,
+            api_surfaces,
+            cluster_analysis,
+            cluster_results,
+            cfg_graphs,
+        )
 
     @trace
```
Svilen-Stefanov, line 328: This is a bit subtle: `generate_scope_relations` here is called with cluster_results/cfg_graphs=None, so the relation phase has no explicit data dependency on the apply phase, it reaches into `self._scope_relation_contexts[scope_id]` that `update_scope` populated as a side effect (and raises IncrementalScopeContextMissingError if it's missing). So the "apply before generate" ordering is enforced by a runtime raise rather than by construction, and it only works because `touched_scopes` (refresh|removed) and the context write condition (refresh|new|removed) happen to line up. Since `_apply_incremental_scope_recursively` already threads `touched_scopes` up through `RecursiveScopeUpdateResult`, could we thread the per-scope contexts up the same way and pass them in explicitly? That'd make the ordering unrepresentable-if-violated and let us drop both the instance dict and those never-used optional args.

static_analyzer/incremental_orchestrator.py
```diff
@@ -155,29 +158,31 @@ def _restore_inbound_edges(
                 refs = []
```
Svilen-Stefanov, line 165: blocker-ish?: this swallows any exception from the LSP `references()` call into `refs = []` at debug level, and an empty refs then means `_edge_reference_call_sites` returns nothing, so the cached cross-boundary edge is silently dropped. So a transient LSP hiccup during warm-start would make a real, previously-validated edge disappear from the incremental result with no visible signal, which cuts against the "add/preserve the edges" goal. Should a genuine LSP failure here (vs a legitimately-empty references result) at least be surfaced, or the edge kept-pending rather than dropped? Feels risky to treat "the server errored" the same as "no references exist".

Svilen-Stefanov (changes requested):
Leaving the comments I have so far + some from some agents I started. Take a look, I hope they clean up things a bit and fix some issues.

ivanmilevtues:
Addressed the collected quick-win and nit review items in commit `fd78616`.

### Correctness fixes

- Normalize every NOOP operation by clearing generated name, description, and key-entity metadata before validation.
- Canonicalize empty scoped cluster IDs to the root scope.
- Resolve qname-only file references relative to the repository.
- Honor `CODEBOARDING_DISABLE_CACHE_REUSE` on both static-analyzer construction paths.
- Skip fileless relation endpoints instead of persisting empty-path file and method-index entries.
- Deep-copy sub-analysis file-index entries and methods while building the unified root index.

### Cleanup and performance

- Added shared scope-operation contracts for action sets, cluster refs, member qnames, and component-name normalization.
- Reused one source cache across scope file-index rebuilds and one `SourceInspector` across incremental edge restoration.
- Replaced per-endpoint method scans with per-file qualified-name lookups.
- Removed the dead scope-context accessor, duplicate symbol tokenization, and one-line relation-edge wrapper.
- Strengthened generic repair and validation typing, and narrowed relation validators to their structural protocol.
- Moved file-index tests into `tests/diagram_analysis/test_file_index.py` and applied the smaller annotation/docstring cleanups.

### Validation

- Focused changed-area suite: **185 passed**.
- `mypy .`: **clean across 319 source files**.
- Full non-integration suite excluding only the environment-contaminated package-discovery check: **1,809 passed, 28 skipped**, **81.48% coverage**.
- The unfiltered required command also reached **81.48% coverage** with **1,810 tests passing**; its sole failure was `tests/test_pyproject_packages.py`, which discovers pre-existing Python packages under `temp/repos/` as undeclared project packages.

static_analyzer/reference_resolver.py
```diff
@@ -37,20 +46,57 @@ def fix_source_code_reference_lines(self, analysis: AnalysisInsights) -> Analysi
...
+            has_existing_file = reference.reference_file is not None and self.reference_file_exists(
+                reference.reference_file
+            )
+            has_complete_location = (
+                has_existing_file
+                and reference.reference_start_line is not None
+                and reference.reference_end_line is not None
+            )
+            resolved = has_complete_location or self.resolve_reference(reference, file_candidates)
+            if not resolved and not has_existing_file:
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616. Extension candidates are now built from the repository-relative full path, and a regression test covers qname-only resolution without file candidates.

tests/static_analyzer/test_reference_resolver.py
```diff
@@ -89,6 +89,35 @@ def test_resolve_reference_uses_matching_file_candidate(self):
...
+
+        def get_reference(_language, qname):
+            if qname == canonical_qname:
+                return node
+            raise ValueError("not found")
+
+        def get_loose_reference(_language, qname):
+            if qname == "OCR.extract_text":
+                return canonical_qname, node
+            return None, None
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616. The test and both resolver callbacks now have explicit parameter and return annotations matching the resolver contract.

agents/repair.py
```diff
@@ -0,0 +1,255 @@
...
+        if not needs_routing_repair:
+            continue
+
+        component_id = _resolve_unambiguous_component_id(operation, context)
+        if component_id is not None:
+            operation.component_id = component_id
+            if operation.action == ScopeOperationAction.NOOP:
+                operation.name = None
+            routed_operations += 1
+    return routed_operations
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616. NOOP operations now always clear name, description, and key_entities independently of routing repair, with coverage for an already-valid component_id.

agents/validation.py
```diff
@@ -72,6 +87,60 @@ class RelationValidationTarget(Protocol):
...
+        errors.append(f"Missing cluster_refs: {_format_cluster_ref_list(missing)}")
+    if extra:
+        errors.append(f"Unexpected cluster_refs: {_format_cluster_ref_list(extra)}")
+    if duplicates:
+        errors.append(f"Duplicate cluster_refs: {_format_cluster_ref_list(duplicates)}")
+    return ValidationResult(is_valid=not errors, feedback_messages=errors)
+
+
+def _cluster_ref_from_scoped_ref(ref: ScopedClusterRef) -> ClusterRef:
+    return ClusterRef(ref.language, ref.cluster_id, ref.scope_id)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616. Scoped cluster refs now canonicalize an empty scope_id to ROOT_SCOPE_ID, with a regression test for the root comparison path.

diagram_analysis/diagram_generator.py
```diff
@@ -246,21 +252,36 @@ def _changed_files_for_static_analysis(self) -> set[Path] | None:
...
+        if self.force_full_analysis:
+            logger.info("Force full analysis: skipping static analysis cache")
+        return get_static_analysis(
+            self.repo_location,
+            skip_cache=self.force_full_analysis,
+            source_sha=self.source_sha,
+            cache_dir=self.output_dir,
+            changed_files=self._changed_files_for_static_analysis(),
+        )
+
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616. The new-analyzer path now applies CODEBOARDING_DISABLE_CACHE_REUSE to skip_cache and logs the override, matching the injected-analyzer path.

tests/agents/test_details_agent.py
```diff
@@ -382,13 +383,13 @@ def test_run(self, mock_fix_ref, mock_validation_invoke):
 
         api_response = ComponentApiSurfaces(api_surfaces=[])
         relation_response = ComponentRelations(components_relations=[])
-        mock_validation_invoke.side_effect = [cluster_response, final_response, api_response, relation_response]
+        mock_invoke_repair_validate.side_effect = [cluster_response, final_response, api_response, relation_response]
         mock_fix_ref.return_value = final_response
 
         analysis, subgraph_results = agent.run(self.test_component)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in fd78616 by renaming the unused unpacked value to _subgraph_results.

agents/repair.py
ivanmilevtues: Fixed in fd78616. EXISTING_COMPONENT_ACTIONS and scoped-ref conversion now live once in agents/scope_operations.py and are imported by repair and validation.

agents/repair.py
ivanmilevtues: Fixed in fd78616. Cluster-member flattening and component-name normalization are now shared through agents/scope_operations.py instead of being duplicated across repair and planning.

agents/incremental_agent.py
ivanmilevtues: Fixed in fd78616. The dead lazy accessor was removed and both call sites now use the initialized _scope_relation_contexts field directly.

ivanmilevtues: Fixed in fd78616. The dead lazy accessor was removed and both call sites now use the initialized _scope_relation_contexts field directly.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed in fd78616. _refresh_files_index now creates one SourceCache and passes it through every root/sub-analysis build_files_index call; the shared identity is covered by a test.

tests/agents/test_incremental_planning_agent.py
ivanmilevtues: Partially addressed in fd78616: a decide_scope_update test now exercises the real repair -> validation path for a NOOP carrying extraneous metadata. The differing-label relation case remains part of the broader relation-freshness work.

agents/agent.py
ivanmilevtues: Fixed in fd78616. _parse_invoke and the repair/validation pipeline now preserve generic result and context types instead of using bare type/list parameters.

agents/agent.py
ivanmilevtues: Fixed in fd78616. Validator callables and contexts are now generically typed end to end; relation validators were also narrowed to their structural protocol. mypy is clean across 319 files.

agents/relation_edges.py
ivanmilevtues: Fixed in fd78616. The one-line merge_relation_edges wrapper was removed and the two callers invoke Relation._merge_edges directly.

agents/relation_edges.py
ivanmilevtues: Addressed the requested duplication item in fd78616: _symbol_tokens was removed and all resolver call sites now reuse internal_references.reference_tokens. The broader dead-code and ticket candidates in this grouped comment were intentionally left unchanged.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed in fd78616. Root file-index union now deep-copies both newly inserted FileEntry objects and methods merged from sub-analyses, with regression coverage for both aliasing paths.

agents/relation_edges.py
ivanmilevtues: Fixed in fd78616. index_relation_endpoints now skips endpoints without a reference_file, so no empty-path file bucket or methods_index row is persisted; serialization coverage was updated accordingly.

agents/details_agent.py
ivanmilevtues: Yes, endpoint indexing also runs in full analysis. It is an in-memory O(methods + relation endpoints) pass with no extra LLM, LSP, or file reads, so the expected increase is small.

agents/relation_edges.py
ivanmilevtues: Removed the newly stranded `_collect_method_entries_from_static_analysis` in this PR. I kept the broader models/validators/I/O wrappers/test-only prompt API sweep out of scope and tracked it separately in #410.

PR #409, opened 2026-07-12
feat(core): fail fast on LLM auth errors instead of retrying blind

Svilen-Stefanov:
## Problem

`AuthenticationError` is the **3rd most common crash** in PostHog (118 in 90d, 14 distinct users) — almost entirely bad/expired/placeholder LLM API keys, not an analysis defect. Splitting the data shows two problems: a broad first-run misconfiguration cohort (June: 10 users across vscode/oss/sync/github_action) and a narrow automation cohort (July: one `oss` user's dead key firing ~hourly for 4 days).

Today a rejected 401 is mishandled at three layers:

1. **Retried like a transient error** — `Agent._invoke`'s `classify()` only special-cases HTTP 404 and `ResourceExhausted`; a 401 falls through to the generic 5×-with-backoff path.
2. **Swallowed per-component** — the two broad `except Exception` blocks in `diagram_generator.py` catch the auth error, log it, and continue to the next component with the same dead key. The worst run emitted **47 `analysis_started` + 47 `analysis_completed` + 38 `$exception`** under one `run_id`.
3. **Raw traceback on the CLI** — `main()` had no top-level handler; the analysis call wasn't guarded (only `bootstrap_environment` was), so a mid-run 401 surfaced as an unhandled traceback + exit 1.

## Fix (Phase A — Core)

- **New `agents/llm_errors.py`** — `LLMAuthError` (typed, terminal; carries `provider`, masked `key_tail`, and `telemetry_properties`) + provider-agnostic `detect_auth_error()` recognizing 401/403 across openai/anthropic/google/AWS and message-only re-wraps (including the `Invalid GitHub OIDC token` signature).
- **`agents/agent.py`** — both `classify()` paths detect a 401 and **raise `LLMAuthError` immediately**. Kills the retry storm and bypasses the silent `"Could not get response from the agent."` fallback for auth failures.
- **`diagram_analysis/diagram_generator.py`** — the two broad `except` blocks **re-raise `LLMAuthError`**, so a dead key aborts the whole run instead of degrading component-by-component.
- **`main.py`** — `_dispatch` catches `LLMAuthError`, prints one actionable line to stderr (no traceback), and exits with a **distinct code `2`** so automation/CI can branch on "fix your key."
- **Telemetry** — props ride on the exception, so the `$exception` event (and the dashboard's structured `error` columns, previously null) now carry `error_type=auth`, provider, status, and masked key tail.
- **`agents/llm_config.py`** — `current_provider_key_context()` returns provider + masked key tail, never the secret.

## Tests

- `tests/agents/test_llm_errors.py` — detection across provider exception shapes, non-auth pass-through, existing-error pass-through, telemetry props, message truncation, provider/key-tail masking.
- `tests/agents/test_agent.py::...::test_invoke_auth_error_not_retried` — a 401 fails on the **first** attempt (call count 1, no sleep) as a typed `LLMAuthError`.
- `tests/test_main.py::TestMainAuthErrorHandler` — the CLI exits with code 2 on auth error and does **not** swallow other exceptions.

All green: `black --check` clean, `mypy` 0 errors on changed files, full agents/telemetry/diagram_analysis suites pass.

## Scope

Phase A only — Core fail-fast. Follow-ups (separate PRs): **Phase B** maps `LLMAuthError` to a dedicated JSON-RPC error code in the wrapper (so the vscode extension doesn't depend on stderr truncation) and routes auth failures to the extension's existing "Open Settings" modal. GitHub-OIDC `gha_proxy` failures are a separate proxy-token workstream.

## Summary by CodeRabbit

- **Bug Fixes**
  - Detects LLM credential rejections consistently across supported providers.
  - Authentication failures now fail fast by terminating retries during both invocation and structured parsing.
  - Diagram analysis stops immediately on rejected/auth-related LLM failures.
  - Command-line runs now exit cleanly with a dedicated code and a user-friendly message (no traceback), including provider details with masked key info.

- **Tests**
  - Added unit tests for auth-error detection, retry prevention, key masking, and top-level CLI exit behavior.

### DIFF
--- a/agents/llm_errors.py
+++ b/agents/llm_errors.py
@@ -17,6 +17,12 @@ from __future__ import annotations
 
 import re
 
+# Process exit code for a rejected key. Distinct from 1 (generic failure) so the
+# OSS CLI and the wrapper subprocess both signal "fix your key" the same way, and
+# callers/CI can branch on it. Lives here (not in main.py) so the wrapper can
+# import it without pulling in Core's whole CLI module.
+EXIT_AUTH_ERROR = 2
+
 # Class names, across SDKs, that always mean "credentials were rejected".
 # openai/anthropic/cerebras raise ``AuthenticationError``; google raises
 # ``Unauthenticated``/``PermissionDenied``; Bedrock (botocore) surfaces
--- a/agents/llm_errors.py
+++ b/agents/llm_errors.py
@@ -1,0 +1,119 @@
+"""Provider-agnostic detection and typing of LLM authentication failures.
+
+A rejected API key (HTTP 401, or a provider's equivalent auth/permission
+error) is *permanent* for the run: retrying it wastes minutes of backoff and
+floods telemetry with identical ``$exception`` events. The agent retry loop and
+the CLI both need to recognize one when they see it, regardless of which SDK
+raised it.
+
+``detect_auth_error`` maps a raw provider exception to an :class:`LLMAuthError`
+(or ``None`` when it isn't an auth failure). The typed error carries the
+provider, a masked key tail, and the provider's own message, and forwards them
+as ``telemetry_properties`` so the PostHog ``$exception`` event — and the
+dashboard's structured ``error`` columns — populate instead of staying null.
+"""
+
+from __future__ import annotations
+
+import re
+
+# Class names, across SDKs, that always mean "credentials were rejected".
+# openai/anthropic/cerebras raise ``AuthenticationError``; google raises
+# ``Unauthenticated``/``PermissionDenied``; Bedrock (botocore) surfaces
+# ``AccessDeniedException``/``UnrecognizedClientException`` in the error text.
+_AUTH_TYPE_NAMES = {
+    "AuthenticationError",
+    "Unauthenticated",
+    "PermissionDeniedError",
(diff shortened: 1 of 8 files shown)


agents/llm_errors.py
```diff
@@ -0,0 +1,125 @@
...
+
+from __future__ import annotations
+
+import re
+
+# Process exit code for a rejected key. Distinct from 1 (generic failure) so the
+# OSS CLI and the wrapper subprocess both signal "fix your key" the same way, and
+# callers/CI can branch on it. Lives here (not in main.py) so the wrapper can
+# import it without pulling in Core's whole CLI module.
+EXIT_AUTH_ERROR = 2
```
Svilen-Stefanov, line 24: This is really a hack - we need a better way to communicate such errors.

ivanmilevtues: Yea I see, I also asked chatgpt for more info on ther return types and the errors around the langchain API, however there is no unification and came up with a similar approach, it also mentioned that pure 401 status code validation is not enough on its own. So I think what is done here is probably the best we can for the moment.

ivanmilevtues (approved):
Looks good to me this will save everyone quite some time so nice.

PR #422, opened 2026-07-18
Surgical incremental updates: member-granular change signal + copy-forward (ProgramGraph)

Svilen-Stefanov:
## What this does

Makes an incremental analysis touch **only the components whose code actually changed**, instead of re-writing the whole diagram. On a 5-file edit that genuinely touched 2 architectural components, the number of modified top-level components drops from **6/6 → 2/6**, method-owner stability rises to **98.4%**, and all **29 invariants stay clean**.

## The problem

Editing a few method bodies lit up *every* high-level component, with heavy relation churn. Verified root cause: the "modified" decision was **file-granular** — a component was marked changed when its whole member-file set intersected the changed *file paths*. Because a file's methods legitimately disperse across many components, one body edit dirtied them all. (The per-method `content_hash` that would fix this already lived in `analysis.json` but was never diffed.) On top of that, the re-clustering reshuffled unchanged methods and re-generated their descriptions/relations.

Measured: of 6 top-level components, only **2** owned a genuinely body-changed method; the other 4 were spurious.

## The fix (two changes)

1. **Member-granular change signal** — a component is dirty only when *its own* members' `content_hash` changed (or a member was added/removed), not when its member-file set intersects the changed files. Reuses the per-method `content_hash` already in `analysis.json` (previously written but never diffed). Computed independently of the graph fingerprint, so a body-only edit is detected.
2. **Copy-forward unchanged components** — after the re-partition, pin every unchanged method to its baseline owner and restore unchanged components' name/description/relations verbatim, excluding them from refresh. Relations then follow the deterministic `rebuild_global_relations`. Genuinely-changed and new components still re-cluster normally, so large changes are unaffected.

## Evidence (CodeBoarding repo, depth 3)

**Small change** (5 files, 2 components genuinely changed):

| | before | after |
|---|---|---|
| depth-1 components modified | 6/6 | **2/6** |
| method-owner stability | 97.7% | **98.4%** |
| invariants | — | **29/29, 0 findings** |

**Large change** (102-file restructure): 6/6 correctly modified, **0 invariant findings**, incremental-vs-full-analysis fidelity NMI **0.888**, **0 stale** methods — copy-forward does not harm the big-change case.

**markitdown** (generalization): 1/6 → 1/6 (93% stability) — no regression. The changed file spans a *single* component in markitdown's cohesive converter architecture (vs CodeBoarding's cross-cutting agent files that span 4), so there is no method-dispersion spill to reduce and the fix correctly stays a **no-op**. The spill exists, and is reduced, where files disperse across components.

Builds on `feat/infomap-program-graph-incremental`; also includes a surgical child-scope reconcile (gate + reconcile instead of re-cluster). Unit tests added for member-granular dirtiness and unchanged-component preservation; existing suites green; black + mypy clean.

## Follow-up: edge drift (added in later commits)

After the above, only the genuinely-changed component's *methods* moved, but relations still churned. Two more fixes, assessed from the baseline first (file-dispersion was already minimal, so the problem was edges, not clustering):

- **Preserve relations between unchanged components** — `rebuild_global_relations` re-derived (and re-labelled) every edge each save; now edges between two unchanged components carry their baseline label verbatim.
- **Preserve the full subtree of unchanged components** — child sub-component boundaries of an untouched component are kept byte-identical, so the node→component map (and the deep relations) don't drift.

Result on the small change: depth-1 modified **1/6** (the single genuinely-changed component), method stability **100%**, spurious edge changes **161 → 23**, invariants 29/29 clean. Large change: 6/6 (all components genuinely own changed methods — verified), 0 findings.

## Summary by CodeRabbit

* **New Features**
  * Improved pipeline now uses a unified program graph across clustering, relation extraction, health checks, and diagram generation.
  * Deterministic hierarchical clustering helps keep cluster identities stable across incremental runs.
  * Member-level change detection enables more precise structural diffs.

* **Bug Fixes**
  * Incremental updates now better preserve unchanged component metadata and relations, updating only impacted areas.
  * Validation is stricter for component name/description edits; unresolved relation endpoints are dropped with warnings.
  * Added stronger scope containment checks to prevent invalid ownership across nested scopes.

---

## Review round (commit 208b60ad)

Addressed the CodeRabbit + Codex review across the full diff (verified each — several were incorrect); see the pinned review-response comment for the per-finding breakdown. Highlights:

- **Incremental correctness gap-fills**: body-only edits now fire `has_changes` (were short-circuited); `source_cluster_ids` restored on copy-forward; missing baseline content-hash marks the member dirty instead of a false no-change; partial cluster baselines raise `IncrementalCacheMissingError`; orphaned child scopes are rejected.
- **Static-analyzer fixes**: `program_graph.merge` copy-before-add, alias-map registration, metadata-free CALL-edge counting, zero-based cross-language id collision, async-diagnostics wait on warm start, v3 cache-tag gating.
- **CI**: integration tests migrated to the ProgramGraph API; the `call_graph_nodes`/`call_graph_edges` metrics now measure the call subgraph, matching the existing fixtures (no per-language regeneration). Telemetry no longer logs a false "unhealthy" error from a stale payload key.

**Re-validation (post-review, depth 3):**

| | depth-1 modified | method-owner stability | method-exclusivity violations |
|---|---|---|---|
| CodeBoarding small (5-file edit) | **1/6** | **100%** | 0 |
| CodeBoarding large (102-file restructure) | 6/6 (all genuinely changed) | **98.4%** | 0 |
| markitdown (before → after) | 0/5 → **0/5** (no regression) | 100% | 0 |

Unchanged from the pre-review results. The large run's first attempt hit the pre-existing, nondeterministic `#408` planner `cluster_refs` fragility (unrelated to this diff — it doesn't touch that validation) and succeeded on retry. Full unit suite green (1803 passed), mypy + black clean.

---

## Update — method-granular edge preservation (`e9105e67`)

**What it adds.** Same change as #423 (the edge-preservation code is identical between the two engines): the incremental save rebuilds a changed component's call edges at **method granularity** — keep the baseline edge for a method whose body did not change, take a fresh edge only from a body-changed or newly-added method, drop edges whose endpoint method was deleted — instead of re-deriving the whole component's edges wholesale.

**Effect on this engine — honest, and mixed.** Measured on the benchmark probes:

| | E1 — body edit, must be silent | E8 — symbol removal |
|---|---|---|
| before | *clean* | **crash** (`ScopeContainmentError`) |
| after | `LL_COMPONENT_ADDED 1`, `EDGE_ADDED 3`, `EDGE_REMOVED 1`, `EDGE_EVIDENCE_CHANGED 19` | completes; `EDGE_ADDED 3`, `MEMBERSHIP_MOVED 3` |

So the change makes **E8 complete** (it previously aborted) but **regresses E1** from clean to violated. The reason: this engine (ProgramGraph / Infomap) re-clusters aggressively — a one-line body edit can add a whole component (`LL_COMPONENT_ADDED 1`) — and that **clustering churn** overwhelms the edge preservation.

**The key finding behind this.** The underlying LSP call-edge **extraction is deterministic** (687 raw edges, 100% across six runs); the non-determinism that churns component edges is **clustering / membership assignment**, not edge derivation. Edge preservation (this commit) can only mask churn where the clustering is stable — which holds on #423's CallGraph but not on this engine's more aggressive re-partitioning. The real remaining work is membership stability, not edges.


Svilen-Stefanov:
## Review response (commit 208b60ad)

Went through all 28 automated comments (Codex + CodeRabbit), verifying each against the code — several were incorrect. Summary of every finding and its resolution.

### Fixed — incremental correctness
- **cluster_delta `changed_cluster_ids`** (Codex, CodeRabbit): a pure body-only edit left `changed_cluster_ids` empty, so `ClusterDelta.has_changes` was False and the workflow short-circuited without re-detailing the edited component. Now folds any surviving cluster owning a body-changed member into `changed_cluster_ids`, and emits dropped clusters for a language whose files were all deleted.
- **Relation preservation ignores membership churn** (Codex, CodeRabbit): a component that only gained/lost a member (no body-hash change) was treated as unchanged, so `_preserve_unchanged_global_relations` restored a stale baseline edge. It's now flagged as changed via a baseline-vs-live member-key comparison.
- **Canonical vs raw qname mismatch** (CodeRabbit): `changed_member_qnames` now emits both the canonical graph id and the raw persisted qname, so the `cluster_delta` (canonical) and `diagram_generator` (raw) consumers each intersect correctly.
- **`source_cluster_ids` not restored on copy-forward** (Codex): an unchanged component kept the re-partition's cluster ids while its membership was pinned to baseline. Now captured in the membership baseline and restored alongside `source_group_names`.
- **Missing baseline content hash silently skipped** (CodeRabbit): removed the empty-hash skip in `changed_member_qnames`; a missing baseline hash can't prove a non-empty method unchanged, so the inequality now marks it dirty instead of reporting a false no-change.
- **Partial cluster baseline** (CodeRabbit): `snapshot_from_static_analysis` records languages that have code but no cluster snapshot; the caller now raises `IncrementalCacheMissingError` instead of reclustering the missing language and losing stable ids.
- **Orphaned child scope** (CodeRabbit): `assert_scope_containment` now rejects a `sub_analyses` entry with no parent component instead of silently skipping it.

### Fixed — transplant / static analyzer
- `program_graph.merge` deep-copies nodes/edges before adding (`add_edge` mutates endpoints during alias resolution).
- `result_converter` registers non-canonical names in the graph alias map so `resolve_symbol_id`/`has_symbol` follow them.
- `infomap_clustering` counts metadata-free CALL edges via `log1p(max(count, 1))` instead of dropping them.
- `cluster_helpers` cross-language reindex shifts past the highest used id (fixes zero-based-id collisions).
- warm-start path waits for async diagnostics before merging (rust/C# publish late).
- `analysis_cache.get(expected_sha=None)` now validates a present v3 tag.
- `get_method_invocations` skips languages without a ProgramGraph instead of aborting.
- `cluster_methods_mixin` captures prior method owners before `cluster()` overwrites the seed.
- test-quality: `test_cfg_tools` models an empty `ProgramGraph` (not `None`); `test_analysis_result` parameterizes the graph language.

### False positive
- **`ProgramGraph.symbols` "is a method, call it with `()`"** (CodeRabbit, flagged Critical): it is a `@property` returning a dict (`program_graph.py:238`). Adding `()` would call a dict and crash. No change.

### Deferred (real, but heavier than a review-round patch — each needs its own change + tests)
- Reindexed cluster ids not written back to the graph snapshot; `merge` not restoring `module_members` for re-added symbols (snapshot-consistency, incremental warm-start layer).
- Python `from . import x` / `from pkg import mod` capture only the module prefix; JS `require()` / PHP `include` import coverage — both need resolver-aware handling, not blanket AST/regex widening.
- Reverse-import index for added/deleted targets; per-scope hierarchy-level + id-watermark persistence; tsconfig-baseUrl alias resolution.
- Exception-contract rename (`IncrementalProgramGraphUnavailableError` → `BaselineUnavailableError`): the suggested rename creates an import cycle (both live in modules that import `static_analyzer`); fixed the misleading docstring/log wording, deferred the relocation.

### CI
- `integration-tests` were failing on the removed CallGraph API — migrated to `get_program_graph`/`.source`/`.target`/`iter_reference_nodes`/`.occurrences`. The `call_graph_nodes`/`call_graph_edges` metrics now measure the call subgraph (`call_node_ids`/`call_edges`), which matches the existing fixtures exactly, so **no per-language fixture regeneration is needed**. Python suite verified green locally.
- Also fixed a telemetry bug: `track_lsp_result` read the removed `call_graph` payload key, logging a false "unhealthy" error on every analysis.

Regression tests added for each incremental fix; full unit suite green, mypy + black clean.

PR #426, opened 2026-07-20
Move the integration suite to a private repository

Svilen-Stefanov:
## What

Moves `tests/integration` (141 files) and `tests/INTEGRATION_TESTS.md` out of this repo into the private [CodeBoarding/CodeBoarding-tests](https://github.com/CodeBoarding/CodeBoarding-tests). `integration-tests.yml` becomes a dispatcher that runs the private suite against the PR's commit and mirrors the result.

## Why

The integration fixtures encode the exact expected static-analysis output — reference counts, call-graph shape, package dependencies — for all eight languages. Public, that's a target anyone can read and tune against.

The dispatcher discards the private run's output, so no test name or fixture value reaches this repo's public logs. All you see here is pass/fail plus a link only maintainers can open.

## Fork PRs deliberately don't run it

GitHub withholds secrets from fork-PR workflows. Every workaround — `pull_request_target`, a `workflow_run` hop — ends with contributor-authored code executing in-process with the suite. Since the tests must import and execute the engine to test it at all, that code can simply read them; an approval gate raises the cost of stealing the suite but doesn't prevent it.

So: same-repo PRs, push to `main`, cron and manual dispatch run the suite. Fork PRs skip it with an explanatory note and are covered by the push-to-`main` run right after merge. To check one before merging, read the diff and dispatch `run-integration-tests.yml` with that PR's head SHA.

## What stays public

The Node-less bootstrap checks are install/`tool_registry` coverage, not test coverage, so they split into `bootstrap-verification.yml` and keep running on every PR including forks — no gate, immediate feedback.

## Auth

No new credential. The dispatcher mints a short-lived installation token from the existing CodeBoarding App, narrowed to `repositories: CodeBoarding-tests` and `permission-actions: write`. It has no `contents` permission, so leaking it out of this public workflow would not let anyone download the suite, and it expires in an hour.

## Also

- Drops the 11 pytest markers and the mypy exclude that only the moved suite used.
- Makes `pytest` an explicit dev dependency (it was only transitive via `pytest-cov`), since the private suite runs on this repo's venv. `uv.lock` regenerated.
- `CONTRIBUTING.md` §5e explains the arrangement for contributors; fixes a stale `python setup.py` reference while there.

## Two things worth your call, Ivan

1. **The tests remain in this repo's git history.** Deleting from HEAD doesn't remove them — they're recoverable via `git log`. Fixing that needs a history rewrite and force-push, which breaks every fork and clone. Not done here.
2. **Integration legs now run under the runner's system Node** rather than the embedded `nodeenv`, since the uninstall steps moved to the bootstrap workflow. Small fidelity loss; closing it means having `bootstrap-verification.yml` actually spawn an LSP under the embedded runtime.

## Testing

Public suite 1833 passed / 81.88% coverage (gate 70). Private suite collects the same 167 tests; `-m "integration and python_lang"` passes 9 / skips 2 against real LSP runs. Matrix builder exercised across all 9 event permutations — reproduces the original 9-leg smoke subset and 22-leg full matrix, no empty matrices.

## Summary by CodeRabbit

* **New Features**
  * Reworked CI to run integration tests via a centralized private suite dispatch, including workflow output with the run status and results link; added manual full-matrix dispatch support with improved concurrency/cancellation behavior.
* **Bug Fixes**
  * Updated the pre-commit test step to run `pytest` directly under `uv`, removing the previous integration exclusion while keeping the same coverage failure threshold.
* **Documentation**
  * Updated contributor guidance for the private integration suite and fork behavior; removed outdated in-repo integration/benchmark documentation.
* **Chores**
  * Retired large in-repo integration fixtures and related integration test/health assets; adjusted MyPy/pytest configuration accordingly.


diagram_analysis/diagram_generator.py
```diff
@@ -49,7 +49,7 @@
     ClusterSnapshot,
     snapshot_from_static_analysis,
 )
-from diagram_analysis.exceptions import IncrementalCacheMissingError
+from diagram_analysis.exceptions import IncrementalCacheMissingError, ScopeContainmentError
```
ivanmilevtues: revert

diagram_analysis/diagram_generator.py
```diff
@@ -665,6 +666,40 @@ def _build_file_coverage_summary(self) -> FileCoverageSummary | None:
             not_analyzed_by_reason=summary["not_analyzed_by_reason"],
         )
 
+    def _rescope_child_analyses(
```
ivanmilevtues: revert

diagram_analysis/exceptions.py
```diff
@@ -51,3 +51,18 @@ def __init__(self, scope_id: str, issues: list[str]):
         super().__init__(f"Incremental plan for scope {scope_id!r} is invalid: {issue_summary}")
         self.scope_id = scope_id
         self.issues = issues
+
+
+class ScopeContainmentError(RuntimeError):
```
ivanmilevtues: revert

tests/agents/test_method_uniqueness.py
```diff
@@ -412,5 +421,136 @@ def test_shared_file_no_aliases_no_duplicates(self):
         self.assertEqual(total, 6)
 
 
+class TestScopeContainment(unittest.TestCase):
```
ivanmilevtues: remove

tests/diagram_analysis/test_diagram_analysis.py
```diff
@@ -1506,6 +1507,7 @@ def test_empty_incremental_delta_does_not_prune_stable_leaf_components(
         gen.details_agent = Mock()
         gen.incremental_planning_agent = Mock()
         gen.incremental_agent = Mock()
+        gen.incremental_agent._create_strict_component_subgraph.return_value = ("", {}, {})
```
ivanmilevtues: remove

ivanmilevtues (approved):
Please remove some left overs and then it is ready to be merged,

Svilen-Stefanov:
Went through the bot review comments. Note the branch was accidentally cut from `fix/incremental-child-scope-drift`, so it carried two unrelated incremental-fix commits (`diagram_generator.py`, `analysis.py`, `exceptions.py`). I've rebased onto `main` and force-pushed — the PR is now just the test-move + workflow + docs. That makes several comments moot.

**Fixed**

- **Token expiry during long polls** (`integration-tests.yml`, codex P2) — real bug. The old loop ran up to 70 min and masked `gh` failures as "still running", so an expired App token (1 h lifetime) could turn a green run into a false timeout. Rewrote it to cap polling at 55 min (under the token window), distinguish an API failure from an incomplete run, and fail with an explicit "lost contact / token expired" message instead of a silent timeout. Full-matrix runs that need longer should be launched from the private repo directly.

**Moot after the rebase** (these files are no longer in the PR)

- All four codex comments on `diagram_generator.py:692/918` and `analysis.py:145` — they belonged to the incremental-fix commits, not this change.

**Already handled**

- **`persist-credentials: false`** (CodeRabbit, on `bootstrap-verification.yml`) — that workflow was folded back into the private repo's `run-integration-tests.yml`, whose two checkouts already set `persist-credentials: false`. The public dispatcher never checks out code at all.
- **Scheduled trigger** (codex P2) — the weekly cron lives in the private repo's `run-integration-tests.yml` (it self-schedules against engine `main`), so the drift coverage isn't lost; it just moved with the suite rather than staying in the dispatcher.

**Deliberate — not changing, reasoning below**

- **Diagram-pipeline paths should trigger the suite** (codex P2) — the suite tests static-analysis (LSP) output only; nothing in it imports `diagram_analysis`/`codeboarding_workflows` (the hits are strings in a health fixture). The pre-split workflow didn't trigger on those paths either, so this isn't a regression. Adding them would run the whole matrix for changes it can't detect.
- **Test the merge commit, not `head.sha`** (codex P2) — real tradeoff. `head.sha` tests the branch as-is and keeps the run-name correlation simple; the merge ref can be stale/null. Left as `head.sha` for now; happy to switch if you'd rather catch as-merged conflicts pre-merge.

**Your call**

- **Coverage gate 70 vs documented 80** (CodeRabbit) — the `--cov-fail-under=70` predates this PR; I only removed the now-dead `-m "not integration"` filter from that line. Actual coverage is 81.88%, so bumping the gate to 80 to match the docs would pass, but it's a policy change with ~2% headroom and I didn't want to slip it into a test-infra PR. Tell me which way you want it (bump CI to 80, or fix the docs to 70) and I'll do it.

Svilen-Stefanov:
Looked at both.

**Dependabot PRs — fixed.** Correct catch. A Dependabot PR bumping `pyproject.toml`/`uv.lock` matches the path filter and is same-repo, so the old fork check let it through to the mint step — where Dependabot's read-only token has no secret, so `create-github-app-token` fails and the check goes red with nothing the PR can do about it. Added `github.actor != 'dependabot[bot]'` to the `private-suite` guard so those PRs skip the suite (like forks), and broadened the notice job to explain the skip for both. (Dependabot isn't actually enabled on the repo yet, so this is pre-emptive, but it's cheap and correct.)

**Workflow-modification token exposure — real mechanism, but no escalation under the current access model, so documented rather than re-architected.** The reasoning:

- A fork PR can't reach this — forks get no secrets, so they never mint the token. The vector requires a *same-repo* PR, which needs push access to `CodeBoarding/CodeBoarding`.
- Push access here is `ivanmilevtues`, `Svilen-Stefanov`, `brovatten` — and that's the exact same three people with access to `CodeBoarding-tests`. So everyone who could rewrite this workflow to read the private-suite logs can already `git clone` the private repo directly. There's no one the isolation is protecting the tests *from* who could exploit this.
- The token is also `actions`-only with no `contents`, so even in that scenario it exposes run logs/artifacts, not the test source.

Given that, re-architecting to a `workflow_run` base-context job (which would reintroduce the artifact/PR-number correlation we deliberately removed) isn't justified today. I've instead documented the assumption inline on the job: it holds only while push access to this repo stays a subset of `CodeBoarding-tests` access, and if those sets ever diverge, minting + dispatch should move into a `workflow_run` job that runs the base-branch copy of the workflow, out of PR control.

If you'd rather have the stronger control now regardless, the two options are: (a) a GitHub Environment with required reviewers on `private-suite` (gates every run behind a human click — friction on each PR), or (b) the `workflow_run` split. Say the word and I'll do either.

Svilen-Stefanov:
All three valid — addressed.

**1. Token can read logs/artifacts, not just start a run — fixed the comment.** Correct, and it was my own comment overclaiming. `actions: write` includes read, and per the REST docs that downloads run logs and artifacts, so within its ~1h life the token can pull the private suite's output and `actual_*.json` artifacts (test values), just not the source (no `contents`). Reworded the inline comment and the CI-wiring doc to say that plainly. The mitigation is unchanged (the trust-boundary argument from the last round), but the wording no longer pretends the token is read-only-to-start.

**2. client-id/app-id fallback — added.** You're right that this diverged from `codeboarding-sync`, which supports an app-id-only install. The org currently has only `CODEBOARDING_APP_CLIENT_ID` set, so my single-input step worked today, but an app-id config would have failed the gate. Added a second conditional mint step (client-id preferred, app-id fallback) and OR'd the two token outputs — mirroring sync without its `github-actions[bot]` fallback, since this gate must stay fail-closed. Kept it to two small steps rather than sync's full detection block.

**3. Merge commit vs head.sha — switched to `github.sha`.** Agreed, and this also simplified the expression: `github.sha` is already the PR *merge* commit for `pull_request`, the pushed commit for `push`, and the branch tip for schedule/dispatch — so `ENGINE_SHA` is now just `${{ github.sha }}` (was `head.sha || github.sha`). PRs now test what actually merges, catching main+PR combination failures. The private checkout resolves the merge-commit SHA fine since it's reachable from `refs/pull/N/merge` in the public repo.

Branch stays at the same 7-file diff. Still can't run any of this live — the private-Actions billing block is unchanged — so these are verified by inspection (YAML parse, `bash -n`, and tracing the `if`/token/SHA wiring across every event), not by a green run yet.

Svilen-Stefanov:
Both valid — fixed.

**1. `Iv` normalization missing from the app-id fallback.** Right — my app-id step passed `CODEBOARDING_APP_ID` straight to the action's `app-id` input, so an install that stored the *client* id (the `Iv…` value) in `CODEBOARDING_APP_ID` would have handed a client id to `app-id` and failed, exactly the case sync's lines 54-59 recover. Added the same routing: the client-id step now also fires when `CODEBOARDING_APP_ID` starts with `Iv` and passes that value to `client-id`; the app-id step only fires for a non-`Iv` app id. Traced the four configs — client-id-only (today), numeric-app-id-only, client-id-in-app-id (the flagged bug, now routed to client-id), and nothing-set (empty → caught by the fail-closed pre-flight).

**2. Dependabot guard also skipped Dependabot-authored pushes to main.** Good catch — the actor check sat outside the event branch, so an auto-merged update's push to `main` (actor `dependabot[bot]`) was skipped too, and since the notice job is pull_request-only it got no coverage at all — defeating the point of relying on the push run. Scoped the actor check to the pull_request branch, so push/schedule/dispatch always run regardless of actor; only fork and Dependabot *PRs* skip. Verified across all seven event/actor combinations: `push by dependabot` now runs, and exactly one of private-suite / skip-notice still handles each PR.

Same 7-file diff. Verified by inspection (YAML parse + tracing the guard and credential-selection expressions across every case); still can't run live until the private-Actions billing limit is lifted.

PR #430, opened 2026-07-22
Document OpenAI-compatible gateway configuration

binyangzhu000-sudo:
## Summary
- Document `openai_base_url` / `OPENAI_BASE_URL` as the generic way to connect any OpenAI-compatible gateway.
- Explain the matching API-key and model-override settings.
- Remove the vendor-specific provider implementation in favor of the existing generic path, following maintainer feedback.

## Validation
- Final PR diff contains only `README.md`.
- `git diff --check`
- Verified all source and test files match current `main`.


user_config.py
```diff
@@ -162,22 +166,42 @@ def load_user_config(path: Path = CONFIG_PATH) -> UserConfig:
...
+            text[body_start:],
+            flags=re.MULTILINE,
+        )
+        body_end = body_start + next_section.start() if next_section else len(text)
+        section_body = text[body_start:body_end]
+        if re.search(rf"^\s*#?\s*{re.escape(key)}\s*=", section_body, flags=re.MULTILINE):
+            return
+        injected = text[:body_start] + commented_line + "\n" + text[body_start:]
+    else:
+        injected = text.rstrip() + f"\n\n[{section}]\n" + commented_line + "\n"
```
Svilen-Stefanov: Flagging this one regardless of what we decide about the provider itself, since it would bite existing users.

blocker-ish: won't this write a duplicate `[provider]` table and permanently break someone's config? The `else` branch fires whenever the regex doesn't match, and that isn't the same thing as "the section doesn't exist". I ran the new function against a few valid TOML files:

- `[ provider ]` (spaces inside the brackets)
- `["provider"]` (quoted table name, same table per the spec)
- a `[provider]` header as the last line with no trailing newline

All three come out as `TOMLDecodeError: Cannot declare ('provider',) twice`. And `bootstrap_environment` does `ensure_config_template()` then `load_user_config()` back to back, so the run that corrupts the file is also the run that dies, and there's no self heal from there, you'd have to hand edit `~/.codeboarding/config.toml` to get the CLI back.

The old code had the same hole but only for `[llm]`, and only when `context_window` appeared nowhere in the file. This widens it to `[provider]`, which every existing config already has, so I think it's newly reachable.

Can we ask tomllib whether the table exists rather than asking a regex? Roughly:

```python
try:
    table_exists = section in tomllib.loads(text)
except tomllib.TOMLDecodeError:
    table_exists = False   # Why: already broken, don't make it worse
...
if header is None:
    if table_exists:
        return  # Why: header is there in a shape we can't locate, appending would declare it twice
```

and a test whose post condition is that `tomllib.load(path)` still succeeds, rather than a substring check. That's the property we actually care about.

agents/constants.py
```diff
@@ -36,3 +38,7 @@ class ModelCapabilities:
         "kimi": "moonshotai",
         "glm": "z-ai",
     }
+
+    PROVIDER_CONTEXT_WINDOWS: ClassVar[dict[tuple[str, str], tuple[int, int]]] = {
+        ("atlascloud", "deepseek-ai/deepseek-v4-pro"): (1_000_000, FALLBACK_OUTPUT),
```
Svilen-Stefanov: Where do these numbers come from? Atlas Cloud's own catalog disagrees with both. `GET https://api.atlascloud.ai/v1/models` is unauthenticated and returns:

```json
{"id": "deepseek-ai/deepseek-v4-pro", "context_length": 1048576, "max_output_length": 393216}
```

so 1,048,576 / 393,216 rather than 1,000,000 / 64,000. The output one is off by about 6x. The PR description mentions a live `/v1/models` check, and both of those fields sit in the same JSON object as the `id`, so I think the right values were already there.

Worth saying even if we don't go the provider route, because it points at why I'm wary of the table shape in general: hand copied per model windows go stale silently, and this one covers 1 of the 122 models Atlas exposes (not even `qwen/qwen3.5-flash`, which this PR also sets as the parsing model).

Svilen-Stefanov (commented):
Thanks for the PR, and for the tests and the config.toml migration handling, that's more care than these usually get.

Before going line by line I want to raise the bigger question, because I think it changes what this PR should be.

Atlas Cloud already works with CodeBoarding today, without any code changes. It's OpenAI compatible, so openai_base_url = "https://api.atlascloud.ai/v1" in ~/.codeboarding/config.toml (or OPENAI_BASE_URL in the shell) plus your Atlas key and an agent_model/parsing_model is all it takes. The litellm entry is another route to the same thing, it's selected by base URL alone.

So I think the real gap you ran into isn't a missing provider, it's that nothing tells you any of that. Our README's config block doesn't mention openai_base_url at all, and "set OPENAI_API_KEY to your Atlas Cloud key" is a genuinely odd thing to expect someone to work out. That part I'd take a PR for happily: add openai_base_url to the README config block with a line saying it's how you point at any OpenAI compatible gateway, and it stops being a per vendor problem. We should probably also ask ourselves whether openai and litellm are the right names for that path, since neither of them says "generic OpenAI compatible endpoint" out loud, but that's on us to sort out, not something this PR should have to solve.

What I'm hesitant about is a per vendor entry in LLM_PROVIDERS. It's a wider surface than it looks: the provider list, the secrets map, the config template, the template migration, a context window table, tests, and then the VS Code extension each keep their own provider lists too. 
@binyangzhu000-sudo have you tried using the mentioned above approach and did it not work out or you were unaware of it?

@ivanmilevtues do you have a view here? My instinct is to land the docs fix and keep LLM_PROVIDERS for providers that genuinely need their own client.

I've deliberately not done a full line by line pass, since the above is the question to settle first. Two things I did want to flag either way, inline. If we do end up going the provider route I'll come back with the rest (mostly consistency: the validation and the context window table are applied to atlascloud only while the other providers keep the old shape).

ivanmilevtues:
Looks good to me it is a simple docs change now

ivanmilevtues (commented):
Looks good to me a simple docs change which was missing.

PR #432, opened 2026-07-24
Deterministic incremental analysis: stop redrawing the diagram on a one-line change

Svilen-Stefanov:
Builds on #423. Two halves: what we improved in **full** analysis, and what we found and fixed in **incremental**.

---

## Full analysis

#423 made full analysis deterministic — Leiden picks the components, the LLM only names them. That worked, but it left components too big to read that the engine refused to expand, so diagrams bottomed out too early. Three causes.

### 1. We threw away the evidence before making the decision

The "meta-graph" is how we decide whether a component has separable structure. It counts calls **between different clusters** — an edge inside one cluster doesn't count:

```python
if src_cid != dst_cid:
    edge_weights[(src_cid, dst_cid)] += 1
```

But before building it, we ran `merge_clusters` to squeeze the leaf clusters down to a cap of 50. Say Leiden found 120 leaf clusters, and clusters `A` and `B` — with 40 calls between them — both got merged into super-cluster `S`. Those 40 calls are now *internal* to `S`, so they never enter the meta-graph.

Then we asked "does this component split?" using that gutted meta-graph. The coupling evidence had already been eaten by the merge. No edges means modularity 0.0, which always reads as "cohesive, keep it as one box."

It was circular — merging *is* a grouping decision, taken before and independently of the grouping decision we were trying to make. Fix: drop the pre-merge, build the meta-graph over the raw leaf clusters.

### 2. Absorption was rich-get-richer

Leftover singletons need a home. The old loop took the **smallest** group, found its nearest neighbour, and merged it in:

```python
result[merge_idx].update(result[smallest_idx])
```

The problem: `merge_idx` **grows each round**, and "nearest" was scored against that growing set. A group spanning 30 files overlaps with almost any orphan; a group spanning 2 files rarely does. Graph distance behaves the same — a bigger group touches more of the graph, so it's closer to everything.

So round 1's winner is favoured in round 2, and more in round 3. You end up with one enormous group and a few stragglers — on graph-viewer, **45 of 50 clusters in a single group**.

Fix: absorb by BFS distance over the meta-graph (a real path, not a growing proxy), and break ties toward the **smaller** group, which pushes back against the feedback loop instead of amplifying it.

### 3. The expansion gate ignored size

The old gate was one constant for every component:

```python
separable = modularity >= 0.15   # always 0.15
```

Modularity measures how *cleanly* something splits, not how *badly* it needs splitting. A 400-method component whose parts are genuinely interwoven scores low and stays one unreadable box. A tidy 30-method component scores high and gets split. Exactly backwards from what a reader needs.

The gate now scales with size, and oversized components skip the test entirely:

```python
load = max(methods / 120, files / 12)
if load >= 1.0: return True          # too big to read: split regardless
required = 0.15 * max(0.0, 1.0 - load)
```

**Real example** — `2.2 'ANSI Styling & Output Dispatcher'` from the click test: 38 methods, 7 files, measured subgraph modularity **0.1364**.

| | calculation | verdict |
|---|---|---|
| old gate | `0.1364 >= 0.15` | **no** — stays one 38-method leaf |
| new gate | `load = max(38/120, 7/12) = 0.583` → `required = 0.15 × 0.417 = 0.0625`; `0.1364 >= 0.0625` | **yes** — splits into two |

Same component, same call structure. The old bar was just set without reference to how big the box was.

### Results (4 repos, depth still 3)

| repo | components | depth-3 components | largest top-level | too-big-but-unexpandable |
|---|---|---|---|---|
| click (colorama) | 20 → **46** | 7 → **26** | 243 → **131** | 0 → 0 |
| click (get_strerror) | 16 → **33** | 0 → **13** | 179 → **139** | 0 → 0 |
| codeboarding (local-server) | 30 → **51** | 9 → **28** | 269 → **223** | 3 → **0** |
| codeboarding (model-env) | 16 → **33** | 2 → **15** | 193 → 193 | 3 → **0** |

Dead-ends go to **zero everywhere**. Depth stays 3; users can still expand further on demand.

---

## Incremental analysis

#423 aimed at full analysis, and incremental regressed. Deleting **one function** rewrote ~100% of relations. Four causes.

### 1. An LLM was still deciding the structure

An agent emitted create/delete operations and picked each component's clusters, so a change that touched nothing could still redraw the tree. Replaced with a deterministic planner (`diagram_analysis/scope_plan.py`) that carries every surviving component forward. The LLM keeps naming; it loses structure.

### 2. We anchored on cluster ids, which aren't stable

Each scope's leaf clusters are re-derived from its subgraph on every run, so their numeric ids get reshuffled precisely when a change lands — exactly when we need them to hold still. A component would look brand new despite owning the same code.

Now anchored on **method names**, which survive renumbering. Top-level components carry forward with byte-identical cluster sets.

### 3. A no-op "update" wasn't free

The planner emitted an `UPDATE_COMPONENT` for every survivor, even ones that hadn't moved. That is expensive: it puts the component in `refresh_ids`, which reruns the LLM relation step for the **entire scope**, and strips and restores its clusters, marking every owner changed. One edit therefore relabelled the whole diagram.

An unchanged component now produces **no operation at all**. And when a scope does regenerate, a relation whose two endpoints both stayed put keeps its previous wording but always takes the **fresh** call edges — wording is sticky, structure never is.

### 4. Module-level variables dirtied their whole file

The call graph holds module-level variables (`_ansi_re = re.compile(...)`) as nodes, because other code references them. The saved index doesn't — clustering never assigns a bare variable to a component. So for `click/_compat.py`, the graph has 56 symbols and the index has 53.

We detect module-level edits by hashing everything *outside* the known method spans. One side excised 56 spans, the other 53 — so those three variable lines were cut on one side and not the other, and the hashes **could never match**. Every file containing a module-level variable was reported changed on every run, marking 26 of 33 components dirty and regenerating their relations.

Comparing a baseline against its own commit, where nothing changed at all:

| | phantom changed methods | files falsely flagged |
|---|---|---|
| before | 3 | `_compat.py` |
| after | **0** | **none** |

### Results (small suite, no judge)

| | #423 as merged | this branch |
|---|---|---|
| criteria violated (4 tests) | 11 | **8** |
| engine time (4 tests) | 1494s | **527s** |
| I2 relations changed | 89% | **38%** |
| I2 relation re-wordings | 42 | **23** |
| I3 relations changed | 101% | **49%** |
| I6 relation re-wordings | 36 | **10** |

I6 also flips `introduced_invariants` and `no_phantom_files` from violated to **met** — a deleted file is no longer referenced.

Previous main sits at ~20% relation churn on I2, but across a 110-component tree; in absolute re-wordings this branch is at 23 (I2) and 10 (I6) versus main's 25 and 43. Main isn't clean either: it duplicated 128 of 532 methods into a second owner, and still ships the deleted function in its index — which every branch tag correctly drops.

---

## Known remaining issue

One component per test still loses its children (`2.2` on I2, `5.3` on I3) despite byte-identical membership. Cause identified, **not** fixed here: `analysis_json.py` skips a sub-analysis whenever a save-time `can_expand` recompute flips false, silently discarding an already-analysed subtree. Regrafting it in measurement takes I2 moved-methods 38 → 0 and dangling relations 26 → 0. Next PR.

Also cleaned up: ~4,700 lines removed net — the LLM planning agent and its prompts across six providers, the scope-operation validator and repair chain, and assorted dead symbols.

## Summary by CodeRabbit

* **New Features**
  * Incremental diagram analysis now deterministically plans component updates, preserving unchanged identities and relations.
  * Expansion behavior is more deterministic, using CFG structure with leaf/load-based limits.
  * Diagram analysis now serializes `key_entities` reference files as repo-relative paths.
* **Bug Fixes**
  * Prevented cluster ID collisions across languages during incremental analysis.
  * Improved copy-forward preservation for untouched components and relation reconciliation for moved items.
* **Documentation**
  * Updated `--depth-level` and related API/docs to clarify the safety-valve cap and “leaf ceiling” on-demand expansion behavior.


docs/summary-interactive.md
```diff
@@ -73,28 +73,32 @@ Existing cluster IDs are **preserved** (stable). New clusters get fresh IDs. Thi
...
-The prompt includes a **tool usage policy**: keep reads small and targeted (single representative qname per cluster), stop as soon as confidence is high.
-
-**Tool access**: `read_source_reference` only — the full ReAct toolkit is NOT attached. This constrains the agent from speculatively reading large code blocks.
-
-**Fast-path parse**: `_direct_pydantic_parse` in `agent.py` tries to parse the LLM response directly as the target Pydantic model before invoking the trustcall extractor — saves a second LLM round-trip when the agent emits clean JSON.
-
-**Validation**: Up to 3 retries via `validate_cluster_coverage` if the LLM doesn't cover all cluster ids.
-
-**Output**: `ClusterAnalysis` with `cluster_components` — each has `name`, `cluster_ids`, `description`, `parent_id`. Every affected cluster id must appear exactly once.
+### Step 4: `plan_scope_update()` — no LLM
```
ivanmilevtues, line 76: I am unsure if we need this doc and in general the docs/ package you may remove them as we are not keeping them upto date

agents/abstraction_agent.py
```diff
@@ -35,6 +35,7 @@
 from static_analyzer import StaticAnalysisFatalError
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.cluster_helpers import build_all_cluster_results
+from static_analyzer.constants import Language
```
ivanmilevtues, line 38: I think unused import?

agents/agent_responses.py
```diff
@@ -654,9 +661,17 @@ def assign_relation_ids(analysis: AnalysisInsights) -> None:
...
         else:
             name_to_id[c.name] = c.component_id
+    resolved: list[Relation] = []
     for relation in analysis.components_relations:
-        relation.src_id = name_to_id.get(relation.src_name, "")
-        relation.dst_id = name_to_id.get(relation.dst_name, "")
+        src_id = name_to_id.get(relation.src_name, "")
+        dst_id = name_to_id.get(relation.dst_name, "")
+        if not src_id or not dst_id:
+            logger.info(f"Dropping relation with unresolved endpoint: '{relation.src_name}' -> '{relation.dst_name}'")
```
ivanmilevtues, line 669: Note: this is okay for now, however I think in the future it will be good to show external deps (fine for now)

agents/abstraction_agent.py
```diff
@@ -88,7 +89,10 @@ def step_clusters_grouping(self, cluster_results: dict[str, ClusterResult]) -> C
         final-analysis step.
```
ivanmilevtues, line 89: We should remove the @trace on this method, it doesn't invoke LLMs, so nothing to trace here.

agents/cluster_methods_mixin.py
```diff
@@ -170,6 +113,7 @@ class ClusterMethodsMixin:
     def deterministic_cluster_grouping(
         self,
         cluster_results: dict[str, ClusterResult],
+        cfg_graphs: dict[str, nx.DiGraph],
```
ivanmilevtues, line 116: cfg_per_lang is probably better name

agents/cluster_methods_mixin.py
```diff
@@ -179,17 +123,19 @@ def deterministic_cluster_grouping(
...
+        component's own subgraph when splitting a component, the whole repo at the
+        top level. Handing it the repo graph for a component scope makes the split
+        disagree with the separability gate, which reads the subgraph.
         """
-        cfg_graphs = {
-            lang: self.static_analysis.get_cfg(Language(lang)).clustering_networkx() for lang in cluster_results
-        }
-        groups = supercluster_leaf_ids(cluster_results, cfg_graphs, low, high)
-        node_lookup, file_lookup = _leaf_cluster_lookups(cluster_results)
+        groups, _modularity = supercluster_leaf_ids(cluster_results, cfg_graphs, low, high)
```
ivanmilevtues, line 132: _modularity is unsued, I will check if we can fail directly with mypy or black for these. I will do it now at the background.

ivanmilevtues: Also you can do the cfg per_lang grouping as it used to be done here, and completely drop the step from the agent, just call it here. I think it is better if the agent keeps only (mostly) LLM related code/logic in its own methids.

diagram_analysis/cluster_delta.py
```diff
@@ -563,6 +584,7 @@ def _delta_for_language(
     old_clusters: dict[int, ClusterSnapshotEntry],
     diff_files: set[str] | None = None,
     repo_dir: Path | None = None,
```
ivanmilevtues, line 586: nullable types

agents/cluster_methods_mixin.py
```diff
@@ -550,108 +303,58 @@ def _expand_to_method_level_clusters(self, cfg: CallGraph, cluster_result: Clust
...
+                continue
+            subgraph_cfgs[lang] = sub_cfg
+
+            seeded_snapshot = scoped_snapshot_from_lineage(sub_cfg, source_cluster_id_prefix)
+            if seeded_snapshot:
+                sub_cluster_result = _delta_for_language(
+                    str(lang), sub_cfg.clustering_networkx(), seeded_snapshot
+                ).cluster_results
+            else:
+                sub_cluster_result = sub_cfg.cluster()
```
ivanmilevtues, line 345: I am a bit puzzled here, we are still using Leiden as expected but we still have the clustering on different levels of abstraction, this can be dangerous (and proly is) as if we get into one of the different branches we lose "accuracy" -> for a bigger change we might go one level up in abstraction and the result of the incremental will be flaky.

agents/cluster_methods_mixin.py
```diff
@@ -550,108 +303,58 @@ def _expand_to_method_level_clusters(self, cfg: CallGraph, cluster_result: Clust
...
+
+            cluster_results[lang] = self._expand_to_method_level_clusters(sub_cfg, sub_cluster_result)
+
+        reindex_across_languages(cluster_results)
```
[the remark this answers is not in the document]
ivanmilevtues: What it says here seems reasonable.

agents/details_agent.py
```diff
@@ -94,7 +97,12 @@ def step_clusters_grouping(
         deterministically from the subgraph structure, not by the LLM.
         """
         logger.info(f"[DetailsAgent] Super-clustering subgraph for component: {component.name}")
-        return self.deterministic_cluster_grouping(subgraph_cluster_results, SUBCOMPONENTS_MIN, SUBCOMPONENTS_MAX)
+        return self.deterministic_cluster_grouping(
```
ivanmilevtues, line 100: we can drop the details_agent method here (also the @trace) and call just the deterministic_cluster_grouping either in the next step where it is needed or in the run method.

agents/incremental_agent.py
```diff
@@ -316,14 +323,29 @@ def generate_scope_relations(
...
         api_surfaces = self.step_api_surfaces(scope, scope_name)
-        return self.step_relation_analysis(
+        # Snapshot before regeneration: step_relation_analysis replaces the scope's relations
+        # wholesale, so every edge between two untouched components would come back reworded.
+        baseline_by_pair = {
+            (relation.src_id, relation.dst_id): relation.model_copy(deep=True)
+            for relation in scope.components_relations
+            if relation.src_id and relation.dst_id
+        }
+        rels = self.step_relation_analysis(
```
ivanmilevtues, line 489: > Just a comment nothing to do here
I think this is probably some patchwork. Just as comment for later. For me we should think later on, how to do this better, I still believe there is a way to do it with trustcall style, where we have a patching tool which can only modify components/relations which are not frozen and we'd freeze those before we run the incrmental analysis.

agents/planner_agent.py
```diff
@@ -26,47 +18,66 @@
...
-    modularity = subgraph_peak_modularity(cluster_results, cfg_graphs)
-    separable = modularity >= min_modularity
+    if all(cr.strategy == METHOD_LEVEL_STRATEGY for cr in cluster_results.values()):
+        # One synthetic cluster per method: the meta-graph is the raw call graph, whose
+        # modularity is far higher than any real clustering's and not comparable to the
+        # threshold. Too few natural clusters to separate means there is nothing to split.
+        logger.debug("[Planner] subgraph has no natural cluster structure; keeping as leaf")
+        return False
+    _groups, modularity = supercluster_leaf_ids(cluster_results, cfg_graphs, SUBCOMPONENTS_MIN, SUBCOMPONENTS_MAX)
+    required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
```
ivanmilevtues, line 76: I saw this formula in the comments of the whole PR, if it works it works.

diagram_analysis/diagram_generator.py
```diff
@@ -473,35 +468,6 @@ def _incremental_changed_component_ids(
     return changed
 
 
-def _preserve_unchanged_global_relations(
```
ivanmilevtues, line 476: This is now coverd by the base_pairs makes sesnse.

diagram_analysis/diagram_generator.py
```diff
@@ -581,28 +551,93 @@ def process_component(
...
+        schema), fall back to the structural default of expanding rather than
+        aborting the run.
+
+        Memoized on the component's member set: traversal asks once per component
+        and every save asks again for the whole tree, and the answer depends on
+        nothing else. A component whose membership changed (including one pruned
+        by ``_strip_ignored``) gets a different key and is re-evaluated.
         """
         assert self.details_agent is not None
+        load = leaf_load(component)
```
ivanmilevtues, line 570: Okay, now this explains a lot this method is not using the correct thing, to decide if we are separable or not.

You have added the component_is_separable in planner_agent.py so we should directly delete this method here and use that whenever relevant. We also have component_is_separable and get_expandable_components.

We need only one of those methods as API for the diagram_generator, everything else should become private + w eshould drop all these random methods that we have like this one (_component_separable)

diagram_analysis/diagram_generator.py
```diff
@@ -581,28 +551,93 @@ def process_component(
...
+        return separable
+
+    def _expandable_ids_for_tree(
+        self,
+        root_analysis: AnalysisInsights,
+        sub_analyses: dict[str, AnalysisInsights],
+    ) -> tuple[list[str] | None, dict[str, list[str]] | None]:
+        """The run's own expandable sets for the root scope and each sub-scope.
+
+        Persisting these keeps a component the separability gate kept as a leaf from
```
ivanmilevtues, line 600: this comment is longer than the code.
I am also unsure if the diagram generator is the place to do this.  It is fine to stay here but we do quite a few lines of patchwork before saving that are scattered everywhere ;d. (i.e. normalizing paths to methods and rewiring the relations because of the JSON formats for saving.)

diagram_analysis/diagram_generator.py
```diff
@@ -972,13 +1004,16 @@ def submit_component(comp: Component, lvl: int):
...
                             self._strip_ignored(analysis, sub_analyses)
+                            expandable_ids, sub_expandable_ids = self._expandable_ids_for_tree(analysis, sub_analyses)
                             save_analysis(
                                 analysis=analysis,
                                 output_dir=Path(self.output_dir),
                                 sub_analyses=sub_analyses,
                                 repo_name=self.repo_name,
                                 repo_dir=self.repo_location,
                                 source_tree_hash=self._source_tree_hash(),
+                                expandable_component_ids=expandable_ids,
```
ivanmilevtues, line 1015: Rework the API of the save_analysis func, these were here even before but were nullable.

diagram_analysis/run_context.py
```diff
@@ -57,16 +61,11 @@ def _load_existing_run_id(repo_dir: Path) -> str | None:
     if not (repo_dir / ".git").exists():
         logger.info("Repo not yet cloned at %s; skipping run_id cache lookup", repo_dir)
         return None
-    final_cache = FinalAnalysisCache(repo_dir)
-    cluster_cache = ClusterCache(repo_dir)
-    final_latest = final_cache.load_most_recent_run()
-    cluster_latest = cluster_cache.load_most_recent_run()
```
ivanmilevtues, line 63: Makes sense as now the clusters are deterministic!

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+    owner map to its own files keeps the two apart. Cluster ids are already disjoint
+    across languages, so the per-language results merge without collision.
+
+    Both sides are normalized to repo-relative posix before the file match: ``cluster_to_files``
+    carries the CFG's absolute paths (the static-analysis cache expands them under the repo)
+    while ``file_methods`` is persisted relative, so a raw comparison would never match and
+    strand method anchoring on the ``source_cluster_ids`` fallback — the very cluster-id
+    anchoring this function exists to replace.
+
+    A component can be cluster-backed yet hold no methods — a data-only cluster, which
```
ivanmilevtues, line 67: A bit unclear why would we have a cluster with no methods? It will have what in it? If it is empty wouldn't it non-present

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,172 @@
...
+    for a component left with nothing. Names and descriptions are deliberately absent —
+    ``update_scope`` leaves the existing wording alone.
+
+    A component that comes out of the grouping holding exactly what it already held gets
+    **no operation at all**. An operation is not free: ``update_scope`` puts its target in
+    ``refresh_ids``, which reruns the LLM relation analysis for the whole scope, and
+    ``_remove_reassigned_clusters`` strips and restores the referenced clusters. Emitting
+    one per survivor therefore relabels every relation in the tree on a one-line diff.
+    """
+    combined = combine_cluster_results(cluster_results)
```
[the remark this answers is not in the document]
ivanmilevtues: Probably a good idea to check this.

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+        # still holds methods the clustering failed to represent live code (the subgraph
+        # builder guarantees one cluster per live method), so fail loud rather than save a
+        # scope with stale membership and hide the missed change.
+        still_populated = [
+            component.component_id
+            for component in scope.components
+            if component.component_id and any(group.methods for group in component.file_methods)
+        ]
+        if still_populated:
+            raise IncrementalClusteringError(scope_id, still_populated)
```
ivanmilevtues, line 146: hmm unsure if we have to raise here? What happens if there is a component which got mostly deleted and few of the remaining methods got reassigned to a different component (cluster)?
Is this impossible, to me it seems reasonable if we have a component which had a single util method let's say we delte everything but the util. Then we will raise while proly the util item will go into some other place.

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+        for component in scope.components
+        if component.component_id
+        for cluster_id in component.source_cluster_ids
+    }
+    owner: dict[int, str] = {}
+    for cluster_result in cluster_results.values():
+        language_files = {
+            normalize_repo_path(path, repo_dir) for files in cluster_result.cluster_to_files.values() for path in files
+        }
+        owner_of_method: dict[str, str] = {
```
ivanmilevtues, line 85: This name is confusing as it says owner_of_method i thought it is owner->method but it is infact the opposite -> method_name:component_owner
method_component_map maybe is a better name.

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+            for group in component.file_methods
+            # A language without a file index (cluster_to_files empty) falls back to every
+            # method, preserving single-language behaviour; the split only matters when two
+            # languages carry colliding qnames, and then both have file indexes.
+            if not language_files or normalize_repo_path(group.file_path, repo_dir) in language_files
+            for method in group.methods
+        }
+        for cluster_id, members in cluster_result.clusters.items():
+            tally = Counter(owner_of_method[member] for member in members if member in owner_of_method)
+            if tally:
```
ivanmilevtues, line 98: when would tally be false? isn't an empty coutner also true idk

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+                # Ties go to the lowest component id, so the mapping is run-independent.
+                owner[cluster_id] = min(tally.items(), key=lambda claim: (-claim[1], claim[0]))[0]
```
ivanmilevtues, line 100: I think there should be almost no ties, maybe it is a good idea to log if there is a tie. In general this is proly the best way that I can think of for assigning ownership (matching) but it will be insane to track backwards to understand why things happen so proly good idea to log the ties.

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+            ]
+        )
+    combined_cfg: nx.DiGraph = nx.compose_all(list(cfg_graphs.values())) if cfg_graphs else nx.DiGraph()
+
+    is_root = scope_id == ROOT_SCOPE_ID
+    low = TOP_LEVEL_COMPONENTS_MIN if is_root else SUBCOMPONENTS_MIN
+    high = TOP_LEVEL_COMPONENTS_MAX if is_root else SUBCOMPONENTS_MAX
+
+    previous = previous_ownership(scope, cluster_results, scope_id, repo_dir)
+    grouping = anchored_grouping(combined, combined_cfg, previous, low, high)
```
ivanmilevtues, line 166: I'd be honest didn't quite get this ;d

ivanmilevtues: Leiden then runs on that much smaller weighted graph to find groups of leaf clusters that are strongly connected. Those groups become the top-level components; the LLM only gives them names and descriptions.

I am adding this as explanation, I think that we are doing here the class clustering with the same idea behind it

diagram_analysis/scope_plan.py
```diff
@@ -0,0 +1,274 @@
...
+                continue
+            qualified = CodeBoardingClusterIds.qualify_local_id(
+                CodeBoardingClusterIds.from_graph_id(cluster_id), prefix
+            )
+            if qualified in claimed_ids:
+                owner[cluster_id] = claimed_ids[qualified]
+    return owner
+
+
+def plan_scope_update(
```
ivanmilevtues, line 110: It seems like this is a very important method, i'd spend some time simplifying it so that we know what is happeneing and how things work

static_analyzer/cluster_helpers.py
```diff
@@ -596,61 +271,134 @@ def _seeds_from_partition(
...
+            ranked = [(0, sizes[idx], idx) for idx in range(len(seeds))]
+        target = min(ranked)[2]
+        seeds[target].add(cid)
+        sizes[target] += method_count.get(cid, 0)
 
 
-def _cluster_packages(cid: int, cluster_result: ClusterResult) -> set[str]:
-    """Directories of the files a leaf cluster touches (its 'package')."""
-    return {os.path.dirname(path) for path in cluster_result.cluster_to_files.get(cid, set())}
+# ---------------------------------------------------------------------------
```
ivanmilevtues, line 362: this comment is inaccurate :D

static_analyzer/cluster_helpers.py
```diff
@@ -707,82 +435,145 @@ def supercluster_leaf_ids(
...
+        for owner, _weight in tally.most_common():
+            if owner not in taken:
+                taken.add(owner)
+                owners[index] = owner
+                break
+    return owners
+
+
+@dataclass(frozen=True)
+class AnchoredGrouping:
```
ivanmilevtues, line 496: I think this might be more than a dataclass and just have the anchoring methods within it. We can add a package clustering within the static_analyzer packange and split a bit more the cluster_helpers, cluster_relations and so on.

ivanmilevtues (approved):
Main comment is that we are still using multitude of clustering levels of abstractions, which I though (might be me misunderstanding) is set to leiden with methods and that's it. This might be the reason for some flakiness down the road or things wokring unexpectedly further we now have this meta graph which is supposed to do that level of abstraction.

i think that is my potential blocker as I am puzzled what is used when ;d.

the other thing that would be great if we simplify the planner method as it is loaded and I couldn't really properly understand what is happeneing. Also the AnchoredGrouping is tuff to understand.

But the real comment is on not always using either the meta_graph or the abstraction with Leiden.

Thanks for the effort!

ivanmilevtues:
Handled the review comments from `CodeBoarding/CodeBoarding#432` in commit `d36e531`.

What I fixed:
- Removed redundant `step_clusters_grouping` from both `agents/abstraction_agent.py` and `agents/details_agent.py` and inlined deterministic grouping in `run()` per request.
- Renamed `owner_of_method` → `method_component_map`, simplified tie handling, and added ownership-tie logs in `diagram_analysis/scope_plan.py`.
- Made provisional component naming de-dup within a scope (`_provisional_name(..., used_names)`), and added regression tests.
- Fixed singleton-absorption modularity accounting in `static_analyzer/cluster_helpers.py` + regression test.
- Updated tests to match API simplifications and new behavior.

Validation:
- Ran `pytest tests/static_analyzer/test_cluster_helpers.py tests/diagram_analysis/test_scope_plan.py tests/agents/test_abstraction_agent.py tests/agents/test_details_agent.py -q`
- Result: `47 passed`.

If you want, I can also reply to/resolve individual comments from the review thread with this mapping.

ivanmilevtues:
### Incremental eval rerun · `9463272f`

Small incremental suite, **no judge**, run against fresh full baselines produced by the same PR head. The four full baselines passed **12/12** criteria before the incremental runs started.

| test | criteria | engine time | methods moved | relations changed | semantic rewordings | remaining violation |
|---|---:|---:|---:|---:|---:|---|
| I2 `referenced-symbol-deleted` | 6/7 | 281s | 2 / 531 (0.4%) | 33 / 98 (34%) | 26 | 4 unjustified relations added |
| I3 `capability-retired-no-unit-lost` | 6/7 | 282s | 0 / 577 | 13 / 91 (14%) | 12 | 1 unjustified relation added |
| I4 `body-edit-no-call-change` | 6/7 | 16s | 0 / 679 | 0 / 64 | 0 | backing-edge evidence changed 14 times despite no call change |
| I6 `whole-module-deleted` | 6/7 | 502s | 0 / 910 | 1 / 146 (0.7%) | 0 | expected low-level component removal was not emitted |

| suite summary | result |
|---|---:|
| criteria | **24 met, 4 violated (86%)** |
| violated criterion | `structural_operations` in all four tests |
| incremental engine time | **1,081s total** |
| incremental wall time | **520s** with 4 workers |
| full-baseline wall time | **206s** with 4 workers |

Positive signals: `blast_radius`, `file_delta`, `anchor_fidelity`, `introduced_invariants`, `no_phantom_files`, and `no_ignored_files` all passed in every incremental test. The judge was intentionally skipped to match the PR description's small-suite results.

PR #434, opened 2026-07-26
Ground relation edges in the static CFG and stop ungrounded edge relabels

Svilen-Stefanov:
Stacks on #432 (feat/size-driven-expansion). Targets the residual **edge** churn in incremental analysis: relations between two *unchanged* components were being marked added/modified/deleted with no code change to back them.

## What it does
- **Ground edges in the static CFG** (`ground_relation_edges`): for a statically-backed pair, `all_edges` = the deterministic CFG cross-component call set; the LLM contributes wording, never edges. An edge the LLM invents or re-spells each run can no longer appear, vanish, or duplicate.
- **Carry unchanged-method edges forward** (`_reconcile_unchanged_edges`): within a pair, trust the fresh rebuild only for edges that touch a changed/added/deleted method; take every edge between two byte-identical methods from the baseline. A call lives in its source method's body, so an unchanged source cannot have changed what it calls.
- **Carry the label forward on unchanged edges, not endpoints**: a pair is re-worded only when its own backing edges moved, not when an endpoint component was merely flagged changed.
- **Drop statically-backed component self-loops** (`drop_internal_self_relations`): a `src_id == dst_id` relation is an intra-component call with no cross-component connection to draw at the current depth. The **CFG edge is untouched**, so expanding the component re-materialises it as a real cross-component relation once its endpoints split into different children. Edgeless runtime/config self-relations are left alone.

## Results (CodeBoarding-evals, small variant, engine-matched baseline)
Ungrounded backing-edge changes per test:

| test | before | after |
|---|---|---|
| capability-retired (optional-dep removal) | 5 | **0** |
| whole-module-deleted | 0 | **0** |
| body-edit-no-call-change | 0 | **0** |
| referenced-symbol-deleted | 5 | 6 |

Benchmark 25/28 → **27/28** criteria (3 `structural_operations` violations → 1). Three of four small tests reach zero ungrounded edge churn once base and head share the engine.

The one remaining test (`referenced-symbol-deleted`) deletes a widely-referenced symbol, which forces re-clustering; its residual is dominated by `is_static=False` LLM edges landing on re-clustered pairs. Next lever: method-keyed (not pair-keyed) carry-forward for non-static edges, plus full-vs-incremental clustering determinism.

## Tests
- `tests/static_analyzer/test_global_relations.py::TestSelfLoopSuppression` — self-loop dropped + CFG edge untouched, re-surfaces cross-component on expansion, edgeless self-relation preserved.
- `test_cluster_relations.py`, `test_global_relations.py`, `test_incremental_copy_forward.py`, `test_incremental_agent.py` updated for grounding + carry-forward.
- Full suite: 1384 passed.


static_analyzer/cluster_relations.py
```diff
@@ -54,6 +54,59 @@ def build_global_node_to_component_map(
...
+    """Whether two edge-endpoint qualified names denote the same symbol.
+
+    The CFG names a symbol canonically (``src.pkg.mod.Cls.m``); the LLM often writes a
+    non-canonical variant of the same symbol — a ``Cls:m`` class separator, or a module
+    path missing the source-root prefix (``pkg.mod.Cls.m``). Normalise the separator and
+    accept a suffix match so a highlight is tied to its real CFG edge rather than kept as a
+    drifting duplicate.
+    """
+    a, b = a.replace(":", "."), b.replace(":", ".")
+    return a == b or a.endswith("." + b) or b.endswith("." + a)
```
ivanmilevtues, line 67: this can produce false positives i.e. cli.markitdown.markitdown and cli.markitdown (for the package).

static_analyzer/engine/result_converter.py
```diff
@@ -20,9 +22,16 @@ def convert_to_codeboarding_format(
     symbol_table: SymbolTable,
     result: LanguageAnalysisResult,
     adapter: LanguageAdapter,
+    ignore_manager: RepoIgnoreManager | None = None,
```
ivanmilevtues, line 25: I woul suggest to have this as non-nullable?

static_analyzer/engine/result_converter.py
```diff
@@ -34,6 +43,28 @@ def convert_to_codeboarding_format(
     language = adapter.language
     call_graph = CallGraph(language=language)
 
+    def is_ignored(file_path: object) -> bool:
+        return ignore_manager is not None and ignore_manager.should_ignore(Path(str(file_path)))
```
ivanmilevtues, line 47: if the ignore manager is not nullable we can have this as a direct call.

static_analyzer/engine/result_converter.py
```diff
@@ -34,6 +43,28 @@ def convert_to_codeboarding_format(
...
+
+    def is_ignored_qname(qname: str) -> bool:
```
ivanmilevtues: You can add this method as nullable paramed one to the ignore_manager.is_ignored_sym(sybol_table.symbols.get(qname)) -> if none false else it checks with the path.

static_analyzer/engine/result_converter.py
```diff
@@ -34,6 +43,28 @@ def convert_to_codeboarding_format(
...
+    def keep_hierarchy(hierarchy: dict | None) -> dict:
+        # result.hierarchy is built from the whole symbol table, which includes ignored
+        # files, so drop ignored classes (and ignored super/subclass references) before
+        # they reach getClassHierarchy.
+        filtered: dict = {}
+        for qname, info in (hierarchy or {}).items():
+            if is_ignored_qname(qname):
+                continue
+            kept = dict(info)
+            for relation in ("superclasses", "subclasses"):
```
ivanmilevtues: superclasses and subclasses should be constants

static_analyzer/engine/result_converter.py
```diff
@@ -34,6 +43,28 @@ def convert_to_codeboarding_format(
...
+        sym = symbol_table.symbols.get(qname)
+        return sym is not None and is_ignored(sym.file_path)
+
+    def keep_hierarchy(hierarchy: dict | None) -> dict:
+        # result.hierarchy is built from the whole symbol table, which includes ignored
+        # files, so drop ignored classes (and ignored super/subclass references) before
+        # they reach getClassHierarchy.
+        filtered: dict = {}
+        for qname, info in (hierarchy or {}).items():
+            if is_ignored_qname(qname):
```
ivanmilevtues: Can you check the following? 
I thought that the way we use the ignore manager is by adding it already in the scanner.py and then we pass only the non-ignored files downstream for static_analysis.

Now thinking about it, this might make the static analysis flaky as we don't resolve some references of files which are ignore (let's consider this corner case) and we also are presenting warnings for thing which might be used in the ignored parts of the codebase.

agents/incremental_agent.py
```diff
@@ -314,6 +314,7 @@ def generate_scope_relations(
         scope: AnalysisInsights,
         scope_name: str,
         context: ScopeRelationContext,
+        changed_members: set[str] | None = None,
```
ivanmilevtues, line 361: what qodo said :D

agents/relation_edges.py
```diff
@@ -68,29 +85,81 @@ def _relation_backing_survives(relation: Relation, live_qnames: set[str]) -> boo
     )
 
 
+def _backing_edge_pairs(relation: Relation) -> set[tuple[str, str]]:
+    edges = relation.all_edges or relation.key_edges
+    return {(edge.source.qualified_name, edge.target.qualified_name) for edge in edges}
+
+
+def _relation_edges_unmoved(rebuilt: Relation, previous: Relation) -> bool:
```
ivanmilevtues, line 97: I'd use the source_inspector to check if the calls are the same i.e. if a parameter was changed. Maybe it is too much unsure, as if that is the case then the caller method for sure is modified.

agents/relation_edges.py
```diff
@@ -68,29 +85,81 @@ def _relation_backing_survives(relation: Relation, live_qnames: set[str]) -> boo
...
+    dropped when either endpoint is no longer live — a whole-file deletion never reaches
+    ``changed_members``, so liveness is what excludes its phantom edges.
+    """
+
+    def split(fresh_edges: list[RelationEdge], baseline_edges: list[RelationEdge]) -> list[RelationEdge]:
+        fresh_grounded = [e for e in fresh_edges if _edge_touches_changed_method(e, changed_members)]
+        fresh_by_pair = {(e.source.qualified_name, e.target.qualified_name): e for e in fresh_edges}
+        baseline_stable: list[RelationEdge] = []
+        for edge in baseline_edges:
+            if _edge_touches_changed_method(edge, changed_members):
```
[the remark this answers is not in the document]
ivanmilevtues: Yea I think in general, we need to add a rename/moved status for these cases.

ivanmilevtues (approved):
Okay I did a review, I am still unsure if I fully grasp the changes. I will put on a +1.

I will continue reading it again later after the last PR.

Main comment is that ignore.py is used a bit in a patchwork way imo. We need to filter the ignored files at one place, I wrote somewhere a comment that we are doing that in the scanner.py and then the static_anlaysis should analyze only non-ignored files (which might be an issue if you ignore files which are used within the codebase - resulting in inaccurate warnings and bad resolutions in general (blind spots in the CFG))

static_analyzer/cluster_relations.py
Svilen-Stefanov: Good catch. Fixed it so each LLM key binds to the one static edge it names most specifically, and a bare/partial name that fits several equally (your `cli.markitdown` case) is dropped rather than stamped onto every sibling. Added a test for it.

static_analyzer/engine/result_converter.py
Svilen-Stefanov: Agree, made it non-nullable. Turned out the incremental path (`incremental_orchestrator`) wasn't passing a manager at all, so it was silently skipping this filtering, fixed that too.

static_analyzer/engine/result_converter.py
Svilen-Stefanov: Yep, dropped the local wrapper now that it's non-nullable.

static_analyzer/engine/result_converter.py
Svilen-Stefanov: I tried this, but once it's non-nullable it's just `should_ignore(sym.file_path)` with a rename, so it ends up a trivial passthrough on the manager. Went with a direct `should_ignore` + a small local None check in the converter instead (the None comes from the `symbol_table` lookup, which felt more like a converter concern). Ok with that?

static_analyzer/engine/result_converter.py
Svilen-Stefanov: Done, they're constants now (`SUPERCLASSES` / `SUBCLASSES`).

static_analyzer/engine/result_converter.py
Svilen-Stefanov: Yeah, I think you're right that it's a bit scattered. For now I made the incremental path consistent with the full one (both pass the `ignore_manager` here). The bigger idea of filtering once in scanner.py and only handing non-ignored files downstream is nicer, but the corner case you mention is real: if someone ignores a file that's actually imported or used, we'd lose those references and get blind spots in the CFG plus misleading warnings. So I don't think we can just drop them from static analysis blindly, we'd want to still resolve them but not surface them. Can I make a ticket so we design that properly rather than squeeze it into this PR?

agents/incremental_agent.py
Svilen-Stefanov: Fair, dropped the Optional. Prod always passes a set (empty when nothing changed), so None wasn't buying us anything on the agent methods.

agents/relation_edges.py
Svilen-Stefanov: I think you talked yourself into the right answer here :D if a call's args changed then the caller body changed too, so it's already in `changed_members` and we take the fresh edge. So I don't think we need the source_inspector for this. Lmk if you had a case in mind where the body wouldn't change.

agents/relation_edges.py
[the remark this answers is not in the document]
Svilen-Stefanov: Agree, a proper rename/moved status would be cleaner than leaning on liveness for these. Bit bigger than this PR though, I'll make a ticket.

Svilen-Stefanov:
Thanks Ivan! Went through the comments. Tightened the grounding match so a bare leaf can't stamp two edges, made `ignore_manager` non-nullable and actually passed it from the incremental path (it was being skipped there), and dropped the `changed_members` Optional. On the bigger 'filter in one place' point I left a longer reply inline, I think you're right and it's worth a ticket.

PR #435, opened 2026-07-26
perf(incremental): skip and parallelize scope relation regeneration

Svilen-Stefanov:
Stacks on #434, rebased onto `02d5de47`.

## Why

In an incremental run, relation regeneration dominates wall time: every touched scope makes two sequential LLM calls (`step_api_surfaces` then `step_relation_analysis`), and the scopes are walked one after another with no executor. A full re-analysis of the same repository can finish faster than the incremental update it is supposed to be cheaper than, because the full path fans out over its workers and this loop does not.

## What

**A scope where nothing changed skips both LLM calls.** `generate_scope_relations` rebuilds that scope's edges from the live CFG when `context.changed_ids` is empty — the `changed_ids & settled` set the caller narrows to after the copy-forward pass has proved some components byte-identical. For such a scope `preserve_unchanged_relations` discards the LLM's output anyway: a rebuilt pair that exists in the baseline gets its baseline label restored, and a fresh pair touching nothing changed is dropped. Gating on `changed_members` instead was rejected as unsafe — it misses new components, module-level edits and membership churn.

**Independent scopes regenerate concurrently.** `generate_all_scope_relations` runs them through a `ThreadPoolExecutor` with one isolated `IncrementalAgent` per worker thread. That isolation is required, not defensive: `step_relation_analysis` writes the scope's clustering onto `self.toolkit.context.cluster_analysis`, which `ComponentBridgeEdgesTool` reads mid-invocation, so a shared agent would let two scopes clobber each other. Per-worker clones share only the read-only static analysis, `executor.map` preserves order so the log stays deterministic, and each worker's token and tool counters are folded back through `RunStats.merge` even when a scope raises. One scope or none runs inline.

Relation validation runs `validate_relations` in the abstraction, details and incremental steps, so an LLM relation with a hallucinated or degenerate `key_edge` is retried rather than saved. A repair pass that canonicalises hallucinated edge names, so the full validation costs fewer retries, is a follow-up.

## Validation

Small incremental benchmark (`codeboarding-evals`), this branch and the commit it sits on, both `--baseline fixed` so base and head share the engine that produced the baseline:

| test | `02d5de47` | this branch |
|---|---|---|
| I2 referenced-symbol-deleted | 6 met / 1 violated · 292s | 6 met / 1 violated · **83s** |
| I3 capability-retired-no-unit-lost | 6 met / 1 violated · 393s | 6 met / 1 violated · **103s** |
| I4 body-edit-no-call-change | 7 met / 0 violated · 37s | 7 met / 0 violated · 37s |
| I6 whole-module-deleted | 7 met / 0 violated · 658s | 7 met / 0 violated · **42s** |
| **suite** | **26 met / 2 violated · 659s** | **26 met / 2 violated · 105s** |

Every criterion lands the same way; the suite is 6.3× faster and I6 alone 15.7×.

Both violations are engine-independent. I2 is violated in all six runs across the two engines. I3 passes only when a run emits zero unjustified edge operations, and that is a coin flip: over four runs each, `02d5de47` met it once and this branch not at all, with overlapping churn — 0–3 edges added there, 1–3 here. Runs are shared in CodeBoarding/CodeBoarding-evals#9.

**What the skip does to the document, measured.** With each scope's decision logged, 2 of 12 scopes skip the LLM on I3. One of them drops a baseline runtime relation: a skipped scope's LLM-only pairs can only return through `preserve_unchanged_relations`' restore branch, which additionally requires `_relation_backing_survives`, and that relation's invented edge (`src.click.types.convert_type -> src.click.types.ParamType`) names no cluster member. `rebuild_global_relations` restores it into the saved document from the baseline global set, so no criterion moves; the nested sub-analysis loses it.

That check is load-bearing rather than an oversight. Carrying the scope's own relations forward instead — so an unchanged scope re-grounds what it already had — resurrects a deleted module's edges and re-injects `local_app.py` into the file index through `index_relation_endpoints`, taking I6 from 7 met to 5 met / 2 violated. Measured as `fix-on-fixed` in the evals PR.

## Tests

Full unit suite green: 1930 passed, 28 skipped. `tests/agents/test_monitoring.py` covers `RunStats.merge`; `tests/agents/test_incremental_agent.py` covers the skip gate.

## Summary by CodeRabbit

* **Performance**
  * Improved incremental relation regeneration speed with parallel regeneration across multiple touched scopes.
  * Avoids unnecessary LLM/API-surface and relation-analysis work when no changes are detected; rebuilds relations deterministically from the live CFG.

* **Bug Fixes**
  * Ensures relation IDs, source reference lines, and endpoint indexing are correctly rebuilt after regeneration.
  * Preserves unchanged relations when selectively regenerating edges.

* **Monitoring**
  * Added thread-safe merging for run statistics, combining tokens and per-tool counts/errors/latency.

* **Tests**
  * Updated incremental relation-generation test to include changed IDs.
  * Added coverage for `RunStats.merge()` aggregation behavior.

* **Documentation**
  * Refreshed select docstrings in scope planning utilities.


agents/abstraction_agent.py
```diff
@@ -175,7 +175,10 @@ def step_relation_analysis(
         relation_result = self._invoke_validate(
             prompt,
             ComponentRelations,
-            validators=[validate_relations],
+            # Retry only to recover unknown endpoint names; unsupported/degenerate relations are
+            # dropped downstream (merge_relations + reference cleanup), so evidence re-rolls change
+            # nothing in the saved output.
+            validators=[validate_relation_component_names],
```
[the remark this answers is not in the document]
ivanmilevtues: The key_edges is a way for the LLM to add an edge if it is unresolved i.e. (not static backed because of bad resolution or cross technology communication like REST or something else). I think you probably added this to make the incremental faster as it probably was taking forever to pass the validation.

I would rather put the other thing back and in fact use self._invoke_repair_validate as it might be able to repair few of the edges if their names are hallucinated:
_invoke_repair_validate

we have in the repair.py a method for that:
def repair_key_entities

agents/details_agent.py
```diff
@@ -215,7 +215,8 @@ def step_relation_analysis(
         relation_result = self._invoke_validate(
             prompt,
             ComponentRelations,
-            validators=[validate_relations],
+            # Retry only to recover unknown endpoint names; bad relations are dropped downstream.
+            validators=[validate_relation_component_names],
```
ivanmilevtues: probably the same comment as the one in the abstraction_agent.py:
```The key_edges is a way for the LLM to add an edge if it is unresolved i.e. (not static backed because of bad resolution or cross technology communication like REST or something else). I think you probably added this to make the incremental faster as it probably was taking forever to pass the validation.

I would rather put the other thing back and in fact use self._invoke_repair_validate as it might be able to repair few of the edges if their names are hallucinated:
_invoke_repair_validate

we have in the repair.py a method for that:
def repair_key_entities```

File: agents/incremental_agent.py [hunk 3]
```diff
@@ -316,29 +328,38 @@ def generate_scope_relations(
...
-        rels = self.step_relation_analysis(
-            scope,
-            scope_name,
-            api_surfaces,
-            cluster_analysis,
-            context.cluster_results,
-            context.cfg_graphs,
-        )
+
+        if context.changed_ids:
```
ivanmilevtues, line 387: smart, great time gain

agents/incremental_agent.py
```diff
@@ -316,29 +328,38 @@ def generate_scope_relations(
...
+                api_surfaces,
+                cluster_analysis,
+                context.cluster_results,
+                context.cfg_graphs,
+            )
+        else:
+            # Nothing in this scope changed, so preserve_unchanged_relations below would discard any
+            # LLM output anyway; skip both round-trips and rebuild edges from the live CFG.
+            scope.components_relations = []
+            self._attach_static_relations(scope, context.cfg_graphs)
```
ivanmilevtues, line 402: maybe run this outside of the if/else and pull it out of the step_relation_analysis as I firstly thought why do we do static_relations only for unchanged things.

agents/incremental_agent.py
```diff
@@ -366,23 +387,84 @@ def generate_all_scope_relations(
...
+            tasks.append((ROOT_SCOPE_ID, root_analysis))
         for scope_id in sorted(relation_contexts.keys() - {ROOT_SCOPE_ID}):
             sub = sub_analyses.get(scope_id)
             if sub is not None:
-                rels = self.generate_scope_relations(sub, scope_id, relation_contexts[scope_id], changed_members)
-                if rels:
-                    all_llm_rels.append((scope_id, rels))
+                tasks.append((scope_id, sub))
+
+        if len(tasks) <= 1:
```
ivanmilevtues, line 445: Wouldn't the parallel generation spin just one thread in the case of one task, I think this is a bit of an overoptimisation and later down the line if we change how we do the tasks we will have two places to modify. Personally I think it's just better to have a single self._generation_scope_relations() and it works with parallel.

agents/incremental_agent.py
```diff
@@ -366,23 +387,84 @@ def generate_all_scope_relations(
...
+        worker_local = threading.local()
+        workers: list[IncrementalAgent] = []
+        workers_lock = threading.Lock()
+
+        def run_one(task: tuple[str, AnalysisInsights]) -> tuple[str, list[Relation]]:
+            scope_id, scope = task
+            worker = getattr(worker_local, "agent", None)
+            if worker is None:
+                worker = self._clone_for_worker()
+                worker_local.agent = worker
```
ivanmilevtues, line 481: This is comment is correct unsure how relevant it is in reality. We will get innacurate reports for LLM usage so it won't break funcitonality.

It is probably easier to make the agent safe for parallel usage. I think most of it is readonly operations so unsure why it is dangerous to use them as is. (I suppose only the monitoring part isn't) + we don't have memory so I don't see what would the issue be

agents/incremental_agent.py
```diff
@@ -366,23 +387,84 @@ def generate_all_scope_relations(
...
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
+                results = list(executor.map(run_one, tasks))
+        finally:
+            # Merge even if a scope raised, so a failed run still reports the token/tool
+            # usage the workers already incurred.
+            for worker in workers:
+                self.agent_stats.merge(worker.agent_stats)
+        return results
+
+    def _clone_for_worker(self) -> "IncrementalAgent":
```
ivanmilevtues, line 498: as per my comment on codex's comment, I think we don't need this but rather to make whatever is not threads safe, safe and reuse the same agent. Imo it is readonly

monitoring/stats.py
```diff
@@ -25,6 +25,19 @@ def reset(self):
             self.tool_errors = defaultdict(int)
             self.tool_latency_ms = defaultdict(list)
 
+    def merge(self, other: "RunStats") -> None:
```
ivanmilevtues, line 28: This method is used only in tests, it should go in the test file rather than here.

ivanmilevtues (changes requested):
I think overall this is a great change which is speeding up things quite a lot.

My real comment is around using the repair functionality rather than weakening the validation in all 3 agents. My other comment is to simplify the incremental agent by not spawning more of it but seeing what is the issue of using it in parallel (I think there shouldn't be one).

agents/abstraction_agent.py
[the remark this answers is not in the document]
Svilen-Stefanov: Yeah, fair. I dropped the evidence half of `validate_relations` mostly for speed, the retries were taking ages on incremental. Put it back now. Small snag on the repair though: `repair_key_entities` canonicalizes the component `key_entities`, not the relation `key_edges`, so it doesn't actually touch the hallucinated edges here. Were you thinking a separate repair for the edges (canonicalize the endpoints through the reference resolver, like `repair_key_entity_references` does for entities), or just restore `validate_relations` and lean on `remove_unresolved_references` downstream? I'd lean towards a small edge repair so we keep the validation without paying the retry cost. wdyt?

agents/details_agent.py
Svilen-Stefanov: Same as the abstraction_agent thread, put `validate_relations` back here too. Let's settle the repair mechanism over there and I'll mirror it.

agents/incremental_agent.py
Svilen-Stefanov: 🙏

agents/incremental_agent.py
Svilen-Stefanov: Ah yeah, I get why it reads like static only runs for the unchanged path. The catch is `step_relation_analysis` already does the static grounding internally (assign ids + `build_static_relations`), and it's shared with the abstraction/details agents, so pulling it fully out means changing it there too. I can restructure so `_attach_static_relations` runs once after the if/else and `step_relation_analysis` stops doing it, just want to make sure it doesn't ripple into the full run. Let me have a look if it's clean.

agents/incremental_agent.py
Svilen-Stefanov: True, two places to touch is annoying. The `<=1` branch was just to skip the threadpool + agent setup on trivial changes, but the parallel path handles a single task fine (just one worker), so happy to collapse it. It's a bit tied to the clone question below, if we drop the clone this gets a lot simpler.

agents/incremental_agent.py
Svilen-Stefanov: Agree the monitoring is the only genuinely unsafe bit, and it's inaccurate rather than broken. But it's not fully readonly unfortunately: `step_relation_analysis` writes `cluster_analysis`/`cluster_results`/`cfg_graphs` onto `self.toolkit.context` (the shared `RepoContext`), and `ComponentBridgeEdgesTool` reads them back mid tool-call. So two scopes on the same agent would clobber each other's clustering context, that's why I cloned. To reuse a single agent I'd make those three fields thread-local and put a lock on the monitoring callback. Happy to go that way if you'd prefer one agent, just flagging it's a bit more than 'it's readonly' since we mutate the toolkit context per scope. Thread-local context or keep the clone?

agents/incremental_agent.py
Svilen-Stefanov: Same thread as above, the only real blocker to reusing one agent is the `toolkit.context` mutation (+ the monitoring lock). If you're good with making those thread-local I'll drop the clone and the merge with it.

monitoring/stats.py
Svilen-Stefanov: It's actually used in prod right now, that's how the clone worker stats get folded back into the parent for `llm_usage`. But if we reuse one agent (the thread above), the clones disappear and `merge` goes with them, or moves to the tests. So kind of depends on the clone decision, I'll move/remove it once we settle that.

Svilen-Stefanov:
Thanks Ivan! Agree on both. I've put `validate_relations` back in all 3 agents so the evidence check is in again (that also covers the codex P1). Left some thoughts inline on the repair bit and the single-agent vs clone question, want to align before I go further.

agents/abstraction_agent.py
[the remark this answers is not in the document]
ivanmilevtues: yea we can do a small validation + repair and just keep the retries at 0/1 so we don't pay the cost

ivanmilevtues (approved):
Metric | PR 434 control | PR 435 | Result
-- | -- | -- | --
Criteria met | 30 | 30 | unchanged
Criteria violated | 4 | 4 | unchanged
Incremental accuracy | 88% | 88% | unchanged
Sum of per-test wall time | 1,780s | 356s | 80% lower
Effective speedup | — | 5.0× | substantial
Parallel suite elapsed time | — | 133s | about 2m 13s

Test | PR 434 | PR 435 | Reduction | Quality
-- | -- | -- | -- | --
I2 Referenced symbol deleted | 319s | 84s | 74% | 6 met / 1 violated
I3 Capability retired | 429s | 74s | 83% | 6 met / 1 violated
I4 Body-only edit | 94s | 34s | 64% | 7 met / 0 violated
I6 Whole module deleted | 759s | 35s | 95% | 7 met / 0 violated
I13 MarkItDown expansion | 179s | 129s | 28% | 4 met / 2 violated

PR 434 control is pr434-b05c1e9-from-fixed. PR 434’s merged source matches that engine behavior; the merge only added regenerated .codeboarding documentation afterward.

PR #436, opened 2026-07-27
Honor live .codeboardingignore in function_size health check

Svilen-Stefanov:
## What
The `function_size` health check (E1) filtered files with `RepoIgnoreManager.should_skip_file`, the static helper that matches the **hard-coded default template** instead of the user's live `.codeboardingignore`. It was the only caller of that default-template path in the pipeline.

## Why it matters
When a repo customizes `.codeboardingignore` to opt a template-covered directory back in (`tests/`, `examples/`, `mocks/`, `e2e/`, ...), those functions appear in the rendered architecture and in `file_coverage`, but were silently excluded from the `function_size` metric — under-counting the denominator and missing large-function findings relative to what's shown. Every other layer (CFG discovery, `strip_ignored`, `file_coverage`, clustering) already honors the live file; this brings the health metric in line.

## Change
- Thread the live `RepoIgnoreManager` (built from `repo_path`, always supplied by the generator) from `run_health_checks` → `_collect_checks_for_language` → `check_function_size`, and filter with `should_ignore`.
- Because the call graph is already ignore-filtered upstream, this filter is a no-op for what's rendered — and it additionally excludes any warm-start "ghost" node from the metric.

## Tests
- Updated the exclusion tests to the new contract: a template-ignored `__tests__` file is still excluded; an **opted-in** test dir is now scored; `ignore_manager=None` scores everything.
- `black --check`, `mypy`, and `pytest tests/health/` (41 passed) all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Function-size health checks now respect the repository’s current ignore settings.
  * Files and directories explicitly included by ignore configuration are correctly evaluated.
  * Checks run without ignore configuration now evaluate all files consistently.
  * Health report paths are now consistently shown relative to the repository root.

### DIFF
--- a/health/checks/function_size.py
+++ b/health/checks/function_size.py
@@ -1,4 +1,5 @@
 import logging
+from pathlib import Path
 
 from health.models import (
     FindingEntity,
@@ -25,13 +26,15 @@ def collect_function_sizes(call_graph: CallGraph) -> list[float]:
     return sizes
 
 
-def check_function_size(call_graph: CallGraph, config: HealthCheckConfig) -> StandardCheckSummary:
+def check_function_size(
+    call_graph: CallGraph, config: HealthCheckConfig, ignore_manager: RepoIgnoreManager | None = None
+) -> StandardCheckSummary:
     """E1: Check function/method sizes across the call graph.
 
-    Flags functions that exceed line count thresholds. Large functions are
-    harder to understand, test, and maintain.
-
-    Excludes test and infrastructure files as they have different size norms.
+    Flags functions that exceed line count thresholds. File exclusion follows the
+    live ``.codeboardingignore`` via ``ignore_manager`` so the metric measures exactly
+    what the rendered architecture contains. ``ignore_manager`` is None only when no
+    repo root is available; the call graph is already ignore-filtered upstream regardless.
     """
     findings: list[FindingEntity] = []
     total_checked = 0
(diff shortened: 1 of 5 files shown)


health/checks/function_size.py
```diff
@@ -25,13 +26,15 @@ def collect_function_sizes(call_graph: CallGraph) -> list[float]:
     return sizes
 
 
-def check_function_size(call_graph: CallGraph, config: HealthCheckConfig) -> StandardCheckSummary:
+def check_function_size(
+    call_graph: CallGraph, config: HealthCheckConfig, ignore_manager: RepoIgnoreManager | None = None
```
ivanmilevtues: Probably this can be non-nullable

health/runner.py
```diff
@@ -217,11 +219,14 @@ def run_health_checks(
         return None
 
     repo_root = str(repo_path) if repo_path is not None else None
+    # Live .codeboardingignore so function_size scores exactly what the architecture renders,
+    # rather than the hard-coded default template. None only when the caller omits a repo root.
+    ignore_manager = RepoIgnoreManager(Path(repo_path)) if repo_path is not None else None
```
ivanmilevtues: I think repo_path should never be nullable you shouldn't be able or run health/repository analysis with nullable.

The wrapper disables the buttons, so this should never happen allegedly so we can make them non-nulalble.

ivanmilevtues (approved):
please make the things non-nullable ;d

Issue #65, opened 2025-09-18
LLM prompts optimizations for Claude


ivanmilevtues:
- Following https://github.com/CodeBoarding/CodeBoarding/issues/56 we need to optimize the prompts for each programming language.
- For this task in particular, the person has to optimize the prompts for the Gemini model.

SiddhantGahankari:
Hi @ivanmilevtues , Could you please assign this issue to me.

ivanmilevtues:
Hey @SiddhantGahankari, you can join our discord if you have any questions happy to help: https://discord.gg/T5zHTJYFuy

PR #441, opened 2026-07-31
fix(relations): ground relation edges in symbols and gate churn on changed methods

Svilen-Stefanov:
Closes the report on `CodeBoarding-action` PR #65: a one-file commit produced **five modified edges**, most showing no call site, one deleted edge implying `build_cta.py` belonged to a component it was never in, and added edges naming the changed component but not the changed methods.

Two independent causes.

## 1. Edges were validated against files, not symbols

`RELATION_ANALYSIS_MESSAGE` tells the model not to limit relations to static calls, so it authors `key_edges` freely and `ground_relation_edges` promotes them verbatim when no static edge backs the pair. The only gate, `keep_relation_edge`, asked whether the referenced **file** existed — true of almost any plausible-looking name in the repo.

It now resolves both endpoints as **symbols**:

- both resolve → keep
- neither resolves → drop
- one resolves, the other looks like an internal reference → drop; otherwise keep, since a genuine call into a dependency has no node

Resolution is memoised per analysis, and `looks_internal_reference` caches its token set — it went quadratic once it was on this path.

`prune_ungrounded_edges` then repairs attribution. An edge whose endpoints live in components other than the ones its relation names is **re-filed** to the relation that declares its true pair, not deleted: measured on a stored analysis, deleting them orphaned 25 of 28 real cross-component calls, which existed nowhere else in the document. Deletion is correct only when both endpoints sit inside one component, where there is no cross-component fact to keep — and that is the shape the original report was about.

`drop_reverse_duplicates` collapses `A -> B` and `B -> A`, grounded side winning, ties broken deterministically. Four of the five reported edges were pure direction flips. Both passes run on full and incremental alike; applying them only to incremental made a body-edit commit report five phantom removals.

## 2. Churn was gated on files and on re-clustering, not on changed methods

An edge is a fact about two methods, so the commit has to have touched one of them. Four places substituted a coarser signal:

| where | substituted | now |
|---|---|---|
| `diagram_generator` | scope lost **any** member → drop **every** relation in it | drop only relations whose edges name a departed method |
| `_restore_unchanged_metadata` | any live/baseline membership difference is drift | membership-derived fields (`source_cluster_ids`, key entities) restored separately from prose; an entity survives if its qualified name is still live |
| `_incremental_changed_component_ids` | membership churn alone marks a component changed | the churn must involve a method the commit touched |
| `NOOP` plan step | always added its component to `refresh_ids` | refreshes only when the cluster ids actually moved |

Re-partitioning moves untouched methods between components on nearly every run, which is why each of these fired constantly.

Relation add/remove is gated on a set narrowed to components owning a changed method, plus components genuinely new or gone — deliberately **distinct** from the set used to decide what to re-derive, which must stay broader.

Relabelling requires the edge's **source** to be a changed method. Attributing a call to its caller is what makes "this edge changed" mean "this code changed". (Requiring *both* ends would be stricter, but a caller rewriting how it calls an unchanged callee is a real edge change; the source rule keeps it and the UI can still choose to show only both-ends changes.)

## Measured

Four incremental tests, each engine against **its own** baseline. A shared `--baseline main` carries the baseline's edges forward wherever the commit left the methods alone — 76 of 80 ungrounded edges survived a single update — and hides the entire effect.

| | main | this branch |
|---|---|---|
| relation churn | 35 | **12** |
| add/remove where neither endpoint changed | 9 | **2** |
| relabels with no changed source | 10 | **0** |
| `source_cluster_ids` changed | 22 | **0** |
| `key_entities` changed | 7 | **1** |
| ungrounded edges, fresh full analysis | 87 | **12–18** |
| incremental criteria met | 83% | **97%** |
| full criteria met | 75% | **88%** |

Per test (`additions/removals/modifications`, and how many had neither endpoint changed):

```
referenced-symbol-deleted   main 4/3/18 (2 unjust)  →  4/1/3  (0 unjust)
capability-retired          main 2/4/1  (6 unjust)  →  0/1/0  (1 unjust)
body-edit-no-call-change    main 0/0/0             →  0/0/0
whole-module-deleted        main 0/3/0  (1 unjust)  →  0/3/0  (1 unjust)
```

## Known gaps

- The two remaining unjustified removals are both LLM-authored (`is_static=false`) relations. Not chased — at 2 of 12 they are no longer the dominant signal.
- `structural_operations` is unstable at n=1 on `referenced-symbol-deleted`; repeated runs of one engine produce different operation *sets*. Don't read a single run of that test as a regression.

## Companion

Checks and criteria that measure this are on `CodeBoarding-evals` main (`0735726`): `relations.edge_endpoints_are_known_symbols`, `relations.edges_cross_declared_components`, and the `edge_grounding` / `ungrounded_edge_churn` / `relation_churn` criteria. Compare `incremental/relation-stability-fix` against `incremental/relation-stability-main`.

1959 tests, black and mypy clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved incremental analysis to preserve unaffected component metadata and relation descriptions.
  * Corrected relation detection to focus on changes affecting callers and relevant components.
  * Removed unsupported, misattributed, and duplicate reverse relations while preserving valid external references.
  * Improved handling of renamed, repartitioned, or removed methods and components.

* **Performance**
  * Added caching to speed up repeated reference and node resolution.

* **Tests**
  * Expanded coverage for incremental updates, relation validation, edge ownership, duplicate removal, and external references.

### DIFF
--- a/agents/incremental_agent.py
+++ b/agents/incremental_agent.py
@@ -40,6 +40,11 @@ from agents.prompts import (
     get_system_message,
 )
 from agents.relation_edges import index_relation_endpoints, preserve_unchanged_relations
+from static_analyzer.cluster_relations import (
+    build_node_to_component_map,
+    build_owner_index,
+    prune_ungrounded_edges,
+)
 from agents.scope_ids import ROOT_SCOPE_ID
 from agents.validation import ValidationContext, validate_relations
 from diagram_analysis.file_index import build_files_index
@@ -138,11 +143,17 @@ class IncrementalAgent(ClusterMethodsMixin, CodeBoardingAgent):
             if operation.action == ScopeOperationAction.NOOP:
                 component = components_by_id.get(operation.component_id or "")
                 if component is not None:
-                    component.source_cluster_ids = CodeBoardingClusterIds.sort(
+                    merged = CodeBoardingClusterIds.sort(
                         set(component.source_cluster_ids) | set(_operation_source_cluster_ids(scope_id, operation))
                     )
-                    if component.component_id:
-                        refresh_ids.add(component.component_id)
+                    # NOOP means the planner asked for nothing. Only a cluster set that actually
+                    # moved is a reason to refresh; marking every NOOP component refreshed put
+                    # untouched siblings into `changed_ids`, which is what let their relations
+                    # be added and removed for a commit that never reached them.
+                    if merged != component.source_cluster_ids:
+                        component.source_cluster_ids = merged
+                        if component.component_id:
+                            refresh_ids.add(component.component_id)
                 continue
 
             component = components_by_id.get(operation.component_id or "")
@@ -383,6 +394,17 @@ class IncrementalAgent(ClusterMethodsMixin, CodeBoardingAgent):
             for relation in scope.components_relations
             if relation.src_id and relation.dst_id
         }
+        # Also keyed by NAME. Ids are assigned per run by `assign_relation_ids`, so a
+        # re-partition can hand the same two components different numbers; the id lookup then
+        # misses and the pair is treated as brand new, which re-rolls its wording. Measured on
+        # `referenced-symbol-deleted`: 9 statically-backed relations whose call set was
+        # byte-identical came back re-worded for exactly this reason. Names survive
+        # renumbering, so they are the fallback identity.
+        baseline_by_names = {
+            (relation.src_name, relation.dst_name): relation.model_copy(deep=True)
+            for relation in scope.components_relations
+            if relation.src_name and relation.dst_name
+        }
 
         if context.changed_ids:
             cluster_analysis = _cluster_analysis_for_scope(scope, scope_name, context.cluster_results)
@@ -404,19 +426,48 @@ class IncrementalAgent(ClusterMethodsMixin, CodeBoardingAgent):
 
         if baseline_by_pair:
             live_ids = {component.component_id for component in scope.components if component.component_id}
+            # A relation is ADDED or REMOVED because code changed, not because a partitioner
+            # moved untouched methods around. `context.changed_ids` carries both — it is the
+            # right scope to re-derive, and the wrong reason to change the edge set — so the
+            # preservation gate sees only components the COMMIT reached. Components that are
+            # new or gone keep their ids here: they are changes the commit does account for.
+            commit_changed = {
+                component.component_id
+                for component in scope.components
+                if component.component_id
+                and any(
+                    method.qualified_name in (changed_members or set())
+                    for group in component.file_methods
+                    for method in group.methods
+                )
+            } | (set(context.changed_ids) - live_ids)
             live_qnames = {
                 qualified_name
                 for cluster_result in context.cluster_results.values()
                 for members in cluster_result.clusters.values()
                 for qualified_name in members
             }
+            for relation in scope.components_relations:
+                pair = (relation.src_id, relation.dst_id)
+                if pair not in baseline_by_pair:
+                    by_name = baseline_by_names.get((relation.src_name, relation.dst_name))
+                    if by_name is not None:
+                        baseline_by_pair[pair] = by_name
             scope.components_relations = preserve_unchanged_relations(
                 scope.components_relations,
                 baseline_by_pair,
-                set(context.changed_ids),
+                commit_changed,
                 live_ids,
                 live_qnames,
                 changed_members,
+            )
+            # Preservation runs after the grounding filters and re-injects baseline edges, so
+            # the assembled list is filtered once more — otherwise a baseline's invented or
+            # mis-attributed edges survive every update that leaves their methods alone.
+            scope.components_relations = prune_ungrounded_edges(
+                scope.components_relations,
+                build_owner_index(build_node_to_component_map(scope)),
+                self.reference_resolver.keep_relation_edge,
             )
             rels = scope.components_relations
         return rels
--- a/agents/incremental_agent.py
+++ b/agents/incremental_agent.py
@@ -142,18 +142,9 @@ class IncrementalAgent(ClusterMethodsMixin, CodeBoardingAgent):
                 continue
             if operation.action == ScopeOperationAction.NOOP:
                 component = components_by_id.get(operation.component_id or "")
-                if component is not None:
-                    merged = CodeBoardingClusterIds.sort(
-                        set(component.source_cluster_ids) | set(_operation_source_cluster_ids(scope_id, operation))
-                    )
-                    # NOOP means the planner asked for nothing. Only a cluster set that actually
-                    # moved is a reason to refresh; marking every NOOP component refreshed put
-                    # untouched siblings into `changed_ids`, which is what let their relations
-                    # be added and removed for a commit that never reached them.
-                    if merged != component.source_cluster_ids:
-                        component.source_cluster_ids = merged
-                        if component.component_id:
-                            refresh_ids.add(component.component_id)
+                if component is None:
+                    continue
+                self._sync_noop_component_cluster_ids(scope_id, component, operation, refresh_ids)
                 continue
 
             component = components_by_id.get(operation.component_id or "")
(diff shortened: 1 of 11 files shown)


agents/incremental_agent.py
```diff
@@ -138,11 +143,17 @@ def update_scope(
             if operation.action == ScopeOperationAction.NOOP:
                 component = components_by_id.get(operation.component_id or "")
                 if component is not None:
```
ivanmilevtues: put all this into a private method with a name that describes what we do as I don't get it at the moemnt.

If there is noop we are merging cluster ids and then if the merged ids are different than what we have at the moment we are switching them to the new ones. so to me that sounds it is safe to always put the component source_cluster_ids to merged no?

And last thing is, if we found out that merged !=component.source_luster_ids then we are adding it for refresh.

I need you to rewrite this.
merged cluster_ids = ...
if merged_cluster_ids != component.source_cluster_ids:
  refresh_ids.add()
component.source_cluster_ids = merged # always thelatest version
and for component None just continue upfront.

agents/incremental_agent.py
```diff
@@ -404,20 +426,49 @@ def generate_scope_relations(
 
         if baseline_by_pair:
             live_ids = {component.component_id for component in scope.components if component.component_id}
+            # A relation is ADDED or REMOVED because code changed, not because a partitioner
+            # moved untouched methods around. `context.changed_ids` carries both — it is the
+            # right scope to re-derive, and the wrong reason to change the edge set — so the
+            # preservation gate sees only components the COMMIT reached. Components that are
+            # new or gone keep their ids here: they are changes the commit does account for.
+            commit_changed = {
```
ivanmilevtues: split this into 2 things A = {}
then
C = A |(set(...)-live_ids)

agents/incremental_agent.py
```diff
@@ -404,20 +426,49 @@ def generate_scope_relations(
...
+                    for method in group.methods
+                )
+            } | (set(context.changed_ids) - live_ids)
             live_qnames = {
                 qualified_name
                 for cluster_result in context.cluster_results.values()
                 for members in cluster_result.clusters.values()
                 for qualified_name in members
             }
+            for relation in scope.components_relations:
```
ivanmilevtues, line 456: write a comment on what does this do

agents/incremental_agent.py
```diff
@@ -404,20 +426,49 @@ def generate_scope_relations(
...
                 changed_members,
             )
+            # Preservation runs after the grounding filters and re-injects baseline edges, so
+            # the assembled list is filtered once more — otherwise a baseline's invented or
+            # mis-attributed edges survive every update that leaves their methods alone.
+            scope.components_relations = prune_ungrounded_edges(
+                scope.components_relations,
+                build_owner_index(build_node_to_component_map(scope)),
+                self.reference_resolver.keep_relation_edge,
+            )
```
ivanmilevtues, line 478: that is fine. resolved.

static_analyzer/internal_references.py
```diff
@@ -28,14 +29,39 @@ def parent_qualified_name(qualified_name: str) -> str:
     return parent.split("(", 1)[0]
 
 
+_TOKEN_CACHE: dict[int, tuple[ReferenceType, set[str], set[str]]] = {}
```
ivanmilevtues: Shouldn't use token here on its own sounds like LLM related things. It is partial matching for symbols. Better naming schema please.

agents/relation_edges.py
```diff
@@ -107,7 +109,70 @@ def _relation_edges_unmoved(rebuilt: Relation, previous: Relation) -> bool:
 
 
 def _edge_touches_changed_method(edge: RelationEdge, changed_members: set[str]) -> bool:
-    return edge.source.qualified_name in changed_members or edge.target.qualified_name in changed_members
+    """Whether a code change can account for this call appearing or disappearing.
```
ivanmilevtues: no need for this long comment on a single check I think the name is good enough no?

agents/relation_edges.py
```diff
@@ -107,7 +109,70 @@ def _relation_edges_unmoved(rebuilt: Relation, previous: Relation) -> bool:
...
+    edges, so the arrow means something and must not be touched. A runtime/config relation has
+    no call to fix it, so the model picks an orientation afresh each run — and a swap keyed on
+    the ordered pair reads as one relation deleted and another added. On the reported case
+    (CodeBoarding-action#65) two of the five changed edges were exactly that: the same two
+    component pairs with their arrows reversed and relabelled.
+
+    Applied only when the fresh relation carries no backing edges at all, because an edge is
+    what would make the direction a claim rather than a phrasing — flipping a relation that has
+    one would point its edges the wrong way.
+    """
```
ivanmilevtues, line 128: I agree

agents/relation_edges.py
```diff
@@ -107,7 +109,70 @@ def _relation_edges_unmoved(rebuilt: Relation, previous: Relation) -> bool:
...
+            "dst_id": flipped.dst_id,
+            "src_name": flipped.src_name,
+            "dst_name": flipped.dst_name,
+            "relation": flipped.relation,
+            "evidence": flipped.evidence,
+        }
+    )
+
+
+def _only_change_backed_edges(relation: Relation, changed_members: set[str]) -> Relation:
```
ivanmilevtues: rename to _filter_edges_touched_by_change

agents/relation_edges.py
```diff
@@ -176,25 +241,45 @@ def touches_change(src_id: str, dst_id: str) -> bool:
     kept: list[Relation] = []
     rebuilt_pairs: set[tuple[str, str]] = set()
     for relation in rebuilt_relations:
+        if (relation.src_id, relation.dst_id) not in baseline_by_pair:
```
ivanmilevtues, line 228: Why would this even flip what is the root cause of the flip if it is not an LLM call?

agents/relation_edges.py
```diff
@@ -176,25 +241,45 @@ def touches_change(src_id: str, dst_id: str) -> bool:
...
+        # ones that show WHERE one component reaches the other. They are the only evidence that
+        # the connection itself changed, so they are what decides whether it may be re-worded.
+        # Gating on the endpoint components' change flags instead re-words far more: a component
+        # is flagged for any edit to a file it merely co-owns, so deleting one function re-worded
+        # 17 of the relations in `referenced-symbol-deleted` whose calls had not moved at all.
+        # A pair with no supporting calls has nothing that could have moved, so its wording is
+        # kept too — an edgeless runtime relation is prose, and re-rolling prose is the churn.
+        if (
+            not touches_change(*pair)
+            or _relation_edges_unmoved(relation, previous)
```
ivanmilevtues, line 264: if touches_change is false is it possible for _relation_edges_unmoved to be false?

diagram_analysis/diagram_generator.py
```diff
@@ -344,16 +364,36 @@ def _restore_unchanged_metadata(
                 continue
             final_keys = _member_keys(component)
             owns_changed_file = any(group.file_path in changed_files for group in component.file_methods)
-            if final_keys != meta.member_keys or (meta.member_qnames & changed_members) or owns_changed_file:
+            # Membership counts as unchanged when the only members that came or went are ones
```
ivanmilevtues: way too much text just explain in a single line maybe 2 at most.

static_analyzer/cluster_relations.py
```diff
@@ -67,6 +67,180 @@ def _qnames_match(a: str, b: str) -> bool:
...
+    25 of 28 real calls — the mis-attribution is worth repairing, the call is not worth losing.
+
+    An edge whose true pair has no relation to move to stays where it is: inventing a relation
+    here would be authoring architecture in a filter, and silently deleting the call would be
+    worse than leaving it mis-filed where a check can still flag it.
+    """
+    by_pair: dict[tuple[str, str], Relation] = {(relation.src_id, relation.dst_id): relation for relation in relations}
+    additions: dict[tuple[str, str], list[RelationEdge]] = {}
+
+    def rehome(edge: RelationEdge) -> tuple[str, str] | None | bool:
```
ivanmilevtues: this method name makes 0 sense.

ivanmilevtues (approved):
few minor comments or not so minor I will address and merge after that

agents/relation_edges.py
ivanmilevtues: It only flips for an ungrounded relation regenerated by the LLM. With no backing CFG edges, direction is prose rather than a static claim, so the model can emit the same pair in the reverse order; this restores the baseline orientation.

agents/relation_edges.py
ivanmilevtues: Yes. An untouched pair can still be rebuilt with a different edge set because of re-attribution. The first condition intentionally short-circuits that mismatch and keeps its baseline wording.

PR #443, opened 2026-08-01
No component is expanded into exactly one sub-component

Svilen-Stefanov:
Three defects in series, each of which hides the next. Found while building a delete-component test for the evals; the third was found by that test once it existed.

## A level with one box in it

A component expanded into a single sub-component adds a level that explains nothing — parent and child describe the same code, so a reader opens a box to find one box. **Measured over 574 stored analyses: 120 occurrences**, in two shapes — a childless only child (114) and a chain restating itself down three levels (6). The worst case was `5 Input & Stream Handling → 5.1 URI Resolver → 5.1.1 URI Scheme Handlers`, three nested boxes describing the same 1 file and 2 methods.

`absorb_single_child_components` collapses them at `finalize_for_save`, so it covers the full, incremental and partial flows and repairs a legacy baseline rather than rejecting it.

The repair is a prefix re-rooting. Every identifier in the tree — component id, `sub_analyses` key, scoped cluster id, relation endpoint — is the dotted path to a position, so moving the absorbed child's subtree onto the parent's path is one rewrite of `"<child_id>."` → `"<parent_id>."` applied everywhere. That renumbers the promoted siblings for free and cannot collide: a parent with exactly one child has no other child to collide with.

## An emptied component kept its box

`_cluster_backed_empty_component_ids` shields a component that is empty but still cluster-backed — one awaiting re-detail — from the prune. It was captured **after** `remove_deleted_files`, which clears methods and key entities but not `source_cluster_ids`, so it also shielded everything the scrub had just emptied and the deleted unit kept its box.

Moved to the same side of the scrub as `_capture_baseline_member_keys`, which is where this file already draws that distinction.

## A deleted symbol kept its pointers

`fix_key_entities_refs` is gated on the scopes an update touched, while `_reconcile_child_scope` strips departed membership from scopes it does *not* touch. A component emptied by that route kept key entities naming deleted symbols, and `prune_empty_components` reads a non-empty `key_entities` as a component still having something to describe.

So the empty box survived, the parent kept two children, and there was no single-child level left for the absorption to find. `drop_dangling_key_entities` enforces the invariant whole-tree and unconditionally. **Measured over the eval baselines and corpus: 557 of 557 key entities in healthy documents name a method some component owns**, so enforcing it cannot remove a legitimate reference.

Relations were already covered by two guards and needed no change — verified, 0 edges cited a deleted symbol.

## Verification beyond unit tests

Replayed over every stored analysis in CodeBoarding-evals: all defective documents repair with no scope key orphaned, no component id off its scope's dotted path, and each round-trips through `analysis.json` byte-identically. That last check matters — the store writes children only for a component declared expandable, so a botched re-parenting would delete a subtree *silently* rather than fail.

End to end on the real engine, one baseline through both builds: the e2e test `last-sibling-is-absorbed-into-its-parent` (CodeBoarding-evals) goes from failing on both to passing here, with `2.1` becoming a leaf, `2.1.1` absorbed and `2.1.2` pruned.

## Tests

- **23** in `test_tree_shape.py` — both absorption shapes, deep subtree re-rooting, the root scope and its empty-analysis guard, chain collapse to a fixpoint, relation re-pointing and self-loop drops, idempotency, the `analysis.json` round-trip, and the dangling-entity drop end to end.
- **3** in `test_incremental_copy_forward.py` pin the capture order, including one asserting the **old** order keeps the empty box, so the bug cannot return quietly.
- Two existing fixtures were themselves degenerate trees and were given siblings.

1962 pass; black and mypy clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagram cleanup by removing dangling entities and empty components.
  - Simplified single-child scopes while preserving descendant structure and relationships.
  - Corrected relation handling when underlying call connections are removed or reassigned.
  - Preserved cluster and scope lineage during incremental updates.
  - Improved consistency across repeated and incremental diagram processing, including edge cases involving root and nested scopes.

## Validation

Compared e2e17c2 with 8614ded.

- E1 component deletion: pass
- E2 single-child absorption: pass
- E3 component creation: known existing failure, unchanged
- E5 relation removal: pass

The latest change fixes unrelated relation churn in E5. Previously, the unchanged call:

Command.main → get_completion_class

was incorrectly re-filed during the incremental update, changing the aggregate Command Execution & Context Management → Command Composition & Discovery relation from 6 to 7 backing edges.

After 8614ded, the relation remains unchanged at 6 backing edges, and backing-edge changes without a code cause dropped from 3 to 0. The intended getchar → get_best_encoding relation is still removed correctly.

### DIFF
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -1196,6 +1196,8 @@ class DiagramGenerator:
         self,
         root_analysis: AnalysisInsights,
         sub_analyses: dict[str, AnalysisInsights],
+        *,
+        absorb: bool = True,
     ) -> None:
         """Prepare an analysis tree for its authoritative save.
 
@@ -1203,16 +1205,26 @@ class DiagramGenerator:
         flows. All steps are idempotent and
         safe with an empty ``sub_analyses`` (rebuild is a root-only pass).
 
-        Absorption runs first so the relation rebuild sees the corrected tree, and
+        Absorption runs early so the relation rebuild sees the corrected tree, and
         so a legacy baseline carrying the defect is repaired rather than rejected.
+        ``absorb`` is False for a save that will not persist the static-analysis
+        artifact: the collapse rewrites the tree AND the cluster lineage, and a save
+        that writes only one of the two leaves the next run seeding a collapsed tree
+        from the partition it had before. The next full or incremental save repairs it.
         """
-        drop_dangling_key_entities(root_analysis, sub_analyses)
-        cfgs = (
-            [self.static_analysis.get_cfg(lang) for lang in self.static_analysis.get_languages()]
-            if self.static_analysis
-            else []
-        )
-        self._rebase_baseline(absorb_single_child_components(root_analysis, sub_analyses, cfgs))
+        drop_dangling_key_entities(root_analysis, sub_analyses, self.repo_location)
+        # Before the collapse, not after. Absorbing a childless only child deletes its
+        # scope, and the deletion is only lossless because the parent already owns
+        # everything it owned. On a tree where that does not hold, absorbing first would
+        # discard the child's extra methods and leave nothing for the check to find.
+        assert_scope_containment(root_analysis, sub_analyses)
+        if absorb:
+            cfgs = (
+                [self.static_analysis.get_cfg(lang) for lang in self.static_analysis.get_languages()]
+                if self.static_analysis
+                else []
+            )
+            self._rebase_baseline(absorb_single_child_components(root_analysis, sub_analyses, cfgs))
         self.rebuild_global_relations(root_analysis, sub_analyses)
         self._strip_ignored(root_analysis, sub_analyses)
         assert_scope_containment(root_analysis, sub_analyses)
@@ -1266,7 +1278,7 @@ class DiagramGenerator:
         rewriting those would drop the ``static_analysis.sha`` tag (cold-starting
         the next incremental) and desync the sidecar from ``source_tree_hash``.
         """
-        self.finalize_for_save(root_analysis, sub_analyses)
+        self.finalize_for_save(root_analysis, sub_analyses, absorb=persist_side_artifacts)
         if persist_side_artifacts:
             source_tree_hash = self._source_tree_hash()
         else:
@@ -1570,7 +1582,7 @@ class DiagramGenerator:
             # component emptied by a route that never entered `refresh_ids` — and whose entities
             # `fix_key_entities_refs` therefore never re-resolved — would otherwise be kept alive
             # by pointers into symbols the commit deleted.
-            drop_dangling_key_entities(root_analysis, sub_analyses)
+            drop_dangling_key_entities(root_analysis, sub_analyses, self.repo_location)
             removed_ids = prune_empty_components(root_analysis, sub_analyses, protected_empty_ids)
             if removed_ids:
                 apply_result.refresh_ids -= removed_ids
--- a/diagram_analysis/diagram_generator.py
+++ b/diagram_analysis/diagram_generator.py
@@ -396,7 +396,8 @@ def _restore_unchanged_metadata(
             # on "owns a changed file" re-authored them for every component that merely shares a
             # file with the edit: measured on `referenced-symbol-deleted`, 11 components changed
             # cluster ids and 3 changed key entities for a commit that deleted one function.
-            component.source_cluster_ids = list(meta.source_cluster_ids)
+            if final_keys or not meta.member_keys:
+                component.source_cluster_ids = list(meta.source_cluster_ids)
             # An entity survives unless its symbol is GONE. Editing a method's body does not
             # make the class it lives in a worse choice of key entity — dropping on "changed"
             # rather than "deleted" is why `File` and `LazyFile` vanished from three components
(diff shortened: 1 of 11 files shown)


diagram_analysis/diagram_generator.py
```diff
@@ -1125,17 +1131,67 @@ def finalize_for_save(
         self,
         root_analysis: AnalysisInsights,
         sub_analyses: dict[str, AnalysisInsights],
+        *,
+        absorb: bool = True,
```
ivanmilevtues: is this actually used as false somewhere + why do we use the * so muych

diagram_analysis/tree_shape.py
```diff
@@ -0,0 +1,243 @@
...
+
+logger = logging.getLogger(__name__)
+
+#: ``sub_analyses`` is keyed by component id and the root analysis has no key.
+#: Empty string, not ``ROOT_SCOPE_ID``: this is the *prefix* sense of the root,
+#: the one ``agents.cluster_ids.prefix_for_scope`` returns for it.
+ROOT_SCOPE = ""
+
+
+def drop_dangling_key_entities(
```
ivanmilevtues: This has to be part of validation not a post cleanup I think.

diagram_analysis/tree_shape.py
```diff
@@ -0,0 +1,243 @@
...
+    root_analysis: AnalysisInsights,
+    sub_analyses: dict[str, AnalysisInsights],
+    repo_dir: Path,
+) -> list[str]:
+    """Drop key entities naming a symbol no component owns; return what went.
```
ivanmilevtues: This comments here are 2 chapters of a book like wut

diagram_analysis/tree_shape.py
```diff
@@ -0,0 +1,243 @@
...
+    owned = {
+        (group.file_path, method.qualified_name)
+        for scope in [root_analysis, *sub_analyses.values()]
+        for component in scope.components
+        for group in component.file_methods
+        for method in group.methods
+    }
+    owned_names = {qualified_name for _path, qualified_name in owned}
+
+    def is_live(entity: SourceCodeReference) -> bool:
```
ivanmilevtues: hate annonymous methods please make into a priavte one or as a method to SourceCodeReference.

diagram_analysis/diagram_generator.py
```diff
@@ -1125,17 +1131,67 @@ def finalize_for_save(
...
+        ``absorb`` is False for a save that will not persist the static-analysis
+        artifact: the collapse rewrites the tree AND the cluster lineage, and a save
+        that writes only one of the two leaves the next run seeding a collapsed tree
+        from the partition it had before. The next full or incremental save repairs it.
         """
+        drop_dangling_key_entities(root_analysis, sub_analyses, self.repo_location)
+        # Before the collapse, not after. Absorbing a childless only child deletes its
+        # scope, and the deletion is only lossless because the parent already owns
+        # everything it owned. On a tree where that does not hold, absorbing first would
+        # discard the child's extra methods and leave nothing for the check to find.
```
ivanmilevtues: yea this sounds like an althogether error, a child shouldn't own more than its parent

ivanmilevtues: This seems like patchwork all toghether double check and add assert run evals and see if it fails or works well.

diagram_analysis/diagram_generator.py
```diff
@@ -1125,17 +1131,67 @@ def finalize_for_save(
...
+            }
+            rebased: dict[tuple[str, str], Relation] = {}
+            for (src_id, dst_id), relation in self._baseline_global_relations.items():
+                src, dst = moved(src_id), moved(dst_id)
+                # Mirrors _reroot_relations: a pair that collapses onto itself is not an edge.
+                if src and dst and src != dst:
+                    # The endpoints move with the key. `preserve_unchanged_relations` restores this
+                    # object itself for a pair the rebuild dropped, so a relation left carrying the
+                    # absorbed child's id would be saved pointing at a component that is gone.
+                    rebased.setdefault((src, dst), relation.model_copy(update={"src_id": src, "dst_id": dst}))
```
ivanmilevtues: This seems legit, we should double check with this

ivanmilevtues: We should use ids there as well

diagram_analysis/diagram_generator.py
```diff
@@ -1157,7 +1213,7 @@ def finalize_and_save(
         rewriting those would drop the ``static_analysis.sha`` tag (cold-starting
         the next incremental) and desync the sidecar from ``source_tree_hash``.
         """
-        self.finalize_for_save(root_analysis, sub_analyses)
+        self.finalize_for_save(root_analysis, sub_analyses, absorb=persist_side_artifacts)
```
ivanmilevtues: yea I think we should drop the variabnle and always absorb. In theory you shound't be able to expand if there are no underlying clusters.

diagram_analysis/diagram_generator.py
```diff
@@ -1331,6 +1387,11 @@ def generate_analysis_incremental(
         # membership baseline (captured post-scrub below) so the restore passes never re-inject a
         # deleted method. Also drives the empty-delta path, which returns before that capture.
         self._baseline_member_keys = _capture_baseline_member_keys(root_analysis, sub_analyses)
+        # Also before the scrub. Why: this shields an empty-but-cluster-backed component from the
+        # prune, and `remove_deleted_files` clears methods and key entities but not
+        # `source_cluster_ids` — so asked afterwards it shields the components the scrub itself
+        # just emptied, and a deleted unit keeps its box.
+        protected_empty_ids = _cluster_backed_empty_component_ids(root_analysis, sub_analyses)
```
ivanmilevtues: why would a cluster backed component be empty? We are doing the cleaning to early I suppose then

ivanmilevtues (changes requested):
It is not ready for merge, I will take a look tmrw. There are few comments that make a lot of sense.

I think my main concern is that this is lot's of patchwork especially around this empty component but cluster backed thing, That seems super scetchy.

ivanmilevtues (approved):
Great

PR #445, opened 2026-08-01
Mark automated traffic as internal on every telemetry event

Svilen-Stefanov:
## What

`source` has always said which invoker a run came from. What it could not say is whether that invoker is **a person using the product**, so a dashboard measuring usage had to spell out `source not in ('tests', 'evals')` — correct until the next internal source is added, and silently wrong from that moment on. That is how automated traffic comes to be counted as usage.

Every event now also carries `internal`, derived from `source` in one place, so a product metric filters on one condition and stays correct when another internal source appears.

Automated runs still emit — an eval run is real signal about analysis quality and dropping it would lose that. They just must never be counted as usage.

## Also fixes an ordering hazard the change surfaced

The origin used to merge **first**:

```python
properties={"source": source, **(properties or {})}
```

which left an event model free to overwrite `source` just by having a field of that name. No schema does today — but `internal` is the property the metrics filter on, so a payload that could overwrite it is a payload that could hide automated traffic inside usage. It now merges last, in both `capture` and `capture_exception`.

## Every source is accounted for

| source | emitted by | internal |
|---|---|---|
| `oss` | `main.py` | no |
| `vscode` | the extension | no |
| `github_action` | `github_action.py` | no |
| `core` | any other embedding (default) | no |
| `tests` | `tests/conftest.py` | **yes** |
| `evals` | the benchmark harness's `_entry.py` | **yes** |

`core` stays product traffic on purpose: defaulting the other way would quietly drop a real integration's runs.

## Verification

Six tests in `test_telemetry_service.py` — a product source, every internal source, an undeclared source, an exception carrying the origin, and a caller property failing to overwrite it. They exercise `ProductTelemetry` directly because `test_telemetry_events.py` stubs `capture` out and the stamping happens below that stub.

1936 tests pass; mypy and black clean. `TELEMETRY.md` documents both properties.

## Summary by CodeRabbit

* **Documentation**
  * Documented telemetry source categories, including product, development, testing, and evaluation environments.
  * Clarified how automated test and evaluation activity is identified in telemetry data.

* **Bug Fixes**
  * Standardized origin information across product and exception telemetry events.
  * Prevented event metadata from being incorrectly overridden by supplied values.
  * Improved classification of internal activity to support more accurate usage reporting.

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -53,8 +53,17 @@ never personal data.
 
 Every event also carries:
 
-- `source` — `"oss"` for the open-source CLI, `"vscode"` when invoked by the
-  extension, or `"core"` for other embeddings.
+- `source` — who invoked the run: `"oss"` for the open-source CLI, `"vscode"`
+  when invoked by the extension, `"github_action"` in CI, `"tests"` for
+  CodeBoarding's own test suite, `"evals"` for the benchmark harness, or
+  `"core"` for other embeddings.
+- `internal` — `true` when that source is **not** a person using the product
+  (`tests`, `evals`). Automated runs still emit, because an eval run is real
+  signal about analysis quality, but they must never be counted as usage. It is
+  derived from `source` in one place (`telemetry/service.py`) so a product
+  metric filters on one condition and stays correct when another internal
+  source is added — a hand-kept list of source values is right until it is
+  quietly not.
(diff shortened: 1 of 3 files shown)


ivanmilevtues (approved):
Thanks for adding this, it shuld reduce noise in telemetry

PR #448, opened 2026-08-03
Bound language-server memory so large repos stop being OOM-killed

Svilen-Stefanov:
## The problem

The [`CodeBoarding/server` review job](https://github.com/CodeBoarding/server/actions/runs/30814553550/job/91688982643) exits 143 — SIGTERM, the runner OOM-killed after 20 minutes. Target is bitwarden/server: 3,926 analyzed `.cs` files, 69MB of source, one `.slnx` over 70 projects.

Answering `textDocument/references` across a solution that size makes Roslyn materialize a compilation per project and keep every one. csharp-ls grows **linearly in positions queried, with no ceiling**, measured over an identical replay:

```
pos 1000/7500: rss=2.7GB      pos 5000/7500: rss=9.4GB
pos 2000/7500: rss=3.5GB      pos 6000/7500: rss=10.6GB
pos 3000/7500: rss=6.4GB      pos 7000/7500: rss=11.3GB
pos 4000/7500: rss=7.0GB      FINAL rss=11.7GB  (growth 10.9GB)
```

The real run queries 15,229 positions. A 16GB runner dies about halfway. C# is the only language in this shape — Java already caps `-Xmx` by file count and 50% of RAM.

## The fix

`LSPRecycler` restarts the server when its process tree crosses a budget (40% of RAM; `CODEBOARDING_LSP_MEMORY_BUDGET_MB` overrides). A restart returns it to *exactly* its post-load size:

```
[pre-restart] 40 batches, rss=6.0GB
RESTART: 21s, rss 6.0GB -> 324.7MB (reclaimed 5.6GB, load baseline was 325.2MB)
```

Gated on the new `LanguageAdapter.workspace_owns_documents`, true only for C#: csharp-ls answers position queries for files we never `didOpen`, so a fresh process is equivalent to the one it replaced. Servers that only know documents we pushed must not be recycled — they would come back empty.

Three details that came out of measurement, not design:

- **Check before every batch, not every N.** One batch of solution-wide queries can allocate gigabytes; a 10-batch interval overshot the budget by 6GB.
- **Shrink the batch above half the budget.** A batch is one round-trip we cannot interrupt, so its own growth is the floor under any bound.
- **Disarm if a bare reload already exceeds the budget.** Otherwise: restart → still over → restart, forever, on exactly the huge repos this protects.
- **Warm the reference index after each restart.** A loaded workspace is not a warm index; without this the first real batch burns its deadline building one (timeout warnings tripled).

## Also fixed

| | before | after |
|---|---|---|
| `SourceInspector` caches (69MB repo) | 3.4GB retained | **1.2GB**, no wall-clock cost |
| `should_ignore` | 178µs × 100k+ calls | **13.3µs** (13x) |
| LSP for a language with no files | started, held all run (~350MB) | skipped |
| Memory visibility in CI | none | per-phase + 60s trend |

`SourceInspector` kept four unbounded per-file caches. The parse tree is needed exactly once per file — to build that file's usage index, which is cached separately and outlives it — so it is now an LRU bounded by a node budget, and raw bytes are not retained at all.

## Result

Bitwarden with the budget set to a 16GB-runner equivalent: **peak 7.6GB, held**, ~20% faster than baseline at the same position, and it runs straight through the region where CI died (same `Timeout waiting for references request` warnings).

## Verification

Private integration suite — 8 real repos, pinned reference/node/edge counts:

```
codeboarding_python  PASSED    wordpress_php      PASSED
mockito_java         PASSED    lodash_javascript  PASSED
prometheus_go        PASSED    clap_rust          FAILED
excalidraw_typescript PASSED   serilog_csharp     PASSED
```

`serilog_csharp` is a C# repo small enough never to recycle and it matches exactly — that is the guarantee: **under budget, results are unchanged**. `clap_rust` fails only because the host has no `cargo`, which is the pre-existing fail-fast path.

864 unit tests, mypy and black clean.

## Trade-off worth knowing

csharp-ls is deterministic *within* a process but not across processes: the same 2,000 queries run twice in one process are identical; after a restart 6 differ (+0.03% references, always *more*). So over budget the choice is a 0.03% result difference versus no result at all. Under budget nothing restarts.

## Not done, deliberately

`didOpen` → `didClose` per file (~1.1GB, but the first recycle already drops every overlay), packing the usage index into ints (~0.3GB, obfuscates the code), and lazy per-language LSP startup (breaks the VSCode live path). Phase 2 is 91% LSP wait and 9% Python, so `find_containing_symbol`'s linear scan was left alone.

Full writeup with every claim linked to the log that produced it, plus seven re-runnable probe scripts, lives in the workspace at `memory-investigation/`.

## Summary by CodeRabbit

* **Performance**
  * Improved analysis speed by caching ignore checks and parsed source information.
  * Added memory-aware processing to reduce batch sizes and restart language services when necessary.
* **Reliability**
  * Language services recover more gracefully from memory pressure and transient startup issues.
  * Projects without analyzable source files complete successfully without unnecessary services.
* **Bug Fixes**
  * Improved consistency when evaluating relative and absolute ignored paths.
  * Preserved analysis results when cached syntax data is evicted.
  * Improved handling of language-service document queries and startup failures.


Svilen-Stefanov:
## End-to-end validation: the full run completed

Bitwarden/server, budget set to what a 16GB runner derives (6.2GB):

```
Phase 2 (edges): recycled 26 time(s); 1539 batches ran shrunk under memory pressure
Phase 2 total (build edges): 5581.6s, 48041 edges
Phase 3 (hierarchy): 98.0s
Phase 4 (package deps): 1.2s
Pipeline complete: 3926 files, 35979 symbols, 48041 edges, 4805 hierarchy entries
mem[CSharp] PEAK total=10.6GB (python=1.3GB, children=9.3GB) | top: csharp-ls=9.1GB
Converted 15246 nodes, 48041 edges (0 skipped) for CSharp
TOTAL wall time: 5849.7s, peak tree RSS: 10.6GB
```

**97 minutes, 10.6GB peak, complete result.** The same repo previously hit 11.9GB at 45% of Phase 2 and was still climbing when the runner killed it.

## Correction to the PR description

I wrote "peak 7.6GB, held" above — that was a mid-run reading, not the final peak. **The real peak is 10.6GB.** It fits a 16GB runner and the run finishes, but with less margin than that number implied. Two contributors I under-weighted:

**The overshoot has a fat tail.** Most of the 26 recycles fired within 100MB of the budget, but the worst overshot by 2.9GB:

```
 11 x  LSP holding 6.3GB      2 x  LSP holding 7.0GB
  2 x  LSP holding 6.4GB      1 x  LSP holding 7.5GB
  1 x  LSP holding 6.5GB      1 x  LSP holding 7.8GB
  1 x  LSP holding 6.7GB      1 x  LSP holding 8.3GB
  2 x  LSP holding 7.1GB          (csharp-ls peaked at 9.1GB)
```

That tail is one un-interruptible batch: some symbols carry ~24,000 references and a single query can allocate hundreds of megabytes, so a fixed `PRESSURED_BATCH_DIVISOR` of 10 is not aggressive enough at the ceiling.

**Python reached 1.3GB**, not the ~400MB visible earlier — it grows as more files get parsed, matching the isolated 1.2GB worst case.

So the practical sizing rule is **budget + ~3GB (LSP tail) + ~1.4GB (Python)**.

## Suggested follow-up (deliberately not in this PR)

Size the batch in proportion to remaining headroom rather than a fixed divisor, so it reaches 1 as the budget is approached. That should cut the tail substantially. I left it out because the 97-minute run above validates the divisor version and re-validating costs another 97 minutes — happy to do it as a follow-up if you'd rather have the tighter bound before merge.

ivanmilevtues:
LGTM, tests pass. Few simplification in the code merging.

PR #451, opened 2026-08-05
Fix C# workspace readiness timeouts with explicit solution loading

ivanmilevtues:
## Summary

Fix C# analyses that stall in the initial `textDocument/documentSymbol` workspace-readiness probe by:

- preserving the exact discovered `.sln`, `.slnx`, or `.csproj` path through engine configuration;
- restoring that exact target and launching `csharp-ls --solution <path>` for solutions;
- upgrading the bundled `csharp-ls` from 0.24.0 to 0.25.0 for its faster, parallelized solution target-framework discovery;
- capturing and logging LSP stderr instead of discarding workspace-load diagnostics;
- including the timed-out method, elapsed limit, process state, PID, and RSS in timeout errors; and
- propagating C# analysis failures as fatal errors instead of continuing to a misleading downstream "No component groups found" result.

## Expected root cause

The affected run completes `dotnet restore` and the LSP initialization handshake, then blocks for 1,800 seconds on request 2, `textDocument/documentSymbol`. In `csharp-ls`, that request waits for Roslyn's solution workspace to finish loading. Therefore the reported C# LSP startup duration measures process startup and initialization—not workspace readiness.

The expected contributors are:

1. CodeBoarding 0.24.0 discarded the exact discovered solution filename and launched plain `csharp-ls`, requiring the server to search for and choose among visible solutions.
2. csharp-ls 0.24.0 performs serial target-framework probing before `MSBuildWorkspace.OpenSolutionAsync`; this is expensive for large multi-project solutions. Version 0.25.0 parallelizes the remaining probes and reads target frameworks directly.
3. LSP stderr was discarded, preventing the timeout from being localized to framework probing, MSBuild workspace loading, a build host, or a specific project.
4. The C# exception was treated as a recoverable per-language error, allowing later architecture generation to replace the actionable timeout with an unrelated empty-component error.

This PR addresses those conditions, but does not claim that csharp-ls 0.25.0 alone guarantees every large solution—including Rapidata—will load within the 1,800-second cap. Rapidata must be rerun to validate that outcome.

## Preconditions: behavior before this change

- The scanner retained only the directory containing a solution, not the exact solution path.
- C# restore independently selected the first matching solution/project in that directory.
- The server command was plain `csharp-ls`; solution selection was implicit.
- csharp-ls 0.24.0 could initialize quickly while Roslyn continued loading the workspace asynchronously.
- The first `documentSymbol` request could consume the full readiness timeout before symbol discovery began.
- Server stderr was sent to `/dev/null`.
- Timeout errors contained only a JSON-RPC request number.
- A C# timeout could be swallowed, after which small sibling languages were analyzed and downstream code reported "No component groups found".
- The LSP recycler and reference-memory optimization were never reached because they are constructed after symbol discovery.

## Postconditions: behavior after this change

- The exact scanner-selected `.sln`, `.slnx`, or `.csproj` is preserved in `EngineConfig`.
- `dotnet restore` uses the same exact target selected for analysis.
- Solution launches use `csharp-ls --solution <exact-path>` and log the selected path.
- New installs use csharp-ls 0.25.0.
- LSP stderr is continuously drained and included in normal CodeBoarding logs.
- A request timeout identifies the LSP method, configured duration, process state, PID, and RSS.
- A C# workspace/symbol failure terminates analysis with the original error context; it cannot silently become a downstream empty-architecture result.
- Other languages retain their existing graceful-degradation behavior.

## Public-repository reproduction

Reproduced against [`space-wizards/space-station-14`](https://github.com/space-wizards/space-station-14) at commit `f95b1461ad406ae6aff4305d88211ad744caf7d7` with recursive submodules:

- 11,300 C# files;
- 100 project files;
- 17 solution files;
- root `SpaceStation14.slnx` containing 50 project entries; and
- successful restore in 12.67 seconds.

Pre-fix csharp-ls 0.24.0, launched without `--solution`:

- initialized in 0.827 seconds;
- discovered 17 solutions;
- did not answer the initial pre-open `documentSymbol` probe within a deliberately shortened 5-second timeout;
- reached no bulk `didOpen`, symbol traversal, reference warmup, recycler, or references phase;
- eventually answered the same request after 33.204 seconds when allowed to continue; and
- used approximately 921 MiB RSS at completion.

With csharp-ls 0.25.0 and `--solution SpaceStation14.slnx`:

- workspace load completed in 20.073 seconds versus 23.746 seconds;
- the same `documentSymbol` response completed in 29.204 seconds versus 33.204 seconds.

The five-second failure is a controlled reproduction of the mechanism, not a claim that this fixture naturally exceeds the production 1,800-second timeout. The comparison is also sequential and may benefit from warm caches.

Upstream context: [csharp-language-server issue #242](https://github.com/razzmatazz/csharp-language-server/issues/242) and [csharp-ls 0.25.0 release](https://github.com/razzmatazz/csharp-language-server/releases/tag/0.25.0).

## Verification

- `pytest -q tests/static_analyzer/test_csharp_config_scanner.py tests/static_analyzer/test_csharp_adapter.py tests/static_analyzer/test_static_analyzer_start_clients.py tests/static_analyzer/test_lsp_client.py tests/test_tool_registry.py` — 262 passed.
- Final affected-module rerun — 156 passed.
- `mypy .` — no issues in 312 source files.
- Black checks and `git diff --check` passed.
- Full coverage run — 2,017 passed, 28 skipped, 83.05% coverage. One unrelated environment-dependent tool-registry test failed because the test orb had no system `dotnet`; the focused tool-registry suite passed.

## Summary by CodeRabbit

- **New Features**
  - Improved C# project detection and analysis for solutions and project files, including explicit solution selection.
  - Updated C# language-server support to version 0.25.0.

- **Bug Fixes**
  - Improved diagnostics for language-server timeouts by including process status and memory details.
  - C# analysis failures now correctly stop analysis when configured to treat missing results as fatal.
  - Improved handling of language-server error output.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -42,6 +42,7 @@ class EngineConfig:
     adapter: LanguageAdapter
     project_path: Path
     source_files: list[Path] = field(default_factory=list)
+    project_file: Path | None = None
 
 
 class StaticAnalysisFatalError(RuntimeError):
@@ -135,7 +136,9 @@ def _create_engine_configs(
                             f"Creating engine config for CSharp ({csharp_config.project_type}) at: "
                             f"{csharp_config.root.relative_to(repository_path)}"
                         )
-                        configs.append(EngineConfig(adapter, csharp_config.root))
+                        configs.append(
+                            EngineConfig(adapter, csharp_config.root, project_file=csharp_config.project_file)
+                        )
                 else:
                     logger.info("No C# projects detected")
 
@@ -250,8 +253,8 @@ class StaticAnalyzer:
                 t_start = time.monotonic()
                 # Allow adapters to prepare the project before LSP startup
                 # (e.g. ``dotnet restore`` so csharp-ls sees framework refs).
-                adapter.prepare_project(project_path)
-                command = adapter.get_lsp_command(project_path)
+                adapter.prepare_project(project_path, engine_config.project_file)
+                command = adapter.get_lsp_command(project_path, engine_config.project_file)
                 init_options = adapter.get_lsp_init_options(self.ignore_manager)
                 extra_env = adapter.get_lsp_env(project_path)
                 # Node-based LSPs spawn child ``node`` processes by name; on
@@ -620,6 +623,8 @@ class StaticAnalyzer:
             except StaticAnalysisFatalError:
                 raise
             except Exception as e:
+                if adapter.fail_on_empty_symbols:
+                    raise StaticAnalysisFatalError(f"{adapter.language} analysis failed: {e}") from e
                 logger.error(f"Error during engine analysis for {adapter.language}: {e}")
                 track_lsp_result(
                     language=adapter.language_enum.value,
--- a/static_analyzer/csharp_config_scanner.py
+++ b/static_analyzer/csharp_config_scanner.py
@@ -21,12 +21,17 @@ class CSharpProjectConfig:
         self,
         root: Path,
         project_type: str,  # "solution", "project", or "none"
+        project_file: Path | None = None,
     ):
         self.root = root
         self.project_type = project_type
+        self.project_file = project_file
 
     def __repr__(self) -> str:
-        return f"CSharpProjectConfig(root={self.root}, project_type={self.project_type})"
+        return (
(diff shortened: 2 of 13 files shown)


ivanmilevtues:
irrelevant//

PR #452, opened 2026-08-05
Investigation: csharp-ls hangs forever on multi-targeted C# monorepos (customer C# analysis failure)

ivanmilevtues:
## What

Documents the root cause of the customer C# analysis failure (`StaticAnalysisFatalError: No component groups found` after `Timeout waiting for LSP response to request 2`) and adds a self-contained reproduction harness.

- `docs/development/csharp-ls-multitarget-hang.md` — full investigation write-up: failure chain, csharp-ls code-level mechanism, measurements, upstream status, fix option space
- `tests/repro/csharp_ls_multitarget_hang/gen_repo.py` — synthetic C# monorepo generator (`--multitarget` is the hang trigger)
- `tests/repro/csharp_ls_multitarget_hang/lsp_probe.py` — minimal LSP driver that surfaces csharp-ls's load narration and times the exact "request 2" from production logs

## Root cause (short version)

csharp-ls (our pinned 0.24.0, and still current main) computes a workspace target framework via `compatibleTfmSet`, which folds per-project TFM sets as **lazy nested sequences with no per-step dedup**. Each project using `<TargetFrameworks>` (plural) roughly doubles the enumeration forced by the final `Set.ofSeq` — O(2^N). At ~32+ multi-targeted projects the solution load exceeds our 1800s probe cap; at the customer's scale it would take years. The C# result comes back empty, only trivial JS/TS nodes remain, and the pipeline fail-fasts.

## Reproductions

| Repro | Result |
|---|---|
| Synthetic, 50 single-target projects | loads in 16.5s |
| Synthetic, 100 single-target projects | 27s (linear) |
| Synthetic, 25 multi-target projects | 33s |
| Synthetic, 50 multi-target projects | **hang, killed at 850s** |
| **OSS: open-telemetry/opentelemetry-dotnet** (`OpenTelemetry.Extended.slnx`, 78 projects / 44 multi-targeted) | **hang, no response in 420s** |

Quick check for whether a given repo will hit this: `grep -rl "<TargetFrameworks>" --include="*.csproj" | wc -l` ≥ ~30.

## Non-goals

No fix here — per team agreement we reproduce first and decide on fixes separately. The doc lists the option space (upstream one-line dedup fix; surfacing csharp-ls narration in our logs; stall detection instead of a silent 1800s wait).

Draft until we agree on next steps.

### DIFF
--- a/docs/development/csharp-ls-multitarget-hang.md
+++ b/docs/development/csharp-ls-multitarget-hang.md
@@ -1,0 +1,166 @@
+# Investigation: csharp-ls hangs forever on multi-targeted C# monorepos
+
+**Date:** 2026-08-05
+**Status:** Root cause identified and reproduced (synthetic + OSS). Fix not yet decided — this document records the evidence.
+
+## TL;DR
+
+A customer's C# analysis fails with `StaticAnalysisFatalError: No component groups found` because
+**csharp-ls (0.24.0, our pinned C# language server) never finishes loading their solution**. The hang
+is an exponential-time target-framework computation in csharp-ls that is triggered by projects using
+`<TargetFrameworks>` (plural, i.e. multi-targeting). With ~30 or more multi-targeted projects in one
+solution the load takes hours-to-years, so CodeBoarding's sync probe times out after 1800s, the C#
+result comes back empty, and the pipeline (correctly, per our fail-fast policy) refuses to build an
+architecture out of nothing.
+
+The bug is in csharp-ls, was introduced around 0.21, and is **still present in csharp-ls main** as of
+August 2026 — bumping our pin does not fix it.
+
+## Customer failure signature
+
+From the customer's GitHub Actions log (a private .NET monorepo: 7,592 C# files, `.slnx` solution,
+microservices layout). Everything before static analysis is healthy — `dotnet restore` succeeds and
+csharp-ls answers `initialize`:
+
+```
+18:03:07 INFO  [static_analyzer:274] CSharp LSP start: 18.3s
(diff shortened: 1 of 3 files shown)


ivanmilevtues:
Fix is up in #453, scoped to the hang itself.

The mechanism this doc identified holds up. Two things worth adding to the record:

**The fold is in every release, not just our pin.** I diffed `Roslyn/Solution.fs` across tags — `Seq.fold compatibleTfmsOfTwoSets firstSet |> Set.ofSeq` is byte-identical in 0.24.0, 0.25.0, 0.26.0 and current `main`. So the pin bump in #451 does not address this one.

**The fix does not need upstream.** Pointing MSBuild's `CustomAfterMicrosoftCommonTargets` and `CustomAfterMicrosoftCommonCrossTargetingTargets` at a generated targets file that leaves each project on one framework makes the fold linear, because the blowup is driven purely by how many projects contribute a multi-element set. The SDK imports that file after the project body, so it covers the `Directory.Build.props` variant from the 2026-08-05 update as well — which matters, since that variant is invisible to a csproj grep.

Measured against this branch's repro scripts (csharp-ls 0.24.0, SDK 10.0.301, Linux):

| Repro | Before | After |
|---|---|---|
| 40 projects, `<TargetFrameworks>` in csproj | no load in 180s | 9.5s |
| 40 projects, `<TargetFrameworks>` in props only | no load in 150s | 9.7s |
| opentelemetry-dotnet @ `core-1.17.0` (36 of 70) | no load in 180s | 13.1s |

Both reproductions are now integration tests in CodeBoarding-tests#5, so the doc's harness has a permanent home. Whether this PR still merges as documentation is your call — the write-up is the reason the fix was findable.

PR #456, opened 2026-08-06
fix(csharp): stop the workspace TargetFramework inverting project conditions

RapidPoseidon:
## WHAT

csharp-ls never answers the first `textDocument/documentSymbol` sync probe on a solution whose projects don't all target the same framework. The probe times out; left running past the deadline, csharp-ls dies of `Out of memory.`

The cause is a **self-referencing project graph that csharp-ls creates itself**, and it is independent of solution size. This fixes it by demoting the offending MSBuild property.

## WHY

`Roslyn/Solution.fs` folds every project's frameworks into one **global** MSBuild `TargetFramework` and passes it to `MSBuildWorkspace.Create`:

```
Roslyn.Solution: Will use MSBuild props: map [(TargetFramework, net10.0)]
```

A global property **outranks** a project's own `<TargetFramework>`. So in a solution of 281 `net10.0` projects plus 5 `netstandard2.0` Roslyn analyzer/generator projects, those 5 are evaluated as `net10.0`.

That inverts conditions keyed on `$(TargetFramework)`. The idiomatic guard that stops analyzer projects from referencing themselves —

```xml
<ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
  <ProjectReference Include="...Analyzers.csproj" OutputItemType="Analyzer" ... />
```

— flips for exactly the `netstandard2.0` projects it exists to exclude. Measured on the affected repo with `dotnet msbuild -getItem:ProjectReference`:

| `Rapidata.Generators.csproj` (declares `netstandard2.0`) | ProjectReferences |
|---|---|
| as declared | **1** (its one real sibling) |
| under `-p:TargetFramework=net10.0` | **4, including itself** |

`OpenSolutionAsync` with `LoadMetadataForReferencedProjects = true` then chases that self-reference and the resulting sibling cycles, and grows without bound.

### Why #448 and #454 don't cover this

- **#454** injects a `.targets` override guarded by `Condition="'$(TargetFrameworks)' != ''"`. The affected repo has **zero** projects using `<TargetFrameworks>` (plural) — all 286 are single-target. The guard cannot fire. The damage here happens *because* projects are single-target: the fold has nothing to disambiguate, yet its result is still applied as a global property.
- **#448** targets the right subsystem but the wrong phase. This blows up during **solution load**, before any `references` query, so `LSPRecycler` never gets a turn. Bounding the heap would make the OOM arrive *sooner*, not later.
- **Raising the probe timeout is not the fix.** The load never converges; a higher cap converts a timeout into an OOM crash.

## HOW

`TreatAsLocalProperty="TargetFramework"` on a `Directory.Build.props` injected via `DirectoryBuildPropsPath` — the props analogue of the `DirectoryBuildTargetsPath` mechanism #454 already uses. That demotes the global property so each project sees its declared framework again. The injected props chains the repo's own `Directory.Build.props`, and honours a pre-existing `DirectoryBuildPropsPath` the same way #454 does for targets.

Multi-target projects still need a framework picked for them, so the targets file reinstates the workspace value when a project declares only `TargetFrameworks`.

**One deliberate behaviour change worth your attention:** the *fold* of a multi-target project down to one framework stays behind `_MULTI_TARGET_PROJECT_THRESHOLD` (now gated on a `CodeBoardingFoldMultiTargetFrameworks` env flag, so #454's behaviour is unchanged for repos above and below the threshold). But the **injection itself is now unconditional**, because the bug does not depend on project count — a 24-project reproduction hangs under the old `> 25` gate. Demoting the property is a no-op unless the workspace value contradicts a declared framework, i.e. it only takes effect in the broken case. This is why `test_project_env_leaves_small_workspaces_unchanged` became `test_project_env_leaves_small_workspaces_multi_targeting`.

## HOW TESTED

Reproduction added at `tests/static_analyzer/scripts/csharp_mixed_framework_repro.py` — it generates a synthetic solution with this shape, so **you don't need the affected private repo**:

```bash
uv run python tests/static_analyzer/scripts/csharp_mixed_framework_repro.py /tmp/repro --projects 40
cd /tmp/repro && dotnet restore Synthetic.slnx
# then point CodeBoarding at /tmp/repro
```

`--variant name-guard` keys the same guard on `$(MSBuildProjectName)`, which a global framework cannot invert. It is the control that isolates the cause to the property rather than to solution size:

| Synthetic solution, 44 projects, single variable changed | Result |
|---|---|
| guard on `$(TargetFramework)` (idiomatic) | never finishes — killed at 540s, 2.4GB still climbing |
| guard on `$(MSBuildProjectName)` — **control** | ✅ loads in **33.7s** |

Full CodeBoarding static analysis on the synthetic repro, with this change:

| | Before | After |
|---|---|---|
| 44 projects / 324 files | probe never returns | ✅ **52.6s** total, probe 27.8s, 1664 symbols |
| 24 projects / 164 files (below old threshold) | stuck >7 min in "Waiting for LSP server indexing" | ✅ **32.3s** total, probe 18.4s, 844 symbols |

On the real 286-project / 7585-file solution that prompted this, driving csharp-ls 0.24.0 over stdio with the same handshake the engine uses (`probe_before_open`), using the exact artifacts this code generates:

| | Before | After |
|---|---|---|
| `documentSymbol` probe (request 2) | never answers | ✅ **`PROBE OK in 131.5s`** |
| RSS | 774MB@90s → 7.3GB@754s, still climbing | peak **1.39GB**, bounded |
| left running past the cap | `Out of memory.`, server dead at ~43 min | loads, `Finished loading solution` at 128.7s |

Also verified no #454 regression: a `TargetFrameworks=netstandard2.0;net8.0;net10.0` project still resolves to a single `net10.0`, and a normal `net10.0` project still receives all four of its analyzer `ProjectReference`s.

`uv run pytest` — 2019 passed, 28 skipped. `uv run black . --check` and `uv run mypy` clean. Two unit tests added for the new env contract.

## Note for upstream

The root cause is arguably csharp-ls's: `resolveDefaultWorkspaceProps` emits a global `TargetFramework` even when no project multi-targets, so there is nothing to disambiguate. A fix there would make this workaround unnecessary, but you pin `csharp-ls 0.24.0`, so this keeps large mixed-framework solutions analyzable today. Happy to file that upstream too if useful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved C# and F# workspace loading for projects targeting multiple frameworks.
  * Large workspaces now receive optimized multi-targeting configuration, while smaller workspaces retain their existing behavior.
  * Preserved custom build configuration paths and environment settings during analysis.
  * Improved handling of workspace-level framework settings to provide more reliable project discovery and analysis.
* **Tests**
  * Added coverage for mixed-framework solutions and large-workspace configurations.

### DIFF
--- a/static_analyzer/engine/adapters/csharp_adapter.py
+++ b/static_analyzer/engine/adapters/csharp_adapter.py
@@ -36,7 +36,7 @@ _SINGLE_TARGET_FRAMEWORK_TARGETS = r"""<Project>
   <Import
     Project="$(CodeBoardingDirectoryBuildTargetsPath)"
     Condition="Exists('$(CodeBoardingDirectoryBuildTargetsPath)')" />
-  <PropertyGroup Condition="'$(TargetFrameworks)' != ''">
+  <PropertyGroup Condition="'$(TargetFrameworks)' != '' and '$(CodeBoardingFoldMultiTargetFrameworks)' == 'true'">
     <CodeBoardingTargetFrameworks>$([System.Text.RegularExpressions.Regex]::Replace('$(TargetFrameworks)', '\s', ''))</CodeBoardingTargetFrameworks>
     <CodeBoardingTargetFrameworks>$([System.Text.RegularExpressions.Regex]::Replace('$(CodeBoardingTargetFrameworks)', ';+', ';'))</CodeBoardingTargetFrameworks>
     <CodeBoardingTargetFrameworks>;$([System.String]::Copy('$(CodeBoardingTargetFrameworks)').Trim(';'));</CodeBoardingTargetFrameworks>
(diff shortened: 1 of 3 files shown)


ivanmilevtues:
#457 is based on this and now merged

PR #457, opened 2026-08-06
fix(csharp): handle mixed single- and multi-target workspaces

ivanmilevtues:
## Summary

- retain PR #456's `TreatAsLocalProperty` fix for workspace-wide `TargetFramework` overrides
- preserve Roslyn's normal outer/inner-build expansion for genuine multi-target projects
- keep the existing large-workspace framework fold separately gated

## Why

PR #456's targets fallback assigned the entire semicolon-delimited `TargetFrameworks` value to singular `TargetFramework` when no workspace framework was selected. That poisons normal multi-target evaluation and caused the C# integration regression. Removing only that fallback fixes the regression without weakening the analyzer self-reference fix.

This incorporates and corrects #456 because its source branch is in a contributor fork and cannot be updated from this repository branch.

## Validation

- focused C# adapter tests: 5 passed
- Black and mypy checks passed
- committed mixed-framework integration fixture: analyzer self-reference absent and normal multi-target outer build preserved
- fixture csharp-ls document-symbol probe passed
- existing real-repository C# document-symbol regression passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved C# project loading for large mixed-target-framework workspaces.
  * Preserved custom build configuration paths while applying workspace framework settings.
  * Stabilized generated build configuration updates to prevent incomplete files.

* **Tests**
  * Added coverage for small and large workspace framework behavior.
  * Added a utility for reproducing mixed-framework C# solution loading issues.

### DIFF
--- a/static_analyzer/engine/adapters/csharp_adapter.py
+++ b/static_analyzer/engine/adapters/csharp_adapter.py
@@ -44,10 +44,6 @@ _SINGLE_TARGET_FRAMEWORK_TARGETS = r"""<Project>
     <TargetFrameworks>$([System.String]::Copy('$(CodeBoardingTargetFrameworks)').Trim(';').Split(';')[0])</TargetFrameworks>
     <TargetFrameworks Condition="'$(CodeBoardingPreferredTargetFramework)' != '' and $(CodeBoardingTargetFrameworks.Contains(';$(CodeBoardingPreferredTargetFramework);'))">$(CodeBoardingPreferredTargetFramework)</TargetFrameworks>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == '' and '$(TargetFrameworks)' != ''">
-    <TargetFramework Condition="$(TargetFrameworks.Contains('$(CodeBoardingWorkspaceTargetFramework)'))">$(CodeBoardingWorkspaceTargetFramework)</TargetFramework>
-    <TargetFramework Condition="'$(TargetFramework)' == ''">$(TargetFrameworks)</TargetFramework>
-  </PropertyGroup>
 </Project>
 """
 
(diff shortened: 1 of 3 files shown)


ivanmilevtues (commented):
All tests are passing ran them locally because of github's incident

RapidPoseidon:
Follow-up data, because the obvious next inference is wrong and would cost you time.

Driving **csharp-ls 0.24.0 standalone — no CodeBoarding in the loop at all**, so none of this repo's injected props/targets apply, the unbounded growth still happens. That reads like "the leak is independent of the `TargetFramework` inversion, so it must be a csharp-ls scale bug." It isn't. The inversion is caused by csharp-ls's *own* global property, so taking CodeBoarding out of the loop leaves the bug fully intact — and it explains why the same 1800s timeout appeared on 0.13.4 and 0.13.5, which predate #454 entirely.

Controlled matrix, all standalone csharp-ls 0.24.0, all ~304 projects, all single-target, restored not built. Only the `Directory.Build.props` guard key and the injection differ:

| 304-project solution | Injection | Result | Peak RSS |
|---|---|---|---|
| homogeneous `net10.0`, no analyzer projects | none | ✅ probe 177.3s | 852MB |
| mixed TFM, guard on `$(MSBuildProjectName)` | none | ✅ probe 183.3s | 919MB |
| mixed TFM, guard on `$(TargetFramework)` | none | ❌ unbounded, 3.6GB @855s climbing | — |
| mixed TFM, guard on `$(TargetFramework)` | **`main` @ #457** | ✅ **probe 135.7s** | 943MB |

Scale is not the trigger: 304 projects load in ~180s. A *mixed-framework* 304-project solution also loads fine, as long as the guard isn't keyed on `$(TargetFramework)`. Flip only that key and the same binary on the same solution grows without bound; apply #457 and it loads. So #457 addresses the whole runaway growth, not a slice of it — a per-project-scoping fallback for "large" solutions would be solving a problem that doesn't exist.

On the real 286-project/7585-file solution, `main` @ #457 gives `Finished loading solution` at 128.7s and the probe answering in **131.5s**, peak RSS 1.39GB bounded — against unbounded growth to 12.5GB and `Out of memory.` at ~40min unpatched. The 12.5GB is ~75% of a 16GB cgroup, i.e. exactly the default .NET GC heap hard limit, so it is a steady ~340MB/min climb into the ceiling rather than one freak allocation.

Reproduction is committed at `tests/static_analyzer/scripts/csharp_mixed_framework_repro.py`; `--variant name-guard` is the control row above.

Two notes:

- Your correction to #456 was right — my targets fallback assigned the whole semicolon-delimited `TargetFrameworks` to singular `TargetFramework`. My multi-target check passed only because the workspace framework happened to be in the list, so the broken branch never fired. Letting Roslyn do its own inner-build expansion is cleaner than what I wrote.
- The root cause is still arguably upstream: `resolveDefaultWorkspaceProps` emits a global `TargetFramework` even when no project multi-targets, so there is nothing to disambiguate. Happy to file that on `razzmatazz/csharp-language-server` if you'd like it tracked there — it would eventually let this workaround be deleted.

#459 is a separate follow-up: it makes this class of failure visible, since the 1800s cap plus `stderr=DEVNULL` meant nobody upstream had ever seen the actual `Out of memory.`

PR #469, opened 2026-08-13
refactor(clustering): extract ClusteringService, disentangle clustering from agents

ivanmilevtues:
## Summary

Static analysis, clustering, and the LLM agents were entangled: leaf and subgraph clustering ran inside the agents via `ClusterMethodsMixin`, and the orchestrator reached into agent internals for the separability gate and the incremental scope inputs. The pipeline now has one clear cut per stage:

```
StaticAnalyzer -> StaticAnalysisResults
               -> ClusteringService -> ClusteringResults
               -> agents (name/describe/relate the fixed groups)
```

- **`clustering/service.py`**: `ClusteringService` derives every scope's deterministic cluster structure (leaf clusters, scoped call graphs, Leiden grouping) for the whole project and per component, before any LLM call; it also owns the no-callable-structure fail-fast that lived in `AbstractionAgent.run`.
- **`clustering/assignment.py`**: the deterministic post-LLM pass (`populate_file_methods`, `assemble_one_component_per_group`, `build_static_relations`, ...) as plain functions with explicit inputs; `agents/cluster_methods_mixin.py` is deleted.
- **`cluster_helpers.py` and `cluster_relations.py`** move from `static_analyzer/` to `clustering/`, which also removes the inverted `static_analyzer -> agents` dependency `cluster_relations` carried.
- **`AbstractionAgent.run(clustering)` and `DetailsAgent.run(component, clustering)`** consume `ClusteringResults`; `DiagramGenerator` runs the clustering stage explicitly and no longer calls agent-private helpers.

Also fixes a pre-existing gate failure: the pyproject packages test tripping over locally cloned repos under `temp/`.

This supersedes `refactor/clustering-service`, rebuilt from latest `main` with the `cluster_viz` visualization work and all its references dropped.

## Behavior preservation

Benchmarked against the v1 refactor branch on identical hardware/models (23 tests across full, incremental, and e2e legs): clustering shape is bit-for-bit identical on all 8 repositories — same component counts and hierarchy depth level-for-level. Remaining metric deltas move in both directions at n=1 and trace to LLM run variance, not the refactor.

## Testing

- `uv run mypy .` — clean (322 files)
- `uv run pytest --cov=. --cov-fail-under=80` — 2061 passed, 28 skipped, coverage 83.24%
- `uv run black .` — clean via pre-commit

## Summary by CodeRabbit

* **New Features**
  * Added more consistent project, component, and incremental clustering for architecture diagrams.
  * Improved analysis details with clearer methods, files, control-flow evidence, and code relationships.
  * Added support for combining generated insights with analyzed code structure.

* **Bug Fixes**
  * Improved code-element assignment to components.
  * Prevented duplicate key entities and redundant or unsupported relationships.
  * Added safer handling for small, empty, and disconnected code areas.


agents/abstraction_agent.py
```diff
@@ -31,17 +30,25 @@
...
-from static_analyzer.cluster_helpers import build_all_cluster_results
-from static_analyzer.constants import Language
 from static_analyzer.graph import ClusterResult
 
 logger = logging.getLogger(__name__)
 
 
-class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
+class AbstractionAgent(CodeBoardingAgent):
+    """Names and relates the top-level components fixed by the clustering stage."""
```
ivanmilevtues: this comment is useless we can remove it.

agents/abstraction_agent.py
```diff
@@ -188,43 +180,33 @@ def step_relation_analysis(
...
-        cluster_analysis = self.step_clusters_grouping(cluster_results)
-        if not cluster_analysis.cluster_components:
-            # No leaf clusters means the repo has no callable structure to abstract
-            # (unsupported/empty/no-code). Fail loudly instead of emptying the
-            # architecture and crashing downstream in populate_file_methods.
-            raise StaticAnalysisFatalError(
-                f"No component groups found for {self.project_name}: the static analysis produced "
-                "no callable structure to build an architecture from."
-            )
+        build_static_relations(analysis, self.static_analysis)
```
ivanmilevtues: hmm this is odd, having the static analysis as a property while having the clsutering as a parameter. I suppose we have a single stat analkysis per run, but the same should be also the case for the clustering. I just want to have a clean api for both of these dependencies into our agents.

agents/abstraction_agent.py
```diff
@@ -188,43 +180,33 @@ def step_relation_analysis(
...
+
+    def run(self, clustering: ClusteringResults) -> AnalysisInsights:
+        """Turn the clustering stage's fixed top-level groups into a named architecture."""
+        cluster_analysis = clustering.cluster_analysis
+        cluster_results = clustering.cluster_results
 
-        # Step 2: Name and describe each fixed group into a component (LLM, one component per group)
+        # Step 1: Name and describe each fixed group into a component (LLM, one component per group)
         analysis = self.step_final_analysis(cluster_analysis, cluster_results)
```
ivanmilevtues: final_analysis doesn't fit this anaymore, we should name it as step_analaysis_insight_shell or something liek that.

agents/abstraction_agent.py
```diff
@@ -188,43 +180,33 @@ def step_relation_analysis(
...
+        # Step 2: Assign hierarchical component IDs ("1", "2", "3", ...)
         assign_component_ids(analysis)
-        # Step 4: Resolve cluster IDs deterministically from group names
-        self._resolve_cluster_ids_from_groups(analysis, cluster_analysis)
-        # Step 5: Populate file_methods deterministically from cluster results + orphan assignment
-        self.populate_file_methods(analysis, cluster_results)
+        # Step 3: Resolve cluster IDs deterministically from group names
+        resolve_cluster_ids_from_groups(analysis, cluster_analysis)
+        # Step 4: Populate file_methods deterministically from cluster results + orphan assignment
+        populate_file_methods(analysis, cluster_results, self.repo_dir, self.static_analysis)
```
ivanmilevtues: that sounds like something that maybe should be part of the agent stuff. as it is effectively just an analysis thing no really related to clustering it just ingesting it as a result. same goes for a lot of the methods here.

agents/details_agent.py
```diff
@@ -34,15 +33,25 @@
...
 from static_analyzer.analysis_result import StaticAnalysisResults
-from static_analyzer.cluster_helpers import SUBCOMPONENTS_MAX, SUBCOMPONENTS_MIN
 from static_analyzer.graph import CallGraph, ClusterResult
 
 logger = logging.getLogger(__name__)
 
 
-class DetailsAgent(ClusterMethodsMixin, CodeBoardingAgent):
+class DetailsAgent(CodeBoardingAgent):
+    """Names and relates one component's sub-components fixed by the clustering stage."""
```
ivanmilevtues: this comment is useless we can drop it

diagram_analysis/diagram_generator.py
```diff
@@ -990,6 +997,8 @@ def pre_analysis(self):
 
         self.static_analysis = static_analysis
         self.meta_context = meta_context
+        # Clustering is its own pipeline stage: static analysis -> clustering -> agents.
+        self.clustering_service = ClusteringService(self.repo_location, static_analysis)
```
ivanmilevtues, line 935: as we do the static analysis in the pre_analysis I would in fact also like to get the clustering upfront like the static analysis as well. so you can have it here and then reuse it everywhere with self.clusterin_result

diagram_analysis/diagram_generator.py
```diff
@@ -610,6 +610,7 @@ def __init__(
 
         self.details_agent: DetailsAgent | None = None
         self.static_analysis: StaticAnalysisResults | None = None  # Cache static analysis for reuse
+        self.clustering_service: ClusteringService | None = None
```
ivanmilevtues, line 589: there are way too many comments in the members above let's remove most of them

diagram_analysis/diagram_generator.py
```diff
@@ -725,8 +726,14 @@ def _process_component(
         """Process a single component and return its name, sub-analysis, and new components to analyze."""
         try:
             assert self.details_agent is not None
+            assert self.clustering_service is not None
 
-            analysis, _ = self.details_agent.run(component)
+            # Clustering stage first: the component's subgraph clustering is an input
+            # to the agent, which only names and relates the fixed groups.
+            clustering = self.clustering_service.cluster_component(
+                component, source_cluster_id_prefix=component.component_id
```
ivanmilevtues: feels like the component.component_id is not needed as we already have it in the component right?

agents/abstraction_agent.py
```diff
@@ -146,7 +138,7 @@ def step_final_analysis(
     @trace
     def step_api_surfaces(self, analysis: AnalysisInsights) -> ComponentApiSurfaces:
         logger.info(f"[AbstractionAgent] Analyzing component API surfaces for: {self.project_name}")
-        static_call_evidence = self.build_scope_cfg_string(analysis)
+        static_call_evidence = build_scope_cfg_string(analysis, self.static_analysis)
```
ivanmilevtues: i don't think we noi longer need the self.static_analysis, as we have the cluster result and that contains the static analysis itself, so probably the static analysis has to stay viable only till we do the clustering and then we don't need to persist it as something. From there on we have the cluster reuskts and the services associated tiwh tit.

clustering/service.py
```diff
@@ -0,0 +1,277 @@
...
+    reindex_across_languages,
+    supercluster_leaf_ids,
+)
+from constants import MIN_CLUSTERS_THRESHOLD
+from diagram_analysis.cluster_delta import _delta_for_language
+from diagram_analysis.cluster_snapshot import ClusterSnapshotEntry
+from static_analyzer import StaticAnalysisFatalError
+from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.constants import CALLABLE_TYPES, CLASS_TYPES, Language
+from static_analyzer.graph import METHOD_LEVEL_STRATEGY, CallGraph, ClusterResult
```
ivanmilevtues: cluster result probably has to move the clsuter package right, same for method level strategy. call graph can stay with the static analyzer graph. but the other should probably move here no. Also maybe the clustering should also go in the static analysis package overall and no thave its own here

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,277 @@
...
+        logger.info(
+            f"[ClusteringService] Partitioned {sum(len(g) for g in groups)} leaf clusters "
+            f"into {len(cluster_components)} deterministic groups"
+        )
+        return ClusterAnalysis(cluster_components=cluster_components)
+
+    def component_subgraph(
+        self,
+        component: Component,
+        source_cluster_id_prefix: str = "",
```
ivanmilevtues: that comes from the component so no need of this parameter here. Also I think that this might not be needed as a public method as this is just a substep to cluster the component's subgraph so proabbly we need to leve only that method as public and let it reuse this one as internal method.

clustering/service.py
```diff
@@ -0,0 +1,277 @@
...
+            )
+        return ClusteringResults(
+            cluster_results=cluster_results,
+            cfg_graphs=self.static_analysis.available_cfgs(),
+            cluster_analysis=cluster_analysis,
+        )
+
+    def cluster_component(self, component: Component, source_cluster_id_prefix: str = "") -> ClusteringResults:
+        """Cluster one component's subgraph and group it into sub-component groups."""
+        cluster_results, subgraph_cfgs = self.component_subgraph(component, source_cluster_id_prefix)
```
ivanmilevtues: yes i thionk this component_subgraph has to become a private method. probably also the group_clusters. if not let me know why it has to be public.

Also drop the source_cliuster_id_prefix as it is effectively the component.component_id or something like that and we alwasys take it from the compnent either way.

clustering/assignment.py
```diff
@@ -0,0 +1,590 @@
+"""Deterministic post-LLM assignment: pin the LLM's named components back onto the clusters.
```
ivanmilevtues: so this feels like a service which should be in the agent part, and we need a clean api it should depend clearly on the clustering result or something like that. I want it to be not as random as it is at the moment. Now it seems a bit like a util almost. But this in fact is the enricher for our llm generated analysis, and what it does is based on the clustering/static analysis result it infills data.

clustering/assignment.py
```diff
@@ -0,0 +1,590 @@
...
+
+    if len(final) != len(architecture.components):
+        logger.info(
+            f"[Assignment] Reconciled {len(architecture.components)} LLM components "
+            f"to {len(final)} (one per deterministic group)"
+        )
+    architecture.components = final
+
+
+def ensure_unique_key_entities(analysis: AnalysisInsights) -> None:
```
ivanmilevtues: this sounds like something to go into the repair.py and use it when we prompt and repair the reuslt

ivanmilevtues (commented):
Okay I did a first pass on the changes, most serious requiements are to tighten the API and introdce a bit more object oriented design here so we can have better level of abstracitons and encapsulation.

agents/abstraction_agent.py
ivanmilevtues: Removed in 1ba5e9ee.

agents/abstraction_agent.py
ivanmilevtues: Fixed in 1ba5e9ee: `ClusteringResults` now carries the `StaticAnalysisResults` it was derived from, and it is the agents' single constructor dependency — `AbstractionAgent(..., clustering=...)`, and `run()` takes no arguments at all. No more property-vs-parameter split between the two.

agents/abstraction_agent.py
ivanmilevtues: Renamed `step_final_analysis` -> `step_analysis_shell` in both agents (it produces the named-but-relationless insight shell).

agents/abstraction_agent.py
ivanmilevtues: Agreed — the whole deterministic post-LLM pass moved to the agent side in 1ba5e9ee: `agents/enrichment.py` with `StaticAnalysisEnricher`. The agents now call `enricher.populate_file_methods(analysis)`, `enricher.resolve_cluster_ids(analysis)`, etc., instead of importing loose functions from the clustering package.

agents/details_agent.py
ivanmilevtues: Removed in 1ba5e9ee.

diagram_analysis/diagram_generator.py
ivanmilevtues: Done: `pre_analysis` now derives `self.clustering_result = self.clustering_service.cluster_project()` right after static analysis, and both `generate_analysis` and agent construction reuse it — the project scope is clustered exactly once per run.

diagram_analysis/diagram_generator.py
ivanmilevtues: Trimmed to short one-liners in 1ba5e9ee.

diagram_analysis/diagram_generator.py
ivanmilevtues: Right — dropped. `cluster_component(component)` derives the prefix from `component.component_id` internally, and every `source_cluster_id_prefix` parameter is gone from the service's public API.

agents/abstraction_agent.py
ivanmilevtues: Yes, exactly — the clustering is derived from the static analysis, so `ClusteringResults` now carries it (`clustering.static_analysis`). The agents no longer take or persist a separate static-analysis handle: the base class feeds the toolkit/reference resolver from the clustering result, and the remaining uses (CFG evidence, validation contexts) go through the enricher or `self.clustering`.

clustering/service.py
ivanmilevtues: Both done in 1ba5e9ee: the package folded into `static_analyzer/clustering/`, with `ClusterResult` + `METHOD_LEVEL_STRATEGY` moved into `models.py`; `CallGraph` stays in `graph.py` and imports the models intra-package (its cluster cache made this the clean cut). One tradeoff to be aware of: the clustering subpackage imports agent response models (`Component`, `ClusterAnalysis`) — a contained upward edge; `graph.py` stays clean and there are no runtime cycles (the subpackage `__init__` is deliberately import-free because `graph.py` pulls in `models` during `static_analyzer` init).

static_analyzer/clustering/service.py
ivanmilevtues: Done: `_component_subgraph` is private now, `cluster_component(component)` is the public entry that reuses it, and the prefix is derived from `component.component_id` instead of being a parameter.

clustering/service.py
ivanmilevtues: Both are private now (`_component_subgraph`, `_group_clusters`) — and to answer the question: nothing outside the service actually needed them. The generator's separability probe moved into the service as `component_is_separable(component)` (memo cache included), and the incremental structural diff got a purpose-built `subgraph_clusters(component)`. `source_cluster_id_prefix` is gone from the public API entirely.

clustering/assignment.py
ivanmilevtues: Done: this is now `agents/enrichment.py` with `StaticAnalysisEnricher` — constructed per scope with that scope's `ClusteringResults`, it infills the deterministic data (group pinning, cluster ids, file methods, static relations, CFG evidence) into the LLM-generated analysis. Went with `StaticAnalysisEnricher` as a shorter form of your `StaticAnalysisInformationEnhancer`. The incremental path keeps three scoped module-level functions because its inputs (cluster results, subgraphs, id prefixes) are assembled per operation rather than carried by one `ClusteringResults`.

clustering/assignment.py
ivanmilevtues: Moved to `agents/repair.py` and wired into the `_invoke_repair_validate` chain as `repair_unique_key_entities` in both agents' analysis-shell steps. It also still runs once after reference resolution: at repair time `reference_file` isn't resolved yet, and the tiebreak deciding which component keeps a contested entity needs the resolved files.

agents/enrichment.py
```diff
@@ -0,0 +1,557 @@
...
+    """Deterministic component for a group the LLM failed to name (merged/dropped it)."""
+    symbols = group_symbols(group.cluster_ids, node_lookup)
+    name = symbols[0].split(".")[-1] if symbols else group.name
+    return Component(name=name, description=group.description, key_entities=[])
+
+
+def _scoped_cfg(
+    lang: str,
+    static_analysis: StaticAnalysisResults,
+    cfg_graphs: dict[str, CallGraph] | None,
```
ivanmilevtues: cfg graphs should never be none, so please make it non-nullable and if it is it should just be empty dict

diagram_analysis/diagram_generator.py
```diff
@@ -694,11 +626,15 @@ def _expandable_ids_for_tree(
         if self.details_agent is None:
             return None, None
 
+        def separable(component: Component) -> bool:
+            assert self.clustering_service is not None
+            return self.clustering_service.component_is_separable(component)
```
ivanmilevtues: just use the service directly no need to have this wrapper method.

diagram_analysis/diagram_generator.py
```diff
@@ -694,11 +626,15 @@ def _expandable_ids_for_tree(
...
+
         def expandable_ids(scope: AnalysisInsights, parent_had_clusters: bool = True) -> list[str]:
             ids = [
                 component.component_id
                 for component in get_expandable_components(
-                    scope, parent_had_clusters=parent_had_clusters, separable=self._component_separable
+                    scope, parent_had_clusters=parent_had_clusters, separable=separable
```
ivanmilevtues, line 636: yea here just use self.cllustering_service.component_is_separable

diagram_analysis/diagram_generator.py
```diff
@@ -1730,14 +1678,14 @@ def _build_scope_incremental_inputs(
 def scoped_snapshot_for_component(
```
ivanmilevtues: This method seems as if it has to go to the clusteringService because it is effectively doing cluster wrok no?

diagram_analysis/diagram_generator.py
```diff
@@ -990,6 +931,10 @@ def pre_analysis(self):
 
         self.static_analysis = static_analysis
         self.meta_context = meta_context
+        # Clustering is its own pipeline stage: static analysis -> clustering -> agents.
+        # The project-scope result is derived once here and reused everywhere.
+        self.clustering_service = ClusteringService(self.repo_location, static_analysis)
+        self.clustering_result = self.clustering_service.cluster_project()
```
ivanmilevtues, line 936: This seems like a legitemate concernt. will take a look in a minute.

static_analyzer/clustering/models.py
```diff
@@ -0,0 +1,42 @@
...
+from collections import defaultdict
+from collections.abc import Callable
+from dataclasses import dataclass, field
+
+logger = logging.getLogger(__name__)
+
+# Marker on a ClusterResult whose clusters are synthetic one-method-per-cluster
+# groups, produced when a subgraph had too few natural clusters to assign methods
+# at a useful granularity. Its modularity is not comparable to a real clustering's.
+METHOD_LEVEL_STRATEGY = "method_level_expansion"
```
ivanmilevtues: this should probably go in constants.py as well as other constonstants from the clustring package
we should have clustring/constants.py

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+
+import networkx as nx
+
+from agents.agent_responses import ClusterAnalysis, ClustersComponent, Component
+from static_analyzer.clustering import separability
+from static_analyzer.clustering.cluster_helpers import (
+    SUBCOMPONENTS_MAX,
+    SUBCOMPONENTS_MIN,
+    TOP_LEVEL_COMPONENTS_MAX,
+    TOP_LEVEL_COMPONENTS_MIN,
```
ivanmilevtues: these 4 constaints should go in to the new constants.py

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.constants import CALLABLE_TYPES, CLASS_TYPES, Language
+from static_analyzer.clustering.models import ClusterResult, METHOD_LEVEL_STRATEGY
+from static_analyzer.graph import CallGraph
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ClusteringResults:
```
ivanmilevtues: we have new models.py add it there (for the clustering)

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+    cfg_graphs: dict[str, CallGraph]
+    #: deterministic component groups ("Group i"); the LLM only names them
+    cluster_analysis: ClusterAnalysis
+    #: the static analysis this clustering was derived from
+    static_analysis: StaticAnalysisResults
+    #: component id whose subgraph this scope is; "" for the whole project
+    scope_id: str = ""
+
+
+def scoped_snapshot_from_lineage(cfg: CallGraph, scope_id: str) -> dict[int, ClusterSnapshotEntry]:
```
ivanmilevtues: we had another method fo scoping stuff in diagram.py so this is another proof that that method should come over here, maybe not in this file but still somehwere her

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+
+    return ClusterResult(
+        clusters=new_clusters,
+        cluster_to_files=new_cluster_to_files,
+        file_to_clusters=dict(new_file_to_clusters),
+        strategy=METHOD_LEVEL_STRATEGY,
+    )
+
+
+class ClusteringService:
```
ivanmilevtues, line 44: put the class on top of this file it is annoying tohave to go 157 lines down to get the actual service.py

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+                    str(lang), sub_cfg.clustering_networkx(), seeded_snapshot
+                ).cluster_results
+            else:
+                sub_cluster_result = sub_cfg.cluster()
+
+            cluster_results[lang] = _expand_to_method_level_clusters(sub_cfg, sub_cluster_result)
+
+        reindex_across_languages(cluster_results)
+
+        if source_cluster_id_prefix:
```
ivanmilevtues: i saw that there is only one place where we are sending this as none, i don't like this behavior.we should eitehr always record or not, and amke it clear from the api whenwe are not recording things. so i'd rather change this to a parameter which statest record cluster paths and by default is true and make this the actual method than using a nullable parameter for it.

agents/enrichment.py
```diff
@@ -0,0 +1,557 @@
...
+    def pin_components_to_groups(self, architecture: ComponentArchitecture) -> None:
+        """Force exactly one component per fixed group — the count is Leiden's, not the LLM's.
+
+        The groups (and their membership) are decided deterministically upstream;
+        the LLM only names and describes them. Whatever the LLM returns, we pin the
+        result to one component per group: the LLM's component that claimed a group
+        keeps its name/description/key_entities; any group the LLM merged away or
+        dropped gets a deterministic fallback so the count never drifts.
+        """
+        node_lookup = combine_cluster_results(self.clustering.cluster_results).clusters
```
ivanmilevtues: it feels like this combine_cluster_results, should be somehow a method within the clustering?

agents/enrichment.py
```diff
@@ -0,0 +1,557 @@
...
+        claimant: dict[str, Component] = {}
+        for comp in architecture.components:
+            for group_name in comp.source_group_names:
+                claimant.setdefault(group_name.lower(), comp)
```
ivanmilevtues: i don't get the name claimant. this is a group_component dict so name it like that. src_group_to_component

agents/enrichment.py
```diff
@@ -0,0 +1,557 @@
...
+            logger.info(
+                f"[Enrichment] Reconciled {len(architecture.components)} LLM components "
+                f"to {len(final)} (one per deterministic group)"
+            )
+        architecture.components = final
+
+    def resolve_cluster_ids(self, analysis: AnalysisInsights) -> None:
+        """Resolve source_cluster_ids deterministically from source_group_names via case-insensitive lookup."""
+        group_name_to_ids: dict[str, list[GraphClusterId]] = {
+            cc.name.lower(): cc.cluster_ids for cc in self.clustering.cluster_analysis.cluster_components
```
ivanmilevtues: i would name those cluster_groups polease go ahead and rename it. also the class should ClustersGroup

agents/enrichment.py
ivanmilevtues: cfg graphs should never be none, so please make it non-nullable and if it is it should just be empty dict

static_analyzer/graph.py
```diff
@@ -8,6 +8,7 @@
 import networkx as nx
 import networkx.algorithms.community as nx_comm
 
+from static_analyzer.clustering.models import ClusterResult
```
ivanmilevtues: we should decouple the graph from the cacheing of clustering and in general from saving the results. 

The cluster method in general should go to the clustering service. teh callgraph should become more of a data object itself.

methods like clustering_ntewortkX should be renamed to something liek to_networkX as that is what it does now.

agents/abstraction_agent.py
```diff
@@ -115,28 +104,28 @@ def step_final_analysis(
             )
 
         context = ValidationContext(
-            cluster_results=cluster_results,
-            static_analysis=self.static_analysis,
-            llm_cluster_analysis=llm_cluster_analysis,
+            cluster_results=self.clustering.cluster_results,
```
ivanmilevtues: just pass the cluster_results no need to have both things - > cluster_data = self.clustering

agents/abstraction_agent.py
```diff
@@ -115,28 +104,28 @@ def step_final_analysis(
...
+            static_analysis=self.clustering.static_analysis,
+            llm_cluster_analysis=cluster_analysis,
         )
 
         architecture = self._invoke_repair_validate(
             prompt,
             ComponentArchitecture,
-            repairs=[repair_component_group_names, repair_key_entities],
+            repairs=[repair_component_group_names, repair_key_entities, repair_unique_key_entities],
```
ivanmilevtues, line 109: i see, makes sense.  I want this to be not a patchwork on 3 palces that's why I am thinking on where to put it. I think that probably it has to be the last step of the run method then. This really gives me a headache ;d. put it as last step in the run for now. Keep it in the repaird sthou

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,346 @@
...
+    def subgraph_clusters(self, component: Component) -> tuple[dict[str, ClusterResult], dict[str, CallGraph]]:
+        """The component's scoped leaf clusters and subgraph CFGs, without grouping.
+
+        For consumers that need the raw scoped clustering (e.g. the incremental
+        structural diff) rather than component groups. Records cluster lineage
+        under the component's id, like ``cluster_component``.
+        """
+        return self._component_subgraph(component, component.component_id)
+
+    def component_is_separable(self, component: Component) -> bool:
```
ivanmilevtues, line 134: we have more than one of this method which I hate we have it in separability.py we also have it here. we should end up with just one method which is reused whenever needed,

ivanmilevtues (commented):
second pass of comments.

diagram_analysis/diagram_generator.py
ivanmilevtues: Dropped in 438bb78d — `_expandable_ids_for_tree` binds `separable = self.clustering_service.component_is_separable` directly (one assert for the Optional), no wrapper method.

diagram_analysis/diagram_generator.py
ivanmilevtues: Done — both this site and `_process_component` now pass `self.clustering_service.component_is_separable` straight into `get_expandable_components`.

diagram_analysis/diagram_generator.py
ivanmilevtues: Agreed — it is cluster work. `scoped_snapshot_for_component` moved into the service as `ClusteringService.scoped_snapshot(component, scope_id)` in 438bb78d; the incremental input builder calls it there.

diagram_analysis/diagram_generator.py
ivanmilevtues: The trade-off to be aware of: agents consume the project clustering at construction now, so incremental runs also pay one project-scope Leiden sweep they previously skipped (seconds of deterministic work, not LLM time). If it ever shows up in the eval wall times we can make `clustering_result` a lazily computed cached property — but since agent construction forces it anyway under the constructor-injection design, eager-in-pre_analysis is the honest place for it.

static_analyzer/clustering/models.py
ivanmilevtues: Done in 438bb78d: `static_analyzer/clustering/constants.py` now holds every tuning constant of the stage — `METHOD_LEVEL_STRATEGY`, the `TOP_LEVEL_COMPONENTS_*`/`SUBCOMPONENTS_*` ranges, and the separability sizing knobs (`MIN_METHODS_TO_EXPAND`, `MAX_LEAF_*`, `EXPAND_MODULARITY_THRESHOLD`). `_RESOLUTION_LADDER` stayed private in cluster_helpers as an algorithm detail.

static_analyzer/clustering/service.py
ivanmilevtues: Moved to `clustering/constants.py` in 438bb78d.

static_analyzer/clustering/service.py
ivanmilevtues: Done — `ClusteringResults` lives in `clustering/models.py` now (with a `combined()` method, see the enrichment thread). One nuance: `ClusterResult`'s definition had to stay physically next to `CallGraph` because the pickled cluster cache references `static_analyzer.graph.ClusterResult` (older baselines would stop unpickling) and a full move recreates a `graph <-> models` import cycle; `models.py` re-exports it, so `clustering.models` is the import home for both.

static_analyzer/clustering/service.py
ivanmilevtues: Yes — that method (`scoped_snapshot_for_component`) came over in 438bb78d as `ClusteringService.scoped_snapshot`, right next to `scoped_snapshot_from_lineage`.

static_analyzer/clustering/service.py
ivanmilevtues: Done — `ClusteringService` now opens the file; the module helpers sit below it.

static_analyzer/clustering/service.py
ivanmilevtues: Agreed and done: `_component_subgraph(component, record_cluster_paths: bool = True)`. The prefix is derived from `component.component_id` internally; the separability probe is the one caller passing `False`, and the flag also skips lineage *seeding* so a probe stays fully read-only (same behavior the empty prefix used to give, now stated by the API).

agents/enrichment.py
ivanmilevtues: Made it one in 438bb78d: `ClusteringResults.combined()` unions the per-language results, so the enricher reads `self.clustering.combined().clusters`. The standalone `combine_cluster_results` moved into `clustering/models.py` next to the model for the remaining scoped callers.

agents/enrichment.py
ivanmilevtues: Renamed to `src_group_to_component`.

agents/enrichment.py
ivanmilevtues: Renamed repo-wide in 438bb78d: class `ClustersComponent` -> `ClustersGroup`, field `ClusterAnalysis.cluster_components` -> `cluster_groups`.

agents/enrichment.py
ivanmilevtues: Done — every `cfg_graphs` parameter in enrichment is non-nullable now; `{}` means "fall back to the global CFG" for the scoped helpers, and `build_static_relations` takes the scope's graphs explicitly (callers pass `static_analysis.available_cfgs()` for the whole project).

static_analyzer/graph.py
ivanmilevtues: Done in 438bb78d, in three cuts: (1) the whole `cluster()` algorithm family moved to `clustering/graph_clustering.py` as `cluster_call_graph(cfg)` — health's cohesion check and the helpers call that now; (2) `CallGraph` keeps only data responsibilities: a `cluster_cache` property + `set_cluster_cache` (killing the private-field poke), pruning delegated to `ClusterResult.pruned_to`, and lineage recording; (3) `clustering_networkx()` renamed `to_networkx_with_references()`. The cache *slot* stays on the graph deliberately — it's derived data that must survive `filter_by_nodes`/merge pruning and ship in the pkl, and hosting it elsewhere would break every existing baseline.

agents/abstraction_agent.py
ivanmilevtues: Done via context factories in 438bb78d: the shell step passes `ValidationContext.for_shell(self.clustering)` and the relation step `ValidationContext.for_relations(self.clustering, repo_dir=..., components=...)`; `ComponentRepairContext` now holds the clustering directly (`ComponentRepairContext(self.reference_resolver, self.clustering)`). No more unpacked cluster_results/static_analysis/cluster_analysis trios at the call sites.

agents/abstraction_agent.py
ivanmilevtues: It already is the last step — `ensure_unique_key_entities(analysis)` is the final statement of `run()` in both agents (after reference resolution, so the file-based tiebreak sees resolved paths), and `repair_unique_key_entities` stays in the repair chain as you asked. So: repair chain + one final pass, no third place.

static_analyzer/clustering/service.py
ivanmilevtues: Deduplicated in 438bb78d: the pure math in `separability.py` is now uniquely named `subgraph_is_separable` (still directly unit-testable), and the only component-facing entry is the service's memoized `component_is_separable`, which calls it. One name, one owner each.

Issue #471, opened 2026-08-13
Move qualified-name canonicalization out of CallGraph.add_node into an explicit upstream pass


ivanmilevtues:
## Summary

`CallGraph` currently performs qualified-name canonicalization **incrementally, inside `add_node`**, and maintains a persistent alias lookup table to paper over the names it discarded. This is dedup logic living one layer too low: the graph is repairing bad input from the LSP layer instead of receiving clean input.

Follow-up to the `static_analyzer/cfg` + `static_analyzer/clustering` split, which deliberately left this in place because it touches the engine and merge paths.

## Current situation

`CallGraph` carries two indexes for this (`static_analyzer/graph.py:172-173`):

```python
self._location_index: dict[LocationKey, str] = {}
self._alias_to_canonical: dict[str, str] = {}
```

`add_node` (`static_analyzer/graph.py:180-214`) keys each node by physical location — `(file_path, line_start, line_end, node_type, col_start)` — and when two qualified names collide on one location it keeps the **longest** name, mutates the surviving `Node.fully_qualified_name` in place, flattens any alias chain that pointed at the old name, and rewrites `methods_called_by_me` on every source node that referenced it.

Every read path then has to resolve through `_alias_to_canonical`:

- `has_node` / `_resolve_name` — `graph.py:216-222`
- `add_edge` — `graph.py:224-226`
- `add_reference_edge` — `graph.py:251-252`
- `_carry_reference_edges`, resolving through the *source* graph's alias map — `graph.py:268-270`
- `clustering_networkx` — `graph.py:409-412`

## Why it exists — do not just delete it

Added in 263da33e ("Real-time updates fixes", #287) to fix a user-visible bug:

> Reproduces the nanoclaw bug where the same physical method appears in multiple sibling components due to qualified-name aliases (e.g. `src.index.funcA` and `container.agent-runner.src.index.funcA`) landing in different clusters and hence different components.

The LSP genuinely emits two qualified names for one physical symbol — most commonly two tsconfig projects in a monorepo resolving the same file under different roots. Without dedup the two names cluster separately and the same method renders into two sibling components.

`tests/agents/test_method_uniqueness.py` is the regression test. **Location dedup is load-bearing and must survive this change.** What should not survive is the persistent alias table and the in-place promotion.

## Latent bug this would fix

`static_analyzer/incremental_orchestrator.py:152-156`:

```python
if not call_graph.has_node(src_name) or not call_graph.has_node(dst_name):
    continue

[...]

PR #287, opened 2026-04-10
Real-time updates fixes

ivanmilevtues:
---

<a href="https://app.devin.ai/review/codeboarding/codeboarding/pull/287" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * File dependency discovery: added a way to find and return per-file dependencies via language-aware analysis.

* **Refactor**
  * Improved component method synchronization to prevent duplicate method assignments and enforce correct primary ownership.
  * Enhanced call-graph node deduplication to consolidate aliases and maintain stable node references.

* **Tests**
  * Added comprehensive tests validating method-uniqueness across components, including alias and large-scale scenarios.

PR #472, opened 2026-08-13
refactor(cfg): split graph.py into cfg, clustering and llm_renderers packages

ivanmilevtues:
`static_analyzer/graph.py` had grown to **811 LoC** doing four unrelated jobs: graph structure, a Leiden clustering engine, custody of clustering results, and LLM prompt rendering. Nearly half of it was not graph structure at all.

## Layout

The dependency direction is now one-way — `static_analyzer/cfg` → `clustering` → `agents`:

```
static_analyzer/cfg/           static_analyzer/clustering/       agents/llm_renderers/
  call_graph.py   268            engine.py           242           call_graph.py  106
  edge.py          80            method_cluster_paths.py  84       __init__.py     10
  canonical.py     22            models.py            68
  __init__.py      11            cache.py             52
                                 service.py           34
                                 constants.py         15
                                 __init__.py          21
```

`CallGraph` is down to **268 LoC** and 13 public methods — structure, derivation and `to_networkx`.

## The cluster cache moved off the graph

It was `CallGraph._cluster_cache`: private by name, but written from three production modules and read from two. It is now a `ClusterCache` owned by `LanguageResults`, reached via `StaticAnalysisResults.get_clusters()` / `set_clusters()`. `ClusteringService.cluster(graph)` is pure — it neither mutates the graph nor caches.

Three things follow from that:

- **`filter`/`union` no longer silently prune a partition.** The warm-start path carries and prunes it in one visible place instead of it happening invisibly inside two graph methods.
- **`cluster()` lost the side effect that kept a health check switched off.** `check_component_cohesion` called `cluster()` purely for a partition, and seeding `_cluster_cache` with a current-graph partition would masquerade as a historical snapshot in the next incremental delta. That is gone. **Left unwired** — turning a health check back on is a behaviour change, not a refactor.
- **`adopt()` and `record_scope()` split what the old code conflated.** Setting a language's partition and recording a component sub-partition's lineage looked nearly identical before (`_cluster_cache = r` + `record_cluster_paths(r)` versus `record_cluster_paths(r, prefix)`). Collapsing them into one method would let a component's sub-partition overwrite what `cluster_snapshot` reads. There is a test pinning the distinction.

## API and typing

- `to_networkx(reference_kinds=())` replaces the `to_networkx` / `clustering_networkx` pair. Call edges are always included; the five clustering call sites now pass their kinds explicitly rather than relying on a default hidden inside the graph.
- `reference_edges` is `list[ReferenceEdge]` instead of 3-tuples of strings whose third element was a stringified enum.
- `CLUSTERING_REFERENCE_KINDS` is typed with `EdgeKind` and named for what it holds. The old `CLUSTERING_EDGE_KINDS` excluded `EdgeKind.CALL` despite call edges always being clustered, and existed twice — once as bare strings, once derived.
- `from __future__ import annotations` throughout the new packages; the six quoted forward refs are gone.

## Cache invalidation

`_TAG_VERSION` `v2` → `v3`. Pre-refactor `static_analysis.pkl` baselines become cache misses and surface `BaselineUnavailableError`.

This is deliberate and is why the two `getattr(self, "reference_edges", ())` unpickling guards could be deleted. Without the bump, `pickle` restores `_cluster_cache` into the instance `__dict__` even though the class no longer declares it — the attribute survives, is ignored, and the incremental quietly loses its cluster baseline. That is exactly the silent degradation AGENTS.md §5 forbids.

## Removed

`_common_dot_prefix` (referenced only by its own test), `detect_communities` (a pure pass-through — callers import `leiden_utils.find_partition` directly), both `getattr` unpickling guards, and a dead `new_d` assignment.

## Not in scope

- **Qualified-name canonicalization stays inside `CallGraph.add_node`.** It exists to fix a real user-visible bug (#287, the nanoclaw monorepo case where one physical method landed in two sibling components) and moving it upstream touches `result_converter` and the merge path. Tracked as #471.
- `cluster_helpers.py` and `cluster_relations.py` stay where they are. `cluster_relations.py` imports `agents.agent_responses` to construct agent-layer response objects, so moving it into `clustering/` would bake a layering violation into the new package — it belongs in `agents/`, as its own change.

## Checks

- `2062 passed, 28 skipped`
- `mypy`: no issues in 329 source files
- `black`: clean
- coverage **83.03%** (gate 80%)

Tests were split to match the packages: `test_call_graph.py` (structure), `test_clustering.py` (service + cache), `tests/agents/test_llm_renderers.py`. The `filter_by_files` tests no longer route through `cluster()` just to pick file paths.

## Summary by CodeRabbit

* **New Features**
  * Added clearer call-graph output with detailed method views and automatic summaries for oversized graphs.
  * Improved clustering results, including file mappings, deterministic grouping, fallback handling, and preservation across incremental analysis.
  * Added support for structural relationships such as containment and inheritance in graph analysis.
  * Added public access to per-language clustering results and cache state.
* **Bug Fixes**
  * Improved component scoping and coupling metrics by excluding unrelated or non-call relationships from calculations.
* **Tests**
  * Expanded coverage for rendering, clustering, caching, filtering, and incremental analysis.


agents/llm_renderers/call_graph.py
```diff
@@ -0,0 +1,106 @@
...
+    logger.info(f"[CFG Tool] LLM string: {len(detailed)} characters, size limit: {size_limit} characters")
+    if len(detailed) <= size_limit:
+        return detailed
+
+    logger.info(
+        f"[CallGraph] Control flow graph is too large ({len(detailed)} chars), switching to class-level summary."
+    )
+    class_level = _render_class_level(graph, skip_set)
+    logger.info(f"[CallGraph] Class-level summary: {len(class_level)} characters")
+    return class_level
```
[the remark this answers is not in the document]
ivanmilevtues: that is fine for now. let it be. We have been running it like that forever never saw an issue.

agents/llm_renderers/call_graph.py
```diff
@@ -0,0 +1,106 @@
...
+
+    result = f"Control flow graph with {active_nodes} nodes and {active_edges} edges\n"
+
+    for file_path in sorted(file_nodes):
+        nodes = sorted(file_nodes[file_path], key=lambda n: n.fully_qualified_name)
+        for node in nodes:
+            if node.methods_called_by_me:
+                label = node.entity_label()
+                targets = ", ".join(sorted(node.methods_called_by_me))
+                result += f"{label} {node.fully_qualified_name} calls: {targets}\n"
```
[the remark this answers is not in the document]
ivanmilevtues: I suppose this should be addressed yes.

diagram_analysis/tree_shape.py
```diff
@@ -14,12 +14,12 @@
 def absorb_single_child_components(
     root_analysis: AnalysisInsights,
     sub_analyses: dict[str, AnalysisInsights],
-    cfgs: Sequence[CallGraph] = (),
+    cluster_caches: Sequence[ClusterCache] = (),
```
ivanmilevtues, line 17: hmm the cache shouldn't be used we should always use the CallGraph and have the cache as something hidden behind it, or just used once when laoding it and from then on it should be purely just the cfgs

static_analyzer/analysis_cache.py
```diff
@@ -51,9 +51,11 @@
...
 _LEGACY_CACHE_SUBDIR = "cache"
 # Tag file format prefix; bump if the on-disk pickle layout changes.
-# v2: StaticAnalysisResults switched from dict-of-dicts to LanguageResults
-# dataclass storage. v1 pickles will be treated as cache misses and re-run.
-_TAG_VERSION = "v2"
+# v2: StaticAnalysisResults switched from dict-of-dicts to LanguageResults storage.
+# v3: clustering state moved off CallGraph into LanguageResults.clusters, and
+# reference edges became ReferenceEdge objects rather than 3-tuples.
+# Older pickles are treated as cache misses and re-run.
+_TAG_VERSION = "v3"
```
ivanmilevtues, line 58: nooo we shouldn't change this, we should keep it as is.

static_analyzer/analysis_cache.py
```diff
@@ -439,24 +444,13 @@ def _rederive_inherits_edges(call_graph: CallGraph, class_hierarchies: dict[str,
...
     """
-    existing = {(s, d) for s, d, k in call_graph.reference_edges if k == str(EdgeKind.INHERITS)}
+    existing = {(ref.src, ref.dst) for ref in call_graph.reference_edges if ref.kind is EdgeKind.INHERITS}
     for child, info in class_hierarchies.items():
         for superclass in info.get("superclasses", []):
             if (child, superclass) not in existing:
                 call_graph.add_reference_edge(child, superclass, EdgeKind.INHERITS)
 
 
-def _collect_invalidated_edge(
```
ivanmilevtues, line 449: why have we changed the caching strats here I don't want to break existing caching for now.

static_analyzer/cfg/call_graph.py
```diff
@@ -0,0 +1,243 @@
...
+
+        edge = Edge(self.nodes[src_name], self.nodes[dst_name], [])
+        for call_site in call_sites:
+            edge.add_call_site(dict(call_site))
+        self.edges.append(edge)
+        self._edge_by_key[edge_key] = edge
+
+        self.nodes[src_name].added_method_called_by_me(self.nodes[dst_name])
+
+    def add_reference_edge(self, src_name: str, dst_name: str, kind: EdgeKind) -> None:
```
ivanmilevtues, line 115: this should accept a refernce object and unpack it internally.

static_analyzer/cfg/call_graph.py
```diff
@@ -0,0 +1,243 @@
...
+            for ref in source.reference_edges:
+                # Resolve through the SOURCE's alias map: an endpoint stored under a short
+                # alias must map to the canonical name ``out`` promoted it to, or a call edge
+                # (which add_edge resolves) survives while its reference edge is silently dropped.
+                resolved = ReferenceEdge(source._resolve_name(ref.src), source._resolve_name(ref.dst), ref.kind)
+                if resolved.src in out.nodes and resolved.dst in out.nodes and resolved.src != resolved.dst:
+                    carried.append(resolved)
+        out.reference_edges = list(dict.fromkeys(carried))
+
+    def filter(self, keep_node: Callable[[Node], bool]) -> CallGraph:
```
ivanmilevtues, line 144: put all public methods on top and then the privates below. so that when I scroll i see first the public api. Add this as a rule in the AGENTS.md and the Reviews.md

static_analyzer/cfg/location_key.py
```diff
@@ -0,0 +1,17 @@
+"""Physical-location identity, used to dedup the several qualified names the LSP
+can emit for one symbol. Issue #471 moves the canonicalization itself here."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class LocationKey:
```
ivanmilevtues, line 10: seems like the file should be named location_key.py

static_analyzer/clustering/cache.py
```diff
@@ -0,0 +1,47 @@
...
+        Why: sub-clustering a component produces a partition of that scope, not of
+        the language — adopting it would overwrite the top-level partition that
+        ``cluster_snapshot`` reads.
+        """
+        self.method_paths.record(cluster_result, scope_id)
+
+    def pruned_to(self, surviving_nodes: Mapping[str, Node]) -> ClusterCache:
+        """Return a copy restricted to ``surviving_nodes``, for filter/union of the graph."""
+        return ClusterCache(
+            result=self.result.pruned_to(surviving_nodes),
```
ivanmilevtues: i woudl rename pruned_to, to jsut prune so it is consistent.

static_analyzer/clustering/cache.py
```diff
@@ -0,0 +1,47 @@
...
+            method_paths=self.method_paths.prune(surviving_nodes),
+        )
+
+    def visit_paths(self, fn: Callable[[str], str]) -> None:
```
ivanmilevtues, line 46: the cache shouldn't have such method, it is just a data object no? I am not even sure for the adopt and record scope as they seem to be parts of the clusterresult methods either way, this to me should just be the cache and it should have load/save/update and that is it. I don't like this api. maybe prune as it is update in a way.

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __init__(
+        self,
+        target_clusters: int = ClusteringConfig.DEFAULT_TARGET_CLUSTERS,
+        min_cluster_size: int = ClusteringConfig.DEFAULT_MIN_CLUSTER_SIZE,
+    ) -> None:
+        self.target_clusters = target_clusters
+        self.min_cluster_size = min_cluster_size
+
+    def cluster(self, graph: CallGraph) -> ClusterResult:
```
ivanmilevtues, line 26: we can move most if not all of the engine here and rename the service mayeb to an engine and again keep just cluster as public method and keep the other things as private methods.

static_analyzer/clustering/models.py
```diff
@@ -0,0 +1,68 @@
...
+        return self.clusters.get(cluster_id, set())
+
+    def visit_paths(self, fn: Callable[[str], str]) -> None:
+        self.cluster_to_files = {cid: {fn(path) for path in paths} for cid, paths in self.cluster_to_files.items()}
+        remapped_file_to_clusters: dict[str, set[int]] = defaultdict(set)
+        for path, cluster_ids in self.file_to_clusters.items():
+            remapped_file_to_clusters[fn(path)].update(cluster_ids)
+        self.file_to_clusters = dict(remapped_file_to_clusters)
+
+    def pruned_to(self, surviving_nodes: Mapping[str, Node]) -> ClusterResult:
```
ivanmilevtues: name just prune

ivanmilevtues:
Superseded by a stack of nine reviewable PRs, per the 300/500-line target. Closing in favour of **#474**, which is the base of the chain.

| PR | Diff | What |
|---|---|---|
| #474 → `main` | 358 | Rendering leaves `CallGraph` |
| #475 → 474 | 176 | Clustering result types get their own package (+ cache tag bump) |
| #476 → 475 | 517 | Leiden search leaves `CallGraph` |
| #478 → 476 | 136 | Candidate machinery collapses to one seeded run |
| #479 → 478 | 574 | **Partition moves off `CallGraph` onto `LanguageResults`** |
| #480 → 479 | 266 | `graph.py` becomes the `cfg` package |
| #481 → 480 | 83 | ⚠️ One `to_networkx`, reference-augmented by default |
| #482 → 481 | 113 | ⚠️ Skipped nodes dropped from rendered call targets |
| #483 → 482 | 778 | Review fixes: typed reference edges, ordering, dead code |

Each PR targets the one above it, so GitHub shows only that step's diff. Review top-down; merging #474 auto-retargets #475 onto `main`.

**The two ⚠️ PRs change behaviour** and are labelled as such — #481 shifts orphan cluster assignment, #482 changes what the LLM sees in component-scoped prompts. Both want an eval run before merge.

The twelve inline comments here are anchored to lines that moved, so they could not be carried across. Every one of them is addressed somewhere in the stack:

- module docstring duplication, function-local imports → #483
- `static_analyzer.graph` compat shim → rejected; the package is internal and both cited references are generated artifacts
- `size_limit` on the class-level fallback → closed here as working as intended
- skipped destinations in the renderer → #482
- `add_reference_edge` taking an object, public-methods-first, `pruned_to`→`prune`, `canonical.py`→`location_key.py` → #483 (and #480 for the file rename)
- `ClusterCache` reduced to a data object, `engine` folded into the service → **not in this stack**; queued as follow-up API work
- cache should be hidden behind `CallGraph` → filed as #473

The `_TAG_VERSION` v3 bump stays, per your call on #475 — old baselines invalidate deliberately rather than failing with a `ModuleNotFoundError` swallowed into a silent cache miss.

PR #474, opened 2026-08-14
refactor(agents): move CFG rendering out of CallGraph into llm_renderers

ivanmilevtues:
**Stack 1 of ~8** — decomposing #472 by concern. Base: `main`.

Prompt shape is an agent concern, not a graph one: `CallGraph` should not know how it gets described to an LLM. `llm_str` and its two level helpers become module functions in `agents/llm_renderers`, taking the graph as a parameter.

**Pure move**, with one deliberate exception: the class-level path used to log `[CallGraph] Class-level summary: N characters` twice — once in `llm_str`, once again at the end of `_llm_str_class_level` — and now logs it once. Same logic, same output. `__str__` stays on the class — it is a debugging repr, not a prompt.

`graph.py`: 811 → 717 lines.

### Notes
- `agents.llm_renderers` added to the `pyproject.toml` package list; the existing `test_pyproject_packages` guard caught the omission.
- The four rendering tests moved to `tests/agents/test_llm_renderers.py`, then two were fixed in review: nodes are built with `NodeType.FUNCTION` instead of a bare `12`, and `test_functions_render_as_functions_at_class_level` now uses `size_limit=10` so it actually reaches the class-level renderer (at `100` the 93-char detailed render returned early, and the bare `"Function"` assertion passed on the wrong branch).

### Checks

### Stack
1. **this PR** — rendering leaves the graph
2. #475 — clustering result types move to their own package
3. #476 — Leiden search leaves the graph

## Summary by CodeRabbit

* **New Features**
  * Added clearer call-graph output grouped by file and method.
  * Large call graphs now automatically switch to a concise class-level summary.
  * Skipped nodes are excluded from displayed graph counts while preserving relevant call targets.
  * Control-flow graph results now use consistent call-graph rendering across project and component views.

* **Bug Fixes**
  * Improved readability and usability of graph results for large projects.

* **Tests**
  * Added coverage for detailed rendering, size-based summaries, function labels, and skipped nodes.


agents/llm_renderers/call_graph.py
```diff
@@ -0,0 +1,104 @@
...
+    logger.info(f"[CFG Tool] LLM string: {len(detailed)} characters, size limit: {size_limit} characters")
+    if len(detailed) <= size_limit:
+        return detailed
+
+    logger.info(
+        f"[CallGraph] Control flow graph is too large ({len(detailed)} chars), switching to class-level summary."
+    )
+    class_level = _render_class_level(graph, skip_set)
+    logger.info(f"[CallGraph] Class-level summary: {len(class_level)} characters")
+    return class_level
```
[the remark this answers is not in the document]
ivanmilevtues: we have been operating like this forever and there is no fallback for class level so i'd rather keep it as is. for now and when we start improving we might add this.

tests/agents/test_llm_renderers.py
```diff
@@ -0,0 +1,65 @@
...
+            graph.add_node(Node(f"class{i % 5}.ClassA.method{i}", NodeType.METHOD, "/file.py", i * 10, i * 10 + 5))
+        for i in range(49):
+            graph.add_edge(f"class{i % 5}.ClassA.method{i}", f"class{(i+1) % 5}.ClassA.method{i+1}")
+
+        result = render_call_graph(graph, size_limit=100)
+
+        self.assertIn("class-level summary", result)
+        self.assertIn("Class", result)
+
+    def test_functions_render_as_functions_at_class_level(self):
```
Svilen-Stefanov, line 35: The name says class-level, but do we ever get there? The detailed render for this graph is 94 chars and the limit is 100, so `render_call_graph` returns early on line 22 and `_render_class_level` never runs. The old test even carried a comment saying functions should *stay* in detailed format, so the rename kind of inverts it. Maybe drop the limit to something like 10 and assert on the summary line, or rename it back to something about the `Function` label?

agents/llm_renderers/call_graph.py
```diff
@@ -0,0 +1,104 @@
...
+    detailed = _render_detailed(graph, skip_set)
+    logger.info(f"[CFG Tool] LLM string: {len(detailed)} characters, size limit: {size_limit} characters")
+    if len(detailed) <= size_limit:
+        return detailed
+
+    logger.info(
+        f"[CallGraph] Control flow graph is too large ({len(detailed)} chars), switching to class-level summary."
+    )
+    class_level = _render_class_level(graph, skip_set)
+    logger.info(f"[CallGraph] Class-level summary: {len(class_level)} characters")
```
Svilen-Stefanov, line 29: nit: the description says same log lines, but the class-level path used to log this twice (here and again at the end of `_llm_str_class_level`) and now logs it once. The dedupe is the right call, just worth not leaving "same log lines" in the body in case someone is counting them.

tests/agents/test_llm_renderers.py
```diff
@@ -0,0 +1,65 @@
...
+from agents.llm_renderers import render_call_graph
+from static_analyzer.constants import NodeType
+from static_analyzer.graph import CallGraph
+from static_analyzer.node import Node
+
+
+class TestRenderCallGraph(unittest.TestCase):
+    def test_small_graph_stays_detailed(self):
+        graph = CallGraph()
+        graph.add_node(Node("module.src", 12, "/file.py", 1, 10))
```
Svilen-Stefanov: nit: we import `NodeType` and use it on line 26, but most of the nodes here are still built with a bare `12`. Since the file is new anyway, can we use the enum everywhere?

Svilen-Stefanov (approved):
Nice, this one is easy to check. I diffed the moved bodies against main and they are the same apart from self. becoming graph., the hoisted delimiter local and pulling 2_500_000 out into DEFAULT_SIZE_LIMIT, and llm_str genuinely has no callers left. Good that the pyproject packages guard caught the missing entry.

One renamed test does not reach the path its new name promises, rest is nits.

tests/agents/test_llm_renderers.py
ivanmilevtues: done thanks

tests/agents/test_llm_renderers.py
ivanmilevtues: good catch fixed.

PR #475, opened 2026-08-14
refactor(clustering): move the clustering result types into their own package

ivanmilevtues:
**Stack 2 of ~8** — base: `stack/1-llm-renderers` (#474). Review that one first.

Types only — no algorithm moves here.

`ClusterResult` and `METHOD_LEVEL_STRATEGY` describe the *output* of clustering, and `MethodClusterPaths` tracks each method's scoped lineage. None of the three is graph structure. They become `static_analyzer/clustering`, which `graph.py` now imports from rather than declares.

The Leiden search itself stays on `CallGraph` and moves in the next PR, so this is a pure relocation of type declarations plus the import updates that follow.

**176 changed lines** across 37 files — 35 of those files are a single import line each.

### Checks


static_analyzer/graph.py
```diff
@@ -8,9 +8,10 @@
 import networkx as nx
 import networkx.algorithms.community as nx_comm
 
+from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
```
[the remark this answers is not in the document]
ivanmilevtues: we should just bump the version of the cache because we either way will play with it.

static_analyzer/analysis_cache.py
```diff
@@ -52,8 +52,12 @@
 _LEGACY_CACHE_SUBDIR = "cache"
 # Tag file format prefix; bump if the on-disk pickle layout changes.
 # v2: StaticAnalysisResults switched from dict-of-dicts to LanguageResults
-# dataclass storage. v1 pickles will be treated as cache misses and re-run.
-_TAG_VERSION = "v2"
+# dataclass storage.
+# v3: MethodClusterPaths moved to static_analyzer.clustering. Pickle records a
+# class by module path, so a v2 pkl would raise ModuleNotFoundError on load —
+# the tag gate turns that into a clean, logged miss before anything unpickles.
+# Older pickles are treated as cache misses and re-run.
```
ivanmilevtues: this is because of package changes which the pkl will try to resolve directly.

static_analyzer/analysis_cache.py
```diff
@@ -52,8 +52,12 @@
 _LEGACY_CACHE_SUBDIR = "cache"
 # Tag file format prefix; bump if the on-disk pickle layout changes.
 # v2: StaticAnalysisResults switched from dict-of-dicts to LanguageResults
-# dataclass storage. v1 pickles will be treated as cache misses and re-run.
-_TAG_VERSION = "v2"
+# dataclass storage.
+# v3: MethodClusterPaths moved to static_analyzer.clustering. Pickle records a
+# class by module path, so a v2 pkl would raise ModuleNotFoundError on load —
```
Svilen-Stefanov: just checking on the wording here: the gate only runs on the paths that read the tag, and `_get_unlocked` reads it only `if expected_sha is not None` (line 187). So a bare `get()` skips it and goes straight to the unpickle, where a v2 pkl raises `ModuleNotFoundError` into the broad except and comes out as a warning instead. `load_with_sha` and `get(expected_sha=...)` do gate properly, so this is mostly about the comment promising more than the code does. Maybe say "on the SHA-gated paths"?

static_analyzer/graph.py
```diff
@@ -8,9 +8,10 @@
 import networkx as nx
```
Svilen-Stefanov, line 8: nit: just above this, `field` on line 4 is unused now, the only `field(default_factory=...)` uses in here went with `ClusterResult`. Nothing will catch it either, pre-commit only runs black.

tests/agents/test_abstraction_agent.py
```diff
@@ -185,7 +185,7 @@ def test_step_final_analysis(self, mock_invoke_repair_validate):
         mock_invoke_repair_validate.return_value = mock_response
 
         # Create mock cluster_results
-        from static_analyzer.graph import ClusterResult
+        from static_analyzer.clustering import ClusterResult
```
Svilen-Stefanov: nit: this duplicates the module-level import you added on line 18, and the migration was already touching the line. Can we just delete it (and the comment above it)?

Svilen-Stefanov (approved):
Clean one. The moved types are byte-identical, `method_cluster_paths.py` is a 100% rename, and I grepped for stragglers importing the old paths and found none. Good that the tag bump got reasoned through instead of skipped, the version history next to `_TAG_VERSION` is genuinely useful.

One heads-up rather than a change request: once this ships, every committed `.codeboarding` baseline out there is a guaranteed cache miss for one run, because the pickle references `static_analyzer.method_cluster_paths`. The action catches that and degrades to a full run, but CLI `incremental` users get an `IncrementalCacheMissingError` traceback rather than the clean "run a full analysis" prompt. Worth a line in the release notes?

One nit that has no line to sit on since the file is a pure rename: `record`'s `cluster_result` param in `method_cluster_paths.py:30` is still untyped, and the import cycle that forced that is exactly what this PR removes. Can we annotate it now?

PR #476, opened 2026-08-14
refactor(clustering): move the Leiden search out of CallGraph into the engine

ivanmilevtues:
**Retargeted to `main`.** #475 merged into `stack/1-llm-renderers` after that branch had already delivered to `main`, so the type move never reached `main`. This PR now carries it: the first three commits are #475 (including its review fixes), the last two are the Leiden move and its review fixes.

Rebased onto `main` rather than just repointing the base — #474 landed as a squash, so the original commits are not ancestors of `main` and a bare retarget would have shown #474's changes as new.

Pure move — the algorithm is byte-for-byte what it was.

The search never needed the graph: it reads an exported `nx.DiGraph` and a delimiter, nothing else. Eight private `CallGraph` methods become module functions in `clustering/engine.py` behind `cluster_graph()`, and `CallGraph.cluster()` becomes a four-line call that still owns the cache.

`detect_communities` goes with them. It was a pass-through to `leiden_utils.find_partition`, and leaving a clustering entry point on a module that no longer clusters would defeat the point. `cluster_helpers` and the determinism test call `find_partition` directly now; that test was named for the wrapper, so it is renamed for what it actually exercises.

The partition is still cached on `CallGraph` — that moves next.

`graph.py`: 686 → 447 lines.

### On the size

517 changed lines, over the 500 target. 497 of those are **one move counted on both sides** — 242 lines appearing in `engine.py`, 240 leaving `graph.py`. Read `engine.py` once and confirm `graph.py` lost the same thing. Getting under 500 would mean splitting the algorithm across two PRs, which makes review worse rather than better.

### Also in here
- `_abstract_node_name` had a `"package"` branch the level ladder never passes; it was unreachable in every branch of the stack, so it is removed here rather than deferred. `level` is now a `Level` StrEnum — strategy names are unchanged, so no cached strategy string moves.
- `find_partition`'s docstring lost its subject when this PR deleted `detect_communities`; the orphaned fragment is dropped.
- `cluster_graph` is out of the clustering package `__init__`; its one caller already used the submodule path.
- New `tests/static_analyzer/test_clustering_engine.py` — six tests driving `cluster_graph` with bare `nx.DiGraph`s, which is what the move makes possible and nothing else in the stack does. Covers the connected-components fallback, which had no coverage at all.

### Checks
`2065 passed, 28 skipped` · mypy clean (323 files) · black clean

## Summary by CodeRabbit

* **New Features**
  * Added enhanced graph clustering with raw, class-level, and file-level analysis.
  * Improved clustering reliability with deterministic results and connected-component fallback.
  * Added richer cluster results, file mappings, and method lineage tracking.
* **Bug Fixes**
  * Updated cached analysis data handling to prevent incompatibility after clustering changes.
* **Tests**
  * Added coverage for empty graphs, disjoint clusters, file mappings, opaque names, and fallback behavior.


static_analyzer/clustering/__init__.py
```diff
@@ -1,14 +1,16 @@
...
 
+from static_analyzer.clustering.engine import cluster_graph
 from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
 from static_analyzer.clustering.models import METHOD_LEVEL_STRATEGY, ClusterResult
 
 __all__ = [
     "METHOD_LEVEL_STRATEGY",
     "ClusterResult",
     "MethodClusterPaths",
+    "cluster_graph",
```
ivanmilevtues: this is the public api so cool to see it is a single method for now

static_analyzer/clustering/engine.py
```diff
@@ -0,0 +1,242 @@
...
+        return delimiter.join(parts[:-1])
+    elif level == "file" and len(parts) > 2:
+        return delimiter.join(parts[:-2])
+    elif level == "package" and len(parts) > 3:
+        return parts[0]
+    else:
+        return node_name
+
+
+def _cluster_with_algorithm(graph: nx.DiGraph, algorithm: str) -> list[set[str]]:
```
Svilen-Stefanov, line 96: Do we want to carry `_cluster_with_algorithm` into the new file? Its only caller (line 168) passes the literal `"leiden"`, so the louvain / greedy_modularity branches and the `nx_comm` import exist only to serve a branch that cannot run. And #478 deletes all of it two PRs later, so reviewers here spend their attention on three functions that do not survive the stack. Feels like the collapse belongs in the PR that creates the module?

static_analyzer/clustering/engine.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+def _abstract_node_name(node_name: str, level: str, delimiter: str) -> str:
+    parts = node_name.split(delimiter)
+
+    if level == "class" and len(parts) > 1:
+        return delimiter.join(parts[:-1])
+    elif level == "file" and len(parts) > 2:
+        return delimiter.join(parts[:-2])
+    elif level == "package" and len(parts) > 3:
```
Svilen-Stefanov: Again, the `"package"` branch cannot run, the ladder on line 37 is only `[None, "class", "file"]`. And while we are in here, could `level` be a StrEnum instead of a bare `str`? Then the closed set documents itself and the stringly comparisons go.

static_analyzer/clustering/__init__.py
```diff
@@ -1,14 +1,16 @@
-"""Clustering of call graphs: the result types and method lineage.
+"""Clustering of call graphs: the search, the result types, and method lineage.
 
-``models`` holds the result types; ``method_cluster_paths`` tracks each method's
-scoped cluster ancestry.
+``engine`` is the search over an exported ``nx.DiGraph``; ``models`` holds the
+result types; ``method_cluster_paths`` tracks each method's scoped cluster ancestry.
 """
 
+from static_analyzer.clustering.engine import cluster_graph
```
Svilen-Stefanov: I wonder if `cluster_graph` should be in the package `__init__`. It means every `from static_analyzer.clustering import ClusterResult` now pulls igraph and leidenalg in through `engine` -> `leiden_utils`, and there are six modules importing this package only for the dataclass. The one real caller already imports it from the submodule path, so the re-export is not buying much.

static_analyzer/leiden_utils.py
```diff
@@ -43,7 +43,7 @@ def find_partition[T](
 ) -> list[set[T]]:
     """Run Leiden on the graph from singletons; return list of community sets.
 
-    The ``static_analyzer.graph.detect_communities`` public entry point
+    This is the public entry point
```
Svilen-Stefanov: nit: this reads a bit broken now, "This is the public entry point / forwards here." The subject was `detect_communities`, which this PR deletes. Can we just drop the fragment and keep the sentence about `RBConfigurationVertexPartition`?

static_analyzer/clustering/engine.py
```diff
@@ -0,0 +1,242 @@
...
+import networkx.algorithms.community as nx_comm
+
+from static_analyzer.clustering.models import ClusterResult
+from static_analyzer.constants import ClusteringConfig
+from static_analyzer.leiden_utils import find_partition
+
+logger = logging.getLogger(__name__)
+
+
+def cluster_graph(
```
Svilen-Stefanov, line 29: just checking, is a direct test for `cluster_graph` coming later in the stack? The docstring sells the move as being able to exercise the search on any graph, but everything still drives it through `CallGraph`, so nothing actually does that yet.

Svilen-Stefanov (approved):
Move itself checks out, I compared the eight bodies and they are identical apart from threading `delimiter` through. Good call retargeting the determinism test at `find_partition` rather than dropping it with the wrapper.

We are copying a chunk of unreachable code into a brand new file and then deleting it two PRs later in #478. If we are drawing the box here, could it be created without that?

Nice one with the single `cluster_graph` API.

static_analyzer/clustering/engine.py
ivanmilevtues: fixed in the next PR

static_analyzer/leiden_utils.py
ivanmilevtues: Fixed in fd1ca65a — dropped the fragment, kept the `RBConfigurationVertexPartition` sentence, exactly as suggested.

Worth noting it was this PR that broke it (deleting `detect_communities` orphaned the subject), and it does eventually get repaired — but not until `stack/9-review-fixes`, six PRs downstream. That would have meant six merges carrying "This is the public entry point / forwards here." in the history, so it belongs here.

static_analyzer/clustering/engine.py
ivanmilevtues: Both done in fd1ca65a.

The `"package"` branch was dead in every version of the stack, not just this one — I checked all seven downstream branches and it is still there at the tip, so nothing was ever going to remove it. Gone now.

`level` is a `StrEnum` (`Level.RAW` / `CLASS` / `FILE`). `RAW` is the empty string, so the falsy checks at the two call sites read naturally and the `str | None` goes away entirely. Strategy names are unchanged — `StrEnum` formats as its value, so results still read `leiden_level_class` and no cached strategy string moves.

static_analyzer/clustering/engine.py
ivanmilevtues: No, nothing in the stack ever tested it directly — I checked by module path and function name across all seven downstream branches. The tip does add `tests/static_analyzer/test_clustering.py`, but every clustering test there goes through `ClusteringService().cluster(CallGraph)`; its one bare-`nx` test covers `find_partition`, a different function. So you were right that the docstring was selling something nothing demonstrated.

Added `tests/static_analyzer/test_clustering_engine.py` in fd1ca65a — six tests driving `cluster_graph` with hand-built `nx.DiGraph`s, no `CallGraph` anywhere: clustering a bare digraph, cliques staying intact, the empty graph, the `file_path` attribute driving the file mappings, undelimited node names, and the connected-components fallback.

Two things that fell out of writing it: the fallback path at `engine.py:71-76` had no coverage at all, and the tests pass unchanged against the stack tip (verified in a scratch worktree), so they survive the collapse in the next PR.

static_analyzer/clustering/__init__.py
ivanmilevtues: Dropped in fd1ca65a — `__all__` is now just the three types.

One correction though: the igraph/leidenalg part does not actually hold. `static_analyzer/__init__.py:19` imports `CallGraph`, and `graph.py` imports `clustering.engine` → `leiden_utils`, so importing *any* `static_analyzer.*` module already pulls both in regardless of this `__init__`. I verified — `import static_analyzer` alone loads igraph and leidenalg.

Removed it anyway on the half that does hold: the re-export was not buying anything, since `cluster_graph`'s only caller (`graph.py:9`) already imports it from the submodule path. Also 14 production modules import the package for `ClusterResult`, not six.

static_analyzer/clustering/engine.py
ivanmilevtues: Confirmed deferred — `stack/4-clustering-collapse` (`3d1084fb`, "collapse the candidate machinery to one seeded Leiden run") deletes `_cluster_with_algorithm` outright, along with the `louvain`/`greedy_modularity` branches and the now-unused `nx_comm` import, and replaces `_try_all_algorithms` with `_leiden_candidate`. Its follow-up `542b6c5d` also retires the "can be exercised on any graph" module docstring.

Taking your point on drawing the box without the dead code: I did remove the other unreachable branch (`"package"`) here rather than deferring it, because unlike this one it was never scheduled for removal anywhere in the stack.

ivanmilevtues:
Review fixes pushed in fd1ca65a. Answering the approval note as well:

> We are copying a chunk of unreachable code into a brand new file and then deleting it two PRs later in #478. If we are drawing the box here, could it be created without that?

Split it by whether the dead code was ever scheduled to die:

- **`_cluster_with_algorithm` (louvain / greedy_modularity / `nx_comm`)** — left as-is. `stack/4-clustering-collapse` is dedicated to exactly this collapse, so removing it here would duplicate that PR's whole diff and conflict on rebase.
- **The `"package"` branch in `_abstract_node_name`** — removed here. It is unreachable in *every* branch of the stack including the tip, so no later PR was going to clean it up.

Also in this commit:
- `level` is now a `StrEnum` (`Level.RAW` / `CLASS` / `FILE`); strategy names are unchanged, so no cached strategy string moves.
- `find_partition`'s orphaned docstring fragment is dropped — this PR is what broke it, and it otherwise stayed broken until `stack/9-review-fixes`.
- `cluster_graph` is out of the package `__init__`.
- New `tests/static_analyzer/test_clustering_engine.py` — six tests driving `cluster_graph` with bare `nx.DiGraph`s, which is the thing the move makes possible and nothing else in the stack does.

`1475 passed, 24 skipped` · mypy clean · black clean.

PR #478, opened 2026-08-14
refactor(clustering): collapse the candidate machinery to one seeded Leiden run

ivanmilevtues:
**Stack 4 of ~9** — base: `stack/3-cluster-engine` (#476).

The search carried plumbing for choosing between algorithms it never chose between. `_try_all_algorithms` only ever called leiden; `_cluster_with_algorithm`'s `louvain` and `greedy_modularity` branches were unreachable; and the candidate list never held more than one element per level.

Collapsed: `_leiden_candidate` returns a single scored `Candidate`, and `_map_candidates_to_original` loses its loop. `cluster_graph` takes keyword-only options so a caller cannot silently transpose `delimiter` and `target_clusters`.

`engine.py`: 242 → 221 lines. **136 changed lines.**

### Why this is its own PR

This restructure was bundled inside the original #472 alongside the file split, where it read as relocation. It isn't — it changes the shape of the search, so it gets reviewed on its own.

### Behaviour is preserved

The old code appended nothing to `all_candidates` when Leiden raised; the new one appends a zero-score candidate. Neither can win, because selection requires `best_score > 0.0`, so an all-levels-failed run still falls through to connected components exactly as before. The only observable difference is a debug log line on the failure path.

### Checks

## Summary by CodeRabbit

* **Improvements**
  * Updated graph clustering to use a consistent seeded approach across raw, class, and file levels.
  * Improved scoring and fallback behavior when clustering encounters failures.
  * Added strategy information to clustering results for clearer reporting.
* **Compatibility**
  * Updated clustering configuration usage to require named parameters, reducing ambiguity and improving reliability.

### DIFF
--- a/static_analyzer/clustering/engine.py
+++ b/static_analyzer/clustering/engine.py
@@ -1,8 +1,4 @@
-"""Seeded-Leiden clustering with a level-up search over an exported graph.
-
-Operates on ``nx.DiGraph`` only — it never sees a ``CallGraph``, so the search
-can be exercised on any graph. ``ClusteringService`` does the export.
-"""
+"""Seeded-Leiden clustering with a level-up search over a graph the caller exports."""
 
 from __future__ import annotations
 
--- a/static_analyzer/clustering/engine.py
+++ b/static_analyzer/clustering/engine.py
@@ -1,21 +1,25 @@
-"""Clustering search over an exported graph.
+"""Seeded-Leiden clustering with a level-up search over an exported graph.
 
-Operates on ``nx.DiGraph`` only — it never sees a ``CallGraph``, so the search can
-be exercised on any graph. The caller does the export.
+Operates on ``nx.DiGraph`` only — it never sees a ``CallGraph``, so the search
+can be exercised on any graph. ``ClusteringService`` does the export.
 """
+
+from __future__ import annotations
 
 import logging
 from collections import defaultdict
 from enum import StrEnum
 
 import networkx as nx
-import networkx.algorithms.community as nx_comm
 
 from static_analyzer.clustering.models import ClusterResult
 from static_analyzer.constants import ClusteringConfig
 from static_analyzer.leiden_utils import find_partition
 
 logger = logging.getLogger(__name__)
+
+# (communities, strategy name, score)
+Candidate = tuple[list[set[str]], str, float]
 
 
 class Level(StrEnum):
@@ -28,22 +32,23 @@ class Level(StrEnum):
 
 def cluster_graph(
     nx_graph: nx.DiGraph,
-    delimiter: str,
+    *,
+    delimiter: str = ClusteringConfig.QUALIFIED_NAME_DELIMITER,
     target_clusters: int = ClusteringConfig.DEFAULT_TARGET_CLUSTERS,
     min_cluster_size: int = ClusteringConfig.DEFAULT_MIN_CLUSTER_SIZE,
 ) -> ClusterResult:
-    """Cluster the graph using a try-all-then-level-up approach.
+    """Cluster ``nx_graph``, levelling up until the partition covers enough of it.
 
-    Flow: try all algorithms at each abstraction level (raw, class, file).
-    If coverage >= 50% at any level, stop and return the best result.
-    Falls back to connected components if everything fails.
+    Scores a Leiden partition at each abstraction level (raw, class, file) and stops
+    at the first to reach ``MIN_COVERAGE_RATIO``, else keeps the best-scoring level.
+    Falls back to connected components if every level scores zero.
     """
     if nx_graph.number_of_nodes() == 0:
         logger.warning("No nodes available for clustering.")
         return ClusterResult(strategy="empty")
 
     total_nodes = nx_graph.number_of_nodes()
-    all_candidates: list[tuple[list[set[str]], str, float]] = []
+    all_candidates: list[Candidate] = []
 
     for level in Level:
         work_graph = nx_graph
@@ -52,30 +57,23 @@ def cluster_graph(
             if work_graph.number_of_nodes() == 0:
                 continue
 
-        candidates = _try_all_algorithms(work_graph, min_cluster_size, total_nodes)
-
+        candidate = _leiden_candidate(work_graph, min_cluster_size, total_nodes)
         if level:
-            candidates = _map_candidates_to_original(
-                candidates, nx_graph, level, delimiter, min_cluster_size, total_nodes
-            )
+            candidate = _map_candidate_to_original(candidate, nx_graph, level, delimiter, min_cluster_size, total_nodes)
 
-        all_candidates.extend(candidates)
+        all_candidates.append(candidate)
 
-        # Check if best coverage at this level is good enough
-        if candidates:
-            best = max(candidates, key=lambda c: c[2])
-            best_coverage = _coverage(best[0], min_cluster_size, total_nodes)
-            logger.info(f"Level {level or 'raw'}: best={best[1]} score={best[2]:.3f} coverage={best_coverage:.3f}")
-            if best_coverage >= ClusteringConfig.MIN_COVERAGE_RATIO:
-                break
+        communities, strategy, score = candidate
+        coverage = _coverage(communities, min_cluster_size, total_nodes)
+        logger.info(f"Level {level or 'raw'}: best={strategy} score={score:.3f} coverage={coverage:.3f}")
+        if coverage >= ClusteringConfig.MIN_COVERAGE_RATIO:
+            break
 
-    # Pick overall best
     if all_candidates:
         best_communities, best_strategy, best_score = max(all_candidates, key=lambda c: c[2])
         if best_score > 0.0:
             return _build_result(best_communities, best_strategy, min_cluster_size, nx_graph)
 
-    # Absolute fallback: connected components
     logger.warning("All clustering strategies scored 0, falling back to connected components")
     components = list(nx.connected_components(nx_graph.to_undirected()))
     return _build_result(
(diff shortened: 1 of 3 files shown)


static_analyzer/clustering/engine.py
```diff
@@ -1,74 +1,73 @@
-"""Clustering search over an exported graph.
+"""Seeded-Leiden clustering with a level-up search over an exported graph.
 
-Operates on ``nx.DiGraph`` only — it never sees a ``CallGraph``, so the search can
-be exercised on any graph. The caller does the export.
+Operates on ``nx.DiGraph`` only — it never sees a ``CallGraph``, so the search
+can be exercised on any graph. ``ClusteringService`` does the export.
```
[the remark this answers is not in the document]
ivanmilevtues: yea will fix this not a big issue i think.

static_analyzer/clustering/engine.py
```diff
@@ -1,74 +1,69 @@
...
-        # Check if best coverage at this level is good enough
-        if candidates:
-            best = max(candidates, key=lambda c: c[2])
-            best_coverage = _coverage(best[0], min_cluster_size, total_nodes)
-            logger.info(f"Level {level or 'raw'}: best={best[1]} score={best[2]:.3f} coverage={best_coverage:.3f}")
-            if best_coverage >= ClusteringConfig.MIN_COVERAGE_RATIO:
-                break
-
-    # Pick overall best
     if all_candidates:
```
Svilen-Stefanov: Is `if all_candidates:` ever false now? `_LEVELS[0]` is `""` so the `continue` never fires, and `_leiden_candidate` always returns a candidate even when Leiden throws. Feels like the last piece of the machinery this PR set out to collapse, could we drop the guard and unpack `max(...)` directly, leaving just the `best_score > 0.0` gate?

static_analyzer/clustering/engine.py
```diff
@@ -154,55 +136,49 @@ def _abstract_at_level(graph: nx.DiGraph, level: str, delimiter: str) -> nx.DiGr
...
+    analysis.json must reproduce on the next run.
     """
-    candidates: list[tuple[list[set[str]], str, float]] = []
+    communities: list[set[str]] = []
     try:
-        communities = _cluster_with_algorithm(graph, "leiden")
-        score = _score_clustering(communities, min_cluster_size, total_nodes)
-        candidates.append((communities, "leiden", score))
-        logger.debug(f"leiden: score={score:.3f}, clusters={len(communities)}")
+        communities = find_partition(graph, seed=ClusteringConfig.CLUSTERING_SEED)
```
Svilen-Stefanov, line 165: `_abstract_at_level` goes to the trouble of aggregating `edge_weights`, but we call `find_partition` without `weight="weight"`, so at class and file level a pair joined by 50 calls weighs the same as one joined by 1. `cluster_helpers.py:219` does pass it. Is one of the two dead? Happy to leave it for now since it predates the PR, but a one-line note would save the next person the trace.

static_analyzer/clustering/engine.py
```diff
@@ -1,74 +1,69 @@
...
+
+# (communities, strategy name, score)
+Candidate = tuple[list[set[str]], str, float]
+
 
 def cluster_graph(
     nx_graph: nx.DiGraph,
-    delimiter: str,
+    *,
+    delimiter: str = ClusteringConfig.QUALIFIED_NAME_DELIMITER,
```
Svilen-Stefanov: nit: do we want a default here? No caller relies on it, and constants.py:75 explicitly anticipates a per-language delimiter. If that lands and someone forgets the argument, we would split on `"."` for a `::` language and silently hand back the raw level instead of blowing up, which is the thing AGENTS.md section 5 asks us not to do. Keyword-only without a default still prevents the transposition you were worried about.

static_analyzer/clustering/engine.py
```diff
@@ -154,55 +136,49 @@ def _abstract_at_level(graph: nx.DiGraph, level: str, delimiter: str) -> nx.DiGr
...
     try:
-        communities = _cluster_with_algorithm(graph, "leiden")
-        score = _score_clustering(communities, min_cluster_size, total_nodes)
-        candidates.append((communities, "leiden", score))
-        logger.debug(f"leiden: score={score:.3f}, clusters={len(communities)}")
+        communities = find_partition(graph, seed=ClusteringConfig.CLUSTERING_SEED)
     except Exception as e:
-        logger.debug(f"Algorithm leiden failed: {e}")
-    return candidates
+        logger.debug(f"Leiden failed: {e}")
```
Svilen-Stefanov: I wonder if `debug` is right here. If `find_partition` throws at every level we ship connected-components clusters that look like a perfectly normal result, and at the default log level there is nothing saying why. `warning` with `exc_info=True` would at least leave a trace, the catch itself is fine.

static_analyzer/clustering/engine.py
```diff
@@ -1,74 +1,69 @@
...
 from static_analyzer.constants import ClusteringConfig
 from static_analyzer.leiden_utils import find_partition
 
 logger = logging.getLogger(__name__)
 
+# Abstraction levels tried in order; "" is the raw method-level graph.
+_LEVELS: tuple[str, ...] = ("", "class", "file")
+
+# (communities, strategy name, score)
+Candidate = tuple[list[set[str]], str, float]
```
Svilen-Stefanov: nit: could `Candidate` be a NamedTuple? Then `c.score` replaces `c[2]` on line 63 and the comment doing the type's job goes away. `strategy` can also only ever be `"leiden"` now, so we could hardcode the label instead of round-tripping it.

Svilen-Stefanov (approved):
Nice, lgtm. Left a couple of nits / questions.

static_analyzer/clustering/engine.py
ivanmilevtues: Right on both counts — `Level.RAW` is `""` so the `continue` cannot fire before the first append, and `_leiden_candidate` returns a scored candidate even when Leiden throws. The guard was unreachable.

Dropped in 98b7169d; `max(...)` is unpacked directly and `best_score > 0.0` is the only gate left, with a one-line note saying why the list is never empty.

static_analyzer/clustering/engine.py
ivanmilevtues: Both done in 98b7169d.

`Candidate` is a `NamedTuple` carrying `communities`, `level`, `score` — so `c.score` replaces `c[2]` and the `# (communities, strategy name, score)` comment is gone.

On the label: rather than hardcoding `"leiden"` in the tuple I derived the whole strategy name from the level (`strategy` property — `"leiden"` for RAW, `f"leiden_level_{level}"` otherwise). That drops `strategy` from the round-trip entirely and let `_map_candidate_to_original` lose a parameter, since the level now travels with the candidate.

Names are byte-identical (`leiden`, `leiden_level_class`, `leiden_level_file`) — they are persisted in analysis.json, so I added a test pinning them rather than trusting the refactor.

static_analyzer/clustering/engine.py
ivanmilevtues: Agreed, changed to `warning` with `exc_info=True` in 98b7169d, and the message names the level that failed.

Your reasoning is the comment I left on it: at `debug`, a graph that throws at every level ships connected-components clusters that look like a perfectly ordinary result with nothing at the default log level explaining why.

static_analyzer/clustering/engine.py
ivanmilevtues: Good call, default removed in 98b7169d — keyword-only, no default.

Checked before changing it: the only production caller is `graph.py:371` (`CallGraph.cluster`), which passes `delimiter=self.delimiter` explicitly, and the engine tests all pass it too. So nothing relied on the default.

And the failure mode you describe is real rather than hypothetical: `_abstract_node_name` splits on the delimiter and returns `node_name` unchanged when the split yields too few parts, so a `::` language run with the `"."` default would not error — it would quietly produce a raw-level partition at every level. Exactly the silent degradation AGENTS.md §5 rules out.

static_analyzer/clustering/engine.py
ivanmilevtues: Confirmed, and it is the engine side that ignores them: `cluster_helpers.py:219` passes `weight="weight"`, `_leiden_candidate` calls `find_partition(graph, seed=...)` with no `weight=`. So `_abstract_at_level` aggregates counts that nothing reads, and at class/file level a pair joined by 50 calls does weigh the same as one joined by 1.

Neither is dead — they are just inconsistent. Taking your suggestion and leaving the behaviour alone with a note at the aggregation in 98b7169d:

```
# These weights are recorded but never read: _leiden_candidate calls find_partition
# without ``weight=``, so class and file levels cluster unweighted. Passing them would
# change every existing partition, so it needs its own change.
```

Flagging the reason it is not a one-line fix: passing `weight="weight"` changes the partition at class and file level for every repository, which moves cluster IDs and the strategy that wins. That wants its own PR with the eval numbers, not a rider on a refactor that is otherwise behaviour-preserving.

ivanmilevtues:
Review fixes in 98b7169d. All five inline comments addressed — replies on each thread.

| Comment | Resolution |
|---|---|
| `if all_candidates:` unreachable | Guard dropped, `max(...)` unpacked directly, `best_score > 0.0` is the only gate |
| `Candidate` as NamedTuple | Done, and the strategy label is now derived from the level rather than round-tripped |
| `delimiter` default | Removed; keyword-only without a default |
| Leiden failure logged at `debug` | `warning` with `exc_info=True`, message names the level |
| `edge_weights` never reaching `find_partition` | Left as-is with a note, per your suggestion |

The codex-bot P1 about `ClusteringService` in the module docstring was already fixed by this PR's own second commit (`739d1ac8`), which replaces it with the terse one-liner.

**Note this PR now targets `main`, not `stack/3-cluster-engine`.** #476 merged and its branch was deleted, which auto-closed this PR; it has been reopened and rebased onto `main`. The rebase had one real conflict — #476 landed a `Level` StrEnum over the same levels ladder this PR collapses into `_LEVELS`. Resolved by keeping the collapse and the StrEnum: `for level in Level`, with `_LEVELS` dropped as redundant.

Strategy names are unchanged throughout (`leiden`, `leiden_level_class`, `leiden_level_file`) — they are persisted in analysis.json, so there is now a test pinning them.

`1475 passed, 24 skipped` · mypy clean · black clean.

Issue #473, opened 2026-08-14
Hide ClusterCache behind CallGraph access


ivanmilevtues:
Follow-up from the #472 review discussion (decision D4 — deferred, not dropped).

## The ask

> the cache shouldn't be used we should always use the CallGraph and have the cache as something hidden behind it, or just used once when loading it and from then on it should be purely just the cfgs

Today `StaticAnalysisResults.get_clusters(language)` is reachable from anywhere, and callers across `diagram_analysis/`, `agents/` and `static_analyzer/` reach for the cache directly rather than working from the loaded objects. The cache should be an I/O boundary — read once at load, written once at save — and invisible in between.

## Why it was deferred

It needs an audit of every `get_clusters()` caller and a decision on what replaces each one. That is a different question once `ClusterCache` has already been reduced to a dumb data object (`load`/`save`/`update`), which is queued ahead of it. Scoping this before that lands would produce the wrong answer.

## Scope when picked up

- Inventory every `get_clusters()` / `set_clusters()` call site and classify: genuine load/save boundary, or leaked internal access.
- Decide the replacement surface — most likely the partition travels with the `CallGraph` the caller already holds.
- `tree_shape.py` is the concrete example that prompted this.

## Depends on

- `ClusterCache` reduced to `load`/`save`/`update` (`adopt` + `record_scope` collapsing into one `update`).

PR #479, opened 2026-08-14
refactor(clustering): move the partition off CallGraph onto LanguageResults

ivanmilevtues:
**Stack 5 of ~9** — base: `stack/4-clustering-collapse` (#478). **The behavioural one; read this one slowly.**

A partition is a *result about* a graph, not part of it. `_cluster_cache` and `method_cluster_paths` become a `ClusterCache` owned by `LanguageResults`, reached through `StaticAnalysisResults.get_clusters` / `set_clusters`. `ClusteringService.cluster()` replaces `CallGraph.cluster()` and is pure — it neither mutates the graph nor caches.

**324 insertions / 250 deletions across 21 files.**

### Three consequences worth reviewing

**`filter`/`union` no longer silently prune a partition.** The warm-start path in `_run_warm_start` now carries and prunes it in one visible place. Losing a cluster baseline would be a visible deletion rather than an invisible one.

**`ControlFlowGraph.merge` stops merging lineage**, for the same reason — that work moved into the explicit prune above. Note this leaves `MethodClusterPaths.merge` with no callers; I've left it rather than widen this PR, and it belongs in the API cleanup.

**`check_component_cohesion` loses the side effect that kept it switched off.** `cluster()` used to seed `_cluster_cache` with a current-graph partition, which would masquerade as a historical snapshot in the next incremental delta. The service cannot do that. Still left unwired — turning a health check back on is a behaviour change, not a refactor.

### One inverted test

`test_cluster_is_cached` becomes `test_clustering_does_not_mutate_the_graph`. The guarantee genuinely inverted: from "the second call returns the same object" to "the second call returns an equal, freshly built one".

`graph.py`: 447 → 392 lines.

### Checks

## Summary by CodeRabbit

* **Improvements**
  * Improved clustering consistency across languages and analysis scopes.
  * Preserved cached clustering results more reliably during incremental updates and warm starts.
  * Enhanced snapshot generation to handle empty or partially available clustering data gracefully.
  * Improved component grouping and lineage handling after analysis changes.
* **Documentation**
  * Clarified cache requirements, fallback behavior, and clustering behavior.

### DIFF
--- a/health/checks/cohesion.py
+++ b/health/checks/cohesion.py
@@ -2,6 +2,7 @@ import logging
 
 from health.models import FindingEntity, FindingGroup, HealthCheckConfig, Severity, StandardCheckSummary
 from static_analyzer.graph import CallGraph
+from static_analyzer.clustering import ClusteringService
 
 logger = logging.getLogger(__name__)
 
@@ -16,15 +17,13 @@ def check_component_cohesion(call_graph: CallGraph, config: HealthCheckConfig) -
     cluster than inside it, suggesting the grouping may not reflect
     actual code organization.
 
-    NOT wired into ``health.runner`` today: the ``cluster()`` call below seeds
-    ``CallGraph._cluster_cache`` with a current-graph partition, which would
-    masquerade as a historical snapshot in the next incremental delta. Before
-    re-enabling, either reset ``_cluster_cache`` after the call or compute the
-    partition without touching the cache.
+    Not wired into ``health.runner`` today, but clustering here is now free of
+    side effects: ``ClusteringService`` returns a partition without touching any
+    language's ``ClusterCache``.
     """
     warning_entities: list[FindingEntity] = []
 
-    cluster_result = call_graph.cluster()
+    cluster_result = ClusteringService().cluster(call_graph)
     if not cluster_result.clusters:
         return StandardCheckSummary(
             check_name="component_cohesion",
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -327,7 +327,7 @@ class StaticAnalyzer:
         """Gracefully shut down all engine LSP server processes. Idempotent.
 
         Persists the latest ``_cached_results`` to the pkl on the way down so
-        downstream mutations (``CallGraph._cluster_cache`` populated by the
+        downstream mutations (the language's ``ClusterCache``, populated by the
         abstraction agent) reach disk in one save instead of two. Save errors
         are logged but never block teardown.
         """
@@ -657,8 +657,9 @@ class StaticAnalyzer:
         ``update_cfg_for_changed_files`` along with the language's portion of the
         cached state, and put the merged result back into a fresh
         ``StaticAnalysisResults``. Merging (rather than a full re-LSP) is what
-        preserves the cached CFG's ``_cluster_cache``, so the next incremental
-        run still finds a cluster baseline.
+        preserves the language's ``ClusterCache`` (carried across by ``select``,
+        keeping only surviving nodes), so the next incremental run still finds a
+        cluster baseline.
 
         Changed-file source: ``self.changed_files`` when set at construction
         (git-free — e.g. the wrapper's fingerprint diff), else ``git diff`` via
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -691,7 +691,7 @@ class StaticAnalyzer:
                 surviving = results.get_cfg(language).nodes
             except ValueError:
                 surviving = {}
-            results.set_clusters(language, cached_results.get_clusters(language).prune(surviving))
+            results.set_clusters(language, cached_results.get_clusters(language).select(surviving))
             self._collect_diagnostics_for(adapter, engine_client, analysis)
             track_lsp_result(
                 language=adapter.language_enum.value,
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -326,10 +326,9 @@ class StaticAnalyzer:
     def stop_clients(self) -> None:
         """Gracefully shut down all engine LSP server processes. Idempotent.
 
-        Persists the latest ``_cached_results`` to the pkl on the way down so
-        downstream mutations (the language's ``ClusterCache``, populated by the
-        abstraction agent) reach disk in one save instead of two. Save errors
-        are logged but never block teardown.
+        Persists ``_cached_results`` on the way down so the abstraction agent's
+        ``ClusterCache`` writes reach disk in one save. Save errors are logged
+        but never block teardown.
         """
         if not self._clients_started:
             return
@@ -656,10 +655,8 @@ class StaticAnalyzer:
         Per language: determine the changed-file list, hand it to
         ``update_cfg_for_changed_files`` along with the language's portion of the
         cached state, and put the merged result back into a fresh
-        ``StaticAnalysisResults``. Merging (rather than a full re-LSP) is what
-        preserves the language's ``ClusterCache`` (carried across by ``select``,
-        keeping only surviving nodes), so the next incremental run still finds a
-        cluster baseline.
+        ``StaticAnalysisResults``. The cached ``ClusterCache`` is grafted on either
+        way, so a language that fell back to a full re-LSP still leaves a baseline.
 
         Changed-file source: ``self.changed_files`` when set at construction
         (git-free — e.g. the wrapper's fingerprint diff), else ``git diff`` via
@@ -685,9 +682,9 @@ class StaticAnalyzer:
                 )
 
             self._absorb_into_results(results, language, analysis)
-            # Carry the cached partition onto the merged graph — nodes the re-LSP dropped
-            # fall out of the pruned copy. Without this the warm start would hand the next
-            # incremental run an empty cluster baseline.
+            # Both branches, including the full re-LSP: the partition describes cached_sha,
+            # which stays a valid delta baseline however the graph was rebuilt. select()
+            # drops whatever the re-LSP no longer has.
             try:
                 surviving = results.get_cfg(language).nodes
             except ValueError:
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -684,6 +684,14 @@ class StaticAnalyzer:
                 )
 
             self._absorb_into_results(results, language, analysis)
+            # Carry the cached partition onto the merged graph — nodes the re-LSP dropped
+            # fall out of the pruned copy. Without this the warm start would hand the next
+            # incremental run an empty cluster baseline.
+            try:
+                surviving = results.get_cfg(language).nodes
+            except ValueError:
+                surviving = {}
+            results.set_clusters(language, cached_results.get_clusters(language).prune(surviving))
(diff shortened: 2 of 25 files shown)


static_analyzer/language_results.py
```diff
@@ -162,10 +162,12 @@ class LanguageResults:
...
     source_files: SourceFiles = field(default_factory=SourceFiles)
+    clusters: ClusterCache = field(default_factory=ClusterCache)
 
     def visit_paths(self, fn: Callable[[str], str]) -> None:
         self.cfg.visit_paths(fn)
         self.hierarchy.visit_paths(fn)
         self.references.visit_paths(fn)
         self.dependencies.visit_paths(fn)
         self.source_files.visit_paths(fn)
+        self.clusters.visit_paths(fn)
```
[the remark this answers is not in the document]
ivanmilevtues: we haven;t releasd v3 so it is fine to keep it as is

static_analyzer/language_results.py
```diff
@@ -162,10 +162,12 @@ class LanguageResults:
     references: References = field(default_factory=References)
     dependencies: PackageDependencies = field(default_factory=PackageDependencies)
     source_files: SourceFiles = field(default_factory=SourceFiles)
+    clusters: ClusterCache = field(default_factory=ClusterCache)
```
ivanmilevtues, line 165: this i don't like why do we have the cache here it should be the actual Clusters. I think it is probably because we have the caching and the lagnugage results coupled. If that is the case this should be again added in the Github issue for caching if not already there. (https://github.com/CodeBoarding/CodeBoarding/issues/473)

static_analyzer/clustering/cache.py
```diff
@@ -0,0 +1,47 @@
...
+from collections.abc import Callable, Mapping
+from dataclasses import dataclass, field
+
+from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
+from static_analyzer.clustering.models import ClusterResult
+from static_analyzer.node import Node
+
+
+@dataclass
+class ClusterCache:
```
ivanmilevtues, line 19: mostly extracting the cahce as a first step there will be a follow-up where we will make it hidden from the API (https://github.com/CodeBoarding/CodeBoarding/issues/473)

static_analyzer/__init__.py
```diff
@@ -684,6 +684,14 @@ def _update_cached_results(
...
 
             self._absorb_into_results(results, language, analysis)
+            # Carry the cached partition onto the merged graph — nodes the re-LSP dropped
+            # fall out of the pruned copy. Without this the warm start would hand the next
+            # incremental run an empty cluster baseline.
+            try:
+                surviving = results.get_cfg(language).nodes
+            except ValueError:
+                surviving = {}
+            results.set_clusters(language, cached_results.get_clusters(language).prune(surviving))
```
ivanmilevtues: i think it is a bit confusing here it feels like we will prune away the surviving nodes, but we are in fact filtering to those if that is the case that this method should better be named filter.

static_analyzer/language_results.py
Svilen-Stefanov, line 165: Doesn't this change the on-disk pickle layout? `LanguageResults` gains `clusters` and `visit_paths` dereferences it unconditionally, but `_TAG_VERSION` stays v3 (analysis_cache.py is untouched here), so a pkl written by the previous stack branch still passes the gate and gets unpickled. I tried it: pickle restores dataclasses through `__dict__` so the field is absent rather than defaulted, and the load dies with `AttributeError: 'LanguageResults' object has no attribute 'clusters'` inside `_absolutize`, which the broad except turns into a warning and a None. So the outcome is fine (a miss and a re-index, not corrupt state), it just gets there by accident of a bare except instead of the clean pre-unpickle miss the tag exists for, and the warning reads like a bug. The file's own comment says bump when the layout changes, and #475 bumped for a strictly smaller change four commits ago. Bump to v4?

tests/static_analyzer/test_analysis_cache_inmem.py
```diff
@@ -187,32 +187,34 @@ def test_diagnostics_merge_with_new_winning(self) -> None:
 
 
 class TestClusterCachePreservation(unittest.TestCase):
-    """``_cluster_cache`` must survive warm-start invalidation/merge.
+    """The partition must survive warm-start invalidation/merge, pruned to the survivors."""
```
Svilen-Stefanov, line 191: These do not test production anymore I think. Every test in the class now builds a `ClusterCache` and calls `.select(...)` itself, so the assertions hold whatever the code does. The invariant they were guarding moved to `static_analyzer/__init__.py:694`, and if you delete that line the whole file still passes green (so does test_warmstart_changed_files.py, it only asserts which files get re-LSP'd). Could we add one against `_update_cached_results` instead?

static_analyzer/__init__.py
```diff
@@ -684,6 +684,14 @@ def _update_cached_results(
...
 
             self._absorb_into_results(results, language, analysis)
+            # Carry the cached partition onto the merged graph — nodes the re-LSP dropped
+            # fall out of the pruned copy. Without this the warm start would hand the next
+            # incremental run an empty cluster baseline.
+            try:
+                surviving = results.get_cfg(language).nodes
+            except ValueError:
+                surviving = {}
+            results.set_clusters(language, cached_results.get_clusters(language).select(surviving))
```
Svilen-Stefanov, line 692: just checking the `changed_files is None` case: that is the branch where change detection failed and we do a full re-LSP, and we now graft the cached partition onto a graph rebuilt from scratch. Before this, that language finished the warm start with no partition at all. Probably fine (the partition describes `cached_sha`, which is still a valid delta baseline no matter how the graph got rebuilt) but it is a behaviour change the body does not mention. If it is deliberate, worth saying so in the comment above?

static_analyzer/clustering/service.py
```diff
@@ -0,0 +1,33 @@
...
+class ClusteringService:
+    """Clusters a ``CallGraph``.
+
+    Pure: it neither mutates the graph nor caches anything. Callers that want to
+    keep a partition store it in the ``ClusterCache`` on their ``LanguageResults``.
+    """
+
+    def __init__(
+        self,
+        target_clusters: int = ClusteringConfig.DEFAULT_TARGET_CLUSTERS,
```
Svilen-Stefanov: nit: nobody passes either of these, all four call sites are `ClusteringService().cluster(...)` and the tests patch `cluster` rather than construct with arguments. So it is a stateless class holding two ints no one sets. Either a module-level `cluster(graph, target_clusters=..., min_cluster_size=...)`, or give it a caller that actually configures it?

health/checks/cohesion.py
```diff
@@ -16,15 +17,13 @@ def check_component_cohesion(call_graph: CallGraph, config: HealthCheckConfig) -
     cluster than inside it, suggesting the grouping may not reflect
     actual code organization.
 
-    NOT wired into ``health.runner`` today: the ``cluster()`` call below seeds
-    ``CallGraph._cluster_cache`` with a current-graph partition, which would
-    masquerade as a historical snapshot in the next incremental delta. Before
-    re-enabling, either reset ``_cluster_cache`` after the call or compute the
-    partition without touching the cache.
+    Not wired into ``health.runner`` today, but clustering here is now free of
```
Svilen-Stefanov, line 20: This ends up saying the opposite of `health/runner.py:103-108` in the same commit: that one still says the check is skipped because it calls `CallGraph.cluster()` and would seed `_cluster_cache`, and both of those are deleted here. Same stale name in `diagram_analysis/exceptions.py:11` and `static_analyzer/__init__.py:330,660`. Can we sweep them?

tests/diagram_analysis/test_cluster_snapshot.py
```diff
@@ -38,22 +37,24 @@ def _build_graph(node_specs: list[tuple[str, str]]) -> CallGraph:
     return graph
 
 
-def _build_static(graphs: dict[str, CallGraph]) -> StaticAnalysisResults:
+def _build_static(graphs: dict[str, CallGraph], partitions: dict[str, ClusterResult] = {}) -> StaticAnalysisResults:
```
Svilen-Stefanov: nit: mutable default. Harmless as written since it is only read, but we have no ruff/bugbear to catch it if someone later mutates it. `| None = None` or just make it required, two of the three callers pass it anyway.

Svilen-Stefanov (approved):
LGTM.

Two things to point out: the cache-preservation tests no longer touch production code, and I think this changes the on-disk pickle layout.

static_analyzer/clustering/service.py
ivanmilevtues: fair done

health/checks/cohesion.py
ivanmilevtues: cleaned the comments, only one cluster method is left in the clusterservice.
caching is still imperfect but that will be addressed with follow up prs based on the issue i opened

static_analyzer/__init__.py
ivanmilevtues: this runs for both branches, we caryy the pruned result and then .adopt the rest so the output is the same, just more explicit in its enrichment mechanism

tests/diagram_analysis/test_cluster_snapshot.py
ivanmilevtues: required now.

static_analyzer/language_results.py
ivanmilevtues: yes it has to stay v3, that is not released yet.

tests/static_analyzer/test_analysis_cache_inmem.py
ivanmilevtues: Confirmed by deleting the line rather than reading it: with the graft at `__init__.py:692` replaced by a `pass`, **all 27 tests in this file and `test_warmstart_changed_files.py` still pass**. You were right that nothing was guarding it.

Added `TestUpdateCachedResultsCarriesPartition` in 3d90e586 — two tests driving the real `_update_cached_results`, using the same `object.__new__(StaticAnalyzer)` harness as `test_warmstart_changed_files.py`. Cached state holds `{1: {a.foo}, 2: {b.qux}}`, the re-LSP returns a graph where only `b.qux` survives, and both assert the result carries `{2: {"b.qux"}}` — cluster 1 pruned, cluster 2 kept.

Mutation-checked in both directions, since a test that cannot fail is the thing you were objecting to:

```
graft deleted   → 2 failed, 23 passed   AssertionError: {} != {2: {'b.qux'}}
graft restored  → 25 passed
```

The second test covers the `changed_files is None` / full-re-LSP branch from your other comment, so that path is now pinned by a test rather than resting on a comment.

What is real in them: `_update_cached_results`, `_changed_files_for_language` choosing the branch, `_extract_language_dict`, `_absorb_into_results`, and `ClusterCache.select` over a real `CallGraph`. Stubbed: `update_cfg_for_changed_files`, `_run_full_analysis`, diagnostics and telemetry. So it binds the wiring, not the LSP path.

I left `TestClusterCachePreservation` in place — it is still useful coverage of `select` / `invalidate_files` / `merge_results` / `union` / `filter` — but you are right that its name and docstring ("must survive warm-start invalidation/merge") claim a layer it does not touch. Happy to rename it to say it covers the primitives if you would rather the file be explicit about that.

PR #480, opened 2026-08-14
refactor(cfg): graph.py becomes the cfg package

ivanmilevtues:
**Stack 6 of 9** — retargeted to `main`; #479 is merged.

What's left of `graph.py` is graph structure, so it moves to `static_analyzer/cfg` and splits along its three concerns: `call_graph.py` holds `CallGraph`, `edge.py` the edge types, `location_key.py` the physical-location identity used to dedup LSP aliases.

**50 files, 266 changed lines** — 47 of those files are a one-line import update. Git sees `call_graph.py` as a rename of `graph.py`, so the diff shows what actually moved rather than 300 lines deleted and 300 added.

### Checks
`2068 passed, 28 skipped` · mypy clean (328 files) · black clean

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated control-flow graph package with public support for call graphs, edges, edge types, and source-location identifiers.
  - Improved edge handling with call-site metadata, endpoint access, deduplication, and path transformation.
  - Included the new graph package in distributed builds.

- **Refactor**
  - Updated analysis tools, health checks, agents, diagnostics, and tests to use the reorganized graph interfaces without changing behavior.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -16,7 +16,7 @@ from static_analyzer.engine.lsp_client import LSPClient
 from static_analyzer.engine.result_converter import convert_to_codeboarding_format
 from static_analyzer.engine.source_inspector import SourceInspector
 from static_analyzer.engine.utils import uri_to_path
-from static_analyzer.graph import CallGraph
+from static_analyzer.cfg import CallGraph
 from static_analyzer.incremental_orchestrator import update_cfg_for_changed_files
 from static_analyzer.java_config_scanner import JavaConfigScanner
 from static_analyzer.lsp_client.diagnostics import FileDiagnosticsMap
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -7,6 +7,7 @@ from repo_utils.git_ops import get_changed_files_since
 from repo_utils.ignore import RepoIgnoreManager
 from static_analyzer.analysis_cache import StaticAnalysisCache
 from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.cfg import CallGraph
 from static_analyzer.constants import Language
 from static_analyzer.csharp_config_scanner import CSharpConfigScanner
 from static_analyzer.engine.adapters import get_adapter
@@ -16,7 +17,6 @@ from static_analyzer.engine.lsp_client import LSPClient
 from static_analyzer.engine.result_converter import convert_to_codeboarding_format
 from static_analyzer.engine.source_inspector import SourceInspector
 from static_analyzer.engine.utils import uri_to_path
-from static_analyzer.cfg import CallGraph
 from static_analyzer.incremental_orchestrator import update_cfg_for_changed_files
 from static_analyzer.java_config_scanner import JavaConfigScanner
 from static_analyzer.lsp_client.diagnostics import FileDiagnosticsMap
--- a/static_analyzer/analysis_cache.py
+++ b/static_analyzer/analysis_cache.py
@@ -32,7 +32,7 @@ from filelock import FileLock
 from constants import STATIC_ANALYSIS_PKL, STATIC_ANALYSIS_SHA
 from repo_utils.path_utils import to_absolute_path, to_relative_path
 from static_analyzer.analysis_result import AnalysisData, InvalidatedAnalysis, InvalidatedEdge
-from static_analyzer.graph import CallGraph, Edge, EdgeKind
+from static_analyzer.cfg import CallGraph, Edge, EdgeKind
 from static_analyzer.lsp_client.diagnostics import FileDiagnosticsMap
 from static_analyzer.node import Node
 
--- a/static_analyzer/analysis_cache.py
+++ b/static_analyzer/analysis_cache.py
@@ -53,8 +53,8 @@ _LEGACY_CACHE_SUBDIR = "cache"
 # Tag file format prefix; bump if the on-disk pickle layout changes.
 # v2: StaticAnalysisResults switched from dict-of-dicts to LanguageResults
 # dataclass storage.
-# v3: MethodClusterPaths moved to static_analyzer.clustering. Older pickles are
-# treated as cache misses and re-run.
+# v3: MethodClusterPaths moved to static_analyzer.clustering, then CallGraph and the
+# edge types to static_analyzer.cfg. Older pickles are treated as cache misses and re-run.
 _TAG_VERSION = "v3"
 
 
--- a/static_analyzer/cfg/__init__.py
+++ b/static_analyzer/cfg/__init__.py
@@ -1,0 +1,11 @@
+"""Call-graph structure: nodes, call edges, reference edges and subgraph derivation.
+
+``call_graph`` holds the graph itself, ``edge`` the edge types, and ``location_key``
+the physical-location identity used to dedup LSP qualified-name aliases.
+"""
+
+from static_analyzer.cfg.call_graph import CallGraph
+from static_analyzer.cfg.edge import Edge, EdgeKind
+from static_analyzer.cfg.location_key import LocationKey
+
+__all__ = ["CallGraph", "Edge", "EdgeKind", "LocationKey"]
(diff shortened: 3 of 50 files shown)


static_analyzer/cfg/call_graph.py
```diff
@@ -14,79 +16,6 @@
...
-        return normalized
-
-    def visit_paths(self, fn: Callable[[str], str]) -> None:
-        for site in self._call_sites:
-            if "file" in site:
-                site["file"] = fn(str(site["file"]))
-        self._call_site_keys = {tuple(sorted(site.items())) for site in self._call_sites}
-
-
 class CallGraph:
```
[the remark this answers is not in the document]
ivanmilevtues: we haven't released v3 so this is fine.

static_analyzer/cfg/location_key.py
```diff
@@ -0,0 +1,15 @@
+"""Physical-location identity, used to dedup the several qualified names the LSP
+can emit for one symbol. Issue #471 moves the canonicalization itself here."""
```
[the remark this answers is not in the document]
ivanmilevtues: fair we should mention it there not here and keep this just a code comment.

static_analyzer/analysis_cache.py
```diff
@@ -32,7 +32,7 @@
 from constants import STATIC_ANALYSIS_PKL, STATIC_ANALYSIS_SHA
 from repo_utils.path_utils import to_absolute_path, to_relative_path
 from static_analyzer.analysis_result import AnalysisData, InvalidatedAnalysis, InvalidatedEdge
-from static_analyzer.graph import CallGraph, Edge, EdgeKind
+from static_analyzer.cfg import CallGraph, Edge, EdgeKind
```
Svilen-Stefanov, line 35: Same question as #475 but pointing the other way: this moves `CallGraph` / `Edge` / `EdgeKind` / `LocationKey` from `static_analyzer.graph` to `static_analyzer.cfg`, which is exactly the rationale #475's commit message used to justify v2 to v3 ("pickle records a class by module path"), yet the tag stays v3. I reproduced it, a pkl saved at #479 fails at #480 with `ModuleNotFoundError: No module named 'static_analyzer.graph'`, swallowed into one warning and a miss. Small blast radius since main and the committed baseline are both still v2, so the only pkls carrying v3 *and* the old module path come from intermediate builds of this stack (us, branch CI, or a baseline re-sync landing on main between #475 and #480). Still feels like we should either bump it or, if it stays v3 deliberately, extend the comment at 56-59 so the next person does not think a `static_analyzer.graph` shim would be enough.

static_analyzer/cfg/location_key.py
```diff
@@ -0,0 +1,15 @@
+"""Physical-location identity, used to dedup the several qualified names the LSP
```
Svilen-Stefanov: nit: the package layout sentence from `cfg/__init__.py` gets restated in all three siblings, and this 15-line file says it twice. AGENTS.md asks for it once in the package `__init__`. The class docstring on line 9 already carries the meaning.

static_analyzer/cfg/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Call-graph structure: nodes, call edges, reference edges and subgraph derivation.
+
+``call_graph`` holds the graph itself, ``edge`` the edge types, and ``location_key``
+the physical-location identity used to dedup LSP qualified-name aliases.
+"""
+
+from static_analyzer.cfg.call_graph import CallGraph
+from static_analyzer.cfg.edge import Edge, EdgeKind
+from static_analyzer.cfg.location_key import LocationKey
```
Svilen-Stefanov: nit: `LocationKey` is in `__all__` but nothing outside `static_analyzer/cfg/` imports it (only the def and two uses in `call_graph.py`). `__all__` is the package contract, so advertising an internal dedup key invites someone to depend on its shape. Drop it?

static_analyzer/__init__.py
```diff
@@ -16,7 +16,7 @@
 from static_analyzer.engine.result_converter import convert_to_codeboarding_format
 from static_analyzer.engine.source_inspector import SourceInspector
 from static_analyzer.engine.utils import uri_to_path
-from static_analyzer.graph import CallGraph
+from static_analyzer.cfg import CallGraph
```
Svilen-Stefanov: nit: renaming in place left the previously alphabetical `static_analyzer.*` block unsorted, `cfg` now sits after `engine.*`. We have no isort hook so ordering by hand is the only signal we get.

static_analyzer/cfg/call_graph.py
```diff
@@ -1,11 +1,13 @@
...
+
 import logging
 from collections.abc import Callable, Collection, Hashable, Mapping, Sequence
-from dataclasses import dataclass
-from enum import StrEnum
 from types import MappingProxyType
 
 import networkx as nx
 
+from static_analyzer.cfg.edge import Edge, EdgeKind
```
Svilen-Stefanov, line 9: just checking, the body's "one real fix fell out of the move" (dropping the `MethodClusterPaths` import that made the package graph circular) is not in this diff is it? `git show origin/stack/5-cluster-cache:static_analyzer/graph.py` already has no clustering import, it went in #479. Maybe the body was drafted before the rebase?

Svilen-Stefanov (approved):
LGTM.

Two things: the tag question again, and a couple of bits that #483 comes back and rewrites which could just be done here.

static_analyzer/analysis_cache.py
ivanmilevtues: yes it stays at v3 as we have not released anything with v3, no need to bump version on every commit here.

static_analyzer/cfg/call_graph.py
ivanmilevtues: Right on all of it — the fix went in #479, not here. Confirmed two ways: `dc11abfc` (#479's squash) is what removed the import, and the only `MethodClusterPaths` in this PR's commit is a *context* line in `test_tree_shape.py`, sitting next to a `graph` → `cfg` import update. The diff never contained the change.

Not a rebase artifact either, which was my first guess too: stack/6 forked from the old stack/5 tip, and that already included the removal, so the claim was never true for this diff. The commit message of `c011ddf8` carries it as well as the PR body.

Section removed from the body in c8538468. The layering claim it made is still true, it just is not this PR's doing.

static_analyzer/cfg/__init__.py
ivanmilevtues: Dropped from `__all__` and from the package imports entirely in c8538468 — `LocationKey` has no user outside `call_graph.py`, which imports it directly from `static_analyzer.cfg.location_key`.

static_analyzer/cfg/location_key.py
ivanmilevtues: Removed in c8538468. One correction to the detail: only `location_key.py` restated the layout — `edge.py` and `call_graph.py` already carry one-line docstrings about their own contents. The class docstring is doing the work, as you said.

static_analyzer/__init__.py
ivanmilevtues: Fixed in c8538468 — `static_analyzer.cfg` now sits after `analysis_result` where it belongs.

static_analyzer/analysis_cache.py
ivanmilevtues: Good repro — the `ModuleNotFoundError: No module named 'static_analyzer.graph'` is real, and you are right that this move is the same class of change as the one v3 was minted for.

Staying on v3 deliberately. It is unreleased: the committed baseline is `v2` on both `main` and this branch (`.codeboarding/static_analysis.sha`), so no artifact anyone has pairs v3 with the old module paths — only intermediate builds of this stack, as you say. Same call as on #479's tag thread.

Extended the v3 note in c8538468 to name the `cfg` move alongside the `clustering` one, so the next person sees both moves under the one tag. Kept it to a line rather than spelling out the shim reasoning — a `static_analyzer.graph` shim would not help anyway, since the tag gates the load before anything unpickles.

PR #481, opened 2026-08-14
feat(cfg): one to_networkx, reference-augmented by default

ivanmilevtues:
**Stack 7 of 9** — base: `stack/6-cfg-package` (#480). ⚠️ **Behaviour change — deliberate.**

`to_networkx` and `clustering_networkx` differed only in which reference kinds they folded in, so they collapse into `to_networkx(reference_kinds=...)` with `DEFAULT_REFERENCE_KINDS` (`CONTAINS`, `INHERITS`) as the default. Every clustering call site drops its explicit argument; the two coupling checks opt out with `()`.

**83 changed lines** across 12 files.

### Two callers change what they see

**`health` fan-in/fan-out are protected.** They pass `()`. Folding `CONTAINS` in would give every method a phantom +1 fan-in from its owning class and corrupt E2/E3.

**`_build_undirected_graphs` in `cluster_methods_mixin` now sees the same graph the partition came from.** It was building orphan-placement graphs from call edges alone, while the clusters it placed nodes *into* were computed on a reference-augmented graph. So constructors and interface methods — exactly the symbols the reference kinds exist to reconnect — resolved as disconnected and never found a nearest cluster.

**This shifts orphan assignments on real repos.** It needs an eval run before merge, and results should be read knowing that I2 membership flips roughly 1 run in 4 from `set[Path]` iteration order, independent of this change.

### Also
The `getattr` guard on `reference_edges` goes away — the tag bump in #475 means no pre-reference-edge pickle can reach this code.

### Checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency and determinism across graph-based clustering, diagram analysis, and subgraph grouping.
  * Added control over which reference relationships are included in generated graph representations.
  * Ensured coupling analysis focuses on call relationships and avoids unrelated containment links.
  * Prevented reference relationships from being included when rebuilding selected graph snapshots.

* **Tests**
  * Updated clustering and graph conversion coverage for default, filtered, and call-only graph outputs.

### DIFF
--- a/health/checks/coupling.py
+++ b/health/checks/coupling.py
@@ -18,7 +18,9 @@ def collect_coupling_values(call_graph: CallGraph) -> tuple[list[float], list[fl
     Returns:
         A tuple of (fan_out_values, fan_in_values).
     """
-    nx_graph = call_graph.to_networkx()
+    # Call edges only: a CONTAINS edge from a class to its methods is not a call, and
+    # folding it in would give every method a phantom +1 fan-in.
+    nx_graph = call_graph.to_networkx(reference_kinds=())
     fan_out_values: list[float] = []
     fan_in_values: list[float] = []
 
@@ -96,7 +98,7 @@ def check_fan_in(call_graph: CallGraph, config: HealthCheckConfig) -> StandardCh
     total_checked = 0
     threshold = config.fan_in_max
 
-    nx_graph = call_graph.to_networkx()
+    nx_graph = call_graph.to_networkx(reference_kinds=())  # call edges only; see collect_coupling_values
     for node_name in nx_graph.nodes:
         node = call_graph.nodes.get(node_name)
         if node and (node.is_class() or node.is_data()):
--- a/health/checks/coupling.py
+++ b/health/checks/coupling.py
@@ -98,7 +98,8 @@ def check_fan_in(call_graph: CallGraph, config: HealthCheckConfig) -> StandardCh
     total_checked = 0
     threshold = config.fan_in_max
 
-    nx_graph = call_graph.to_networkx(reference_kinds=())  # call edges only; see collect_coupling_values
+    # Call edges only: a CONTAINS edge would give every method a phantom +1 fan-in from its class.
+    nx_graph = call_graph.to_networkx(reference_kinds=())
     for node_name in nx_graph.nodes:
         node = call_graph.nodes.get(node_name)
         if node and (node.is_class() or node.is_data()):
--- a/static_analyzer/cfg/call_graph.py
+++ b/static_analyzer/cfg/call_graph.py
@@ -6,7 +6,7 @@ from types import MappingProxyType
 
 import networkx as nx
 
-from static_analyzer.cfg.edge import Edge, EdgeKind
+from static_analyzer.cfg.edge import DEFAULT_REFERENCE_KINDS, Edge, EdgeKind
 from static_analyzer.cfg.location_key import LocationKey
 from static_analyzer.constants import ClusteringConfig
 from static_analyzer.node import Node
@@ -43,7 +43,7 @@ class CallGraph:
         self._alias_to_canonical: dict[str, str] = {}
         # Non-call relationship edges (CONTAINS/INHERITS/TYPEREF/IMPORT), kept off
         # ``self.edges`` so relations and ``methods_called_by_me`` stay call-only.
-        # Merged into the graph only for clustering (``clustering_networkx``).
+        # ``to_networkx`` folds them into the export by default.
         # Each entry: (src_qname, dst_qname, EdgeKind value).
         self.reference_edges: list[tuple[str, str, str]] = []
 
@@ -198,7 +198,14 @@ class CallGraph:
         for edge in self.edges:
             edge.visit_paths(fn)
 
-    def to_networkx(self) -> nx.DiGraph:
+    def to_networkx(self, reference_kinds: Collection[str] = DEFAULT_REFERENCE_KINDS) -> nx.DiGraph:
+        """Export to networkx: always call edges, plus reference edges of the given kinds.
+
+        Why: folding in CONTAINS/INHERITS keeps constructors, dunders and interface
+        methods from being graph-isolated, which is what every structural consumer
+        wants. Pass ``()`` for call edges only — degree-counting metrics need that,
+        since a CONTAINS edge is not a call.
+        """
         nx_graph = nx.DiGraph()
         for node in self.nodes.values():
             nx_graph.add_node(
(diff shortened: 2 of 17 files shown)


static_analyzer/cfg/edge.py
```diff
@@ -25,6 +25,12 @@ class EdgeKind(StrEnum):
     IMPORT = "import"
 
 
+# Folded in on top of call edges by ``CallGraph.to_networkx`` unless a caller opts out.
+# CONTAINS and INHERITS reconnect the symbols the call graph leaves isolated —
+# constructors, dunders, interface methods. No engine emits TYPEREF/IMPORT yet.
+DEFAULT_REFERENCE_KINDS: tuple[EdgeKind, ...] = (EdgeKind.CONTAINS, EdgeKind.INHERITS)
```
Svilen-Stefanov, line 33: Now that this is the value that actually gets read, `ClusteringConfig.CLUSTERING_EDGE_KINDS` (constants.py:92) has no readers left, and it is the one still carrying the whole rationale about TYPEREF / IMPORT and over-merging. So we have a dead constant plus two competing docs for the same decision, and whoever edits that one will wonder why nothing changed. Can we delete it and move the note up here?

static_analyzer/cfg/call_graph.py
```diff
@@ -198,7 +198,14 @@ def visit_paths(self, fn: Callable[[str], str]) -> None:
         for edge in self.edges:
             edge.visit_paths(fn)
 
-    def to_networkx(self) -> nx.DiGraph:
+    def to_networkx(self, reference_kinds: Collection[str] = DEFAULT_REFERENCE_KINDS) -> nx.DiGraph:
```
Svilen-Stefanov: The body says two callers change what they see, but `diagram_analysis/cluster_snapshot.py:58` also called bare `to_networkx()` before this. Looks harmless (it only reads node attributes and the reference loop adds no nodes), but it now resolves and adds an edge per method for a graph whose edges it throws away. Worth passing `reference_kinds=()` there so the intent is explicit, and mentioning it in the description?

static_analyzer/cfg/call_graph.py
```diff
@@ -210,23 +217,13 @@ def to_networkx(self) -> nx.DiGraph:
...
-        # getattr: baselines pickled before reference edges existed lack the attribute.
-        # Resolve endpoints through the alias map so an edge stored under a short alias still
-        # lands on the canonical node (matching how add_edge resolves call edges).
-        for src, dst, kind in getattr(self, "reference_edges", ()):
+        kinds = set(reference_kinds)
+        if not kinds:
+            return nx_graph
+        # Resolve endpoints through the alias map so an edge stored under a short alias
+        # still lands on the canonical node (matching how add_edge resolves call edges).
+        for src, dst, kind in self.reference_edges:
```
Svilen-Stefanov, line 220: nit: we drop the `getattr(self, "reference_edges", ())` guard here but the identical one survives at line 136. If the v3 tag really means no pre-reference-edge pickle can reach us, shouldn't both go? Otherwise the surviving one keeps implying they can.

tests/agents/test_abstraction_agent.py
```diff
@@ -121,7 +121,7 @@ def test_step_clusters_grouping_single_language(self):
         agent = self._make_agent()
         cr, graph = self._clustered_graph(range(1, 13))
         self.mock_static_analysis.get_cfg.return_value.to_networkx.return_value = graph
-        self.mock_static_analysis.get_cfg.return_value.clustering_networkx.return_value = graph
+        self.mock_static_analysis.get_cfg.return_value.to_networkx.return_value = graph
```
Svilen-Stefanov: nit: 123 and 124 are the same statement now, looks like the `clustering_networkx` line got rewritten instead of deleted. Same pair at 144-145, and in test_details_agent.py at 204-205, 401-402 and 410-411.

health/checks/coupling.py
```diff
@@ -96,7 +98,7 @@ def check_fan_in(call_graph: CallGraph, config: HealthCheckConfig) -> StandardCh
     total_checked = 0
     threshold = config.fan_in_max
 
-    nx_graph = call_graph.to_networkx()
+    nx_graph = call_graph.to_networkx(reference_kinds=())  # call edges only; see collect_coupling_values
```
Svilen-Stefanov: nit: this points at `collect_coupling_values`, which nothing calls (only the def and this comment). When someone deletes it the explanation for `reference_kinds=()` goes with it. Maybe spell the reason out inline here instead?

Svilen-Stefanov (approved):
Thanks for these, I hope things will get cleaner after this refactor. Left a couple of comments.

tests/agents/test_abstraction_agent.py
ivanmilevtues: Good catch — all five, fixed in e27298be. The `clustering_networkx` line had been rewritten to `to_networkx` instead of deleted, so both lines ended up identical: two in `test_abstraction_agent.py` (124, 145) and three in `test_details_agent.py` (205, 402, 411). Swept the whole of both files for consecutive duplicates afterwards; none left.

health/checks/coupling.py
ivanmilevtues: Confirmed dead — `collect_coupling_values` has exactly two mentions in the repo, its own `def` and that comment. Spelled the reason inline in e27298be:

```python
# Call edges only: a CONTAINS edge would give every method a phantom +1 fan-in from its class.
nx_graph = call_graph.to_networkx(reference_kinds=())
```

Left the function itself alone — it is exported without an underscore from `health/checks/`, so deleting it is a separate call rather than a rider on this PR.

static_analyzer/cfg/call_graph.py
ivanmilevtues: You are right, both should go. Removed the surviving one in e27298be.

Checked it was safe rather than symmetric-for-its-own-sake: `extra_sources` is typed `CallGraph`, and all four call sites pass either nothing or `other` from `union`, so every `source` has the attribute set in `__init__`. The only way it would not is a pre-reference-edge unpickle — which is exactly what the tag rules out, so the guard was asserting the thing we say cannot happen.

static_analyzer/cfg/call_graph.py
ivanmilevtues: Right, and `cluster_snapshot` was not the only caller the description had missed — so rather than annotate one site, `reference_kinds` now has **no default** and every caller states what it wants (e27298be).

mypy then found the rest for free: five test call sites, plus it made the audit exhaustive instead of my grepping for bare calls. `cluster_snapshot` passes `()` as you suggested — it reads node attributes and discards the edges, so resolving a CONTAINS edge per method was pure waste. The nine clustering call sites pass `DEFAULT_REFERENCE_KINDS` explicitly.

The method docstring is back to one line, since the parameter is now self-documenting at every call site.

static_analyzer/cfg/edge.py
ivanmilevtues: Done in e27298be — `ClusteringConfig.CLUSTERING_EDGE_KINDS` deleted (it had exactly one mention left, its own assignment) and its TYPEREF/IMPORT rationale moved onto `DEFAULT_REFERENCE_KINDS`, which is now the single home for the decision.

PR #484, opened 2026-08-14
perf(engine): bound how many LSP servers are resident at once

Svilen-Stefanov:
> Stacked on #477 — based on `fix/csharp-definitions-edge-strategy` so it shows only the two engine commits. Retargets to `main` once #477 merges.

## The observation

A full run on [abp](https://github.com/abpframework/abp) spawns **28 C# engines**, which looks alarming. It isn't, on its own: abp ships 30 `.sln`/`.slnx` files and the scanner emits one engine per solution. Their file counts sum to **6,297 against 6,284 analyzed** — they partition the repo, nothing is analyzed twice.

The waste is the lifecycle. `start_clients()` starts all 28 up front, and `_run_full_lsp_pass` then walks them **one at a time**. We paid peak memory for full parallelism and got none of the throughput: 12.87GB resident to do strictly sequential work.

## The change

`CODEBOARDING_MAX_CONCURRENT_ENGINES` bounds residency. When set, clients are started inside the full pass and shut down as each engine finishes, with a worker pool sized to the bound. **Unset keeps today's behaviour exactly**, so this is inert until opted into.

## Measured

abp, 6,284 C# files, 12-core host, static analysis only. **Every configuration produced an identical graph** (23,866 nodes / 34,813 edges), so these differ only in cost:

| cap | wall | speedup | peak RSS | servers |
|---|---|---|---|---|
| unbounded (today) | 1,413.3s | 1.00× | 12.87 GB | 28 |
| 1 | 1,394.2s | 1.01× | **4.95 GB** | 1 |
| **4** | **612.1s** | **2.31×** | 7.09 GB | 4 |
| 6 | 697.7s | 2.03× | 7.61 GB | 6 |
| 8 | 676.7s | 2.09× | 8.52 GB | 8 |

cap=1 shows the eager start bought nothing — same wall time, 2.6× less memory. cap=4 is the knee. Past it, summed phase-1 CPU climbs 967s → 1,449s → 1,725s as the servers contend.

## Two bugs the sweep caught

**Restore ordering.** cap=1 first returned **34,806 edges — 7 short of baseline**. `dotnet restore` writes the `obj/` artifacts a *sibling* project's compilation resolves against, so an engine analyzed before its siblings are restored silently loses cross-project edges. Projects are now prepared before any server starts; cap=1 then matched 34,813 exactly. This was only visible because the harness diffed edge counts rather than assuming same-inputs-same-output.

**Validation read live clients.** `_validate_analysis_results` iterated `_engine_clients`, which is empty under a bound — the zero-symbol guard would have silently stopped running. It now reads engine configs.

## Sizing the bound

`recommended_engine_concurrency()` is the smallest of three real limits, not a fitted constant:

- **Work** — never more servers than engines. Most repos have one or two, so this mechanism is inert outside monorepos.
- **CPU** — `cores // 3`. An engine costs about three cores: the server peaks near 1.9 (measured), plus its `dotnet` children and the tree-sitter parsing the same pass runs on the Python side. On 12 cores that predicts 4, which is where abp actually peaked.
- **Memory** — the recycler's own budget divided by observed per-engine footprint. This binds on a large host: 128 cores would allow 42 engines on CPU alone.

The memory term reuses `default_memory_budget()` deliberately. The two mechanisms did not agree before — the recycler lets each server grow to the full budget while nothing accounted for how many were resident, so during the baseline run 28 servers each believed they could reach 12GB.

| host | work | CPU | memory | result | binds on |
|---|---|---|---|---|---|
| 12-core, abp | 28 | 4 | 6 | 4 | CPU |
| 128-core, abp | 28 | 42 | 6 | 6 | memory |
| 128-core, typical repo | 2 | 42 | 6 | 2 | work |
| 2-core CI runner | 28 | 0 | 1 | 1 | floor |

## Why it is not the default yet

`_update_cached_results` — the warm-start path — reads `_engine_clients` directly. Defaulting the bound on would leave that empty, so the loop body would never run and an incremental analysis would **silently return stale results**. That is a worse failure than the problem being solved, so the default flip needs that path to spawn on demand first. `recommended_engine_concurrency()` and its tests are in place for when it does.

2,071 unit tests pass; the 3 `test_cli_parser` failures reproduce on `main` and are unrelated. mypy and black clean.

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for concurrent analysis engines based on environment settings and available system resources.
  - Added per-engine memory budgeting for language servers and call-graph analysis.
  - Improved concurrent analysis startup, cleanup, failure handling, and cancellation.
  - Added clearer errors when live client operations are unavailable.
  - Prevented temporary files from conflicting during parallel C# analysis.

- **Bug Fixes**
  - Improved validation to recognize configured analysis engines even before clients are active.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -17,7 +17,7 @@ from static_analyzer.engine.adapters import get_adapter
 from static_analyzer.engine.call_graph_builder import CallGraphBuilder
 from static_analyzer.engine.language_adapter import LanguageAdapter
 from static_analyzer.engine.lsp_client import LSPClient
-from static_analyzer.engine.lsp_recycler import default_memory_budget
+from static_analyzer.engine.lsp_recycler import default_memory_budget, per_engine_memory_budget
 from static_analyzer.engine.result_converter import convert_to_codeboarding_format
 from static_analyzer.engine.source_inspector import SourceInspector
 from static_analyzer.engine.utils import uri_to_path
@@ -309,9 +309,15 @@ class StaticAnalyzer:
                 failed_languages.append(adapter.language)
                 failed_details.append(f"{adapter.language}: {exc}")
 
-        if attempted and max_concurrent_engines():
+        if not attempted:
+            logger.info(f"No source files for any detected language in {self.repository_path}; no LSP clients started.")
+            self._engine_clients = []
+            self._clients_started = True
+            return
+
+        if max_concurrent_engines():
             logger.info(
-                "Deferring %d engine LSP client(s); the full pass keeps at most %d resident.",
+                "%d engine LSP client(s) will start during analysis, at most %d running at a time.",
                 len(attempted),
                 max_concurrent_engines(),
             )
@@ -324,12 +330,6 @@ class StaticAnalyzer:
                     self._prepare_project_once(engine_config)
                 except Exception:
                     logger.exception(f"Failed to prepare {engine_config.project_path}; the full pass will report it")
-            self._engine_clients = []
-            self._clients_started = True
-            return
-
-        if not attempted:
-            logger.info(f"No source files for any detected language in {self.repository_path}; no LSP clients started.")
             self._engine_clients = []
             self._clients_started = True
             return
@@ -725,8 +725,13 @@ class StaticAnalyzer:
         absorb_lock = threading.Lock()
         spawned: list[str] = []
         spawn_failures: list[str] = []
+        # Keyed by position in the engine list so results merge in configuration
+        # order. The merges replace on key collision, and overlapping configs
+        # (nested solution roots) do collide, so completion order would let two
+        # identical runs keep different nodes and produce different component IDs.
+        completed: dict[int, tuple[Language, dict]] = {}
 
-        def run_one(engine_config: EngineConfig, engine_client: LSPClient | None) -> None:
+        def run_one(engine_config: EngineConfig, engine_client: LSPClient | None, order: int = 0) -> None:
             """Analyze one engine. Owns the client's lifetime when given none."""
             adapter, project_path = engine_config.adapter, engine_config.project_path
             language = adapter.language_enum
@@ -748,7 +753,7 @@ class StaticAnalyzer:
                 duration_ms = round((time.monotonic() - t_lang_start) * 1000)
                 logger.info(f"Engine analysis for {adapter.language} completed in {duration_ms / 1000:.1f}s")
                 with absorb_lock:
-                    self._absorb_into_results(results, language, analysis)
+                    completed[order] = (language, analysis)
                     self._collect_diagnostics_for(adapter, engine_client, analysis)
                     track_lsp_result(
                         language=adapter.language_enum.value,
@@ -782,8 +787,8 @@ class StaticAnalyzer:
 
         cap = max_concurrent_engines()
         if not cap:
-            for engine_config, engine_client in self._engine_clients:
-                run_one(engine_config, engine_client)
+            for order, (engine_config, engine_client) in enumerate(self._engine_clients):
+                run_one(engine_config, engine_client, order)
         else:
             pending = [cfg for cfg in self._engine_configs if cfg.source_files]
             logger.info("Running %d engine(s) with at most %d resident at a time", len(pending), cap)
@@ -793,7 +798,8 @@ class StaticAnalyzer:
             # wait so each in-flight ``run_one`` still shuts its own client down.
             pool = ThreadPoolExecutor(max_workers=cap)
             try:
-                for future in as_completed([pool.submit(run_one, cfg, None) for cfg in pending]):
+                futures = [pool.submit(run_one, cfg, None, order) for order, cfg in enumerate(pending)]
+                for future in as_completed(futures):
                     future.result()
             finally:
                 pool.shutdown(wait=True, cancel_futures=True)
@@ -803,6 +809,10 @@ class StaticAnalyzer:
                     "Failed to start any engine LSP client "
                     f"(attempted: {', '.join(cfg.adapter.language for cfg in pending)}){details}"
                 )
+
+        for order in sorted(completed):
+            language, analysis = completed[order]
+            self._absorb_into_results(results, language, analysis)
 
         summaries = []
         for language in results.get_languages():
@@ -954,7 +964,10 @@ class StaticAnalyzer:
                 f"Diagnostics for {adapter.language}: {len(merged_diags)} files, {total} items "
                 f"(cache={len(cache_diags)}, live={len(live_diags)})"
             )
-        self.collected_diagnostics[adapter.language_enum] = merged_diags
+        # Merge, not replace: a monorepo yields several configs per language, and
+        # replacing kept only whichever finished last -- which under the concurrency
+        # bound is a different one run to run.
+        self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
     def _loc_for_adapter(self, adapter: LanguageAdapter) -> int:
         """Return scanner LOC that should have been covered by this adapter."""
@@ -994,7 +1007,12 @@ class StaticAnalyzer:
         logger.info(f"Analyzing {len(source_files)} {adapter.language} files")
 
         t_build_start = time.monotonic()
-        builder = CallGraphBuilder(engine_client, adapter, project_path)
+        builder = CallGraphBuilder(
+            engine_client,
+            adapter,
+            project_path,
+            memory_budget_bytes=per_engine_memory_budget(max(max_concurrent_engines(), 1)),
+        )
         engine_result = builder.build(source_files)
         logger.info(f"CallGraphBuilder.build() for {adapter.language}: {time.monotonic() - t_build_start:.1f}s")
         if adapter.fail_on_empty_symbols is True and not builder.symbol_table.symbols:
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -2,7 +2,7 @@ import logging
 import os
 import threading
 import time
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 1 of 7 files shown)


static_analyzer/__init__.py
```diff
@@ -244,61 +311,38 @@ def start_clients(self) -> None:
...
+            # Prepare everything now, while nothing is resident: a project must
+            # not be analyzed before its siblings are restored.
+            for engine_config in self._engine_configs:
+                if not engine_config.source_files:
+                    continue
+                try:
+                    self._prepare_project_once(engine_config)
+                except Exception:
+                    logger.exception(f"Failed to prepare {engine_config.project_path}; continuing")
+            self._engine_clients = []
```
Svilen-Stefanov, line 333: blocker-ish: doesn't the warm-start path go empty once the cap is on? `_engine_clients` is empty by design now, so a run that finds a pkl iterates nothing, builds an empty `StaticAnalysisResults`, reports success, and then persists that over the good pkl. So run 2 is empty and run 3 is empty as well.

The docstring above says "silently return stale results", but this is worse than stale, it is empty and it eats the cache on the way out, and nothing in the code enforces the opt-in caveat. Should we make it impossible rather than documented: either raise when the cap is set (a loud error beats a plausible-but-wrong result, per AGENTS.md), or give `_update_cached_results` the same spawn-on-demand treatment `run_one` has? `notify_file_changed`, `get_file_symbols`, `get_adapter_for_file`, `discover_file_dependencies` and `collect_fresh_diagnostics` all have the same hole.

static_analyzer/__init__.py
```diff
@@ -598,37 +700,66 @@ def _run_full_lsp_pass(self) -> StaticAnalysisResults:
...
+                        logger.exception(f"Error shutting down {adapter.language} client for {project_path}")
+
+        cap = max_concurrent_engines()
+        if not cap:
+            for engine_config, engine_client in self._engine_clients:
+                run_one(engine_config, engine_client)
+        else:
+            pending = [cfg for cfg in self._engine_configs if cfg.source_files]
+            logger.info("Running %d engine(s) with at most %d resident at a time", len(pending), cap)
+            with ThreadPoolExecutor(max_workers=cap) as pool:
```
Svilen-Stefanov: With the executor, does a `StaticAnalysisFatalError` still abort early? The context manager waits for everything already submitted before the exception propagates, so where sequentially we bailed within minutes, we now churn through the other 27 engines first. On abp that is potentially an hour of full LSP cost for a run we already knew was doomed. `as_completed` plus `shutdown(wait=False, cancel_futures=True)` on the first exception?

static_analyzer/__init__.py
```diff
@@ -48,6 +52,67 @@ class StaticAnalysisFatalError(RuntimeError):
...
+    if not raw:
+        return 0
+    try:
+        return max(int(raw), 0)
+    except ValueError:
+        logger.warning("Ignoring non-integer %s=%r", MAX_CONCURRENT_ENGINES_ENV_VAR, raw)
+        return 0
+
+
+def recommended_engine_concurrency(engine_count: int) -> int:
```
Svilen-Stefanov, line 87: Is `recommended_engine_concurrency` wired to anything yet? As far as I can tell only the new tests call it, so `CORES_PER_ENGINE` and `ENGINE_MEMORY_FOOTPRINT_BYTES` are inert and the only shipped knob is the hand-set integer.

Also I don't think the memory term does what the docstring claims. It divides a per-server recycle budget by a per-server footprint, so on a big host it yields cap=6 while the recycler independently lets each of those 6 grow to 12GB, which is 72GB permitted rather than the 12GB the two mechanisms are supposed to agree on. Either wire it in once the warm-start hole is closed (unset env var meaning `recommended_engine_concurrency(len(pending))`), or drop it until it is used?

Svilen-Stefanov (commented):
Self-review. The bound is the right shape and the tests around it are good, but I would not merge it as is.

With the cap set, any run that finds a pkl on disk takes the warm-start path over an empty `_engine_clients`, returns an empty result, reports success, and then writes that over the good cache, so the next run is empty too. Second one is that a fatal error no longer aborts early now that the executor drains everything already submitted. And the recommended-concurrency helper has no production caller yet, so its two constants are currently dead weight.

static_analyzer/__init__.py
```diff
@@ -244,61 +311,38 @@ def start_clients(self) -> None:
...
+            # not be analyzed before its siblings are restored.
+            for engine_config in self._engine_configs:
+                if not engine_config.source_files:
+                    continue
+                try:
+                    self._prepare_project_once(engine_config)
+                except Exception:
+                    logger.exception(f"Failed to prepare {engine_config.project_path}; continuing")
+            self._engine_clients = []
+            self._clients_started = True
```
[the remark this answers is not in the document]
Svilen-Stefanov: Confirmed, and worse than "returns and persists": I ran it and the pkl comes back `([], 'cafebabe')`. `_validate_analysis_results` does not save us either, `get_source_files` returns `[]` on the empty result so every language hits the `continue`. `track_lsp_result` never fires at all, since it is inside the skipped loop.

Went with the loud guard rather than spawn-on-demand. `_live_clients(operation)` now fronts all seven readers, the five in `_update_cached_results`/`notify_file_changed`/`get_file_symbols`/`get_adapter_for_file`/`discover_file_dependencies` plus `collect_fresh_diagnostics` and `get_diagnostics_generation`. It keys on "configs have work but nothing is resident" rather than on the env var, so a future second mechanism that empties the list cannot slip past it, and it raises before `_results_need_saving = True` so the good pkl survives untouched.

static_analyzer/__init__.py
```diff
@@ -598,37 +700,66 @@ def _run_full_lsp_pass(self) -> StaticAnalysisResults:
...
+                        loc=self._loc_for_adapter(adapter),
+                        status="success",
+                        duration_ms=duration_ms,
+                        analysis=analysis,
+                        diagnostics=self.collected_diagnostics.get(adapter.language_enum, {}),
+                    )
             except StaticAnalysisFatalError:
                 raise
             except Exception as e:
                 logger.error(f"Error during engine analysis for {adapter.language}: {e}")
```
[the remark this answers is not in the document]
Svilen-Stefanov: Right. `run_one` records spawn successes and failures separately now, and the bounded branch re-raises the same `RuntimeError` the eager path does when nothing started, same message shape. Test covers 3 engines all failing to spawn, plus one where a single engine survives and it correctly does not raise.

static_analyzer/__init__.py
```diff
@@ -244,61 +311,38 @@ def start_clients(self) -> None:
...
+            )
+            # Prepare everything now, while nothing is resident: a project must
+            # not be analyzed before its siblings are restored.
+            for engine_config in self._engine_configs:
+                if not engine_config.source_files:
+                    continue
+                try:
+                    self._prepare_project_once(engine_config)
+                except Exception:
+                    logger.exception(f"Failed to prepare {engine_config.project_path}; continuing")
```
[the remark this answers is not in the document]
Svilen-Stefanov: Partly. The literal scenario does not repro: `get_lsp_env` re-runs the same `resolve_dotnet_sdk`, so a broken SDK fails again and the engine is reported rather than silently degraded. It only bites on a self-healing failure, which is narrow.

Hardened it anyway, but not as "record success only". That makes every pool worker retry `resolve_dotnet_sdk`, and several concurrent `_install_channel` calls into one shared install dir is worse than a single clean failure. So the outcome is recorded and re-raised on the second caller. I also keyed the memo on `(language, project_path)` rather than the path alone, since a Python or TypeScript config rooted at the repo root could otherwise suppress C#'s restore for that same root, which is a wider bug than the one filed here.

static_analyzer/__init__.py
```diff
@@ -598,37 +700,66 @@ def _run_full_lsp_pass(self) -> StaticAnalysisResults:
...
+
+        cap = max_concurrent_engines()
+        if not cap:
+            for engine_config, engine_client in self._engine_clients:
+                run_one(engine_config, engine_client)
+        else:
+            pending = [cfg for cfg in self._engine_configs if cfg.source_files]
+            logger.info("Running %d engine(s) with at most %d resident at a time", len(pending), cap)
+            with ThreadPoolExecutor(max_workers=cap) as pool:
+                for future in [pool.submit(run_one, cfg, None) for cfg in pending]:
```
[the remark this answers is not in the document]
Svilen-Stefanov: Confirmed. `get_servers_dir()` takes no project argument, so all workers hit the same two fixed filenames with the same pid-derived temp path. The interleaving is exactly the one described, the loser's `os.replace` gets FileNotFoundError, `run_one` swallows it, and that solution silently contributes nothing to the run.

Fixed with a thread id in the temp name rather than a lock: the content is a module constant so concurrent replaces are indistinguishable, and no lock means no truncate-mid-publish window either. It is first-run-only, but that is every cold CI job.

static_analyzer/__init__.py
```diff
@@ -48,6 +52,67 @@ class StaticAnalysisFatalError(RuntimeError):
...
+    make an incremental run silently return stale results. Making that path
+    spawn on demand is the prerequisite for flipping the default.
+    """
+    raw = os.environ.get(MAX_CONCURRENT_ENGINES_ENV_VAR, "").strip()
+    if not raw:
+        return 0
+    try:
+        return max(int(raw), 0)
+    except ValueError:
+        logger.warning("Ignoring non-integer %s=%r", MAX_CONCURRENT_ENGINES_ENV_VAR, raw)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Half. `4.0` and `abc` already log a warning, so "silently" is not quite right for the headline example, but negatives genuinely were silent. Also found `=4_0` returns 40, because `int()` accepts underscore separators, which is a wrong-direction failure.

Not raising though. Both other knobs in this subsystem (`lsp_recycler.default_memory_budget`, the PHP adapter) warn and fall back, and this getter runs inside a per-engine loop so raising would fire minutes into a run rather than at startup. Negatives now log the same warning and I renamed the test to `test_negative_is_ignored`. If we do want fail-fast on this knob it belongs next to `load_user_config` at the CLI entry point, where a typo costs zero seconds.

static_analyzer/__init__.py
```diff
@@ -48,6 +52,67 @@ class StaticAnalysisFatalError(RuntimeError):
...
+
+def max_concurrent_engines() -> int:
+    """How many engine LSP servers may be resident at once. 0 disables the bound.
+
+    Why bound it: a monorepo emits one engine per sub-project (abp ships 30 C#
+    solutions). Starting them all up front holds every Roslyn workspace in
+    memory while the full pass uses one at a time — peak 12.9GB across 28
+    servers to do strictly sequential work.
+
+    On a 12-core box abp took 1413s at 28 resident, 1394s at 1, 612s at 4 and
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fair, both were well past the point where the prose outweighs the code (17 lines against 8, and 14 against 3). Trimmed to what you cannot derive from the code: why ~3 cores, why the default is off, where the 2GB figure comes from. The sweep numbers live in the PR body, which is where they belong.

static_analyzer/__init__.py
Svilen-Stefanov: Fixed, and it is the bad version rather than the stale one, I ran it and the pkl really does come back `([], 'cafebabe')`. `_validate_analysis_results` does not catch it either, an empty result has no source files so every language is skipped.

Went with make-it-impossible. `_live_clients(operation)` raises `StaticAnalysisFatalError` whenever there is configured work but nothing resident, and all seven readers go through it, including the five listed here plus `get_diagnostics_generation`. Keyed on the state rather than on the env var so it cannot be bypassed by whatever empties the list next.

static_analyzer/__init__.py
Svilen-Stefanov: Confirmed, and it is as bad as suspected. Ran the real `_run_full_lsp_pass` with 28 fake engines at cap 3: all 28 spawn a client and run a full analysis after the fatal has already been raised.

Fixed, but not with `wait=False`. That one leaves in-flight `run_one` calls mutating `results` and holding servers while the caller races ahead into `_validate_analysis_results` and `stop_clients`, measured 1 of 4 owned-client shutdowns done at the moment of return. It is `as_completed` plus `shutdown(wait=True, cancel_futures=True)` instead, so the queue is dropped but each in-flight call still reaches its finally block. Test asserts fewer than 12 of 12 engines run, and it does fail against the old `with` form.

static_analyzer/__init__.py
Svilen-Stefanov: Both confirmed. Only the tests call it, and the memory claim is false: `default_memory_budget()` is per server (`lsp_recycler.py:64`, and `CallGraphBuilder` passes no `budget_bytes`), so at cap 6 that is 72GB permitted on a box with a 12GB budget, exactly as suspected.

Kept the helper but deleted the two sentences that were not true, and the docstring now says what the term actually does. Did not wire it as the default: even with the guard in place, flipping it would put every repo on the deferred lifecycle, which is a much bigger behaviour change than this PR should carry. So the sizing analysis stays available and the default flip is its own PR, along with passing `default_memory_budget() // cap` into the recycler so the two mechanisms genuinely do agree.

static_analyzer/__init__.py
```diff
@@ -244,61 +295,38 @@ def start_clients(self) -> None:
...
                 failed_details.append(f"{adapter.language}: {exc}")
-                if engine_client is not None:
-                    try:
-                        engine_client.shutdown()
-                    except Exception:
-                        logger.exception(
-                            f"Error shutting down partially-started {adapter.language} client during cleanup"
-                        )
+
+        if attempted and max_concurrent_engines():
```
ivanmilevtues: from what i get attempted will always be true. Attempted is the list of clients we tried to start, if we haven't tried to start anything then max_concurrent_engines will be false by default no?

I also thinkg tha thte logger is wrong, we will start few of the attempted lang servers no?

static_analyzer/__init__.py
ivanmilevtues: Has this comment been addressed @Svilen-Stefanov ?

static_analyzer/__init__.py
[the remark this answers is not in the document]
ivanmilevtues: This could be a reasonable comment unsure if it is at the right location @Svilen-Stefanov

ivanmilevtues (approved):
It is good that the tests are passing, you have few comments of your own and I think one of the P1's from Codex should be checked

Svilen-Stefanov (commented):
I noticed my replies were in pendinging so pushing them now...

The warm-start one turned out to be the empty-and-overwrite case rather than stale, I reproduced it and the pkl comes back as ([], 'cafebabe'). So the caveat is enforced by _live_clients now instead of documented, across all seven readers rather than the six I listed.

recommended_engine_concurrency stays, but its docstring no longer claims the memory term makes the two mechanisms agree on RAM, because it does not. Passing budget_bytes=default_memory_budget() // cap into LSPRecycler is the actual fix for that and it wants its own PR.

The "Why it is not the default yet" section of the description is out of date now, I will rewrite it.

PR #494, opened 2026-08-18
refactor(clustering): move grouping algorithms into package

ivanmilevtues:
## Behavioral impact (upfront)

This PR does **not** introduce a new grouping algorithm. Fresh and anchored grouping already exist on `main`; this PR moves them behind a clearer service boundary.

| Area | Impact |
|---|---|
| Grouping results | **No intended change** for existing production call paths. The same resolution ladder, seed, component ranges, drift budget, modularity selection, absorption, and ownership rules are used. |
| Cluster IDs across languages | **No runtime change.** Both offset loops retain `main`'s `offset += max(cluster_ids) + 1` behavior, including gaps when inputs are 1-based. Regression tests now pin both 0-based disjoint ranges and the existing 1-based ranges. |
| Python API | **Intentional change.** Grouping-specific free functions are replaced by the stateless `GroupingService.group()` and `GroupingService.anchored_group()` API. Arbitrary `low`/`high`/`seed`/`drift_budget` call-site arguments are replaced by the two declared config profiles. |
| Import paths | **Intentional change.** `static_analyzer.constants` is renamed to `static_analyzer.config`; all in-repo imports are updated. Grouping-specific compatibility exports are removed from `cluster_helpers`. |

## Moves and API mapping

| Before | After | What moved / why | Runtime impact |
|---|---|---|---|
| Grouping implementation in `static_analyzer/cluster_helpers.py` | `static_analyzer/clustering/grouping.py` | Meta-graph construction, modularity optimization, leftover absorption, fresh grouping, and anchored grouping now live with the clustering package that owns them. | None intended; implementation logic is preserved. |
| `supercluster_leaf_ids(...)` / `supercluster_by_modularity_peak(...)` | `GroupingService().group(cluster_results, cfg_graphs, subcomponents=...)` | One public entry point combines per-language cluster results and CFGs, then performs fresh grouping. | Existing root calls still use 5–8 components; subcomponent calls still use 3–8. |
| `anchored_grouping(combined, combined_cfg, previous_owner, low, high, ...)` | `GroupingService().anchored_group(cluster_results, cfg_graphs, previous_owner, subcomponents=...)` | Anchored and fresh grouping are two operations on one stateless service; callers no longer pre-combine inputs. | Existing root/subcomponent behavior and the 0.10 drift budget are preserved. |
[...]

PR #495, opened 2026-08-18
feat(clustering): add scope results with communication

ivanmilevtues:
## Goal

Introduce the structural result that the rest of this stack can consume: cluster one exact graph scope once, and retain its per-language leaf clusters, architectural groups, symbol ownership, and cross-group communication together.

This is the contract-building PR. It moves clustering from “a set of helpers whose consumers reconstruct context” toward “one service call that returns the complete structural view of a scope.”

## Behavioral impact (upfront)

This PR adds a new scope-level result and service API. It does **not** yet make the existing full or incremental analysis paths consume that result; those integrations remain in the follow-up PRs in the stack.

| Area | Impact |
|---|---|
| Existing full/incremental analysis | **No intended runtime change.** Existing agents still use their current grouping and recursion paths in this PR. |
| Grouping algorithm | **No new algorithm.** `cluster_scope()` delegates to the `GroupingService` introduced by #494 for fresh or ownership-anchored grouping. |
| Root cluster construction | **No intended result change.** `build_all_cluster_results` and cache synchronization move from `clustering/grouping.py` to `clustering/orchestration.py`; the existing offset and cache-adoption behavior is retained. |
| Scope result | **Intentional new API.** `ClusteringService.cluster_scope(...)` returns one `ClusterScopeResult` containing leaf clusters, groups, complete symbol membership, directed sibling connections, modularity, and regrouping state. |
| Empty leaf clustering | **Fail-fast, matching the previous surface behavior.** A populated language that produces no leaf clusters raises `LeafClustersUnavailableError` instead of returning a successful empty scope. Callers that explicitly request method-level fallback receive synthetic callable/class clusters first. |
| Group IDs | **Preserves established incremental allocation behavior.** Anchored groups retain their previous component IDs; new groups append after the highest surviving sibling index instead of filling retired numeric gaps. |
| Identifier vocabulary | **Intentional type-name cleanup.** Shared `ClusterId`, `ScopeId`, `GroupId`, and `ComponentId` aliases replace overlapping graph/CodeBoarding ID names. |
| Call-site evidence | **Intentional contract tightening.** Connection edges retain normalized `file`, `line`, and optional `column` locations through the shared `CallSiteLocation` type; malformed call-site values fail validation rather than flowing through as arbitrary mappings. |

## Result contract and API mapping

| Before | After | Contract / reason |
|---|---|---|
| `ClusteringService.cluster(graph)` returns one language's `ClusterResult` | Existing method remains; new `cluster_scope(graphs, ...)` returns `ClusterScopeResult` | Keeps leaf clustering available while adding one complete scope-level operation. |
| Consumers separately hold per-language cluster results, grouping output, and CFG evidence | `ClusterScopeResult.leaf_clusters_by_language`, `.groups`, `.connections`, `.modularity`, and `.regrouped` | One result owns the complete structural view needed by hierarchy builders and analysis consumers. |
| Grouping output is `list[set[int]]` plus separately reconstructed members | `ClusterGroup(group_id, cluster_ids, symbol_members_by_language, previous_component_id, children)` | Makes group identity, language → qualified-symbol membership, prior ownership, and future child scope explicit. |
| No typed sibling-communication model | `GroupConnection` contains concrete `ClusterConnectionEdge` calls between `source_group_id` and `target_group_id` | Preserves direction and exact qualified source/target names with call-site evidence. |
| Agent-local `GraphClusterId` / `CodeBoardingClusterId` aliases | Shared top-level `ClusterId`, `ScopeId`, `GroupId`, and `ComponentId` | Gives the static-analysis and agent layers one vocabulary without importing the agent layer from the analyzer. The module is explicitly included in built wheels. |
| Root clustering orchestration lives in `clustering/grouping.py` | `clustering/orchestration.py` | Separates root analysis/cache synchronization from the grouping algorithm while preserving the `cluster_helpers` compatibility export. |
| Generic call-site dictionaries | `CallSiteLocation` (`line`, optional `file`, optional `column`) | Documents and validates the source-location evidence retained on inter-group calls. |

`cluster_scope()` accepts the language graphs for one exact scope and optional precomputed leaf clusters. It gives languages disjoint cluster-ID ranges, applies fresh or previous-ownership-anchored grouping, assigns deterministic group IDs, assigns every callable/class symbol to one group (including symbols omitted by the leaf clustering), and records directed calls crossing sibling groups.

## Compatibility boundary

`ClusteringService.cluster()` and the existing agent flows remain available. General cluster-building/reindexing imports exposed through `static_analyzer.cluster_helpers` continue to resolve after root orchestration moves.

This PR computes **one scope only**. It does not recursively build the component hierarchy and does not yet replace agent-side grouping. #497 owns recursive hierarchy construction; #498 and #499 switch fresh and incremental analysis to consume that hierarchy; #500 removes the superseded paths.

## Validation

- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` — **2,149 passed, 28 skipped**, **83.05% coverage**.
- Focused scope suite — **9 passed**.
- `uv run mypy .` — passed across 334 source files.
- `uv run black --check ...` — passed for all changed Python files.
- Wheel build inspection — `clustering_ids.py` is present in the built distribution.
- `git diff --check` — passed.

## Stack roadmap

|---|---|---|
| Foundation | [#494](https://github.com/CodeBoarding/CodeBoarding/pull/494) | Move the existing grouping algorithms behind the stateless `GroupingService` API without intentionally changing grouping behavior. |
| **This PR** | **#495** | Define and compute a complete structural result for one graph scope, including groups, members, and sibling communication. |
| Next | [#497](https://github.com/CodeBoarding/CodeBoarding/pull/497) | Recursively build those scope results into a deterministic hierarchy using exact induced subgraphs and centralized expansion rules. |
| Follow-up | [#498](https://github.com/CodeBoarding/CodeBoarding/pull/498) | Make fresh/full analysis consume the precomputed hierarchy instead of regrouping inside agents. |
| Follow-up | [#499](https://github.com/CodeBoarding/CodeBoarding/pull/499) | Make incremental analysis build and consume an ownership-anchored hierarchy while preserving stable member ownership. |
| Final cleanup | [#500](https://github.com/CodeBoarding/CodeBoarding/pull/500) | Remove the legacy agent-side grouping and recursive incremental orchestration once both analysis paths use the hierarchy. |

---

### DIFF
--- a/clustering_ids.py
+++ b/clustering_ids.py
@@ -1,0 +1,5 @@
+"""Identifier types shared across clustering layers."""
+
+type GraphClusterId = int
+type CodeBoardingClusterId = str
+type ComponentId = str
--- a/clustering_ids.py
+++ b/clustering_ids.py
@@ -1,5 +1,6 @@
 """Identifier types shared across clustering layers."""
 
-type GraphClusterId = int
-type CodeBoardingClusterId = str
+type ClusterId = int
+type ScopeId = str
+type GroupId = str
 type ComponentId = str
--- a/static_analyzer/cfg/__init__.py
+++ b/static_analyzer/cfg/__init__.py
@@ -5,6 +5,6 @@ the physical-location identity used to dedup LSP qualified-name aliases.
 """
 
 from static_analyzer.cfg.call_graph import CallGraph
-from static_analyzer.cfg.edge import DEFAULT_REFERENCE_KINDS, Edge, EdgeKind, ReferenceEdge
+from static_analyzer.cfg.edge import CallSiteLocation, DEFAULT_REFERENCE_KINDS, Edge, EdgeKind, ReferenceEdge
 
-__all__ = ["DEFAULT_REFERENCE_KINDS", "CallGraph", "Edge", "EdgeKind", "ReferenceEdge"]
+__all__ = ["DEFAULT_REFERENCE_KINDS", "CallGraph", "CallSiteLocation", "Edge", "EdgeKind", "ReferenceEdge"]
--- a/static_analyzer/cfg/edge.py
+++ b/static_analyzer/cfg/edge.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from collections.abc import Callable, Hashable, Mapping, Sequence
 from dataclasses import dataclass
 from enum import StrEnum
+from typing import NotRequired, TypedDict
 
 from static_analyzer.node import Node
 
@@ -40,18 +41,26 @@ class ReferenceEdge:
     kind: EdgeKind
 
 
+class CallSiteLocation(TypedDict):
+    """A one-based source location where a call occurs."""
+
+    line: int
+    file: NotRequired[str]
+    column: NotRequired[int]
+
+
 class Edge:
     def __init__(self, src_node: Node, dst_node: Node, call_sites: Sequence[Mapping[str, Hashable]] = ()) -> None:
         self.src_node = src_node
         self.dst_node = dst_node
-        self._call_sites: list[dict[str, Hashable]] = []
+        self._call_sites: list[CallSiteLocation] = []
         self._call_site_keys: set[tuple[tuple[str, Hashable], ...]] = set()
         for site in call_sites:
             self.add_call_site(site)
 
     @property
-    def call_sites(self) -> list[dict[str, Hashable]]:
-        return [dict(site) for site in self._call_sites]
+    def call_sites(self) -> list[CallSiteLocation]:
+        return [site.copy() for site in self._call_sites]
 
     def get_source(self) -> str:
         return self.src_node.fully_qualified_name
@@ -76,8 +85,22 @@ class Edge:
         self._call_site_keys = {tuple(sorted(site.items())) for site in self._call_sites}
 
     @staticmethod
-    def _normalize_call_site(call_site: Mapping[str, Hashable]) -> dict[str, Hashable]:
+    def _normalize_call_site(call_site: Mapping[str, Hashable]) -> CallSiteLocation:
         normalized = dict(call_site)
         if "file" not in normalized and "file_path" in normalized:
             normalized["file"] = normalized.pop("file_path")
-        return normalized
+        file = normalized.get("file")
+        line = normalized.get("line")
+        column = normalized.get("column")
+        if not isinstance(line, int):
+            raise ValueError("Call sites require a one-based integer line")
+        location = CallSiteLocation(line=line)
+        if file is not None:
+            if not isinstance(file, str):
+                raise ValueError("Call-site files must be strings")
+            location["file"] = file
+        if column is not None:
+            if not isinstance(column, int):
+                raise ValueError("Call-site columns must be one-based integers")
+            location["column"] = column
+        return location
--- a/static_analyzer/clustering/__init__.py
+++ b/static_analyzer/clustering/__init__.py
@@ -7,7 +7,7 @@ state that ``LanguageResults`` owns.
 
 from static_analyzer.clustering.cache import ClusterCache
 from static_analyzer.config import DEFAULT_GROUPING_CONFIG, SUBCOMPONENT_GROUPING_CONFIG, GroupingConfig
-from clustering_ids import ComponentId, GraphClusterId
+from clustering_ids import ClusterId, ComponentId, GroupId, ScopeId
 from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
 from static_analyzer.clustering.models import (
     METHOD_LEVEL_STRATEGY,
@@ -18,7 +18,7 @@ from static_analyzer.clustering.models import (
     ClusterScopeResult,
     GroupConnection,
 )
-from static_analyzer.clustering.service import ClusteringService
+from static_analyzer.clustering.service import ClusteringService, LeafClustersUnavailableError
 
 __all__ = [
     "DEFAULT_GROUPING_CONFIG",
@@ -31,9 +31,12 @@ __all__ = [
     "ClusterResult",
     "ClusterScopeResult",
     "ClusteringService",
+    "ClusterId",
     "ComponentId",
-    "GraphClusterId",
+    "GroupId",
     "GroupConnection",
     "GroupingConfig",
+    "LeafClustersUnavailableError",
     "MethodClusterPaths",
+    "ScopeId",
 ]
--- a/static_analyzer/clustering/__init__.py
+++ b/static_analyzer/clustering/__init__.py
@@ -8,7 +8,15 @@ state that ``LanguageResults`` owns.
 from static_analyzer.clustering.cache import ClusterCache
 from static_analyzer.config import DEFAULT_GROUPING_CONFIG, SUBCOMPONENT_GROUPING_CONFIG, GroupingConfig
 from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
-from static_analyzer.clustering.models import METHOD_LEVEL_STRATEGY, AnchoredGrouping, ClusterResult
+from static_analyzer.clustering.models import (
+    METHOD_LEVEL_STRATEGY,
+    AnchoredGrouping,
+    ClusterConnection,
+    ClusterConnectionEdge,
+    ClusterGroup,
+    ClusterResult,
+    ClusterScopeResult,
+)
 from static_analyzer.clustering.service import ClusteringService
 
 __all__ = [
@@ -17,7 +25,11 @@ __all__ = [
     "SUBCOMPONENT_GROUPING_CONFIG",
     "AnchoredGrouping",
     "ClusterCache",
+    "ClusterConnection",
+    "ClusterConnectionEdge",
+    "ClusterGroup",
     "ClusterResult",
+    "ClusterScopeResult",
     "ClusteringService",
     "GroupingConfig",
     "MethodClusterPaths",
--- a/static_analyzer/clustering/__init__.py
+++ b/static_analyzer/clustering/__init__.py
@@ -7,15 +7,16 @@ state that ``LanguageResults`` owns.
 
 from static_analyzer.clustering.cache import ClusterCache
 from static_analyzer.config import DEFAULT_GROUPING_CONFIG, SUBCOMPONENT_GROUPING_CONFIG, GroupingConfig
+from clustering_ids import ComponentId, GraphClusterId
 from static_analyzer.clustering.method_cluster_paths import MethodClusterPaths
 from static_analyzer.clustering.models import (
     METHOD_LEVEL_STRATEGY,
     AnchoredGrouping,
-    ClusterConnection,
     ClusterConnectionEdge,
     ClusterGroup,
     ClusterResult,
     ClusterScopeResult,
+    GroupConnection,
 )
 from static_analyzer.clustering.service import ClusteringService
 
@@ -25,12 +26,14 @@ __all__ = [
     "SUBCOMPONENT_GROUPING_CONFIG",
     "AnchoredGrouping",
     "ClusterCache",
-    "ClusterConnection",
     "ClusterConnectionEdge",
     "ClusterGroup",
     "ClusterResult",
     "ClusterScopeResult",
     "ClusteringService",
+    "ComponentId",
+    "GraphClusterId",
+    "GroupConnection",
     "GroupingConfig",
     "MethodClusterPaths",
 ]
--- a/static_analyzer/clustering/grouping.py
+++ b/static_analyzer/clustering/grouping.py
@@ -7,15 +7,12 @@ from collections import Counter, defaultdict, deque
 import networkx as nx
 import networkx.algorithms.community as nx_comm
 
-from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.config import (
     DEFAULT_GROUPING_CONFIG,
     SUBCOMPONENT_GROUPING_CONFIG,
     GroupingConfig,
-    Language,
 )
 from static_analyzer.clustering.models import AnchoredGrouping, ClusterResult
-from static_analyzer.clustering.service import ClusteringService
 from static_analyzer.leiden_utils import find_partition
 
 logger = logging.getLogger(__name__)
@@ -50,32 +47,6 @@ class GroupingService:
         combined = combine_cluster_results(cluster_results)
         combined_cfg: nx.DiGraph = nx.compose_all(list(cfg_graphs.values())) if cfg_graphs else nx.DiGraph()
         return _anchored_group(combined, combined_cfg, previous_owner, config)
-
-
-def build_all_cluster_results(static_analysis: StaticAnalysisResults) -> dict[str, ClusterResult]:
-    """Cluster every detected language in one shared cluster-ID namespace."""
-    cluster_service = ClusteringService()
-    cluster_results: dict[str, ClusterResult] = {}
-    offset = 0
-    for lang in static_analysis.get_languages():
-        result = cluster_service.cluster(static_analysis.get_cfg(lang))
-        if offset:
-            result = reindex_cluster_result(result, offset)
-            logger.info(f"[Cluster] {lang}: offset IDs by +{offset} ({len(result.clusters)} clusters)")
-        cluster_results[str(lang)] = result
-        offset += max(result.clusters, default=0) + 1
-
-    _sync_cluster_cache(static_analysis, cluster_results)
-    return cluster_results
-
-
-def _sync_cluster_cache(static_analysis: StaticAnalysisResults, cluster_results: dict[str, ClusterResult]) -> None:
-    """Store each language's partition (with its final, offset IDs) on its cluster cache."""
-    for lang, result in cluster_results.items():
-        try:
-            static_analysis.get_clusters(Language(lang)).adopt(result)
-        except ValueError:
-            logger.warning("Could not sync cluster cache for unknown language %s", lang)
 
 
 def reindex_across_languages(cluster_results: dict[str, ClusterResult]) -> None:
@@ -317,6 +288,11 @@ def _method_counts(cluster_result: ClusterResult) -> dict[int, int]:
 def _modularity(meta_graph: nx.DiGraph, groups: list[set[int]]) -> float:
     """0.0 on an edgeless meta-graph — there is nothing to separate."""
     return nx_comm.modularity(meta_graph, groups, weight="weight") if meta_graph.number_of_edges() else 0.0
+
+
+def score_grouping(cluster_result: ClusterResult, cfg_graph: nx.DiGraph, groups: list[set[int]]) -> float:
+    """Score a concrete grouping against its inter-cluster meta-graph."""
+    return _modularity(_build_meta_graph(cluster_result, cfg_graph), groups)
 
 
 def _optimize_grouping(
--- a/static_analyzer/clustering/grouping.py
+++ b/static_analyzer/clustering/grouping.py
@@ -89,30 +89,6 @@ def reindex_cluster_result(cluster_result: ClusterResult, offset: int) -> Cluste
     )
 
 
-def _build_meta_graph(cluster_result: ClusterResult, cfg_graph: nx.DiGraph) -> nx.DiGraph:
-    """Build a weighted directed graph of calls between clusters."""
-    node_to_cluster: dict[str, int] = {}
-    for cluster_id, nodes in cluster_result.clusters.items():
-        for node in nodes:
-            node_to_cluster[node] = cluster_id
-
-    meta_graph = nx.DiGraph()
-    for cid in cluster_result.clusters:
-        meta_graph.add_node(cid)
-
-    edge_weights: dict[tuple[int, int], int] = defaultdict(int)
-    for src, dst in cfg_graph.edges():
-        src_cid = node_to_cluster.get(src)
-        dst_cid = node_to_cluster.get(dst)
-        if src_cid is not None and dst_cid is not None and src_cid != dst_cid:
-            edge_weights[(src_cid, dst_cid)] += 1
-
-    for (src_cid, dst_cid), weight in edge_weights.items():
-        meta_graph.add_edge(src_cid, dst_cid, weight=weight)
-
-    return meta_graph
-
-
 def group_symbols(cluster_ids: list[int], node_lookup: dict[int, set[str]]) -> list[str]:
     """Qualified names in a group, most top-level first (fewest name segments)."""
     names = {qname for cid in cluster_ids for qname in node_lookup.get(cid, set())}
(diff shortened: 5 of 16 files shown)


static_analyzer/clustering/grouping.py
```diff
@@ -319,6 +290,11 @@ def _modularity(meta_graph: nx.DiGraph, groups: list[set[int]]) -> float:
     return nx_comm.modularity(meta_graph, groups, weight="weight") if meta_graph.number_of_edges() else 0.0
 
 
+def score_grouping(cluster_result: ClusterResult, cfg_graph: nx.DiGraph, groups: list[set[int]]) -> float:
+    """Score a concrete grouping against its inter-cluster meta-graph."""
+    return _modularity(_build_meta_graph(cluster_result, cfg_graph), groups)
```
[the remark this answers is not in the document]
ivanmilevtues: yea we should address that, it is not too big of a deal reorganize as per expectation.

static_analyzer/clustering/models.py
```diff
@@ -75,3 +75,49 @@ class AnchoredGrouping:
     groups: list[set[int]]
     owners: list[str]
     regrouped: bool
+
+
+@dataclass
+class ClusterConnectionEdge:
+    """One concrete call crossing between two structural groups."""
```
ivanmilevtues: this i don't like we can drop the commend and make things a bit more specific:
source_method: string or source_qualifed_name: string
target_method: str or target_qualified_name: string

call_sites i don't really get the dict, isn't this a list of lines of code where we exactly call the symbols.

static_analyzer/clustering/models.py
```diff
@@ -75,3 +75,49 @@ class AnchoredGrouping:
...
+    target_group_id: str
+    edges: list[ClusterConnectionEdge] = field(default_factory=list)
+
+
+@dataclass
+class ClusterGroup:
+    """One deterministic architectural group inside a clustered scope."""
+
+    group_id: str
+    cluster_ids: list[int]
```
ivanmilevtues: I think we have cluster_ids and component_ids as specfici type one is int the other is "{a}.{b}" and such things pelase make sure we use the right types.

static_analyzer/clustering/models.py
```diff
@@ -75,3 +75,49 @@ class AnchoredGrouping:
...
+    members: dict[str, set[str]] = field(default_factory=dict)
```
ivanmilevtues: just members has no context i have no idea what do we have here by members, are these the methods? if so add method_memebrs or maybe symbol_memebrs or what this is, also it is not clear key -> value, what are both?

static_analyzer/clustering/models.py
```diff
@@ -75,3 +75,49 @@ class AnchoredGrouping:
...
+    def qualified_names(self) -> set[str]:
+        return {qualified_name for language_members in self.members.values() for qualified_name in language_members}
+
+
+@dataclass
+class ClusterScopeResult:
+    """The complete partition, grouping, and communication for one graph scope."""
+
+    scope_id: str
+    partitions: dict[str, ClusterResult] = field(default_factory=dict)
```
ivanmilevtues: I am not really sure what is a partition, what is the key here?

I think groups are gropued clusters, clusters are the things from the algorithm so what are partitions?

It is important for me to udnerstand these. Please answer here in a comment so we can imporve the naming and things can get clearer.

static_analyzer/clustering/grouping.py
```diff
@@ -52,32 +49,6 @@ def anchored_group(
         return _anchored_group(combined, combined_cfg, previous_owner, config)
 
 
-def build_all_cluster_results(static_analysis: StaticAnalysisResults) -> dict[str, ClusterResult]:
```
ivanmilevtues, line 55: these are moved. information for reviewers.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +41,190 @@ class ClusteringService:
...
+        *,
+        scope_id: str = _ROOT_SCOPE_ID,
+        partitions: Mapping[str, ClusterResult] = _EMPTY_PARTITIONS,
+        previous_owner: Mapping[int, str] = _EMPTY_OWNERS,
+        method_level_fallback: bool = False,
+    ) -> ClusterScopeResult:
+        """Cluster and group one exact graph scope, retaining sibling communication."""
+        scope_partitions: dict[str, ClusterResult] = {}
+        for language, graph in graphs.items():
+            scope_partitions[language] = partitions[language] if language in partitions else self.cluster(graph)
```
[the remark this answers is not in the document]
ivanmilevtues: I need to double check here, if it is missing, how do we know that it is not just an addition. Meaning that we have just added a new PL with this new change i.e. adding UI so JS suddenly shows up.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +44,191 @@ class ClusteringService:
...
+        high: int = TOP_LEVEL_COMPONENTS_MAX,
+        method_level_fallback: bool = False,
+    ) -> ClusterScopeResult:
+        """Cluster and group one exact graph scope, retaining sibling communication."""
+        scope_partitions: dict[str, ClusterResult] = {}
+        for language, graph in graphs.items():
+            scope_partitions[language] = partitions[language] if language in partitions else self.cluster(graph)
+        if method_level_fallback:
+            scope_partitions = {
+                language: self._expand_to_method_level(graphs[language], partition)
```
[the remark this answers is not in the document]
ivanmilevtues: This sounds like a bug can you write a minimal test which reproduces this and shows the issue. I get it that the result is that our matching will be off and so the clusters will get scrambled.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +44,191 @@ class ClusteringService:
...
+                for language, partition in scope_partitions.items()
+            }
+        reindex_across_languages(scope_partitions)
```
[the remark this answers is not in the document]
ivanmilevtues: How was this before, we should preserve previous behavior if possible, why did we change this if that was the case.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +44,191 @@ class ClusteringService:
...
+        )
+
+    def _assign_members(
+        self,
+        graphs: Mapping[str, CallGraph],
+        partitions: dict[str, ClusterResult],
+        groups: list[ClusterGroup],
+    ) -> None:
+        if not groups:
+            return
```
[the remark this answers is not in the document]
ivanmilevtues: can you check if this is the same behvarior as we used to have. I agree we should probably add a logging message here?
Please answer before we continue thou.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +44,191 @@ class ClusteringService:
...
+            partition = partitions.get(language, ClusterResult())
+            cluster_by_member = {
+                qualified_name: cluster_id
+                for cluster_id, members in partition.clusters.items()
+                for qualified_name in members
+            }
+            undirected = graph.to_networkx(reference_kinds=()).to_undirected()
+            for qualified_name, node in graph.nodes.items():
+                if node.type not in CALLABLE_TYPES | CLASS_TYPES:
+                    continue
```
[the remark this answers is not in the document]
ivanmilevtues: validate if this is behvaior change

static_analyzer/clustering/service.py
```diff
@@ -16,3 +44,191 @@ class ClusteringService:
...
+                if cluster_id not in group_by_cluster:
+                    cluster_id = self._cluster_for_file(node.file_path, partition)
+                if cluster_id not in group_by_cluster:
+                    cluster_id = self._nearest_cluster(qualified_name, partition, undirected)
+                group = group_by_cluster[cluster_id] if cluster_id is not None else groups[0]
+                group.members.setdefault(language, set()).add(qualified_name)
+
+    @staticmethod
+    def _cluster_for_file(file_path: str, partition: ClusterResult) -> int | None:
+        return next(iter(partition.get_clusters_for_file(file_path)), None)
```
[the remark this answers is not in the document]
ivanmilevtues: validate if this is behvaror change

static_analyzer/clustering/grouping.py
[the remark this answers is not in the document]
ivanmilevtues: Confirmed. This is an API-ordering issue, not a behavior change, and it is still present at `4e29cf5`. The check also found that `group_symbols` and `combine_cluster_results` already sit below `_build_meta_graph`, so satisfying the public-first convention means moving all public module functions (including `score_grouping`) above the first private helper, rather than moving only this one.

static_analyzer/clustering/models.py
ivanmilevtues: Confirmed: `source` and `target` come directly from `Edge.get_source()` / `get_destination()`, so they are fully qualified symbol names. `source_qualified_name` and `target_qualified_name` would make that contract clear, and the generic dataclass comment can be removed. `call_sites` is not only a list of line numbers: the CFG contract is a list of normalized source-location mappings containing `file`, `line`, and optionally `column`. It should retain that information (or use a shared typed location model), not be reduced to `list[int]`.

static_analyzer/clustering/models.py
ivanmilevtues: Confirmed. `cluster_ids` are graph-cluster IDs (`GraphClusterId`, represented by `int`). `group_id`, the connection endpoint IDs, and `previous_component_id` are component-style hierarchical IDs (for example `1.3`, represented by `str`). The repo has `GraphClusterId` and `CodeBoardingClusterId` aliases under `agents/cluster_ids.py`, but no distinct `ComponentId`; importing the agent layer into `static_analyzer` would invert the dependency, and `CodeBoardingClusterId` is semantically a source-cluster ID rather than a component ID. The current raw `int`/`str` annotations therefore do lose a real distinction; the fix should put/reuse these aliases at a shared lower-level boundary and add a component-ID type.

static_analyzer/clustering/models.py
ivanmilevtues: The mapping is `language -> set of qualified symbol names`. In this implementation the values contain callable and class-like symbols, not only methods, and `qualified_names` simply flattens those per-language sets. A name such as `qualified_names_by_language` (or `symbol_members_by_language`) would expose both key and value semantics much more clearly than `members`.

static_analyzer/clustering/models.py
ivanmilevtues: Here “partition” is the graph-theory term: each value is the leaf-cluster partition (`ClusterResult`) of one language graph within this scope. The key is the language. `groups` are the next level up and combine numeric leaf-cluster IDs from those per-language results. The distinction is valid, but the field is unnecessarily opaque; `cluster_results_by_language` matches the vocabulary already used throughout the repository and would be clearer than `partitions`.

static_analyzer/clustering/grouping.py
ivanmilevtues: Confirmed as reviewer information: `build_all_cluster_results` and cache synchronization were moved from `clustering/grouping.py` to `clustering/orchestration.py`. `static_analyzer.cluster_helpers` still exports the same compatibility symbol. The implementation is semantically unchanged; only logging style and private-docstring wording changed. The focused helper/scope suites pass (`20 passed`).

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: A missing entry is ambiguous from this API alone, so the concern about a genuinely added language is valid. The existing incremental pipeline resolves that before grouping: `compute_cluster_delta` emits a `ClusterResult` for every current language, including a newly added language, and allocates its IDs above all baseline IDs. Therefore, when `previous_owner` is supplied, a missing partition should be treated as incomplete incremental input; in a fresh call without previous ownership, omission can continue to mean “cluster this graph now.” A blanket rejection for all calls would be wrong, but a conditional contract/check (or an explicit new-language marker) is needed.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: Confirmed with a minimal reproducer on `4e29cf5`. A five-function partition `{42: {a,b,c,d,e}}` with `previous_owner={42: "1.3"}` and `method_level_fallback=True` is expanded to IDs `[0,1,2,3,4]`; the result has group IDs `["1","2","3","4","5"]`, every `previous_component_id` is empty, and `regrouped=True`. The existing nested-scope path expanded first and derived ownership from members afterward, so it did not have this stale-ID seam. A regression test should assert that the group containing those expanded members retains `1.3`; the implementation needs to remap the original owner onto the expanded IDs before anchored grouping.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: This is a behavior change at the new `cluster_scope` boundary, although `reindex_across_languages` itself was not changed. Previously, full analysis reindexed only fresh results (no ownership map), while incremental analysis allocated new IDs above the baseline and/or derived ownership after final reindexing. `cluster_scope` is the first path that accepts `previous_owner` and then changes partition IDs. The reproducer moves retained Python IDs `{1,2}` to `{4,5}` when JavaScript also has `{1,2}`, while owners `7`/`8` remain keyed to `{1,2}` and are consequently attached to JavaScript. Preserve seeded IDs and shift only genuinely new collisions, or remap ownership together with every shifted partition.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: This is not the previous behavior. Full analysis explicitly raised `StaticAnalysisFatalError` when a non-empty repository produced no groups; incremental scope planning raised `IncrementalClusteringError` when a populated scope produced no clusters; child-scope construction automatically expanded coarse/empty natural results to method-level clusters. On the current head, two isolated functions produce an empty partition and `cluster_scope()` returns a successful result with `groups=[]`. A log message alone would preserve the silent bad result; this needs either the appropriate method-level expansion or a specific exception.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: This is not a behavior change relative to the existing persisted component/relations path. Existing `populate_file_methods` initially routes all CFG nodes, but `_build_file_methods_from_nodes` then keeps only `CALLABLE_TYPES | CLASS_TYPES`; static relation construction maps endpoints from those final `file_methods`, so calls involving other symbol kinds were already omitted. The new filter mirrors that established contract (and the PR description promises complete callable/class ownership, not ownership of every CFG symbol). Including data/other edge participants would be a separate behavior expansion, not a compatibility fix for this PR.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: This is also pre-existing behavior, not a change introduced by this PR. The current agent-side `_find_component_by_file` iterates `get_clusters_for_file()` and returns the first mapped component before trying graph distance; when a file belongs to multiple clusters, that is the same arbitrary set choice as `next(iter(...))` here. It is a real semantic limitation worth addressing separately (prefer containment/nearest affinity), but changing it in this contract PR would intentionally change assignment behavior.

clustering_ids.py
```diff
@@ -0,0 +1,5 @@
+"""Identifier types shared across clustering layers."""
+
+type GraphClusterId = int
```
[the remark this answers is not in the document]
ivanmilevtues: that is a great catch. I think we maybe now do not need both at both locations I am not sure, in general i think we should reanme the ComponentId and CodeBoardingClusterId and GraphCliusterId.

So GraphClusterId int clear
ComponentId/CodeBoardingClusterId -> here I think one of these should be a groupId

the whole typenames are not consistent with how they are used. we have groups and components. so lets rediscuss this.

static_analyzer/clustering/service.py
```diff
@@ -16,3 +41,190 @@ class ClusteringService:
...
+    def _allocate_group_ids(scope_id: str, owners: list[str]) -> list[str]:
+        allocated = set(owners) - {""}
+        result: list[str] = []
+        next_index = 1
+        for owner in owners:
+            if owner:
+                result.append(owner)
+                continue
+            while True:
+                candidate = str(next_index) if scope_id == _ROOT_SCOPE_ID else f"{scope_id}.{next_index}"
```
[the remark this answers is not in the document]
ivanmilevtues: so we had lots of these commends on redindexing things and finding ids, we shouldn't change the way things were so pelase keep the same behavior as it used to. Make sure and answer to me in this PR comment what is the result.

ivanmilevtues (commented):
few comments.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: Result: the established `assign_component_ids(..., only_new=True)` behavior is to append after the highest surviving sibling index. The scope allocator now mirrors that behavior exactly rather than filling numeric gaps or adding a separate historical-ID input. For example, when component `1` is retired, component `2` survives, and a new group appears, the IDs are `2` and `3`—the retired `1` is not reused. No cluster reindexing behavior was changed. A regression test covers this case.

clustering_ids.py
```diff
@@ -0,0 +1,6 @@
+"""Identifier types shared across clustering layers."""
+
+type ClusterId = int
```
Svilen-Stefanov: Flagging one more file that's unassigned in the analysis that should be

static_analyzer/cfg/__init__.py
```diff
@@ -5,6 +5,6 @@
 """
 
 from static_analyzer.cfg.call_graph import CallGraph
-from static_analyzer.cfg.edge import DEFAULT_REFERENCE_KINDS, Edge, EdgeKind, ReferenceEdge
+from static_analyzer.cfg.edge import CallSiteLocation, DEFAULT_REFERENCE_KINDS, Edge, EdgeKind, ReferenceEdge
```
Svilen-Stefanov: There are 3 more in this review, including this one.

static_analyzer/clustering/service.py
```diff
@@ -2,17 +2,253 @@
...
+                raise LeafClustersUnavailableError(language)
+        reindex_across_languages(scope_leaf_clusters)
+
+        nx_graphs = {language: graph.to_networkx(DEFAULT_REFERENCE_KINDS) for language, graph in graphs.items()}
+        grouping_service = GroupingService()
+        if previous_owner:
+            grouping = grouping_service.anchored_group(scope_leaf_clusters, nx_graphs, dict(previous_owner))
+            raw_groups = grouping.groups
+            owners = grouping.owners
+            combined = combine_cluster_results(scope_leaf_clusters)
```
Svilen-Stefanov: Do we really need these three lines? `anchored_group` already builds the same `combined` and `combined_cfg` internally, and `_anchored_group` already computes `_modularity(meta_graph, groups)` for the drift check, it just does not hand it back. So we rebuild the meta graph a second time here only to recover a number we already had. Could `AnchoredGrouping` just carry its own modularity? On main the caller built these once and passed them down, so this is new here.

static_analyzer/clustering/service.py
```diff
@@ -2,17 +2,253 @@
...
+                candidate = str(next_index) if scope_id == _ROOT_SCOPE_ID else f"{scope_id}.{next_index}"
+                next_index += 1
+                if candidate not in allocated:
+                    allocated.add(candidate)
+                    result.append(candidate)
+                    break
+        return result
+
+    @staticmethod
+    def _expand_to_method_level(graph: CallGraph, cluster_result: ClusterResult) -> ClusterResult:
```
Svilen-Stefanov: This looks like a copy of `ClusterMethodsMixin._expand_to_method_level_clusters`: same `MIN_CLUSTERS_THRESHOLD`, same sorted callable pass, same class fallback, just without the log line. I had a look at #498 and #500 and both keep the mixin copy as well, so we would be carrying two of these for a while. Can we share one implementation, or at least make a ticket so they do not drift apart?

static_analyzer/cfg/edge.py
```diff
@@ -76,8 +85,22 @@ def visit_paths(self, fn: Callable[[str], str]) -> None:
         self._call_site_keys = {tuple(sorted(site.items())) for site in self._call_sites}
 
     @staticmethod
-    def _normalize_call_site(call_site: Mapping[str, Hashable]) -> dict[str, Hashable]:
+    def _normalize_call_site(call_site: Mapping[str, Hashable]) -> CallSiteLocation:
         normalized = dict(call_site)
```
Svilen-Stefanov: nit: now that we build and return a fresh `CallSiteLocation` at the bottom, this copy and the `pop` only exist for the `file_path` fallback. Could we read `call_site.get("file")` with a `call_site.get("file_path")` fallback straight off the incoming mapping and drop the copy? Otherwise it is an extra dict per call site and the copy is thrown away.

static_analyzer/clustering/__init__.py
```diff
@@ -7,18 +7,36 @@
 
 from static_analyzer.clustering.cache import ClusterCache
 from static_analyzer.config import DEFAULT_GROUPING_CONFIG, SUBCOMPONENT_GROUPING_CONFIG, GroupingConfig
+from clustering_ids import ClusterId, ComponentId, GroupId, ScopeId
```
Svilen-Stefanov: Same thing I mentioned on 494, this list keeps growing. I grepped `from static_analyzer.clustering import` and every call site only pulls `ClusterResult`, `ClusterCache`, `ClusteringService`, `METHOD_LEVEL_STRATEGY` and `MethodClusterPaths`. The four id aliases here, and the scope/group/connection types below, have no importer through the package, everything goes to `clustering_ids` or `.models` directly. Can we only export what is actually consumed?

clustering_ids.py
Svilen-Stefanov, line 3: Flagging one more file that's unassigned in the analysis that should be

static_analyzer/cfg/__init__.py
Svilen-Stefanov, line 8: There are 3 more in this review, including this one.

static_analyzer/clustering/service.py
Svilen-Stefanov, line 90: Do we really need these three lines? `anchored_group` already builds the same `combined` and `combined_cfg` internally, and `_anchored_group` already computes `_modularity(meta_graph, groups)` for the drift check, it just does not hand it back. So we rebuild the meta graph a second time here only to recover a number we already had. Could `AnchoredGrouping` just carry its own modularity? On main the caller built these once and passed them down, so this is new here.

static_analyzer/clustering/service.py
Svilen-Stefanov, line 143: This looks like a copy of `ClusterMethodsMixin._expand_to_method_level_clusters`: same `MIN_CLUSTERS_THRESHOLD`, same sorted callable pass, same class fallback, just without the log line. I had a look at #498 and #500 and both keep the mixin copy as well, so we would be carrying two of these for a while. One of them should go away?

static_analyzer/cfg/edge.py
Svilen-Stefanov, line 89: nit: now that we build and return a fresh `CallSiteLocation` at the bottom, this copy and the `pop` only exist for the `file_path` fallback. Could we read `call_site.get("file")` with a `call_site.get("file_path")` fallback straight off the incoming mapping and drop the copy? Otherwise it is an extra dict per call site and the copy is thrown away.

static_analyzer/clustering/__init__.py
Svilen-Stefanov, line 10: Same thing I mentioned on 494, this list keeps growing. I grepped `from static_analyzer.clustering import` and every call site only pulls `ClusterResult`, `ClusterCache`, `ClusteringService`, `METHOD_LEVEL_STRATEGY` and `MethodClusterPaths`. The four id aliases here, and the scope/group/connection types below, have no importer through the package, everything goes to `clustering_ids` or `.models` directly. Can we only export what is actually consumed?

Svilen-Stefanov (approved):
Also fairly easy to review, thanks! LGTM - left a couple more comments, mostly clean ups and things to look into.

static_analyzer/clustering/service.py
ivanmilevtues: Fixed in #500: `agents/cluster_methods_mixin.py` is deleted. Method-level expansion now has one owner, `ClusteringService._expand_to_method_level`, used by scope clustering when fallback is required.

static_analyzer/clustering/__init__.py
ivanmilevtues: Fixed/superseded in #500: the old ID aliases and unused exports were removed from `static_analyzer/clustering/__init__.py`. The remaining scope/group/connection model exports are imported by current agent, renderer, diagram-analysis, and test code.

static_analyzer/clustering/service.py
ivanmilevtues: Fixed later at the top of the stack in `9f948048`. `AnchoredGrouping` now carries the actual anchored `modularity` alongside `unanchored_modularity`, and `ClusteringService` reads `grouping.modularity` directly instead of rebuilding the combined partition, CFG, and meta-graph.

static_analyzer/cfg/edge.py
ivanmilevtues: Fixed at the top of the stack in `dc1eb960`. `_normalize_call_site` now reads `file` with a direct `file_path` fallback from the incoming mapping, and reads `line`/`column` directly as well; the discarded dictionary copy and mutation are gone. The focused call-graph/call-site suite passes (29 tests), with Black and mypy clean.

ivanmilevtues:

PR #497, opened 2026-08-18
feat(clustering): build recursive scope hierarchy

ivanmilevtues:
## Goal

Build the entire deterministic clustering hierarchy before analysis agents run, rather than discovering and regrouping each component independently while traversing the analysis tree.

Each child scope is computed from the exact induced subgraph owned by its parent group, so hierarchy depth and boundaries come from one structural pipeline.

## Behavioral impact (upfront)

This PR adds recursive hierarchy construction on top of the one-scope result introduced in #495. It does **not** yet make the existing full or incremental analysis pipelines consume that hierarchy; those integrations remain in the follow-up PRs.

| Area | Impact |
|---|---|
| Existing full/incremental analysis | **No intended runtime change.** Existing agents still own hierarchy traversal until #498 and #499 migrate those paths. |
| Hierarchy API | **Intentional new API.** `ClusteringService.cluster_hierarchy(...)` recursively composes `ClusterScopeResult`s up to the requested depth. |
| Child scope boundary | **Intentional contract tightening.** Every child is clustered from the exact per-language induced subgraphs of its parent group, excluding sibling-only symbols and edges. |
| Coarse child partitions | **Preserves the existing fallback behavior.** Child scopes with fewer than `MIN_CLUSTERS_THRESHOLD` natural clusters expand to callable-level clusters, then classes when needed. |
| Incremental ownership during fallback | **Stable component/group IDs are preserved.** Synthetic leaf-cluster IDs may change because fallback creates a new partition, but previous ownership is remapped through the original members after fallback and cross-language reindexing. |
| Non-splitting child results | **No redundant hierarchy levels.** A result with fewer than two non-empty child groups is not attached and does not consume recursion depth. |
| Expansion sizing | **Correctness fix.** The minimum-size gate counts actual `CALLABLE_TYPES` in the exact child graphs instead of counting members retained by natural clustering or class symbols. |
| Expansion policy | **No threshold change.** The existing method/file ceilings and modularity threshold move to `ClusteringConfig` and are shared by hierarchy construction and the existing planner. |
| Modularity metadata | **Intentional naming clarification.** `modularity` scores the actual returned grouping, including ownership anchors; `unanchored_modularity` scores the best grouping without previous ownership and drives structural separability decisions. |

## Result contract and API mapping

| Before | After | Contract / reason |
|---|---|---|
| Callers repeatedly derive child subgraphs and invoke one-scope clustering while traversing analysis components | `ClusteringService.cluster_hierarchy(graphs, max_depth, scope_input)` | Builds the complete structural hierarchy once, before downstream analysis consumes it. |
| Component/file-level traversal can include sibling symbols from shared files | Each child uses `CallGraph.filter_by_nodes(...)` with its parent group's exact language-qualified symbol membership | Makes every recursive scope an exact induced graph boundary. |
| Child partition and prior ownership acquisition are coupled to individual analysis paths | `ClusterScopeInput(leaf_clusters_by_language, previous_owner)` callback | Lets fresh and incremental callers seed each scope without coupling clustering to persisted analysis models. |
| Planner-local size and separability rules | Shared `clustering/expansion.py` rules backed by `ClusteringConfig` | Keeps the hierarchy and remaining planner path on one deterministic policy. |
| Cluster-member counts stand in for scope method counts | Actual callable count from the exact child `CallGraph`s is passed into `scope_is_separable(...)` | Includes singleton callables omitted by natural clustering and excludes classes from the method floor. |
| Fallback replaces local cluster IDs while ownership remains keyed to the old partition | Ownership is projected from original cluster members onto the final fallback/reindexed partition | Prevents numeric ID collisions from assigning unrelated historical component IDs. |
| Anchored grouping exposes a vaguely named `fresh_modularity` | `unanchored_modularity` on `AnchoredGrouping` and `ClusterScopeResult` | Makes clear that this score is the ownership-independent structural baseline, not another group's state. |
| A one-group child can repeat its parent's membership and consume depth | Child scopes require at least two non-empty groups before attachment | Prevents redundant single-child chains and false expansion. |

`cluster_hierarchy()` clusters the root scope, builds exact induced graphs for each group, obtains optional scope-local partitions and ownership anchors through `ClusterScopeInput`, applies method-level fallback to coarse child partitions, and attaches only structurally justified multi-group children until `max_depth` is reached.

For anchored scopes, `modularity` remains the score of the groups that will actually be returned. `unanchored_modularity` is retained separately because expansion should reflect the graph's current structural separability rather than cohesion introduced by a persisted ownership layout.

## Compatibility boundary

`ClusteringService.cluster()` and `cluster_scope()` remain available. The existing planner retains its public behavior and now delegates its shared size/modularity decision to `clustering/expansion.py`.

This PR only **produces** the recursive hierarchy. It does not materialize analysis components from it and does not replace the current full or incremental orchestration. #498 migrates fresh/full analysis, #499 migrates ownership-anchored incremental analysis, and #500 removes the superseded paths.

## Validation

- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` — **2,156 passed, 28 skipped**, **83.14% coverage**.
- Focused hierarchy, scope, and planner suites — **41 passed**.
- `uv run mypy .` — passed across 336 source files.
- `uv run black ...` — passed for all changed Python files.
- Commit pre-commit hook — passed.
- `git diff --check` — passed.

## Stack roadmap

|---|---|---|
| Foundation | [#494](https://github.com/CodeBoarding/CodeBoarding/pull/494) | Move the existing grouping algorithms behind the stateless `GroupingService` API without intentionally changing grouping behavior. |
| Previous | [#495](https://github.com/CodeBoarding/CodeBoarding/pull/495) | Define and compute a complete structural result for one graph scope, including groups, members, and sibling communication. |
| **This PR** | **#497** | Recursively compose scope results into a deterministic hierarchy using exact induced subgraphs and centralized expansion rules. |
| Next | [#498](https://github.com/CodeBoarding/CodeBoarding/pull/498) | Make fresh/full analysis consume the precomputed hierarchy instead of regrouping inside agents. |
| Follow-up | [#499](https://github.com/CodeBoarding/CodeBoarding/pull/499) | Make incremental analysis build and consume an ownership-anchored hierarchy while preserving stable member ownership. |
| Final cleanup | [#500](https://github.com/CodeBoarding/CodeBoarding/pull/500) | Remove the legacy agent-side grouping and recursive incremental orchestration once both analysis paths use the hierarchy. |

---

### DIFF
--- a/agents/planner_agent.py
+++ b/agents/planner_agent.py
@@ -19,6 +19,14 @@ import networkx as nx
 
 from agents.agent_responses import AnalysisInsights, Component
 from static_analyzer.clustering import METHOD_LEVEL_STRATEGY, ClusterResult
+from static_analyzer.clustering.expansion import (
+    EXPAND_MODULARITY_THRESHOLD,
+    MAX_LEAF_FILES,
+    MAX_LEAF_METHODS,
+    MIN_METHODS_TO_EXPAND,
+    scope_is_separable,
+    scope_load,
+)
 from static_analyzer.clustering.grouping import GroupingService
 
 logger = logging.getLogger(__name__)
@@ -26,27 +34,11 @@ logger = logging.getLogger(__name__)
 # Default thresholds
 DEFAULT_MIN_FILES = 1  # Need at least 1 file to have content
 
-# Below this a component holds too little to be worth sub-dividing at all.
-MIN_METHODS_TO_EXPAND = 30
-
-# The size at which a component stops being readable as one box and must be split
-# whatever its call structure says. Measured across the eval corpus: at 12 files /
-# 120 methods every repo's tree comes out with no oversized leaf, while small repos
-# are untouched (they stop at the modularity gate long before this).
-MAX_LEAF_FILES = 12
-MAX_LEAF_METHODS = 120
-
-# Modularity a *small* component's split must reach to be worth making. The bar
-# ramps linearly to zero as the component approaches the leaf ceiling: a large
-# component gets split on weaker structural evidence, because leaving it whole
-# costs the reader more than an imperfect boundary does.
-EXPAND_MODULARITY_THRESHOLD = 0.15
-
 
 def leaf_load(component: Component) -> float:
     """How full a component is against the leaf ceiling; >= 1.0 means too big to leave whole."""
     methods = sum(len(group.methods) for group in component.file_methods)
-    return max(methods / MAX_LEAF_METHODS, len(component.file_methods) / MAX_LEAF_FILES)
+    return scope_load(methods, len(component.file_methods))
 
 
 def component_is_separable(
@@ -74,7 +66,7 @@ def component_is_separable(
         return False
     _groups, modularity = GroupingService().group(cluster_results, cfg_graphs, subcomponents=True)
     required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
-    separable = modularity >= required
+    separable = scope_is_separable(cluster_results, modularity, load, min_methods)
     logger.debug(
         f"[Planner] subgraph modularity={modularity:.4f} (load={load:.2f}, required {required:.4f}) "
         f"-> separable={separable}"
--- a/agents/planner_agent.py
+++ b/agents/planner_agent.py
@@ -19,15 +19,9 @@ import networkx as nx
 
 from agents.agent_responses import AnalysisInsights, Component
 from static_analyzer.clustering import METHOD_LEVEL_STRATEGY, ClusterResult
-from static_analyzer.clustering.expansion import (
-    EXPAND_MODULARITY_THRESHOLD,
-    MAX_LEAF_FILES,
-    MAX_LEAF_METHODS,
-    MIN_METHODS_TO_EXPAND,
-    scope_is_separable,
-    scope_load,
-)
+from static_analyzer.clustering.expansion import scope_is_separable, scope_load
 from static_analyzer.clustering.grouping import GroupingService
+from static_analyzer.config import ClusteringConfig
 
 logger = logging.getLogger(__name__)
 
@@ -45,7 +39,7 @@ def component_is_separable(
     cluster_results: dict[str, ClusterResult],
     cfg_graphs: dict[str, nx.DiGraph],
     load: float,
-    min_methods: int = MIN_METHODS_TO_EXPAND,
+    min_methods: int = ClusteringConfig.MIN_METHODS_TO_EXPAND,
 ) -> bool:
     """Whether a component's own call structure justifies splitting it into sub-components.
 
@@ -65,8 +59,8 @@ def component_is_separable(
         logger.debug("[Planner] subgraph has no natural cluster structure; keeping as leaf")
         return False
     _groups, modularity = GroupingService().group(cluster_results, cfg_graphs, subcomponents=True)
-    required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
-    separable = scope_is_separable(cluster_results, modularity, load, min_methods)
+    required = ClusteringConfig.EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
+    separable = scope_is_separable(cluster_results, modularity, load, total_methods, min_methods)
     logger.debug(
         f"[Planner] subgraph modularity={modularity:.4f} (load={load:.2f}, required {required:.4f}) "
         f"-> separable={separable}"
--- a/static_analyzer/clustering/expansion.py
+++ b/static_analyzer/clustering/expansion.py
@@ -1,0 +1,35 @@
+"""Deterministic rules for deciding whether a clustered scope should split."""
+
+from collections.abc import Mapping
+
+from static_analyzer.clustering.models import METHOD_LEVEL_STRATEGY, ClusterResult
+
+MIN_METHODS_TO_EXPAND = 30
+MAX_LEAF_FILES = 12
+MAX_LEAF_METHODS = 120
+EXPAND_MODULARITY_THRESHOLD = 0.15
+
+
+def scope_load(method_count: int, file_count: int) -> float:
+    """How full a scope is against the leaf ceiling."""
+    return max(method_count / MAX_LEAF_METHODS, file_count / MAX_LEAF_FILES)
+
+
+def scope_is_separable(
+    leaf_clusters_by_language: Mapping[str, ClusterResult],
+    modularity: float,
+    load: float,
+    min_methods: int = MIN_METHODS_TO_EXPAND,
+) -> bool:
+    """Whether a scope's natural clustering and modularity justify a split."""
+    total_methods = sum(
+        len(members)
+        for cluster_result in leaf_clusters_by_language.values()
+        for members in cluster_result.clusters.values()
+    )
+    if total_methods < min_methods:
+        return False
+    if all(cluster_result.strategy == METHOD_LEVEL_STRATEGY for cluster_result in leaf_clusters_by_language.values()):
+        return False
+    required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
+    return modularity >= required
--- a/static_analyzer/clustering/expansion.py
+++ b/static_analyzer/clustering/expansion.py
@@ -2,34 +2,29 @@
 
 from collections.abc import Mapping
 
+from static_analyzer.config import ClusteringConfig
 from static_analyzer.clustering.models import METHOD_LEVEL_STRATEGY, ClusterResult
-
-MIN_METHODS_TO_EXPAND = 30
-MAX_LEAF_FILES = 12
-MAX_LEAF_METHODS = 120
-EXPAND_MODULARITY_THRESHOLD = 0.15
 
 
 def scope_load(method_count: int, file_count: int) -> float:
     """How full a scope is against the leaf ceiling."""
-    return max(method_count / MAX_LEAF_METHODS, file_count / MAX_LEAF_FILES)
+    return max(
+        method_count / ClusteringConfig.MAX_LEAF_METHODS,
+        file_count / ClusteringConfig.MAX_LEAF_FILES,
+    )
 
 
 def scope_is_separable(
     leaf_clusters_by_language: Mapping[str, ClusterResult],
     modularity: float,
     load: float,
-    min_methods: int = MIN_METHODS_TO_EXPAND,
+    method_count: int,
+    min_methods: int = ClusteringConfig.MIN_METHODS_TO_EXPAND,
 ) -> bool:
     """Whether a scope's natural clustering and modularity justify a split."""
-    total_methods = sum(
-        len(members)
-        for cluster_result in leaf_clusters_by_language.values()
-        for members in cluster_result.clusters.values()
-    )
-    if total_methods < min_methods:
+    if method_count < min_methods:
         return False
     if all(cluster_result.strategy == METHOD_LEVEL_STRATEGY for cluster_result in leaf_clusters_by_language.values()):
         return False
-    required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
+    required = ClusteringConfig.EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
     return modularity >= required
--- a/static_analyzer/clustering/models.py
+++ b/static_analyzer/clustering/models.py
@@ -77,7 +77,7 @@ class AnchoredGrouping:
     groups: list[set[ClusterId]]
     owners: list[ComponentId]
     regrouped: bool
-    fresh_modularity: float
+    unanchored_modularity: float
 
 
 @dataclass
@@ -132,6 +132,6 @@ class ClusterScopeResult:
     leaf_clusters_by_language: dict[str, ClusterResult] = field(default_factory=dict)
     groups: list[ClusterGroup] = field(default_factory=list)
     connections: list[GroupConnection] = field(default_factory=list)
-    modularity: float = 0.0
-    fresh_modularity: float = 0.0
+    modularity: float = 0.0  # Score of the actual groups, including ownership anchors.
+    unanchored_modularity: float = 0.0  # Best score without previous ownership anchors.
     regrouped: bool = False
--- a/static_analyzer/clustering/models.py
+++ b/static_analyzer/clustering/models.py
@@ -77,6 +77,7 @@ class AnchoredGrouping:
     groups: list[set[ClusterId]]
     owners: list[ComponentId]
     regrouped: bool
+    fresh_modularity: float
 
 
 @dataclass
@@ -94,6 +95,14 @@ class GroupConnection:
     source_group_id: GroupId
     target_group_id: GroupId
     edges: list[ClusterConnectionEdge] = field(default_factory=list)
+
+
+@dataclass
+class ClusterScopeInput:
+    """Optional precomputed leaf clusters and ownership anchors for one scope."""
+
+    leaf_clusters_by_language: Mapping[str, ClusterResult] = field(default_factory=dict)
+    previous_owner: Mapping[ClusterId, ComponentId] = field(default_factory=dict)
 
 
 @dataclass
@@ -124,4 +133,5 @@ class ClusterScopeResult:
     groups: list[ClusterGroup] = field(default_factory=list)
     connections: list[GroupConnection] = field(default_factory=list)
     modularity: float = 0.0
+    fresh_modularity: float = 0.0
     regrouped: bool = False
--- a/static_analyzer/clustering/service.py
+++ b/static_analyzer/clustering/service.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from collections import defaultdict
+from collections import Counter, defaultdict
 from collections.abc import Callable, Mapping
 from types import MappingProxyType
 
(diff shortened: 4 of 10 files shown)


static_analyzer/clustering/expansion.py
```diff
@@ -0,0 +1,35 @@
+"""Deterministic rules for deciding whether a clustered scope should split."""
+
+from collections.abc import Mapping
+
+from static_analyzer.clustering.models import METHOD_LEVEL_STRATEGY, ClusterResult
+
+MIN_METHODS_TO_EXPAND = 30
```
ivanmilevtues: this should be part of the config.py

static_analyzer/clustering/models.py
```diff
@@ -124,4 +133,5 @@ class ClusterScopeResult:
     groups: list[ClusterGroup] = field(default_factory=list)
     connections: list[GroupConnection] = field(default_factory=list)
     modularity: float = 0.0
+    fresh_modularity: float = 0.0
```
ivanmilevtues: i don't get the concept behind fresh_modularity can you explain and why is it in the clyuters gruop?

static_analyzer/clustering/expansion.py
```diff
@@ -0,0 +1,35 @@
...
+MAX_LEAF_FILES = 12
+MAX_LEAF_METHODS = 120
+EXPAND_MODULARITY_THRESHOLD = 0.15
+
+
+def scope_load(method_count: int, file_count: int) -> float:
```
ivanmilevtues, line 9: theese are moved

static_analyzer/clustering/service.py
```diff
@@ -98,9 +108,83 @@ def cluster_scope(
...
+                continue
+            child_input = scope_input(group.group_id, child_graphs)
+            child = self.cluster_scope(
+                child_graphs,
+                scope_id=group.group_id,
+                partitions=child_input.partitions,
+                previous_owner=child_input.previous_owner,
+                low=SUBCOMPONENTS_MIN,
+                high=SUBCOMPONENTS_MAX,
+                method_level_fallback=True,
```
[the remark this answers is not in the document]
ivanmilevtues: once again a comment on the ids, i just want to make sure that here we are not modifiying the ids as we need to do things.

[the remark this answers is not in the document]
ivanmilevtues: Addressed in `038fa352`. Method-level fallback still creates a new leaf partition, so synthetic leaf cluster IDs can change, but `previous_owner` is now remapped through the original cluster members after fallback and cross-language reindexing. This preserves the stable component/group IDs and prevents numeric cluster-ID collisions from reassigning ownership.

static_analyzer/clustering/service.py
```diff
@@ -113,9 +131,85 @@ def cluster_scope(
...
+            )
+            load = scope_load(method_count, file_count)
+            if load < 1.0 and not scope_is_separable(
+                child.leaf_clusters_by_language,
+                child.fresh_modularity,
+                load,
+            ):
+                continue
+            group.children = child
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in `038fa352`. The hierarchy now requires at least two non-empty child groups before attaching a child scope or recursing, so an anchored one-group result no longer consumes a depth level or creates redundant single-child chains.

static_analyzer/clustering/expansion.py
```diff
@@ -0,0 +1,35 @@
...
+
+def scope_is_separable(
+    leaf_clusters_by_language: Mapping[str, ClusterResult],
+    modularity: float,
+    load: float,
+    min_methods: int = MIN_METHODS_TO_EXPAND,
+) -> bool:
+    """Whether a scope's natural clustering and modularity justify a split."""
+    total_methods = sum(
+        len(members)
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in `038fa352`. The callable count is now computed directly from `CALLABLE_TYPES` in the exact child `CallGraph`s and passed into `scope_is_separable`; class symbols and callables omitted from natural clusters no longer distort the expansion floor.

static_analyzer/clustering/expansion.py
ivanmilevtues: Addressed in `038fa352`. The hierarchy expansion thresholds now live on `ClusteringConfig` in `static_analyzer/config.py`, and both the expansion rules and planner consume that source of truth.

static_analyzer/clustering/models.py
ivanmilevtues: `fresh_modularity` was scope-level metadata on `ClusterScopeResult`, not a field on `ClusterGroup`. It represented the best grouping score without previous-ownership anchors, while `modularity` scores the actual returned grouping, including anchors. The hierarchy needs the unanchored value to judge whether the code structure itself is separable rather than whether the persisted layout is separable. Addressed in `038fa352` by renaming it to `unanchored_modularity` and documenting both fields.

static_analyzer/clustering/expansion.py
ivanmilevtues: Yes. `scope_load` and `scope_is_separable` are the existing planner expansion rules moved and generalized into `static_analyzer/clustering/expansion.py`, so the clustering hierarchy and the existing planner share the same behavior.

static_analyzer/clustering/service.py
```diff
@@ -69,6 +75,15 @@ def cluster_scope(
             scope_leaf_clusters[language] = (
                 leaf_clusters_by_language[language] if language in leaf_clusters_by_language else self.cluster(graph)
             )
+        previous_owner_by_language_member = {
+            language: {
+                qualified_name: previous_owner[cluster_id]
+                for cluster_id, members in cluster_result.clusters.items()
+                if previous_owner.get(cluster_id)
+                for qualified_name in members
```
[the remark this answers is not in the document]
Svilen-Stefanov: This seems new - you might want to take a look at it.

static_analyzer/clustering/service.py
```diff
@@ -113,9 +150,87 @@ def cluster_scope(
...
+                continue
+            child_input = scope_input(group.group_id, child_graphs)
+            child = self.cluster_scope(
+                child_graphs,
+                scope_id=group.group_id,
+                leaf_clusters_by_language=child_input.leaf_clusters_by_language,
+                previous_owner=child_input.previous_owner,
+                method_level_fallback=True,
+            )
+            if sum(bool(child_group.qualified_names) for child_group in child.groups) < 2:
```
Svilen-Stefanov, line 198: blocker-ish: I think this gate reads the wrong grouping. It counts the groups the *anchored* path produced, but the very next gate deliberately passes `child.unanchored_modularity` so the decision stays anchoring independent. When `previous_owner` puts every live cluster on one component, `_anchored_group` builds a single carried group and we bail out right here.

I ran it on the tip with the 121 method single file graph (`scope_load` = 1.0083, so already past the leaf ceiling):

- seeded with `previous_owner={cid: "1.1"}` -> `children is None`, subtree dropped
- exact same graph unseeded -> 3 groups and it keeps recursing

So an incremental run and a full run would produce different depths for the same repo once #498 and #499 land. It also skips the `load >= 1.0` branch just below, which exists precisely to split oversized scopes whatever the structure says. Can we count the unanchored groups here, or move this check below the `load >= 1.0` short circuit?

tests/static_analyzer/test_clustering_hierarchy.py
```diff
@@ -0,0 +1,151 @@
...
+            partition = (
+                partition_for(current, {1: set(current.nodes)}) if scope_id == "root" else split_partition(current)
+            )
+            return ClusterScopeInput(leaf_clusters_by_language={"python": partition})
+
+        result = ClusteringService().cluster_hierarchy({"python": graph}, max_depth=3, scope_input=scope_input)
+
+        self.assertIsNone(result.groups[0].children)
+
+    def test_one_group_child_does_not_consume_a_depth_level(self):
```
Svilen-Stefanov, line 110: Related to my comment on `service.py:198`: I think this test pins that behaviour as correct. The graph here is past the leaf ceiling and the only reason it does not expand is the single carried owner. If we fix the gate, this should flip into asserting that a seeded and an unseeded run of the same graph agree.

tests/static_analyzer/test_clustering_hierarchy.py
```diff
@@ -0,0 +1,151 @@
...
+                partition = split_partition(current)
+            return ClusterScopeInput(leaf_clusters_by_language={"python": partition})
+
+        result = ClusteringService().cluster_hierarchy({"python": graph}, max_depth=2, scope_input=scope_input)
+
+        group_a = next(group for group in result.groups if group.qualified_names == members_a)
+        group_b = next(group for group in result.groups if group.qualified_names == members_b)
+        self.assertEqual(seen_nodes[group_a.group_id], members_a)
+        self.assertEqual(seen_nodes[group_b.group_id], members_b)
+        self.assertEqual(seen_edges[group_a.group_id], set())
```
Svilen-Stefanov, line 67: just checking, does this actually prove the induced graph keeps its own edges? The only edge in the fixture is `a1 -> b1`, which crosses the two groups, so node filtering alone removes it and the assertion passes either way. I tried making `_induced_graphs` throw away every edge of every child graph and the whole suite still passed, 2128 tests. Could we add an edge inside one group and assert that one survives?

tests/static_analyzer/test_clustering_hierarchy.py
```diff
@@ -0,0 +1,151 @@
...
+            current = graphs["python"]
+            partition = (
+                partition_for(current, {1: set(current.nodes)}) if scope_id == "root" else split_partition(current)
+            )
+            return ClusterScopeInput(leaf_clusters_by_language={"python": partition})
+
+        with patch("static_analyzer.clustering.service.scope_is_separable", return_value=False) as separable:
+            ClusteringService().cluster_hierarchy({"python": graph}, max_depth=2, scope_input=scope_input)
+
+        self.assertEqual(separable.call_args.args[3], 31)
```
Svilen-Stefanov, line 143: This patches `scope_is_separable` to False and then asserts on `call_args.args[3]`, so it pins the call shape but never the decision. Two things bother me: a positional index keeps passing silently if the argument order ever changes meaning, and with the gate patched out none of the 16 new tests exercise it. I made `scope_is_separable` return False unconditionally and all 16 still passed, only `test_planner_agent.py` caught it. Can we assert on a keyword and add one test where the real gate decides?

tests/static_analyzer/test_clustering_scope.py
```diff
@@ -110,6 +114,7 @@ def test_previous_owners_become_stable_group_ids(self):
         )
 
         self.assertEqual([group.group_id for group in result.groups], ["2", "4", "7"])
+        self.assertEqual(result.unanchored_modularity, expected_unanchored_modularity)
```
Svilen-Stefanov, line 117: I think this assertion cannot fail. The fixture is three singleton clusters with no edges, so the meta graph is edgeless and `_modularity` returns 0.0 for both values, which means `modularity` and `unanchored_modularity` are both 0.0 here. I replaced `unanchored_modularity` with the anchored `modularity` in `cluster_scope` and all 56 related tests still passed. Could we use a fixture with real edges where the two genuinely differ?

static_analyzer/clustering/service.py
```diff
@@ -113,9 +150,87 @@ def cluster_scope(
...
+        return child_graphs
+
+    @staticmethod
+    def _scope_size(graphs: Mapping[str, CallGraph]) -> tuple[int, int]:
+        files: set[str] = set()
+        method_count = 0
+        for graph in graphs.values():
+            for node in graph.nodes.values():
+                files.add(node.file_path)
+                if node.type in CALLABLE_TYPES:
```
Svilen-Stefanov, line 230: nit, though it might matter for #498: this counts only `CALLABLE_TYPES`, while `leaf_load` on the planner side takes its method count from `component.file_methods`, and `_build_file_methods_from_nodes` keeps `CALLABLE_TYPES | CLASS_TYPES`. So the same scope gets a different load depending on which path measured it, and once #498 swaps the planner for the hierarchy the depth can shift on class heavy repos. Is the callable only count the intended one?

agents/planner_agent.py
```diff
@@ -73,8 +59,8 @@ def component_is_separable(
         logger.debug("[Planner] subgraph has no natural cluster structure; keeping as leaf")
         return False
     _groups, modularity = GroupingService().group(cluster_results, cfg_graphs, subcomponents=True)
-    required = EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
-    separable = modularity >= required
+    required = ClusteringConfig.EXPAND_MODULARITY_THRESHOLD * max(0.0, 1.0 - load)
```
Svilen-Stefanov, line 62: this looks like it only feeds the log line now. The real comparison moved into `scope_is_separable`, which recomputes the same formula itself, so if we retune the ramp in `expansion.py` this will print a `required` that was not the one actually used. Can we drop it, or have `scope_is_separable` return the threshold it used so the log stays honest?

Svilen-Stefanov (changes requested):
Nice to have the hierarchy computed up front, the shape reads much better than the agent side recursion.

One thing I would want sorted before this goes in: the `< 2 groups` gate reads the anchored grouping, so a seeded run and a full run give different trees for the same graph, and it skips the `load >= 1.0` rule. Details inline.

The rest is mostly the new tests. A few of them pass for the wrong reason, I checked by mutating the code they cover and the suite stayed green.

static_analyzer/clustering/service.py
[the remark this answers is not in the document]
ivanmilevtues: Fixed in #500: `_cluster_scope` records previous ownership as a per-language member map before `reindex_across_languages`, then reconstructs each reindexed cluster owner from that language’s members. A colliding local ID can therefore no longer copy ownership across languages.

static_analyzer/clustering/service.py
ivanmilevtues: Resolved by the final hierarchy in #500: `_scope_size` consistently counts only `CALLABLE_TYPES`; class nodes remain structural containers and are intentionally excluded from method load. The old class-inclusive planner `leaf_load` path was removed, so the two competing counts no longer exist.

agents/planner_agent.py
ivanmilevtues: Fixed in #500: the stale planner-side required-modularity calculation and log path were removed. Hierarchy expansion now passes its callable-based scope load directly to `scope_is_separable`, leaving one threshold calculation.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed later in the stack in `c0de73a8`. Every persisted incremental scope now enters with `retain_scope=True`, so neither the anchored one-group gate nor the separability gate can drop its existing subtree. Fresh/new scopes have no previous ownership, so their `child.groups` is the fresh unanchored grouping. The reported seeded 121-method scope therefore no longer exits before the oversized-scope behavior; the retained-scope regression is passing.

tests/static_analyzer/test_clustering_hierarchy.py
ivanmilevtues: Addressed by the retained-scope fix in `c0de73a8` and the test cleanup in `9f948048`. The old test that treated the single carried owner as a reason to drop the subtree was removed. Current coverage instead verifies the production contract: persisted scopes remain expandable, while fresh scope decisions are covered separately through the real separability gate.

tests/static_analyzer/test_clustering_hierarchy.py
ivanmilevtues: Addressed in `9f948048`. The fixture now contains `a1 -> a2` inside group A as well as the cross-group `a1 -> b1` edge. The test asserts that the internal edge survives in A and that the cross-group edge is absent from both induced child graphs.

tests/static_analyzer/test_clustering_hierarchy.py
ivanmilevtues: Addressed in `9f948048`. The mocked assertion now reads the named `method_count` keyword rather than a positional slot, and `test_real_separability_gate_expands_a_modular_fresh_scope` exercises the real gate and verifies that its decision controls expansion.

tests/static_analyzer/test_clustering_scope.py
ivanmilevtues: Addressed in `9f948048`. The test now uses a connected four-cluster fixture whose anchored and unanchored groupings have genuinely different scores, explicitly asserts that the two modularity values differ, and verifies the returned unanchored score against the fresh grouping result.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed on the final #500 stack. ivanmilevtues aligned expansion with the existing main behavior: scopes at or above the leaf ceiling now expand unconditionally, while smaller scopes use both the group count and modularity from the same fresh, unanchored partition. Ownership anchoring no longer controls structural separability; retain_scope remains limited to preserving an already persisted subtree. A regression covers the exact 121-method, single-carried-owner case.

tests/static_analyzer/test_clustering_hierarchy.py
ivanmilevtues: Updated on the final #500 stack. The 121-method seeded case now asserts that the oversized scope expands despite the single carried owner, matching the unseeded behavior and main. The test also confirms that the below-ceiling separability gate is not consulted once load is at least 1.0.

ivanmilevtues:

PR #498, opened 2026-08-18
refactor(analysis): consume precomputed clustering hierarchy

ivanmilevtues:
## Goal

Make the deterministic clustering hierarchy the structural source of truth for full and partial analysis, while leaving the LLM agents responsible for names, descriptions, API surfaces, and relations.

## What this PR addresses

- Splits preparation into two explicit stages:
  - `deterministic_analysis(...)` runs source fingerprinting, static analysis, and clustering;
  - `agent_init()` initializes the LLM-backed agents;
  - `prepare_analysis(...)` composes both stages.
- Builds and indexes the complete `ClusterScopeResult` hierarchy before agent execution.
- Makes `ClusterScopeResult` authoritative for groups, exact per-language graphs, child scopes, and hierarchy indexes.
- Leaves each analysis agent with one public execution method:
  - `AbstractionAgent.run(scope)` for the root;
  - `DetailsAgent.run(scope, component)` for a component.
- Passes precomputed scopes directly to agent analysis; the temporary `ClusterAnalysis` model and `run_scope(...)` API are gone.
- Keeps partial analysis on the same preparation lifecycle: static analysis and clustering still run, and the persisted target component is registered as an exact scope before `DetailsAgent.run(...)`.
- Keeps hierarchy IDs aligned when single-child components are absorbed during final tree normalization.

## Behavior changes

- Full/details analysis no longer regroups when an agent reaches a scope. Group membership and expansion are fixed by deterministic analysis before any LLM call.
- No unrelated output behavior is intentionally changed. Existing naming, detail, API-surface, relation, and serialization stages remain in place.
- Details-agent result caching is removed separately in predecessor #510; static-analysis and metadata caching remain.

## Boundary of this PR

This PR owns full and partial hierarchy consumption. Incremental hierarchy migration and fail-fast baseline handling are isolated in #499. #500 then removes the superseded orchestration paths.

## Stack roadmap

|---|---|---|
| Previous | [#510](https://github.com/CodeBoarding/CodeBoarding/pull/510) | Remove the low-value details-agent cache before changing hierarchy ownership. |
| **This PR** | **#498** | Make full and partial analysis consume the precomputed hierarchy. |
| Next | [#499](https://github.com/CodeBoarding/CodeBoarding/pull/499) | Migrate incremental analysis to an ownership-anchored hierarchy. |
| Final cleanup | [#500](https://github.com/CodeBoarding/CodeBoarding/pull/500) | Remove obsolete fallback and recursive orchestration paths. |

---

### DIFF
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -37,7 +37,7 @@ from static_analyzer.cfg import DEFAULT_REFERENCE_KINDS
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.cluster_helpers import build_all_cluster_results
 from static_analyzer.config import Language
-from static_analyzer.clustering import ClusterResult
+from static_analyzer.clustering import ClusterResult, ClusterScopeResult
 
 logger = logging.getLogger(__name__)
 
@@ -192,12 +192,23 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         assign_relation_ids(analysis)
         self.build_static_relations(analysis)
 
-    def run(self):
-        # Build full cluster results dict for all languages ONCE
-        cluster_results = build_all_cluster_results(self.static_analysis)
+    def run_scope(self, scope: ClusterScopeResult) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
+        """Name and analyze a precomputed root clustering scope."""
+        return self._run_clustered(
+            self.cluster_analysis_from_scope(scope),
+            scope.leaf_clusters_by_language,
+        )
 
-        # Step 1: Deterministically partition leaf clusters into the top-level groups (Leiden, modularity-peak N)
-        cluster_analysis = self.step_clusters_grouping(cluster_results)
+    def run(self) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
+        """Cluster and analyze the root scope directly."""
+        cluster_results = build_all_cluster_results(self.static_analysis)
+        return self._run_clustered(self.step_clusters_grouping(cluster_results), cluster_results)
+
+    def _run_clustered(
+        self,
+        cluster_analysis: ClusterAnalysis,
+        cluster_results: dict[str, ClusterResult],
+    ) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
         if not cluster_analysis.cluster_components:
             # No leaf clusters means the repo has no callable structure to abstract
             # (unsupported/empty/no-code). Fail loudly instead of emptying the
@@ -207,26 +218,20 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
                 "no callable structure to build an architecture from."
             )
 
-        # Step 2: Name and describe each fixed group into a component (LLM, one component per group)
+        # Name and describe each fixed group into a component (LLM, one component per group)
         analysis = self.step_final_analysis(cluster_analysis, cluster_results)
-        # Step 3: Assign hierarchical component IDs ("1", "2", "3", ...)
+        # Assign hierarchical component IDs ("1", "2", "3", ...)
         assign_component_ids(analysis)
-        # Step 4: Resolve cluster IDs deterministically from group names
+        # Resolve cluster IDs deterministically from group names
         self._resolve_cluster_ids_from_groups(analysis, cluster_analysis)
-        # Step 5: Populate file_methods deterministically from cluster results + orphan assignment
+        # Populate file_methods deterministically from cluster results + orphan assignment
         self.populate_file_methods(analysis, cluster_results)
 
-        # Step 6: Analyze component API surfaces
         api_surfaces = self.step_api_surfaces(analysis)
-
-        # Step 7: Discover relations from API surfaces and attach deterministic all_edges
         self.step_relation_analysis(analysis, api_surfaces, cluster_analysis, cluster_results)
 
-        # Step 8: Fix source code reference lines (resolves reference_file paths for key_entities and key_edges)
         analysis = self.reference_resolver.fix_source_code_reference_lines(analysis)
-        # Step 9: Index relation endpoints after reference resolution
         index_relation_endpoints(analysis, self.repo_dir)
-        # Step 10: Ensure unique key entities across components
         self._ensure_unique_key_entities(analysis)
 
         return analysis, cluster_results
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -33,10 +33,7 @@ from agents.validation import (
 )
 from monitoring import trace
 from static_analyzer import StaticAnalysisFatalError
-from static_analyzer.cfg import DEFAULT_REFERENCE_KINDS
 from static_analyzer.analysis_result import StaticAnalysisResults
-from static_analyzer.cluster_helpers import build_all_cluster_results
-from static_analyzer.config import Language
 from static_analyzer.clustering import ClusterResult, ClusterScopeResult
 
 logger = logging.getLogger(__name__)
@@ -79,22 +76,6 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
                 ],
             ),
         }
-
-    @trace
-    def step_clusters_grouping(self, cluster_results: dict[str, ClusterResult]) -> ClusterAnalysis:
-        """Deterministically partition leaf clusters into the top-level component groups.
-
-        Resolution-tuned Leiden picks both the count (modularity peak over
-        ``[5, 8]``) and the membership, so the top-level structure is stable across
-        re-runs — the LLM no longer decides it; it only names them in the
-        final-analysis step.
-        """
-        logger.info(f"[AbstractionAgent] Super-clustering leaf clusters for: {self.project_name}")
-        cfg_graphs = {
-            lang: self.static_analysis.get_cfg(Language(lang)).to_networkx(DEFAULT_REFERENCE_KINDS)
-            for lang in cluster_results
-        }
-        return self.deterministic_cluster_grouping(cluster_results, cfg_graphs)
 
     @trace
     def step_final_analysis(
@@ -192,23 +173,10 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         assign_relation_ids(analysis)
         self.build_static_relations(analysis)
 
-    def run_scope(self, scope: ClusterScopeResult) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
+    def run(self, scope: ClusterScopeResult) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
         """Name and analyze a precomputed root clustering scope."""
-        return self._run_clustered(
-            self.cluster_analysis_from_scope(scope),
-            scope.leaf_clusters_by_language,
-        )
-
-    def run(self) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
-        """Cluster and analyze the root scope directly."""
-        cluster_results = build_all_cluster_results(self.static_analysis)
-        return self._run_clustered(self.step_clusters_grouping(cluster_results), cluster_results)
-
-    def _run_clustered(
-        self,
-        cluster_analysis: ClusterAnalysis,
-        cluster_results: dict[str, ClusterResult],
-    ) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
+        cluster_analysis = self.cluster_analysis_from_scope(scope)
+        cluster_results = scope.leaf_clusters_by_language
         if not cluster_analysis.cluster_components:
             # No leaf clusters means the repo has no callable structure to abstract
             # (unsupported/empty/no-code). Fail loudly instead of emptying the
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -10,9 +10,7 @@ from agents.agent_responses import (
     ComponentApiSurfaces,
     ComponentArchitecture,
     ComponentRelations,
-    ClusterAnalysis,
     MetaAnalysisInsights,
-    assign_component_ids,
     assign_relation_ids,
 )
 from agents.cluster_methods_mixin import ClusterMethodsMixin
@@ -78,17 +76,13 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         }
 
     @trace
-    def step_final_analysis(
-        self, llm_cluster_analysis: ClusterAnalysis, cluster_results: dict[str, ClusterResult]
-    ) -> AnalysisInsights:
+    def step_llm_analysis(self, scope: ClusterScopeResult) -> AnalysisInsights:
         logger.info(f"[AbstractionAgent] Generating final component analysis for: {self.project_name}")
-
-        cluster_str = llm_cluster_analysis.llm_str() if llm_cluster_analysis else "No cluster analysis available."
-
-        group_names = [cc.name for cc in llm_cluster_analysis.cluster_components] if llm_cluster_analysis else []
+        cluster_results = scope.leaf_clusters_by_language
+        group_names = scope.group_names()
 
         prompt = self.prompts["final_analysis"].format(
-            cluster_analysis=cluster_str,
+            cluster_analysis=scope.llm_str(),
         )
 
         if group_names:
@@ -100,7 +94,7 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         context = ValidationContext(
             cluster_results=cluster_results,
             static_analysis=self.static_analysis,
-            llm_cluster_analysis=llm_cluster_analysis,
+            clustering=scope,
         )
 
         architecture = self._invoke_repair_validate(
@@ -114,12 +108,12 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             repair_context=ComponentRepairContext(
                 reference_resolver=self.reference_resolver,
                 cluster_results=cluster_results,
-                llm_cluster_analysis=llm_cluster_analysis,
+                clustering=scope,
             ),
             validation_context=context,
             max_validation_attempts=3,
         )
-        self.assemble_one_component_per_group(architecture, llm_cluster_analysis, cluster_results)
+        self.assemble_one_component_per_group(architecture, scope)
         return AnalysisInsights(
             description=architecture.description,
             components=architecture.components,
@@ -141,13 +135,13 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
         self,
         analysis: AnalysisInsights,
         api_surfaces: ComponentApiSurfaces,
-        cluster_analysis: ClusterAnalysis,
-        cluster_results: dict[str, ClusterResult],
+        scope: ClusterScopeResult,
     ) -> None:
         logger.info(f"[AbstractionAgent] Discovering component relations for: {self.project_name}")
+        cluster_results = scope.leaf_clusters_by_language
         static_call_evidence = self.build_scope_cfg_string(analysis)
-        cfg_graphs = self.static_analysis.available_cfgs()
-        self.toolkit.context.cluster_analysis = cluster_analysis
+        cfg_graphs = scope.graphs_by_language
+        self.toolkit.context.clustering = scope
         self.toolkit.context.cluster_results = cluster_results
         self.toolkit.context.cfg_graphs = cfg_graphs
         prompt = self.prompts["relation_analysis"].format(
@@ -164,7 +158,7 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
                 cfg_graphs=cfg_graphs,
                 repo_dir=str(self.repo_dir),
                 static_analysis=self.static_analysis,
-                llm_cluster_analysis=cluster_analysis,
+                clustering=scope,
                 components=analysis.components,
             ),
             max_validation_attempts=3,
@@ -175,9 +169,8 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
 
     def run(self, scope: ClusterScopeResult) -> tuple[AnalysisInsights, dict[str, ClusterResult]]:
         """Name and analyze a precomputed root clustering scope."""
-        cluster_analysis = self.cluster_analysis_from_scope(scope)
         cluster_results = scope.leaf_clusters_by_language
-        if not cluster_analysis.cluster_components:
+        if not scope.groups:
             # No leaf clusters means the repo has no callable structure to abstract
             # (unsupported/empty/no-code). Fail loudly instead of emptying the
             # architecture and crashing downstream in populate_file_methods.
@@ -187,16 +180,14 @@ class AbstractionAgent(ClusterMethodsMixin, CodeBoardingAgent):
             )
 
         # Name and describe each fixed group into a component (LLM, one component per group)
-        analysis = self.step_final_analysis(cluster_analysis, cluster_results)
-        # Assign hierarchical component IDs ("1", "2", "3", ...)
-        assign_component_ids(analysis)
+        analysis = self.step_llm_analysis(scope)
         # Resolve cluster IDs deterministically from group names
-        self._resolve_cluster_ids_from_groups(analysis, cluster_analysis)
+        self._resolve_cluster_ids_from_groups(analysis, scope)
         # Populate file_methods deterministically from cluster results + orphan assignment
         self.populate_file_methods(analysis, cluster_results)
 
         api_surfaces = self.step_api_surfaces(analysis)
-        self.step_relation_analysis(analysis, api_surfaces, cluster_analysis, cluster_results)
+        self.step_relation_analysis(analysis, api_surfaces, scope)
 
         analysis = self.reference_resolver.fix_source_code_reference_lines(analysis)
         index_relation_endpoints(analysis, self.repo_dir)
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -435,6 +435,9 @@ class ClusterAnalysis(LLMBaseModel):
         description="Grouped clusters into logical components. Multiple cluster IDs can be grouped together if they work as a cohesive unit."
     )
 
+    def group_ids(self) -> dict[str, list[ClusterId]]:
+        return {group.name: group.cluster_ids for group in self.cluster_components}
+
     def llm_str(self):
         if not self.cluster_components:
             return "No clusters analyzed."
--- a/agents/cluster_methods_mixin.py
+++ b/agents/cluster_methods_mixin.py
@@ -40,6 +40,7 @@ from static_analyzer.cfg import CallGraph, DEFAULT_REFERENCE_KINDS
 from static_analyzer.clustering import (
     METHOD_LEVEL_STRATEGY,
     ClusterResult,
+    ClusterScopeResult,
     ClusteringService,
     MethodClusterPaths,
 )
@@ -116,6 +117,25 @@ class ClusterMethodsMixin:
     # These attributes must be provided by the class using this mixin
     repo_dir: Path
     static_analysis: StaticAnalysisResults
+
+    @staticmethod
+    def cluster_analysis_from_scope(scope: ClusterScopeResult) -> ClusterAnalysis:
+        """Adapt a precomputed structural scope to the agents' naming prompt."""
+        combined = combine_cluster_results(scope.leaf_clusters_by_language)
+        return ClusterAnalysis(
+            cluster_components=[
+                ClustersComponent(
+                    name=f"Group {index}",
+                    cluster_ids=group.cluster_ids,
+                    description=_summarize_group(
+                        set(group.cluster_ids),
+                        combined.clusters,
+                        combined.cluster_to_files,
+                    ),
+                )
+                for index, group in enumerate(scope.groups, start=1)
+            ]
+        )
 
     def deterministic_cluster_grouping(
         self,
--- a/agents/cluster_methods_mixin.py
+++ b/agents/cluster_methods_mixin.py
@@ -44,7 +44,6 @@ from static_analyzer.clustering import (
     ClusteringService,
     MethodClusterPaths,
 )
-from static_analyzer.clustering.grouping import GroupingService
 from static_analyzer.node import Node
 
 logger = logging.getLogger(__name__)
@@ -136,45 +135,6 @@ class ClusterMethodsMixin:
                 for index, group in enumerate(scope.groups, start=1)
             ]
         )
-
-    def deterministic_cluster_grouping(
-        self,
-        cluster_results: dict[str, ClusterResult],
-        cfg_graphs: dict[str, nx.DiGraph],
-        *,
-        subcomponents: bool = False,
-    ) -> ClusterAnalysis:
-        """Partition leaf clusters into fixed component groups via resolution-tuned Leiden.
-
-        The count (modularity peak over the configured range) and membership are chosen
-        deterministically, so the structure is stable across re-runs — the LLM no
-        longer decides it. Each group gets a stable ``Group i`` label and a summary
-        of its members; the final-analysis step only names and describes them.
-
-        ``cfg_graphs`` must span exactly the same scope as ``cluster_results`` — the
-        component's own subgraph when splitting a component, the whole repo at the
-        top level. Handing it the repo graph for a component scope makes the split
-        disagree with the separability gate, which reads the subgraph.
-        """
-        groups, _modularity = GroupingService().group(
-            cluster_results,
-            cfg_graphs,
-            subcomponents=subcomponents,
-        )
-        combined = combine_cluster_results(cluster_results)
-        cluster_components = [
-            ClustersComponent(
-                name=f"Group {i}",
-                cluster_ids=sorted(group),
-                description=_summarize_group(group, combined.clusters, combined.cluster_to_files),
-            )
-            for i, group in enumerate(groups, start=1)
-        ]
-        logger.info(
-            f"[{type(self).__name__}] Partitioned {sum(len(g) for g in groups)} leaf clusters "
-            f"into {len(cluster_components)} deterministic groups"
-        )
-        return ClusterAnalysis(cluster_components=cluster_components)
 
     @staticmethod
     def assemble_one_component_per_group(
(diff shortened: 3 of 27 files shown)


ivanmilevtues:
Implemented in `2a2a8fc3` and pushed.

### Architecture cleanup
- `DiagramGenerator.prepare_analysis()` now runs `deterministic_analysis()` followed by `agent_init()`. Static analysis and the clustering hierarchy are complete before LLM-backed analysis starts.
- `AbstractionAgent` has one execution entry point: `run(scope)`.
- `DetailsAgent` has one execution entry point: `run(scope, component)`.
- Agent-side grouping and the `run_scope`/direct-reclustering split were removed; agents now consume precomputed scopes only.

### Behavior assessment
- **Full analysis output behavior:** no change intended. It uses the same hierarchy builder, configured hierarchy depth, expansion flags, and traversal; hierarchy construction has only moved earlier in the lifecycle.
- **Initialization/runtime behavior:** intentionally changed. Static analysis and clustering now complete before LLM initialization/metadata analysis instead of static analysis and metadata analysis running in parallel.
- **Partial analysis:** intentionally changed to the same deterministic boundary. Preparation computes one additional hierarchy level (`depth_cap + 1`) so a component at the current cap can still be expanded, then passes its precomputed scope to `DetailsAgent.run(scope, component)`. A component without a retained child scope no longer silently reclusters inside the agent. The persisted depth cap/output traversal rules are unchanged.
- **Incremental analysis:** no clustering behavior change intended in this PR. It explicitly prepares with `include_clustering=False`, preserving the existing baseline-sensitive cluster-delta path until the incremental hierarchy PR.

### Verification
- Focused agent/generator/workflow tests: 94 passed
- Full suite: 2,160 passed, 28 skipped
- Coverage: 83.26% (80% required)
- Black and mypy: passed

agents/cluster_methods_mixin.py
```diff
@@ -117,44 +117,24 @@ class ClusterMethodsMixin:
...
-            for i, group in enumerate(groups, start=1)
-        ]
-        logger.info(
-            f"[{type(self).__name__}] Partitioned {sum(len(g) for g in groups)} leaf clusters "
-            f"into {len(cluster_components)} deterministic groups"
+    @staticmethod
+    def cluster_analysis_from_scope(scope: ClusterScopeResult) -> ClusterAnalysis:
+        """Adapt a precomputed structural scope to the agents' naming prompt."""
+        combined = combine_cluster_results(scope.leaf_clusters_by_language)
+        return ClusterAnalysis(
```
ivanmilevtues: I am not sure why do we have to restructure the same information into a differnt type I would rather directly use the ClusterScopeResult than keeping the ClusterAnalysis alive. Drop that type and make use of the Clusteranalysis?

can we take a second look and discuss how to simplify this design, i don't like this and ultimately i want this mixin to be gone...

agents/abstraction_agent.py
```diff
@@ -207,26 +186,20 @@ def run(self):
                 "no callable structure to build an architecture from."
             )
 
-        # Step 2: Name and describe each fixed group into a component (LLM, one component per group)
+        # Name and describe each fixed group into a component (LLM, one component per group)
         analysis = self.step_final_analysis(cluster_analysis, cluster_results)
```
ivanmilevtues: rename this to step_llm_analysis.

agents/details_agent.py
```diff
@@ -229,55 +210,33 @@ def step_relation_analysis(
...
+        """Name and analyze one precomputed component scope."""
+        logger.info(f"[DetailsAgent] Processing precomputed component: {component.name}")
+        subgraph_cfgs: dict[str, CallGraph] = {}
+        for language, partition in scope.leaf_clusters_by_language.items():
+            assigned_qnames = {
+                qualified_name
+                for group in scope.groups
+                for qualified_name in group.symbol_members_by_language.get(language, set())
+            } or {qualified_name for cluster in partition.clusters.values() for qualified_name in cluster}
+            subgraph = self.static_analysis.get_cfg(Language(language)).filter_by_nodes(assigned_qnames)
```
ivanmilevtues: it feels like the scope should also have its graph somehow related to it/in it instead of us having to compute it this way no?

diagram_analysis/diagram_generator.py
```diff
@@ -727,16 +780,11 @@ def _process_component(
...
-            parent_had_clusters = bool(component.source_cluster_ids)
-
-            # Get new components to analyze (deterministic, no LLM). The separability
-            # gate keeps cohesive sub-components as leaves rather than splitting them.
-            new_components = get_expandable_components(
-                analysis, parent_had_clusters=parent_had_clusters, separable=self._component_separable
-            )
+            scope = self._preclustered_scopes.get(component.component_id)
+            if scope is None:
+                raise ValueError(f"No precomputed clustering scope for component {component.component_id}")
```
ivanmilevtues, line 852: This is important i have to double check, it feels like we are redoing the indexing 15 times, it should be done just ones when we have the components which is in a post analysis step and then we shouldn't fiddle around with the ids, i feel like. Let's have a quick discussion on how this is done.

codeboarding_workflows/analysis.py
```diff
@@ -99,7 +99,7 @@ def run_partial(
 
     depth_level = int(metadata.get("depth_cap", metadata.get("depth_level", DEFAULT_DEPTH_LEVEL)))
     generator = build_generator(run_paths, run_context, depth_level=depth_level)
-    generator.pre_analysis()
+    generator.prepare_analysis(hierarchy_depth=depth_level + 1)
```
ivanmilevtues: nice i like to see it.

diagram_analysis/diagram_generator.py
```diff
@@ -624,13 +626,49 @@ def __init__(
...
 
     @track_analysis
     def process_component(
         self, component: Component
     ) -> tuple[str, AnalysisInsights, list[Component]] | tuple[None, None, list]:
         return self._process_component(component)
 
+    def _index_clustering_hierarchy(
+        self,
+        hierarchy: ClusterScopeResult,
```
ivanmilevtues: Don't we have such index in the ClusterScopeResult, even if we don't have it I think this should be a part of that class and we should get it ClusterScepeResult.clustering_groups or preclustered_scopes. Either way shouldn't be part of the diagram generator, and I want to know why we need these and discuss if they are needed really or we can reuse something we already have.

diagram_analysis/diagram_generator.py
```diff
@@ -1395,7 +1468,7 @@ def generate_analysis_incremental(
         Raises when no trustworthy baseline or scoped update plan is available.
         """
         if self.details_agent is None or self.incremental_agent is None:
-            self.pre_analysis()
+            self.prepare_analysis(include_clustering=False)
```
ivanmilevtues: this seems like a good catch, probably you can drop the clustering parameter as we will need to do it always no?

static_analyzer/clustering/orchestration.py
```diff
@@ -1,16 +1,36 @@
...
+from static_analyzer.config import Language
 
 logger = logging.getLogger(__name__)
 
 
+def build_clustering_hierarchy(static_analysis: StaticAnalysisResults, max_depth: int) -> ClusterScopeResult:
+    """Build and cache the complete deterministic hierarchy for a full analysis."""
+    root_partitions = build_all_cluster_results(static_analysis)
+
+    def scope_input(scope_id: str, _graphs: Mapping[str, CallGraph]) -> ClusterScopeInput:
```
ivanmilevtues, line 20: I am not a fan of a single line anonymous funcs in general. Also shouldn't this method be behind the interface of the ClsuteringService?

Let's check if it is in fact the case, becuase I think the whole orchestration.py is a compatibility layer, check with the future commits if this is removed later.

static_analyzer/clustering/service.py
```diff
@@ -172,6 +172,19 @@ def cluster_hierarchy(
         self._cluster_children(root, graphs, 1, max_depth, scope_input)
         return root
 
+    @staticmethod
+    def induced_graphs(group: ClusterGroup, graphs: Mapping[str, CallGraph]) -> dict[str, CallGraph]:
```
ivanmilevtues: It feels like this shouldn't be a public method, as it has only internal usage

ivanmilevtues:
Addressed the review batch in `f7eb7ae2`.

### Result
- `ClusterScopeResult` is now the source of truth for full/details analysis: it owns its scoped graphs, prompt-facing group metadata, and hierarchy indexes.
- `AbstractionAgent.run(scope)` and `DetailsAgent.run(scope, component)` consume it directly; the redundant `ClusterScopeResult -> ClusterAnalysis` conversion and duplicate component-ID assignment are gone.
- `DiagramGenerator` now exposes the requested preparation split: `deterministic_analysis()` (fingerprint + static analysis + clustering), `agent_init()`, and `prepare_analysis()` to call both. Deterministic clustering is always included.
- Partial and incremental detailing register exact scopes rooted at persisted/new component IDs before invoking the details agent.
- `ClusteringService._induced_graphs()` is private. The separate orchestration module remains because stacked PRs #499 and #500 still use it to seed/synchronize root cluster-cache state; it is not a compatibility layer removed by those PRs.
- `ClusterAnalysis` remains only on the existing incremental path for compatibility with the next stacked PR; full/details no longer use or construct it.

### Explicit behavior changes
The refactor is intended to preserve full-analysis output, but it deliberately includes these correctness fixes:
1. polyglot nodes with the same qualified name are now kept distinct by language;
2. partial regeneration uses the persisted component's exact file/method membership, so IDs changed by single-child absorption resolve correctly;
3. new incremental components receive an exact preclustered scope before recursive detailing;
4. incremental preparation now always performs deterministic clustering; and
5. an invalid partial component request returns before unnecessary static/agent initialization.

Validation: `uv run black --check .`, `uv run mypy .`, and `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` all pass (2,164 passed, 28 skipped, 83.24% coverage).

diagram_analysis/diagram_generator.py
```diff
@@ -624,13 +625,50 @@ def __init__(
...
+        self._clustering_groups.clear()
+        self._preclustered_scopes.clear()
+
+        def visit(scope: ClusterScopeResult, graphs: dict[str, CallGraph]) -> None:
+            for group in scope.groups:
+                self._clustering_groups[group.group_id] = group
+                if group.children is None:
+                    continue
+                child_graphs = ClusteringService.induced_graphs(group, graphs)
+                self._preclustered_scopes[group.group_id] = (group.children, child_graphs)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `f7eb7ae2`. CFG node ownership is now keyed by `(language, qualified_name)` throughout method assignment, including orphan routing, so same-named Python/Go symbols remain distinct. Added a polyglot reproducer.

diagram_analysis/diagram_generator.py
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `f7eb7ae2`. Partial preparation now builds an exact `ClusterScopeResult` from the persisted component’s rendered file/method membership and roots it at the persisted component ID, so absorbed IDs no longer depend on a fresh uncollapsed hierarchy.

diagram_analysis/diagram_generator.py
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `f7eb7ae2`. Preparation always builds clustering now, and every newly created incremental component below the depth cap gets an exact registered scope before `_generate_subcomponents()` runs. Added ordering coverage for this path.

diagram_analysis/diagram_generator.py
```diff
@@ -1018,15 +1061,39 @@ def pre_analysis(self):
                 json.dump(static_stats, f, indent=2)
             logger.debug(f"Written code_stats.json to {code_stats_file}")
 
-            # Initialize streaming writer (handles timing and run_metadata.json)
+    def agent_init(self) -> None:
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `f7eb7ae2`. The new public lifecycle methods (`deterministic_analysis`, `agent_init`, and `prepare_analysis`) now sit together in the public section before the private helpers.

agents/cluster_methods_mixin.py
ivanmilevtues: Simplified in `f7eb7ae2`: full and details analysis now consume `ClusterScopeResult` directly, including prompt rendering, repair/validation context, and group reconciliation. The conversion to `ClusterAnalysis` is gone. `ClusterAnalysis` remains only for the legacy incremental path until stacked PR #499 migrates that path.

agents/abstraction_agent.py
ivanmilevtues: Renamed to `step_llm_analysis` in `f7eb7ae2` for both full and details analysis.

agents/details_agent.py
ivanmilevtues: Addressed in `f7eb7ae2`. `ClusterScopeResult` now owns `graphs_by_language`, populated by `ClusteringService`, and `DetailsAgent` consumes those exact graphs instead of reconstructing a subgraph from global static analysis.

diagram_analysis/diagram_generator.py
ivanmilevtues: Addressed in `f7eb7ae2`. Each deterministic `ClusterGroup.group_id` is now assigned directly to its reconciled component; the normal full/details `assign_component_ids()` passes were removed. The only remaining ID assignment is the existing incremental-only assignment for genuinely new components.

diagram_analysis/diagram_generator.py
ivanmilevtues: Moved in `f7eb7ae2`. `ClusterScopeResult` now owns `clustering_groups`, `preclustered_scopes`, hierarchy indexing/registration, and absorption rerooting. `DiagramGenerator` only consumes those indexes.

diagram_analysis/diagram_generator.py
ivanmilevtues: Done in `f7eb7ae2`. `include_clustering` was removed; deterministic preparation always includes static analysis and clustering, including incremental initialization.

static_analyzer/clustering/orchestration.py
ivanmilevtues: Checked the stacked PRs before changing this. `orchestration.py` remains in both #499 and #500 and is expanded in #499 to combine cached root partitions with per-scope ownership, so it is not a compatibility layer that disappears. It remains separate from the pure `ClusteringService`; the one-line conditional callback was expanded for readability.

static_analyzer/clustering/service.py
ivanmilevtues: Made private as `ClusteringService._induced_graphs()` in `f7eb7ae2`; it is only used by recursive child clustering.

codeboarding_workflows/analysis.py
ivanmilevtues: Thanks — kept this flow. Partial analysis still performs preparation for a valid target, and `f7eb7ae2` now passes that persisted component into preparation so its exact clustered scope is registered before detailing.

agents/tools/base.py
```diff
@@ -20,7 +20,7 @@ class RepoContext(BaseModel):
     ignore_manager: RepoIgnoreManager
     static_analysis: Optional[StaticAnalysisResults] = None
     changes: ChangeSet | None = None
-    cluster_analysis: ClusterAnalysis = Field(default_factory=lambda: ClusterAnalysis(cluster_components=[]))
+    clustering: ClusterScopeResult | ClusterAnalysis = Field(default_factory=lambda: ClusterScopeResult(scope_id=""))
```
ivanmilevtues, line 23: nooo no double typing wtf, drop the ClusterAnalysis type please. It should be gone.

agents/agent_responses.py
```diff
@@ -435,6 +435,9 @@ class ClusterAnalysis(LLMBaseModel):
         description="Grouped clusters into logical components. Multiple cluster IDs can be grouped together if they work as a cohesive unit."
     )
 
+    def group_ids(self) -> dict[str, list[ClusterId]]:
```
ivanmilevtues, line 438: no but we should completely drop this type and just use the clusterScopeResult instead of it.

diagram_analysis/diagram_generator.py
```diff
@@ -695,6 +786,22 @@ def _expandable_ids_for_tree(
...
+        clustering_groups = self.clustering_hierarchy.clustering_groups if self.clustering_hierarchy else {}
+        if clustering_groups:
+
+            def precomputed_ids(scope: AnalysisInsights) -> list[str]:
+                return [
+                    component.component_id
+                    for component in scope.components
+                    if component.component_id
+                    and (group := clustering_groups.get(component.component_id)) is not None
+                    and group.expandable
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `4a9a6da3`. Expansion now treats an existing generated `sub_analysis` as authoritative, so a newly registered incremental owner is serialized as expandable even when it was not present in the original clustering-group index. The generated subtree is retained.

agents/details_agent.py
```diff
@@ -229,71 +202,29 @@ def step_relation_analysis(
...
+        """Name and analyze one precomputed component scope."""
+        logger.info(f"[DetailsAgent] Processing precomputed component: {component.name}")
+        subgraph_cluster_results = scope.leaf_clusters_by_language
+        analysis = self.step_llm_analysis(component, scope)
 
-        # Step 6: Populate file_methods deterministically from cluster results + orphan assignment
-        # Pass subgraph_cfgs to scope node collection to the component's filtered graph
-        # With method-level expansion, each method has its own cluster -> deterministic assignment
-        self.populate_file_methods(analysis, subgraph_cluster_results, subgraph_cfgs)
+        self._resolve_cluster_ids_from_groups(analysis, scope)
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed structurally in predecessor #510 (`a08e5a00`). The details-agent result cache and run-ID reuse were removed entirely, so this stale cross-scope cache-hit path no longer exists. Static-analysis and metadata caches remain unchanged.

diagram_analysis/diagram_generator.py
```diff
@@ -624,13 +626,102 @@ def __init__(
...
+
+        if self._static_analyzer is not None:
+            logger.info("Using injected StaticAnalyzer (clients already running)")
+            static_analysis = self._get_static_with_injected_analyzer()
+        else:
+            static_analysis = self._get_static_with_new_analyzer()
+
+        self.static_analysis = static_analysis
+        depth = hierarchy_depth if hierarchy_depth is not None else self.depth_level
+        self.clustering_hierarchy = build_clustering_hierarchy(static_analysis, depth)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `4a9a6da3`. Incremental preparation validates and preserves `incremental_base_results` before current clustering can seed or mutate the live cache. A missing/unusable baseline now raises `IncrementalCacheMissingError` rather than comparing the current graph with itself.

agents/tools/base.py
ivanmilevtues: Fixed in stacked #499 (`c0ef96ac`). `ClusterAnalysis` and `ClustersComponent` are deleted, and `RepoContext.clustering` now accepts only `ClusterScopeResult`.

agents/agent_responses.py
ivanmilevtues: Fixed in stacked #499 (`c0ef96ac`). The duplicate `ClusterAnalysis` and `ClustersComponent` response types are gone. Incremental prompts, relation validation, repair, and tool context now use `ClusterScopeResult` directly.

static_analyzer/clustering/orchestration.py
```diff
@@ -1,16 +1,36 @@
...
+        if scope_id == "root":
+            return ClusterScopeInput(leaf_clusters_by_language=root_partitions)
+        return ClusterScopeInput()
+
+    hierarchy = ClusteringService().cluster_hierarchy(
+        static_analysis.available_cfgs(),
+        max_depth,
+        scope_input,
+    )
+    _record_child_scopes(static_analysis, hierarchy)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in stacked #499 (`c0de73a8`). Persisted child partitions are reconstructed from the immutable `incremental_base_results` method-path lineage, not the live paths populated by current hierarchy construction, so current scope recording cannot overwrite the baseline being compared.

diagram_analysis/diagram_generator.py
```diff
@@ -695,6 +796,24 @@ def _expandable_ids_for_tree(
         if self.details_agent is None:
             return None, None
 
+        clustering_groups = self.clustering_hierarchy.clustering_groups if self.clustering_hierarchy else {}
```
Svilen-Stefanov, line 799: blocker-ish: I think this loses can_expand for every unexpanded sibling on a partial run. `run_partial` calls `prepare_analysis(target_component=...)`, which builds a hierarchy holding only that component's subtree, so `clustering_groups` is non-empty but contains none of the root's own components. This branch then keeps a component only if it is in `sub_analyses` or in `clustering_groups`, and a sibling that is a leaf today but was flagged expandable is in neither.

I ran it both ways on the same fixture (root with Target + Sibling, target just expanded):
```
baseline  root expandable ids: ['1', '2']
this PR   root expandable ids: ['1']       clustering_groups keys: ['1.1', '1.2']
```
To be fair to the change, it does keep the dangerous half: a component that already holds a sub-analysis stays expandable via `in sub_analyses`, so we are not discarding analysed subtrees, and analysis_json.py:338 would have shouted if we were. But that warning only fires when there IS a subtree, so a leaf quietly losing its flag is invisible. The user-visible version is you expand one component in the IDE and the others lose their arrow, and since analysis.json is the store it does not come back without a full run.

The old docstring here spelled out this exact trap and got deleted with the code, which is probably how the leaf half of the guard went missing.

codeboarding_workflows/analysis.py
```diff
@@ -131,6 +128,9 @@ def run_partial(
         logger.error(f"Component with ID '{component_id}' not found in analysis")
         return
 
+    generator = build_generator(run_paths, run_context, depth_level=depth_level)
```
Svilen-Stefanov, line 131: This is the call that triggers the can_expand issue I flagged in `_expandable_ids_for_tree`: registering only the target's scope means the save-time recompute can't see the other components. Would it work to register the persisted tree's other root components too, or to keep the previous expandable set when the hierarchy only covers part of the tree?

nit while I'm here: `depth_level + 1` is correct (`_component_depth("1")` is 1, so `_build_component_scope` gets back to `depth_level`) but it reads like a magic offset. One short comment would save the next reader the trip.

diagram_analysis/diagram_generator.py
```diff
@@ -953,79 +1068,39 @@ def _initialize_agents(
...
-        # --- Capture Static Analysis Stats ---
-        static_stats: dict[str, Any] = {"repo_name": self.repo_name, "languages": {}}
-        scanner = ProjectScanner(self.repo_location)
-        loc_by_language = {pl.language: pl.size for pl in scanner.scan()}
-        for language in static_analysis.get_languages():
-            files = static_analysis.get_source_files(language)
-            static_stats["languages"][language] = {
-                "file_count": len(files),
-                "lines_of_code": loc_by_language.get(language, 0),
+    def _register_component_scope(self, component: Component, hierarchy_depth: int) -> None:
```
Svilen-Stefanov, line 1071: blocker-ish: I think this path poisons the next incremental run. `build_scope_hierarchy` is the only hierarchy builder that never calls `_record_scopes`, so the leaf partitions for a component expanded this way are never written into the `ClusterCache` method_paths. `run_partial` then saves with `persist_side_artifacts=False`, so the pkl isn't rewritten either, but `sub_analyses[component_id]` does go into analysis.json.

Next incremental run, that id is in `persisted_scopes`, `scope_input` finds a persisted scope, `_scoped_snapshot` comes back empty and `unclustered_members` is empty, so the lineage check raises `IncrementalCacheMissingError: persisted scope 'X' has no cluster lineage`. So: expand a component in the IDE, then run incremental, and it hard-fails telling you to run a full analysis. That's already our most common user-facing error and this makes a normal UI action reach it.

Should `build_scope_hierarchy` record its scopes too, or should `run_partial` persist the cache after an expansion?

diagram_analysis/diagram_generator.py
```diff
@@ -624,13 +626,112 @@ def __init__(
...
+            monitoring_dir = get_monitoring_run_dir(self.log_path, create=True)
+            self.stats_writer = StreamingStatsWriter(
+                monitoring_dir=monitoring_dir,
+                agents_dict=self._monitoring_agents,
+                repo_name=self.project_name or self.repo_name,
+                output_dir=str(self.output_dir),
+                start_time=self._analysis_start_time,
+            )
+
+    def prepare_analysis(
```
Svilen-Stefanov, line 720: blocker-ish: renaming `pre_analysis` to `prepare_analysis` breaks the wrapper. `codeboarding_pro/analysis_controller.py:175` calls `generator.pre_analysis()` then `generator.process_component(component)`. And it isn't only the rename: `_process_component` now needs a registered scope, so even after renaming it would hit `ValueError: No precomputed clustering scope for component X` for a persisted id. `run_partial` was updated to pass `target_component=` for exactly this, the wrapper wasn't. That's the IDE expand button, so probably worth a companion wrapper PR.

diagram_analysis/diagram_generator.py
```diff
@@ -624,13 +626,112 @@ def __init__(
...
         self._separable_cache: dict[frozenset[tuple[str, str]], bool] = {}
+        self._analysis_start_time = time.time()
 
     @track_analysis
     def process_component(
         self, component: Component
     ) -> tuple[str, AnalysisInsights, list[Component]] | tuple[None, None, list]:
         return self._process_component(component)
 
+    def deterministic_analysis(
```
Svilen-Stefanov, line 637: We lose the static-analysis / meta-agent overlap here. `pre_analysis` used to run them together on a `ThreadPoolExecutor(max_workers=2)`, so the meta LLM call hid behind the LSP pass. Now it's static analysis, then clustering, then the meta call, all sequential, and on a cold meta cache that latency goes straight onto the critical path. Was that deliberate? If the split is what we want structurally, could `agent_init` at least kick the meta call off before `deterministic_analysis` returns?

diagram_analysis/diagram_generator.py
```diff
@@ -727,16 +846,12 @@ def _process_component(
...
-
-            # Get new components to analyze (deterministic, no LLM). The separability
-            # gate keeps cohesive sub-components as leaves rather than splitting them.
-            new_components = get_expandable_components(
-                analysis, parent_had_clusters=parent_had_clusters, separable=self._component_separable
-            )
+            preclustered_scopes = self.clustering_hierarchy.preclustered_scopes if self.clustering_hierarchy else {}
+            scope = preclustered_scopes.get(component.component_id)
+            if scope is None:
+                raise ValueError(f"No precomputed clustering scope for component {component.component_id}")
```
Svilen-Stefanov, line 852: I don't think this raise can ever reach the caller. It's inside the `try` whose `except Exception` two screens down logs and returns `(None, None, [])`, and `_generate_subcomponents` then just does `if comp_name and sub_analysis:`. So a component with no scope silently loses its whole subtree instead of failing loudly, which is the opposite of what AGENTS.md 5 asks for. Should this be raised outside the try, or at least re-raised like `LLMAuthError` is?

nit: this and the `no owned CFG nodes` one below are both bare `ValueError` where we have a taxonomy in `diagram_analysis/exceptions.py`.

diagram_analysis/diagram_generator.py
```diff
@@ -1222,7 +1304,9 @@ def finalize_for_save(
             if self.static_analysis
             else []
         )
-        absorb_single_child_components(root_analysis, sub_analyses, cluster_caches)
+        absorbed_ids = absorb_single_child_components(root_analysis, sub_analyses, cluster_caches)
+        if hierarchy := getattr(self, "clustering_hierarchy", None):
```
Svilen-Stefanov, line 1308: nit: `getattr(self, "clustering_hierarchy", None)` on an attribute that `__init__` always sets. Is a test constructing a partial generator? If so I'd rather fix the test than keep the defensive read here.

diagram_analysis/diagram_generator.py
```diff
@@ -695,6 +796,24 @@ def _expandable_ids_for_tree(
...
+        if clustering_groups:
+
+            def precomputed_ids(scope: AnalysisInsights) -> list[str]:
+                return [
+                    component.component_id
+                    for component in scope.components
+                    if component.component_id
+                    and (
+                        component.component_id in sub_analyses
+                        or ((group := clustering_groups.get(component.component_id)) is not None and group.expandable)
```
Svilen-Stefanov, line 809: Another one from this branch: `.codeboardingignore` no longer reaches can_expand. `_strip_ignored` runs at the top of `finalize_for_save` and the old `get_expandable_components` read the stripped `source_cluster_ids` / `file_methods`, so a component whose files were all ignored correctly came out not expandable. `group.expandable` here comes from the hierarchy, which was built off the raw CFG before any ignoring, so we can now advertise an expand arrow on a component that has nothing left to show. Worth intersecting with the stripped membership?

static_analyzer/clustering/models.py
```diff
@@ -129,9 +131,117 @@ class ClusterScopeResult:
...
+            prefix = f"{child_id}."
+            rerooted_groups: dict[GroupId, ClusterGroup] = {}
+            for group_id, group in self.clustering_groups.items():
+                if group_id == child_id:
+                    continue
+                if group_id.startswith(prefix):
+                    tail = group_id.removeprefix(prefix)
+                    group_id = f"{parent_id}.{tail}" if parent_id else tail
+                    group.group_id = group_id
+                rerooted_groups[group_id] = group
```
Svilen-Stefanov, line 207: just checking, isn't this last-write-wins? If a rerooted grandchild id lands on an id a surviving sibling already holds, `rerooted_groups[group_id] = group` silently replaces it and we lose a group from the index. Same shape in the `preclustered_scopes` loop below. Is there something upstream that guarantees the rerooted ids can't collide, or should this raise if the key is already there?

tests/test_main.py
```diff
@@ -358,7 +364,7 @@ def test_local_source_composes_with_partial_update(self, mock_load_metadata, moc
                     component_id="does-not-matter",
                 )
 
-            mock_generator_class.return_value.pre_analysis.assert_called_once()
+            mock_generator_class.return_value.prepare_analysis.assert_not_called()
```
Svilen-Stefanov, line 367: I think this assertion got flipped and the test stopped testing anything. The diff is `pre_analysis.assert_called_once()` -> `prepare_analysis.assert_not_called()`, but the comment above was only reworded and still says the baseline is there "so run_partial gets past both early-return guards and reaches prepare_analysis (which is what this test asserts on)". It passes because `components=[]` means the component lookup fails and this PR moved `prepare_analysis` after that lookup, so we return before it. Given the docstring is "the local source composes with any scope", shouldn't this assert it IS called, with a component that actually exists in the baseline?

Svilen-Stefanov (changes requested):
The restructuring in the full stack is great, should definitely help decouple the code more and make things simpler. The main things here that I'd like to see addressed before we merge them are the blocker comments.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed on #500 in `f819b215`. Partial saves now preserve persisted `can_expand` IDs for components absent from the target-only hierarchy, while computed groups remain authoritative for the target subtree. Added a Target + Sibling regression test.

codeboarding_workflows/analysis.py
ivanmilevtues: Fixed on #500 in `f819b215`. `run_partial` loads the persisted expandable set and carries it through the target-only save, and the `depth_level + 1` call now has the requested short explanation.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed on #500 in `f819b215`. Selected-scope hierarchy construction now records structural and fallback lineage, and partial analysis persists the updated static-analysis artifact after saving `analysis.json`. Partial expansion also requires an unchanged source-tree hash, including SHA-tagged persistence for injected-analyzer reuse.

diagram_analysis/diagram_generator.py
ivanmilevtues: Tracked as the companion wrapper work in CodeBoarding-wrapper#93: https://github.com/CodeBoarding/CodeBoarding-wrapper/issues/93. The issue covers the removed `pre_analysis` API and migration to target-scope preparation / `run_partial`, plus the moved static-analysis imports and updated tests.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed on #500 in `f819b215`. Full and partial preparation now start metadata analysis before deterministic static analysis/clustering and join afterward, restoring the overlap. Incremental preparation remains deterministic-first so an empty cluster delta stays a zero-LLM path.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed on #500 in `f819b215`. Missing precomputed scopes are checked outside the best-effort LLM catch, both missing-scope paths now raise `ClusteringScopeUnavailableError`, and focused tests assert that the deterministic failures reach the caller.

diagram_analysis/diagram_generator.py
ivanmilevtues: Fixed on #500 in `f819b215`. The defensive `getattr` is removed. The tests that bypass `__init__` now initialize `clustering_hierarchy` explicitly instead of weakening the production contract.

static_analyzer/clustering/models.py
ivanmilevtues: Fixed on #500 in `f819b215`. Rerooting now detects duplicate group and scope IDs before mutating either index and raises rather than silently replacing a surviving entry. Both collision shapes have regression coverage.

tests/test_main.py
ivanmilevtues: Fixed on #500 in `f819b215`. The local-source partial test now supplies a real persisted target component and positively asserts `prepare_analysis(hierarchy_depth=2, target_component=component)` instead of passing through the missing-component early return.

diagram_analysis/diagram_generator.py
ivanmilevtues: Valid regression. `main` handled this at save time: `_strip_ignored` removed ignored `file_methods`, then `get_expandable_components` rejected components with no rendered files. The hierarchy refactor bypassed that final membership gate. Restored the same contract on #500 in `d9385543`: `ClusterGroup.expandable` remains authoritative for structure, but an unexpanded component is advertised only when its post-strip `file_methods` is non-empty. The existing expansion-flags test now covers a structurally expandable component with no rendered files.

ivanmilevtues:

PR #499, opened 2026-08-18
refactor(incremental): consume precomputed clustering hierarchy

ivanmilevtues:
## Goal

Move incremental analysis onto the same precomputed hierarchy as full analysis, while anchoring every persisted scope to its previous component and member ownership.

## What this PR addresses

- Moves full, incremental, and existing-scope hierarchy construction behind `ClusteringService`; deletes the proxy `clustering/orchestration.py` module and keeps low-level operations private.
- Builds one incremental hierarchy up front from live call graphs:
  - the root uses the seeded cluster delta;
  - child scopes use persisted scope-local lineage from `incremental_base_results`;
  - every persisted scope supplies previous ownership anchors and reserves existing component IDs.
- Traverses each precomputed `ClusterScopeResult` once to apply create/update/delete operations.
- Removes the duplicate `ClusterAnalysis` / `ClustersComponent` models. Deterministic clustering results carry no agent names, descriptions, or prompt rendering; agents derive and pass those values explicitly at the prompt boundary.
- Preserves unchanged surviving method ownership when a newly derived leaf cluster mixes members from different existing components, while edited members remain eligible for reassignment only in the language and file that changed.
- Retains persisted child scopes in the traversal even when an edit makes them small or no longer separable.
- Fresh-clusters a newly created child scope with no persisted analysis instead of seeding it from stale lineage under a reused ID.
- Rejects missing or incomplete persisted child lineage with `IncrementalCacheMissingError` instead of silently fresh-clustering the scope.
- Allocates fresh child cluster IDs from one cross-language namespace, reserves IDs from every baseline language even when that language was removed from the live scope, and avoids renumbering already-disjoint sparse IDs.
- Preserves surviving leaf IDs when a low-cardinality child partition expands to one-symbol clusters, allocating additional symbols above the reserved cross-language namespace.
- Restores unchanged descendant analyses verbatim and refreshes only scopes and relations affected by real member or structural changes.
- Constrains warm-start outbound edge repair to changed-to-unchanged boundaries.
- Validates relation CFG evidence against final rendered method ownership, so deterministic ownership repair cannot leave stale cluster evidence behind.
- Drops groups emptied by previous-member ownership repair before incremental scope planning.
- Materializes a repair target for every previous owner that still has live members, even when anchored grouping omitted that owner.
- Reallocates ownerless group IDs after empty repaired groups are removed, keeping hierarchy IDs aligned with incremental component creation.
- Scopes persisted child-lineage completeness checks by language and owning file, so an identical qualified name in another language does not look like missing baseline lineage.

## Explicit behavior changes

- Incremental analysis now fails loudly when either the root baseline or a persisted child scope lacks trustworthy cluster lineage. It no longer silently substitutes a fresh/full clustering result.
- A persisted child analysis continues to be updated even if the current separability gate would not create that child from scratch.
- Edited methods are excluded from previous-member ownership repair only in the language and file that changed, allowing legitimate reassignment without disturbing an identical qualified name in another language.
- A child scope with no persisted analysis starts unseeded, even if stale method-path entries exist under the same reused component ID.
- IDs from deleted members remain reserved within their persisted child scope across every baseline language, so replacement code cannot inherit old component ownership after an entire language disappears.
- Method-level fallback retains each surviving cluster ID on a persisted member and allocates split or replacement symbols above every reserved ID.
- Relation evidence follows the components' final rendered file/method ownership rather than pre-repair cluster membership.
- A group left with no rendered symbols after ownership repair is omitted instead of persisting a fileless component with stale cluster lineage.
- A previous component with surviving live members remains a repair target even when anchored grouping did not emit its owner ID.
- Surviving ownerless groups are reallocated after repair so component IDs and hierarchy group IDs cannot diverge.
- Persisted-lineage completeness is evaluated per language/file rather than globally by qualified name.

Definition-position resolution is unchanged from #498 and `main`; the #499-only ±2-line/selection-width fallback was removed. The service ownership move and agent-boundary rendering refactor are intended to preserve behavior.

No LLM result cache is introduced here. The details-agent cache was removed in predecessor #510; static-analysis and metadata caches remain.

## Test scope

The incremental tests cover distinct ownership, lineage, hierarchy, and fail-fast contracts. The earlier warm-start block was removed; `tests/static_analyzer/test_analysis_cache_inmem.py` has no diff in this PR. Compared with #498, the complete test diff is net +17 lines.

## Boundary of this PR

This PR owns the incremental migration and stability repairs. It removes only the now-redundant clustering orchestration proxy; the broader recursive planner and structural-diff compatibility code remain for deletion in #500.

## Stack roadmap

|---|---|---|
| Previous | [#510](https://github.com/CodeBoarding/CodeBoarding/pull/510) | Remove the low-value details-agent cache. |
| Previous | [#498](https://github.com/CodeBoarding/CodeBoarding/pull/498) | Migrate full and partial analysis to hierarchy consumption. |
| **This PR** | **#499** | Migrate incremental analysis and preserve trusted scope ownership. |
| Next / final | [#500](https://github.com/CodeBoarding/CodeBoarding/pull/500) | Delete remaining migration leftovers and legacy planner paths. |

---


diagram_analysis/diagram_generator.py
```diff
@@ -1447,18 +1464,50 @@ def _apply_incremental_scope_recursively(
...
+            refresh_ids=set(applied.refresh_ids),
+            new_component_ids=set(applied.new_component_ids),
+            removed_ids=set(applied.removed_ids),
+        )
+        if applied.refresh_ids or applied.removed_ids:
+            result.relation_contexts[clustering.scope_id] = applied.relation_context
+
+        for group in clustering.groups:
+            if group.children is None or group.group_id not in sub_analyses:
+                continue
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c0de73a8`. Persisted child scopes are retained in hierarchy traversal even when the current graph is no longer independently separable, so their nested analysis is updated instead of silently skipped.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +502,30 @@ def _incremental_changed_component_ids(
...
+        method_paths = static_analysis.get_clusters(Language(language)).method_paths
+        snapshot = scoped_snapshot_from_lineage(graph, method_paths, scope_id)
+        if snapshot:
+            partition = _delta_for_language(
+                language,
+                graph.to_networkx(DEFAULT_REFERENCE_KINDS),
+                snapshot,
+            ).cluster_results
+        else:
+            partition = clustering_service.cluster(graph)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c0de73a8`. A persisted child with live methods now requires complete scope-local lineage from `incremental_base_results`; missing members raise `IncrementalCacheMissingError` instead of triggering fresh child clustering.

static_analyzer/incremental_orchestrator.py
```diff
@@ -304,6 +308,9 @@ def _definition_nodes(
         return []
     file_path, line, character = location
     target = _most_specific_node_at_position(call_graph, file_path, line, character)
+    # A definition must name a graph declaration, not merely fall inside its body.
+    if target is not None and target.line_start != line + 1:
+        target = None
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `d1020b61` and completed in `63c1a7ff`. Warm-start definition repair accepts adjacent declaration/selection locations in both directions while checking reverse matches against the selected symbol width, so nested body locations remain rejected.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +501,30 @@ def _incremental_changed_component_ids(
...
+    clustering_service = ClusteringService()
+    for language, graph in graphs.items():
+        method_paths = static_analysis.get_clusters(Language(language)).method_paths
+        snapshot = scoped_snapshot_from_lineage(graph, method_paths, scope_id)
+        if snapshot:
+            partition = _delta_for_language(
+                language,
+                graph.to_networkx(DEFAULT_REFERENCE_KINDS),
+                snapshot,
+            ).cluster_results
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c0de73a8`. Child languages share one monotonic allocator seeded above every persisted ID, and already-disjoint sparse language partitions are preserved rather than reindexed.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +502,55 @@ def _incremental_changed_component_ids(
...
+    graphs: Mapping[str, CallGraph],
+    persisted_members: set[str],
+    artifact_dir: Path,
+) -> dict[str, ClusterResult]:
+    """Seed each exact child graph from its persisted scope-local cluster lineage."""
+    snapshots: dict[str, dict[int, ClusterSnapshotEntry]] = {}
+    for language, graph in graphs.items():
+        try:
+            method_paths = baseline_analysis.get_clusters(Language(language)).method_paths
+            snapshot = scoped_snapshot_from_lineage(graph, method_paths, scope_id)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `63c1a7ff`. Allocation now reserves every exact-scope ID present in persisted method-path lineage, including IDs whose only member was deleted, so replacement clusters cannot reuse stale ownership IDs.

static_analyzer/incremental_orchestrator.py
```diff
@@ -526,21 +540,27 @@ def _definition_nodes(
...
-            target = max(
-                same_line,
-                key=lambda node: (
-                    node.is_class(),
-                    node.is_callable(),
-                    len(node.fully_qualified_name),
-                ),
-            )
+        # Some providers return a decorator or annotation line before the declaration.
+        for declaration_line in range(line + 1, line + 4):
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `63c1a7ff`. Adjacent definition resolution now checks declaration/selection offsets in both directions. Reverse matches must also match the selected symbol width, preserving the existing rejection of nested non-graph definitions.

static_analyzer/clustering/repair.py
```diff
@@ -0,0 +1,29 @@
...
+        }
+        for qualified_name, previous_group_id in owner_by_member.items():
+            source = current_group.get(qualified_name)
+            target = group_by_id.get(previous_group_id)
+            if source is None or target is None or source is target:
+                continue
+            source.symbol_members_by_language[language].remove(qualified_name)
+            if not source.symbol_members_by_language[language]:
+                del source.symbol_members_by_language[language]
+            target.symbol_members_by_language.setdefault(language, set()).add(qualified_name)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `63c1a7ff`. Relation evidence is now built from the components’ final rendered file/method ownership and live CFG edges rather than pre-repair cluster membership, so ownership repair and relation validation use the same endpoints.

static_analyzer/clustering/repair.py
```diff
@@ -0,0 +1,29 @@
...
+        current_group = {
+            qualified_name: group
+            for group in groups
+            for qualified_name in group.symbol_members_by_language.get(language, set())
+        }
+        for qualified_name, previous_group_id in owner_by_member.items():
+            source = current_group.get(qualified_name)
+            target = group_by_id.get(previous_group_id)
+            if source is None or target is None or source is target:
+                continue
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c64d6364`. The incremental hierarchy now removes body-edited qualified names from `previous_member_owner` before `ClusterScopeInput` reaches ownership repair. Unchanged survivors remain pinned, while edited methods stay in the freshly clustered group.

diagram_analysis/diagram_generator.py
```diff
@@ -732,6 +732,67 @@ def prepare_analysis(
...
+            raise IncrementalCacheMissingError(self.output_dir)
+        persisted_scopes = {ROOT_SCOPE_ID: root_analysis, **sub_analyses}
+
+        def scope_input(scope_id: str, graphs: Mapping[str, CallGraph]) -> ClusterScopeInput:
+            persisted = persisted_scopes.get(scope_id)
+            leaf_clusters = (
+                root_leaf_clusters
+                if scope_id == ROOT_SCOPE_ID
+                else _incremental_scope_partitions(
+                    baseline_analysis,
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c64d6364`. `scope_input` now returns an unseeded `ClusterScopeInput` immediately for any non-root scope without a persisted analysis. A genuinely new/reused component ID therefore fresh-clusters its child graph and cannot consume stale scope lineage.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +502,64 @@ def _incremental_changed_component_ids(
...
+    scope_id: str,
+    graphs: Mapping[str, CallGraph],
+    persisted_members: set[str],
+    artifact_dir: Path,
+) -> dict[str, ClusterResult]:
+    """Seed each exact child graph from its persisted scope-local cluster lineage."""
+    snapshots: dict[str, dict[int, ClusterSnapshotEntry]] = {}
+    reserved_cluster_ids: set[int] = set()
+    prefix = f"{scope_id}."
+    for language, graph in graphs.items():
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `82209832`. `_incremental_scope_partitions()` now loads method-path lineage for every language recorded in the baseline before processing the live graphs, then reserves all exact-scope local IDs across that complete language set. The regression removes Python from the live scope while retaining baseline ID `1.9` and verifies that fresh TypeScript allocation starts at `10` rather than reusing the deleted Python identity. The full suite passes (2,174 tests on #499).

diagram_analysis/diagram_generator.py
```diff
@@ -732,6 +741,77 @@ def prepare_analysis(
...
+                persisted,
+                leaf_clusters,
+                scope_id,
+                self.repo_location,
+            )
+            member_owner = {
+                language: {
+                    qualified_name: owner
+                    for qualified_name, owner in owner_by_member.items()
+                    if qualified_name not in self._changed_members
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `82209832`. Previous-member ownership exclusions are now built per language by intersecting the live language graph with the content-changed file set, so an identical qualified name in an unaffected language remains anchored. The regression edits `child.py` while Python and TypeScript both expose `child.member`, and verifies only the Python owner is released. The full suite passes (2,174 tests on #499).

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +503,79 @@ def _incremental_changed_component_ids(
...
+        if snapshot:
+            partition = _delta_for_language(
+                language,
+                graph.to_networkx(DEFAULT_REFERENCE_KINDS),
+                snapshot,
+                next_new_id=next_new_id,
+            ).cluster_results
+        else:
+            partition = clustering_service.cluster(graph)
+        partition = clustering_service.expand_to_method_level(graph, partition)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `a1f31a3e`. Method-level expansion now retains each surviving source cluster ID on a member present in the persisted snapshot and allocates every additional one-symbol cluster from the cross-language `next_new_id` watermark. The regression uses surviving ID `4`, deleted/reserved ID `9`, and replacement ID `10` in a below-threshold scope and verifies fallback returns `{4, 10}` rather than renumbering to `{0, 1}`. The full #499 suite passes (2,176 tests, 83.89% coverage).

static_analyzer/clustering/service.py
```diff
@@ -143,6 +212,7 @@ def cluster_scope(
             for group_id, cluster_ids, owner in zip(group_ids, raw_groups, owners, strict=True)
         ]
         self._assign_symbol_members(graphs, scope_leaf_clusters, groups)
+        repair_member_ownership(groups, previous_member_owner)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `c1adba5b`. After previous-member ownership repair, groups with no rendered symbols are now removed before connection building and scope planning, so they cannot create a fileless component carrying stale `source_cluster_ids`. A regression forces a regrouped unowned group to be completely absorbed by the surviving previous component and verifies only that populated component remains.

static_analyzer/clustering/service.py
```diff
@@ -143,6 +212,8 @@ def cluster_scope(
...
+        groups = [group for group in groups if group.qualified_names]
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in `91680462`. After ownership repair, empty groups are removed and IDs are allocated again from the surviving previous owners plus reserved sibling IDs. This keeps ownerless hierarchy group IDs aligned with the IDs assigned during incremental component creation. The consolidated clustering-scope test covers the full sequence: retained owner `1`, materialized owner `2`, and the surviving fresh group compacted to `3`.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +503,84 @@ def _incremental_changed_component_ids(
...
+        if cluster_id.startswith(prefix) and (local_id := cluster_id.removeprefix(prefix)).isdigit()
+    }
+    for language, graph in graphs.items():
+        method_paths = method_paths_by_language.get(language)
+        if method_paths is not None:
+            snapshot = scoped_snapshot_from_lineage(graph, method_paths, scope_id)
+        else:
+            snapshot = {}
+        covered_members = {member for entry in snapshot.values() for member in entry.members}
+        missing_members = (persisted_members & set(graph.nodes)) - covered_members
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in `91680462`. Persisted-lineage completeness now receives members partitioned by language, and the hierarchy builder derives that partition from normalized persisted file paths matched to each live language graph. An identical qualified name in another language no longer causes a false missing-lineage failure; the same-language case still raises `IncrementalCacheMissingError`. A focused test covers both branches.

static_analyzer/clustering/repair.py
```diff
@@ -0,0 +1,29 @@
...
+    group_by_id = {group.group_id: group for group in groups}
+    for language, owner_by_member in previous_member_owner.items():
+        current_group = {
+            qualified_name: group
+            for group in groups
+            for qualified_name in group.symbol_members_by_language.get(language, set())
+        }
+        for qualified_name, previous_group_id in owner_by_member.items():
+            source = current_group.get(qualified_name)
+            target = group_by_id.get(previous_group_id)
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in `91680462`. Ownership repair now materializes a `ClusterGroup` target for every previous owner that still has a live current member but was omitted by anchored grouping, then moves those members into that target before planning. The consolidated clustering-scope test covers an omitted previous owner, removal of the emptied source group, and allocation of the remaining fresh group.

static_analyzer/incremental_orchestrator.py
```diff
@@ -525,22 +539,44 @@ def _definition_nodes(
     if location is None:
         return []
     file_path, line, character = location
+    reported_line = line + 1
```
ivanmilevtues: can we make sure that this is not really changing behavior here?

static_analyzer/incremental_orchestrator.py
```diff
@@ -525,22 +539,44 @@ def _definition_nodes(
...
-            target = max(
-                same_line,
-                key=lambda node: (
-                    node.is_class(),
-                    node.is_callable(),
-                    len(node.fully_qualified_name),
-                ),
-            )
+        # Providers may report either a declaration range or its inner selection range.
+        for offset in (0, 1, -1, 2, -2):
```
ivanmilevtues: this seems like more of something which has to be part of the SourceReferenceResolver rather than here, also i am unsure if it generalizes across PLs

static_analyzer/clustering/models.py
```diff
@@ -112,6 +115,8 @@ class ClusterGroup:
 
     group_id: GroupId
     cluster_ids: list[ClusterId]
+    name: str = ""
```
ivanmilevtues: why do we have anme and desceription here?

These sound like things that are for the llm stuff. We should keep the communication flowing CFG -> Clustering -> agents.

These are not aegnt models. componetns and such things are part of the agentic loop not part of the cfg or clustering.

diagram_analysis/diagram_generator.py
```diff
@@ -732,6 +762,103 @@ def prepare_analysis(
         )
         self.agent_init()
 
+    def _build_incremental_clustering_hierarchy(
```
ivanmilevtues: I don;t like that we have a custom method for this, I would rather prefer it if it is part of the clustering itself. I want to keep all clustering logic behind the same interface where we can also reuse logic where needed and make the diagram_genreator pretty much just an orchetsrator ultimately.

diagram_analysis/diagram_generator.py
```diff
@@ -550,6 +503,84 @@ def _incremental_changed_component_ids(
     return changed
 
 
+def _incremental_scope_partitions(
```
ivanmilevtues: this method again seems as it should be part of the clustering interface and logic, we can have a separate util for incremntal but ultimately the clustering should be the same for both case.

static_analyzer/clustering/orchestration.py
```diff
@@ -13,19 +13,38 @@
...
-    root_partitions = build_all_cluster_results(static_analysis)
+def _unseeded_scope(_scope_id: str, _graphs: Mapping[str, CallGraph]) -> ClusterScopeInput:
+    return ClusterScopeInput()
 
-    def scope_input(scope_id: str, _graphs: Mapping[str, CallGraph]) -> ClusterScopeInput:
-        if scope_id == "root":
-            return ClusterScopeInput(leaf_clusters_by_language=root_partitions)
-        return ClusterScopeInput()
+
+def build_clustering_hierarchy(
```
ivanmilevtues: ultimately i want these to be the public methods for the ClusteringService and not have a separate file orechestration which is just proxy to the ClusteringService.

static_analyzer/clustering/service.py
```diff
@@ -60,13 +63,79 @@ class ClusteringService:
...
+        }
+        for cluster_id, members in partition.clusters.items():
+            if cluster_id in retained_member_by_cluster:
+                continue
+            live_members = set(members) & expanded_member_set
+            if live_members:
+                retained_member_by_cluster[cluster_id] = min(live_members)
+        next_new_id = max(next_new_id, max(partition.clusters, default=-1) + 1)
+
+        def add_symbol(qualified_name: str) -> None:
```
ivanmilevtues: i don't like inline methods rather make them private but not anonymoust they just make fucs supe rlong.

static_analyzer/incremental_orchestrator.py
ivanmilevtues: Addressed in b4c5b429. The implementation was compared directly with #498, local main, and origin/main. The #499-only ±2-line and selection-width heuristic has been removed; `_definition_nodes` now uses the direct position plus exact same-line fallback from the base branch. The changed-to-changed edge exclusion is a separate intentional #499 behavior and remains documented in the PR body.

static_analyzer/incremental_orchestrator.py
ivanmilevtues: Addressed in b4c5b429 by removing the heuristic rather than relocating it. `StaticReferenceResolver` resolves LLM-produced source references, so it is not the right owner for raw LSP definition-location mapping. The final PR keeps the #498/main behavior; any future provider- or language-aware tolerance can be handled separately at the engine/LSP boundary with dedicated cross-language evidence.

static_analyzer/clustering/models.py
ivanmilevtues: Addressed in b4c5b429 and a69bf8f2. `ClusterGroup` and `ClusterScopeResult` no longer carry names, descriptions, or `llm_str()` rendering. Deterministic clustering stays data-only; agent-facing group labels, descriptions, and prompt text are derived explicitly in `agents/llm_renderers/clustering.py` at the agent boundary.

diagram_analysis/diagram_generator.py
ivanmilevtues: Addressed in b4c5b429 and d53720dd. The DiagramGenerator-specific `_build_incremental_clustering_hierarchy` method is gone. `DiagramGenerator` now asks the public `ClusteringService.build_incremental_hierarchy(...)` API directly; partition seeding, persisted ownership, and hierarchy construction are private service implementation details.

diagram_analysis/diagram_generator.py
ivanmilevtues: Addressed in b4c5b429. Incremental scope partitioning moved to private `ClusteringService._incremental_scope_partitions`; callers only request full, incremental, or existing-scope hierarchies. `DiagramGenerator` no longer owns or calls the low-level partition logic.

static_analyzer/clustering/orchestration.py
ivanmilevtues: Addressed in b4c5b429. `static_analyzer/clustering/orchestration.py` has been deleted. The production `ClusteringService` API is limited to `cluster`, `build_full_hierarchy`, `build_incremental_hierarchy`, and `build_scope_hierarchy`; lower-level scope and hierarchy operations are private, and package exports were tightened accordingly.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed in b4c5b429. The nested `add_symbol` function was replaced by the named private `ClusteringService._add_expanded_symbol` method. The one-use ownership-repair helper module was also folded into a private service method to keep this subsystem behind one implementation boundary.

agents/cluster_methods_mixin.py
```diff
@@ -297,8 +270,10 @@ def _create_strict_component_subgraph(
                 continue
             subgraph_cfgs[lang] = sub_cfg
 
-            seeded_snapshot = scoped_snapshot_from_lineage(
-                sub_cfg, self.static_analysis.get_clusters(lang).method_paths, source_cluster_id_prefix
+            seeded_snapshot = ClusteringService._scoped_snapshot(
```
ivanmilevtues: we shouldn't be calling non-public methods, but als in general we shouldnot really call such methods at all these should be precomuted and used directly as params to the agetn methods.

agents/abstraction_agent.py
```diff
@@ -75,14 +86,40 @@ def __init__(
             ),
         }
 
+    def run(self, scope: ClusterScopeResult) -> AnalysisInsights:
```
Svilen-Stefanov, line 89: blocker: CI is red on this head because of this signature change. `run()` went from `tuple[AnalysisInsights, dict]` to `AnalysisInsights`, but three mocks still return `(analysis, {})`, so `generate_analysis` blows up with `AttributeError: 'tuple' object has no attribute 'components'` at diagram_generator.py:1197:
```
FAILED tests/diagram_analysis/test_diagram_analysis.py::TestDiagramGenerator::test_full_analysis_consumes_the_precomputed_hierarchy
FAILED ...::test_generate_analysis_depth_one_preserves_root_expandable_flags
FAILED ...::test_generate_analysis_uses_hierarchy_expandables_for_can_expand
```
(pre-commit run 32573911367, and it reproduces locally). #500 is what fixes those mocks, so this PR isn't independently green. Can we move that test fix down into this PR?

static_analyzer/clustering/service.py
```diff
@@ -51,22 +59,153 @@ def _unseeded_scope(_scope_id: ScopeId, _graphs: Mapping[str, CallGraph]) -> Clu
...
+        max_depth: int,
+        root_leaf_clusters: Mapping[str, ClusterResult],
+        persisted_scopes: Mapping[str, Any],
+        changed_members: Collection[str],
+        changed_files: Collection[str],
+        repo_dir: Path,
+        artifact_dir: Path,
+    ) -> ClusterScopeResult:
+        """Build an anchored hierarchy from persisted ownership and cluster lineage."""
+        from diagram_analysis.exceptions import IncrementalCacheMissingError
```
Svilen-Stefanov, line 86: blocker-ish, and I think this is the real design question for the PR: `static_analyzer` never imported `diagram_analysis` before, on main or on any branch below this one. Now it does four times, all function-level, all to break a genuine cycle (here, plus `_delta_for_language` and `ClusterSnapshotEntry` further down). AGENTS.md forbids function-level imports and REVIEW.md 1/3 say to resolve the cycle rather than hide it.

What convinces me it isn't just style is that the same pressure shows up three more ways in this file: `static_analysis: Any` on five methods (which is why mypy stays green), `_persisted_members` and `_previous_ownership` with no annotations at all, and a second `_ROOT_SCOPE_ID` next to `agents/scope_ids.py`. Should `build_incremental_hierarchy` and friends live in `diagram_analysis` instead, with the service keeping only the pure graph work?

static_analyzer/clustering/service.py
```diff
@@ -166,17 +317,230 @@ def cluster_hierarchy(
...
+    def _incremental_scope_partitions(
+        self,
+        baseline: Any,
+        scope_id: ScopeId,
+        graphs: Mapping[str, CallGraph],
+        persisted_members: Mapping[str, set[str]],
+        artifact_dir: Path,
+    ) -> dict[str, ClusterResult]:
+        # Imported after static-analysis models initialize; cluster_delta reads those models.
+        from diagram_analysis.cluster_delta import _delta_for_language
```
Svilen-Stefanov, line 380: Same thread as above, but this one also reaches into another package's privates: `_delta_for_language` is `_`-prefixed in `diagram_analysis/cluster_delta.py`. If we do keep the code here, could that at least become a public function on the cluster_delta side?

static_analyzer/clustering/service.py
```diff
@@ -166,17 +317,230 @@ def cluster_hierarchy(
...
+        partitions: dict[str, ClusterResult] = {}
+        next_new_id = max(reserved, default=-1) + 1
+        for language in sorted(graphs):
+            graph = graphs[language]
+            snapshot = snapshots[language]
+            partition = (
+                _delta_for_language(
+                    language, graph.to_networkx(DEFAULT_REFERENCE_KINDS), snapshot, next_new_id=next_new_id
+                ).cluster_results
+                if snapshot
```
Svilen-Stefanov, line 425: blocker-ish: this call drops the drift scoping that the root scope gets. `compute_cluster_delta` passes `diff_files` and `repo_dir` into `_delta_for_language`, so at root an added qname whose file is not in the diff is discarded (`# else: not-in analysis, not-in diff -- drift, drop.`). Here both default to None, so we take the `else: added_nodes = raw_added` branch and every live qname missing from the old member union counts as added, changed file or not.

That means LSP drift gets injected into child scopes as genuine additions, re-partitions them, and shows up as component churn at depth on a commit that never touched those files. Root is protected, children are not. Was that deliberate when the call moved out of `diagram_generator`, or should we thread the diff set through?

static_analyzer/clustering/service.py
```diff
@@ -166,17 +317,230 @@ def cluster_hierarchy(
...
+                in {normalize_repo_path(node.file_path, repo_dir) for node in graph.nodes.values()}
+                for method in group.methods
+            }
+            for language, graph in graphs.items()
+        }
+
+    @staticmethod
+    def _previous_ownership(scope, results: Mapping[str, ClusterResult], scope_id: ScopeId, repo_dir: Path):
+        """Recover persisted ownership from stable members before local cluster IDs."""
+        prefix = "" if scope_id == _ROOT_SCOPE_ID else scope_id
```
Svilen-Stefanov, line 482: This is the concrete cost of the layer problem: `prefix = "" if scope_id == _ROOT_SCOPE_ID else scope_id` and the `f"{prefix}.{cluster_id}"` below are exactly `CodeBoardingClusterIds.prefix_for_scope()` and `qualify_local_id()`, which scope_plan.py:77, incremental_agent.py:598 and static_analysis_enricher_mixin.py:207 all still call. It can't import the helper because that would be another static_analyzer to agents import, so we now have the cluster-id grammar written twice.

Good news, I diffed this against the `previous_ownership` it replaces and all the subtle bits survived (per-language `cluster_to_files` restriction, the `if not files` single-language fallback, both sides normalized, the `source_cluster_ids` fallback, the lowest-id tie-break). Only the paragraph explaining why they're there is gone. Could we keep a one-line `Why:` on the per-language restriction? It's the kind of guard someone deletes as redundant.

static_analyzer/clustering/service.py
```diff
@@ -324,6 +667,40 @@ def _assign_symbol_members(
                 group = group_by_cluster[cluster_id] if cluster_id is not None else groups[0]
```
Svilen-Stefanov, line 667: just checking, can't this KeyError? The two guards above test `cluster_id not in group_by_cluster`, but this line tests `cluster_id is not None`, and `_nearest_cluster` returns ids straight out of `cluster_result.clusters` without checking they belong to a group. For a node that is itself a member of an unowned cluster it returns that cluster at distance 0. Calling `_assign_symbol_members` directly with a partition of {1, 2} and groups owning only cluster 1 gives `KeyError: 2`. Is there an invariant that makes every cluster owned by then, or should this be `if cluster_id in group_by_cluster` for consistency?

Separately, `else groups[0]` quietly dumps everything unassignable into component 1, and the `Assigning N orphan node(s)` log went away with the old mixin. `grep orphan` over production code now returns nothing, so we've lost the signal that surfaced the eShop C# case.

agents/repair.py
```diff
@@ -20,13 +20,14 @@ class ComponentRepairTarget(Protocol):
 @dataclass
 class ComponentRepairContext:
     reference_resolver: StaticReferenceResolver
-    clustering: ClusterScopeResult | ClusterAnalysis
+    clustering: ClusterScopeResult
```
Svilen-Stefanov, line 23: Is `clustering` still used? After this PR `repair_component_group_names` reads `context.group_ids` and `repair_key_entities` reads `context.cluster_results`, so I can't find a read of `context.clustering` anywhere. It's a required positional too, so every caller has to fill it. Same story on `ValidationContext.clustering`.

agents/validation.py
```diff
@@ -49,7 +49,8 @@ class ValidationContext:
     valid_component_names: set[str] = field(default_factory=set)  # For file classification validation
     repo_dir: str | None = None  # For path normalization
     static_analysis: StaticAnalysisResults | None = None  # For qualified name validation
-    clustering: ClusterScopeResult | ClusterAnalysis = field(default_factory=lambda: ClusterScopeResult(scope_id=""))
+    clustering: ClusterScopeResult = field(default_factory=lambda: ClusterScopeResult(scope_id=""))
```
Svilen-Stefanov, line 52: Same as in repair.py: I think this field is dead now. `validate_group_name_coverage` uses `context.group_ids` and `validate_relation_evidence` was rewritten onto `_build_component_edge_lookup`, so `clustering` is only mentioned in a docstring and a log line. Can we drop it and the `ClusterScopeResult(scope_id="")` null object with it? (That empty scope_id is also a third sentinel next to `"root"` and a dotted path.)

agents/relation_edges.py
```diff
@@ -74,6 +75,151 @@ def drop_internal_self_relations(relations: list[Relation]) -> list[Relation]:
...
+    return kept
+
+
+def _qualified_names_match(first: str, second: str) -> bool:
+    """Return whether canonical or suffix-qualified names denote the same symbol."""
+    first, second = first.replace(":", "."), second.replace(":", ".")
+    return first == second or first.endswith(f".{second}") or second.endswith(f".{first}")
+
+
+def _is_self_or_descendant(component_id: str, ancestor_id: str) -> bool:
```
Svilen-Stefanov, line 219: Again a duplicate: this is character for character `static_analyzer/cluster_relations.py:97::is_self_or_descendant`, which is already public and is the exact symbol the wrapper imports. Can we just use that one?

agents/relation_edges.py
```diff
@@ -74,6 +75,151 @@ def drop_internal_self_relations(relations: list[Relation]) -> list[Relation]:
     return [relation for relation in relations if not _is_internal_self_relation(relation)]
 
 
+def ground_relation_edges(
+    llm_key_edges: list[RelationEdge], static_edges: list[RelationEdge]
+) -> tuple[list[RelationEdge], list[RelationEdge]]:
+    """Ground an LLM relation's key edges in deterministic static edges."""
+    static_unique = Relation._unique_edges(static_edges)
```
Svilen-Stefanov, line 82: nit: `Relation._unique_edges` is private on the model and we're calling it from here four times (82, 84, 169, 336). Should it just be public, or should this dedupe live on `Relation`?

agents/static_analysis_enricher_mixin.py
```diff
@@ -0,0 +1,225 @@
...
+        self,
+        analysis: AnalysisInsights,
+        scope: ClusterScopeResult,
+    ) -> None:
+        """Populate component files from authoritative scope membership."""
+        source_cache: SourceCache = {}
+        assigned = 0
+        for group in scope.groups:
+            component = analysis.component_by_id(group.group_id)
+            if component is None:
```
Svilen-Stefanov, line 86: This `continue` is silent: if a group has no matching component we drop every symbol it owns from the output, and the `Component symbol coverage: %d/%d` line below still counts them in the denominator, so the log understates it too. Worth at least a warning?

Also the node-collection comprehension right below is copy-pasted verbatim into `IncrementalAgent._patch_scope_file_methods`. Could that be a small `ClusterGroup.nodes(graphs)` helper? I'll make a ticket if you'd rather not do it here.

agents/static_analysis_enricher_mixin.py
```diff
@@ -0,0 +1,225 @@
...
+                    src_id,
+                    dst_id,
+                )
+            ]
+            has_evidence = bool(llm_relation.evidence.strip())
+            if not static_edges and not key_edges and not has_evidence:
+                continue
+            if not static_edges and not key_edges:
+                logger.warning(
+                    "Keeping LLM-only relation without static or key-edge backing: %s -> %s (%s). Evidence: %s",
```
Svilen-Stefanov, line 131: I wonder if this should be dropped rather than warned about. A relation with no static edges and no key edges is exactly the ungrounded case we've been trying to kill, and `ground_relation_edges` feeds it: when there are no static edges it returns `(merged, merged)`, so the LLM's key edges become `all_edges` with nothing backing them. Can we decide these two together?

agents/llm_renderers/clustering.py
```diff
@@ -0,0 +1,69 @@
...
+                scope.graphs_by_language[language].nodes[qualified_name].file_path
+                for language, qualified_names in group.symbol_members_by_language.items()
+                if language in scope.graphs_by_language
+                for qualified_name in qualified_names
+                if qualified_name in scope.graphs_by_language[language].nodes
+            }
+        )
+        parts = [f"{len(group.cluster_ids)} leaf clusters, {len(symbols)} symbols across {len(files)} files."]
+        if files:
+            shown = ", ".join(Path(path).name for path in files[:8])
```
Svilen-Stefanov, line 32: nit: these were named defaults (`max_symbols=12`, `max_files=8`) on `_group_summary` in #498 and became bare literals when the code moved here, plus a new bare `10` for the edge cap below. Small net regression, can we name them?

tests/agents/test_repair.py
```diff
@@ -100,3 +107,20 @@ def test_component_validators_do_not_mutate_repairable_metadata() -> None:
     validate_key_entities(architecture, validation_context)
```
Svilen-Stefanov, line 107: just checking, does this still exercise `validate_group_name_coverage`? The context above is built with `clustering=` but no `group_ids=`, and this PR rewired the validator to read `set(context.group_ids)`, so `expected_group_names` is empty and it takes the early return on the first line. The `"Grop 1"` typo in the fixture looks like it was meant to go through the repair path. Worth passing `group_ids` here so it reaches the real logic? It's also more evidence that `clustering` on these contexts is dead now.

Svilen-Stefanov (changes requested):
This was a bit more though to go through and understand - a mixture of AI generated and "original" comments :D

static_analyzer/clustering/service.py
```diff
@@ -51,22 +59,153 @@ def _unseeded_scope(_scope_id: ScopeId, _graphs: Mapping[str, CallGraph]) -> Clu
...
+            if persisted is None:
+                return ClusterScopeInput(leaf_clusters_by_language=leaf_clusters)
+            cluster_owner, member_owner = self._previous_ownership(persisted, leaf_clusters, scope_id, repo_dir)
+            for language, owner_by_member in member_owner.items():
+                graph = graphs[language]
+                member_owner[language] = {
+                    name: owner
+                    for name, owner in owner_by_member.items()
+                    if name not in changed_members
+                    or normalize_repo_path(graph.nodes[name].file_path, repo_dir) not in normalized_changed_files
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed on the final stack in `727b07cd`: deleted persisted members are checked before any live graph lookup. The later simplification in `ca17c2cd` removed the changed-member ownership filter entirely, so the final code no longer has this failure path.

agents/abstraction_agent.py
ivanmilevtues: Addressed on the final stack in `476574ba`. The stale mocks now return `AnalysisInsights` directly, matching the new `AbstractionAgent.run()` contract, and the full suite is green on the latest #500 tip.

agents/cluster_methods_mixin.py
ivanmilevtues: Addressed on the final stack in `476574ba`. The legacy `cluster_methods_mixin.py` orchestration was removed; agents now consume the precomputed clustering scopes instead of calling those private clustering methods.

static_analyzer/clustering/service.py
```diff
@@ -143,6 +282,18 @@ def cluster_scope(
             for group_id, cluster_ids, owner in zip(group_ids, raw_groups, owners, strict=True)
         ]
         self._assign_symbol_members(graphs, scope_leaf_clusters, groups)
+        self._repair_member_ownership(groups, previous_member_owner)
+        groups = [group for group in groups if group.qualified_names]
```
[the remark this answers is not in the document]
ivanmilevtues: This is intentional under the current architecture contract: component membership and persisted `file_methods` cover callable and class symbols, while variables/constants are not standalone architectural members. A cluster containing architectural symbols is retained through those symbols; a data-only cluster does not represent a persistable component. Supporting data-only components would be a separate product change rather than a regression in this refactor.

static_analyzer/clustering/service.py
ivanmilevtues: Confirmed: both fresh and anchored grouping absorb every structural leaf cluster into exactly one group. I added an explicit assertion at the symbol-assignment boundary on the final #500 tip (`91718a64`), so malformed input now fails with the unowned cluster IDs instead of reaching the later dictionary lookup. The `groups[0]` fallback remains only for callable/class symbols omitted from the leaf partition; those are persisted as explicitly unclustered members.

agents/tools/component_bridge_edges.py
```diff
@@ -27,7 +27,7 @@ class ComponentBridgeEdgesTool(BaseRepoTool):
     args_schema: ArgsSchema = ComponentBridgeEdgesInput
 
     def _run(self, source_group_names: list[str], destination_group_names: list[str]) -> str:
-        if not self.context.clustering.group_ids():
+        if not self.context.cluster_group_ids:
```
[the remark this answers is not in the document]
ivanmilevtues: Confirmed: the broader shared-context race already exists on `main`. Full/deep analysis runs concurrent component work through one `DetailsAgent`, while each scope mutates the same toolkit context. There have been no confirmed user-visible failures from this so far, so the current stack will not grow a speculative concurrency refactor. The proper per-worker agent/toolkit design is tracked in #512 and can be prioritized if this starts producing observable errors.

static_analyzer/clustering/service.py
ivanmilevtues: Confirmed: nested reconstruction does lose the root delta’s diff scoping, and the same gap exists on `main`. Rather than thread another partial special case through this refactor, the complete fix—one normalized rename-aware path set applied to root and every nested `delta_for_language` call—is tracked in #513 with the required nested-scope regression. Resolving this review thread in favor of that focused follow-up.

agents/static_analysis_enricher_mixin.py
ivanmilevtues: Resolved on the final #500 tip (`d39d5384`, simplified in `91718a64`) with the explicit product gate: drop an edge when neither endpoint resolves, or when both endpoints resolve to the same symbol; otherwise keep it. This applies in validation and final relation-edge filtering. Evidence-only runtime/config relations can still survive without manufacturing `all_edges`.

agents/tools/component_bridge_edges.py
```diff
@@ -68,18 +64,12 @@ def _run(self, source_group_names: list[str], destination_group_names: list[str]
...
             for cluster_id in cluster_ids
         }
-
-    def _nodes_for_clusters(self, cluster_ids: set[int]) -> set[str]:
-        nodes: set[str] = set()
-        for cluster_result in self.context.cluster_results.values():
-            for cluster_id in cluster_ids:
-                nodes.update(cluster_result.get_nodes_for_cluster(cluster_id))
-        return nodes
+        return {group.group_id for group in self.context.clustering.groups if set(group.cluster_ids) & cluster_ids}
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `be6a0c2e` on the final #500 tip. The tool context now maps each prompt/component `source_group_name` directly to the authoritative live `ClusterGroup.group_id`, and `getComponentBridgeEdges` uses that identity without reverse-matching leaf-cluster overlap. Leaf `cluster_ids` remain unchanged for Leiden lineage and persistence. Added a regression proving that a repaired group with `cluster_ids=[]` still resolves its real `GroupConnection`, plus full/detail and incremental context checks.

static_analyzer/clustering/service.py
```diff
@@ -166,17 +317,230 @@ def cluster_hierarchy(
...
+        for language, result in results.items():
+            files = {
+                normalize_repo_path(path, repo_dir) for paths in result.cluster_to_files.values() for path in paths
+            }
+            by_member = {
+                method.qualified_name: component.component_id
+                for component in scope.components
+                if component.component_id
+                for group in component.file_methods
+                if not files or normalize_repo_path(group.file_path, repo_dir) in files
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in stacked PR #516, commit 618a989b. Persisted callable/class ownership is now indexed from each language’s normalized live graph rather than inferred from `cluster_to_files`. That member-owner map is reused by ownership repair, so a live callable omitted from Leiden’s structural file maps retains its persisted component. The regression test covers a persisted member outside the structural partition.

static_analyzer/clustering/service.py
```diff
@@ -166,17 +317,230 @@ def cluster_hierarchy(
...
+    def _persisted_members(persisted, graphs: Mapping[str, CallGraph], repo_dir: Path) -> dict[str, set[str]]:
+        if persisted is None:
+            return {}
+        return {
+            language: {
+                method.qualified_name
+                for component in persisted.components
+                for group in component.file_methods
+                if normalize_repo_path(group.file_path, repo_dir)
+                in {normalize_repo_path(node.file_path, repo_dir) for node in graph.nodes.values()}
```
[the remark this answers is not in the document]
ivanmilevtues: Addressed in stacked PR #516, commit 618a989b. Incremental hierarchy preparation now builds one live `normalized file -> qualified names` index per language before walking persisted scopes. Every persisted file group joins against that index, eliminating the repeated graph-wide normalization scan; a regression test verifies live node paths are normalized once.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed in commit f00fb93c. The incremental delta, snapshot, cache exceptions, and hierarchy service now live together under `static_analyzer.clustering`; the service no longer imports `diagram_analysis`, the function-level cycle-breaking imports are gone, and shared ID grammar moved to the neutral `clustering_ids.py` module.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed in commit f00fb93c. The delta implementation now belongs to `static_analyzer/clustering/delta.py`, and `delta_for_language` is a public function imported normally by the clustering service.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed in commit f00fb93c and completed in stacked PR #516, commit 618a989b. Scope prefixes and local-ID qualification now use the shared `CodeBoardingClusterIds` API. Persisted ownership is now built once from normalized live per-language symbols rather than inferred from `cluster_to_files`, while source-ID fallback and deterministic lowest-ID tie-breaking remain intact.

agents/repair.py
ivanmilevtues: Addressed in commit d39d5384. The unused `clustering` field and required `ClusterScopeResult` argument were removed from `ComponentRepairContext`; callers now provide only the cluster results and group IDs the repairs consume.

agents/validation.py
ivanmilevtues: Addressed in commit d39d5384. `ValidationContext.clustering` and its empty-scope null object were removed; validation now carries only `cluster_results`, `group_ids`, and its other direct inputs.

agents/relation_edges.py
ivanmilevtues: Addressed in commit d39d5384. `is_self_or_descendant` is centralized in the neutral `clustering_ids.py` module and both relation assembly and static cluster-relation code import the shared implementation.

agents/relation_edges.py
ivanmilevtues: Addressed in commit d39d5384. `Relation.unique_edges` is now public and all external callers use that public API.

agents/static_analysis_enricher_mixin.py
ivanmilevtues: Addressed in commit d39d5384. A clustering group without a matching component now raises immediately instead of silently dropping its symbols and producing misleading coverage. The two local node projections remain inline rather than adding a helper without a separate domain responsibility.

agents/llm_renderers/clustering.py
ivanmilevtues: Addressed in commit d39d5384. The renderer limits are named `MAX_GROUP_FILES`, `MAX_GROUP_SYMBOLS`, and `MAX_CONNECTION_EDGES` and are used consistently in slicing and overflow labels.

tests/agents/test_repair.py
ivanmilevtues: Addressed in commit d39d5384. The test now passes `repair_context.group_ids` into `ValidationContext`, so the typo reaches the real group-name coverage path instead of taking the empty-groups early return.

ivanmilevtues:

PR #500, opened 2026-08-18
refactor(clustering): remove legacy agent orchestration

ivanmilevtues:
## Goal

Finish the migration by leaving the precomputed `ClusterScopeResult` hierarchy as the only structural orchestration path for standard full, partial, and incremental analysis.

This is a consolidation/deletion PR, not another clustering implementation.

## What this PR addresses

- Removes the old recursive incremental scope updater and child-scope diff/reclustering helpers.
- Removes `plan_scope_update(...)`; incremental planning now accepts only a precomputed `ClusterScopeResult`.
- Removes structural cluster-diff models and classification code used only by the superseded updater.
- Removes standard-workflow agent grouping fallbacks; agents consume precomputed scopes through their single public `run(...)` methods.
- Keeps partial analysis explicit: preparation registers the requested component scope, then `DetailsAgent.run(scope, component)` analyzes it.
- Retains seeded delta computation and changed-member detection used by the hierarchy-driven incremental path.
- Replaces legacy-path tests with direct coverage of hierarchy and partial-analysis boundaries.

## Resulting ownership boundary

- `DiagramGenerator.deterministic_analysis(...)` owns static analysis and hierarchy construction.
- `DiagramGenerator.agent_init()` owns LLM-agent initialization.
- `prepare_analysis(...)` composes those stages.
- `static_analyzer.clustering` owns grouping, membership, hierarchy, expansion, and communication.
- Full, partial, and incremental diagram generation consume `ClusterScopeResult` directly.
- Agents interpret fixed groups and produce semantic analysis; they do not regroup standard workflow scopes.

## Behavior changes

No new behavior is intended in this PR. It removes paths made unreachable by #498 and #499. The explicit incremental fail-fast and persisted-scope behavior changes are documented in #499; details-cache removal is documented in #510.

## Stack roadmap

|---|---|---|
| Previous | [#510](https://github.com/CodeBoarding/CodeBoarding/pull/510) | Remove the low-value details-agent cache. |
| Previous | [#498](https://github.com/CodeBoarding/CodeBoarding/pull/498) | Migrate full and partial analysis to hierarchy consumption. |
| Previous | [#499](https://github.com/CodeBoarding/CodeBoarding/pull/499) | Migrate incremental analysis to an ownership-anchored hierarchy. |
| **This PR / final** | **#500** | Remove obsolete orchestration and leave one structural source of truth. |

There are no additional migration PRs after this one in the current stack.

---


diagram_analysis/scope_plan.py
```diff
@@ -97,65 +93,6 @@ def previous_ownership(
...
-        scope_id,
-        scope,
-        cluster_results,
-        groups,
-        changed_members,
-        regrouped,
-    )
-
-
 def plan_scope_result_update(
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `6fb6fff4`. The architecture summary now names the active deterministic `plan_scope_result_update()` entry point instead of the deleted `plan_scope_update()` compatibility path.

static_analyzer/analysis_cache.py
```diff
@@ -55,8 +55,9 @@
 # dataclass storage.
 # v3: MethodClusterPaths moved to static_analyzer.clustering, then CallGraph and the edge
 # types to static_analyzer.cfg, and reference edges became ReferenceEdge objects.
+# v4: ClusterCache records intentionally unclustered members for each hierarchy scope.
```
Svilen-Stefanov, line 58: Does this mean that if we add a README or a test, we will have a miss?

static_analyzer/analysis_cache.py
```diff
@@ -55,8 +55,9 @@
...
 # Older pickles are treated as cache misses and re-run.
-_TAG_VERSION = "v3"
+_TAG_VERSION = "v4"
```
Svilen-Stefanov, line 60: **Overall on this PR:** Nice cleanup, dropping the structural-diff machinery makes cluster_delta.py much easier to hold in your head. I answered my own analysis_cache question inline (short version: a README won't miss, but v4 misses every existing cache once). The thing I'd still like resolved is that my blocker-ish from #497 about which grouping the `< 2` gate reads is still open, and codex's newest comment on cache.py hasn't been answered yet.

---

Answering my own question above, since I went and checked: no, a README or a test won't cause a miss. The SHA half of this tag is a diff base, not a gate. `load_with_sha()` "never gates on the SHA, it just hands it back", so a new file only widens the re-LSP set, and `hash_repo_source_files` covers docs and configs on purpose so the wrapper reproduces the same map.

The version half is the all-or-nothing one though, and that's what worries me here. `_read_tag_sha_unlocked` returns None when `version != _TAG_VERSION`, so `load_with_sha()` returns None, `incremental_base_results` never gets set, and `build_incremental_hierarchy` raises `IncrementalCacheMissingError`. So every existing user's first incremental run after this ships fails and tells them to run a full analysis, and that's already our most common user-facing error.

The bump does look necessary, I checked: a v3 pickle restores a `ClusterCache` with no `unclustered_members_by_scope` in its `__dict__` and `get_unclustered_members()` raises AttributeError. But could we tolerate v3 with a `__setstate__` or a `getattr` default instead of invalidating everyone? If not, this probably wants a release note.

static_analyzer/clustering/service.py
```diff
@@ -587,10 +606,10 @@ def _cluster_children(
                 not child_input.retain_scope
                 and load < 1.0
                 and not scope_is_separable(
-                    child.leaf_clusters_by_language,
-                    child.unanchored_modularity,
-                    load,
-                    method_count,
+                    leaf_clusters_by_language=child.leaf_clusters_by_language,
+                    modularity=child.unanchored_modularity,
```
Svilen-Stefanov: My blocker-ish from #497 is still open and I think this hunk shows why. This gate deliberately reads `child.unanchored_modularity`, but the `< 2` check three lines up counts `child.groups`, which is what the *anchored* path produced. This PR only wrapped both in `not child_input.retain_scope`, so the two gates still disagree about which grouping they're judging. Shouldn't the count read the unanchored grouping too, or am I misreading what `child.groups` holds after anchoring?

Related, `test_retains_a_persisted_child_scope_when_it_is_no_longer_separable` now pins the `retain_scope` divergence as correct: with the gate patched to False it still asserts expandable + children. That means once a component is persisted it stays expanded forever even if a fresh run would collapse it, so full and incremental disagree on depth for identical source. Is that intended? I couldn't find a test that pins the two agreeing.

static_analyzer/clustering/service.py
```diff
@@ -512,13 +521,23 @@ def _previous_ownership(scope, results: Mapping[str, ClusterResult], scope_id: S
         return owners, member_owners
 
     @staticmethod
-    def _record_child_scopes(static_analysis: Any, scope: ClusterScopeResult) -> None:
+    def _record_scopes(static_analysis: Any, scope: ClusterScopeResult) -> None:
+        for language, partition in scope.leaf_clusters_by_language.items():
+            cache = static_analysis.get_clusters(Language(language))
+            cache_scope_id = "" if scope.scope_id == _ROOT_SCOPE_ID else scope.scope_id
+            if scope.scope_id == _ROOT_SCOPE_ID:
+                cache.adopt(partition)
```
Svilen-Stefanov, line 537: codex's comment about deferring root cache adoption is still unanswered and I'd like to understand it before merge. `cache.adopt(partition)` here runs during `build_incremental_hierarchy`, so inside preparation, long before `finalize_and_save`. If any teardown path flushes the live pkl after a mid-run failure, does the next incremental see an empty delta? @ivanmilevtues can you confirm nothing writes the pkl without a successful save?

static_analyzer/clustering/cache.py
```diff
@@ -36,12 +37,39 @@ def record_scope(self, cluster_result: ClusterResult, scope_id: str) -> None:
...
+
+    def get_unclustered_members(self, scope_id: str = "") -> set[str]:
+        return set(self.unclustered_members_by_scope.get(scope_id, set()))
+
+    def reroot_scope(self, child_id: str, parent_id: str) -> None:
+        """Move a child's structural and unclustered lineage onto its parent."""
+        self.method_paths.reroot_scope(child_id, parent_id)
+        rerooted: dict[str, set[str]] = {}
+        for scope_id, members in self.unclustered_members_by_scope.items():
+            if self._scope_belongs_to(scope_id, child_id):
```
Svilen-Stefanov, line 57: codex's newest comment here is also unanswered. I worked through it and I think it's real: `reroot_scope(child_id="1", parent_id="")` sends scope `"1"` to `moved = ""`, which lands on the same key the root's own unclustered members use, and the root entry itself gets dropped because `_scope_belongs_to("", "")` is True so neither branch keeps it. So absorbing a single top-level child into the root replaces the root's unclustered members with the child's. Worth a test for the `parent_id == ""` case?

static_analyzer/clustering/service.py
```diff
@@ -512,13 +521,23 @@ def _previous_ownership(scope, results: Mapping[str, ClusterResult], scope_id: S
...
+            else:
+                cache.record_scope(partition, scope.scope_id)
+            assigned_members = {
+                member for group in scope.groups for member in group.symbol_members_by_language.get(language, set())
+            }
+            clustered_members = {member for members in partition.clusters.values() for member in members}
+            cache.record_unclustered(assigned_members - clustered_members, cache_scope_id)
         for group in scope.groups:
```
Svilen-Stefanov, line 545: Related to the expand problem I flagged on #498, and this one needs no partial run to hit. `_cluster_children` does `group.expandable = True` and *then* `if depth >= max_depth: continue`, so a group at the cap is expandable with `children` still None. This recursion skips exactly those groups, so nothing records lineage for them, while `analysis_json.py:333` applies no depth cap and happily writes `can_expand: true`.

With the default depth_cap of 3 that means every depth-3 component ships with an expand arrow and no recorded cluster lineage, and expanding one from the UI lands in the same `IncrementalCacheMissingError` state. Should the flag be set only when we actually recorded something, or should the recursion record the cap-level child scope even though it isn't attached?

Svilen-Stefanov (changes requested):
I think this PR looks quite clean, only requesting changes here due to the potential expansion issue, which I'm not quite sure about. 
Thank you again for the iterations on these.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed on the final #500 stack. ivanmilevtues matched main’s expansion contract: load >= 1.0 expands unconditionally; below the ceiling, the minimum group count and modularity both come from the fresh unanchored partition; and retain_scope continues preserving already persisted subtrees. Regressions cover both the 121-method single-owner reproducer and a smaller scope whose anchored result has one group while the unanchored result has multiple groups.

static_analyzer/clustering/service.py
ivanmilevtues: Addressed by `f819b215` on the final stack. The depth cap intentionally stops eager child generation while leaving the component expandable. A partial expansion now verifies the source tree is unchanged, builds the selected component scope directly, records its structural and explicitly unclustered lineage through `build_scope_hierarchy`, saves the child analysis, and then persists the updated static-analysis cache. An unexpanded cap-level component has no persisted child scope yet, so it does not require lineage before that request.

diagram_analysis/diagram_generator.py
```diff
@@ -651,6 +655,52 @@ def agent_init(self) -> None:
...
+            return
+
+        agent_llm, parsing_llm = initialize_llms()
+        self._initialize_meta_agent(agent_llm, parsing_llm)
+        assert self.meta_agent is not None
+        with ThreadPoolExecutor(max_workers=1) as executor:
+            meta_future = executor.submit(
+                self.meta_agent.analyze_project_metadata,
+                skip_cache=self.force_full_analysis,
+            )
```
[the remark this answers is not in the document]
ivanmilevtues: The current implementation preserves the same overlap and failure semantics as `main`: metadata runs in the executor worker while deterministic preparation runs concurrently on the caller thread. `main` also exits a `ThreadPoolExecutor` context on deterministic failure, which waits for the metadata request. Avoiding that wait would be a separate behavior change and cannot reliably cancel an in-flight provider request or its cost, so we are retaining the established parallel flow here.

static_analyzer/analysis_cache.py
ivanmilevtues: Correct: README/test changes do not invalidate the cache because the SHA is a diff base, not a load gate. A cache-format version mismatch does invalidate the existing cache. `9e38ba62` documents the intended contract without tying the README to a particular version: static-analysis caches are versioned, not migrated, so a version upgrade requires one full reindex.

static_analyzer/analysis_cache.py
ivanmilevtues: The remaining actionable items from this summary are now addressed on the final stack. `9e38ba62` documents the cache-version reindex contract and makes the expansion decision match `main`: oversized scopes expand unconditionally, while smaller fresh scopes use both the unanchored group count and unanchored modularity; persisted scopes remain sticky. The root-cache concern is being kept in its dedicated threads because it is a broader pre-existing absorption/cache-model issue and is intentionally not receiving a one-off patch here.

static_analyzer/clustering/service.py
ivanmilevtues: Pushed in `9e38ba62`. The implementation now matches `main`: load >= 1.0 bypasses the separability gates; below that ceiling, both the minimum group count and modularity come from the fresh unanchored partition; `retain_scope` continues preserving already persisted subtrees. Regressions cover both the 121-method single-owner reproducer and the anchored-one/unanchored-multiple case.

static_analyzer/clustering/service.py
ivanmilevtues: Fixed in `9e38ba62`. Clustering preparation now works on detached per-language `ClusterCache` copies. Hierarchy construction, scope recording, and save-time absorption mutate only the pending copies; the live caches are installed only after `analysis.json` saves successfully. A pre-save failure discards pending state, so teardown cannot flush clustering ahead of the authoritative document.

codeboarding_workflows/analysis.py
```diff
@@ -97,6 +98,13 @@ def run_partial(
             f"No baseline analysis.json found in '{run_paths.output_dir}'. Run a full analysis first."
         )
 
+    baseline_source_hash = str(metadata.get("source_tree_hash", ""))
+    current_source_hash = compute_source_tree_hash(run_paths.repo_path)
+    if not baseline_source_hash or current_source_hash != baseline_source_hash:
+        raise BaselineUnavailableError(
+            "Partial analysis requires an up-to-date source baseline. Run incremental analysis before expanding a component."
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `9e38ba62`. A missing `source_tree_hash` is now identified as a legacy/content-unversioned baseline and directs the caller to run a full analysis. A present but stale hash continues to direct the caller to incremental analysis.

codeboarding_workflows/analysis.py
```diff
@@ -129,6 +138,8 @@ def run_partial(
         return
 
     generator = build_generator(run_paths, run_context, depth_level=depth_level)
+    generator.source_sha = current_source_hash
+    # A component at the persisted cap needs one additional level prepared for expansion.
     generator.prepare_analysis(hierarchy_depth=depth_level + 1, target_component=component_to_analyze)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed in `9e38ba62`. Partial expansion already verifies whole-tree hash equality, so it now passes an explicit empty `ChangeSet`. Static analysis receives an empty changed-file set and reuses the verified cache instead of attempting Git detection against a content hash and falling back to a full LSP pass.

static_analyzer/clustering/delta.py
```diff
@@ -597,7 +272,7 @@ def _delta_for_language(
     if not universe:
         return LanguageDelta(language=language, cluster_results=ClusterResult())
 
-    raw_added = live_qnames - old_member_union
+    raw_added = live_qnames - old_member_union - set(known_unclustered_members)
```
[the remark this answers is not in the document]
ivanmilevtues: Fixed at the group/member layer in `9e38ba62`, without fabricating a Leiden cluster. Member changes are classified as added, removed, and modified. Added/removed membership now makes incremental preparation actionable even when the structural `ClusterDelta` is empty, so the deterministic hierarchy is applied and stale group ownership is removed. Modified-only changes retain the zero-LLM path.

static_analyzer/clustering/cache.py
```diff
@@ -36,12 +37,39 @@ def record_scope(self, cluster_result: ClusterResult, scope_id: str) -> None:
...
+                moved = f"{parent_id}{scope_id[len(child_id) :]}" if parent_id else scope_id[len(child_id) :]
+                rerooted[moved.lstrip(".")] = set(members)
+            elif not self._scope_belongs_to(scope_id, parent_id):
```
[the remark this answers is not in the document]
ivanmilevtues: Confirmed, but this sits in the older single-child absorption/cache model rather than the incremental preparation fixes in this commit. `main` already promotes the only root component’s generated children and reroots scoped method paths while retaining the root Leiden partition. Persisting unclustered scope metadata exposes a mismatch between those parallel representations. A correct fix requires consolidating scoped partition and omitted-member lineage; a root-only preservation branch would be patchwork, so this remains intentionally open rather than expanding #500 further.

static_analyzer/clustering/cache.py
ivanmilevtues: The `parent_id == ""` path is reachable when the root container has exactly one component and that component has a generated sub-analysis; finalization removes that redundant architectural level and promotes its children into the root. This absorption behavior already exists on `main`. The new persisted unclustered metadata exposes a mismatch with the older parallel cache representations. Fixing that cleanly requires a scoped-lineage cache refactor, so this thread is intentionally left open rather than adding a root-specific condition to #500.

static_analyzer/clustering/delta.py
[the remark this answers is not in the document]
ivanmilevtues: Confirmed as a separate full/incremental equivalence edge case: the persisted unclustered neighbor is absent from the seeded working graph, so hierarchy fallback can assign it to a group but cannot reproduce the full-run Leiden pair. The member-level trigger in `9e38ba62` deliberately does not fabricate or alter structural clusters. Reconsidering adjacent persisted-unclustered symbols in the seeded frontier would change the clustering algorithm beyond this cleanup’s bounded scope, so this thread remains open.

static_analyzer/clustering/cache.py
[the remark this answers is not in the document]
ivanmilevtues: Addressed in #516 by commit 618a989b. `ClusterCache` now stores each scope’s complete partition and omitted-member lineage in one authoritative record, and generic rerooting promotes the child record—including at root—rather than moving only flattened paths/metadata. The regression test verifies that root absorption replaces both pieces together, without a root-specific preservation branch.

static_analyzer/clustering/cache.py
ivanmilevtues: Addressed in #516 by commit 618a989b. The cache now persists a complete structural partition and omitted-member set per scope. Single-child absorption reroots that complete record through the same prefix remap for root and nested scopes, so the promoted child partition and metadata cannot diverge. Tests cover root promotion, descendant remapping, parent replacement, and missing lineage.

ivanmilevtues:

PR #506, opened 2026-08-20
fix(engine): open .tsx and .jsx with the JSX language id

Svilen-Stefanov:
`typescript-language-server` maps the LSP `languageId` onto tsserver's scriptKind, but every file was opened with the adapter-wide id. A `.tsx` therefore arrived as plain `"typescript"` with JSX parsing off, and tsserver read `<div ...>` as a type assertion.

## What that does to the symbol table

Three failures at once, all measured on CodeBoarding-webview:

- **Flattened.** `CloneButton.tsx` returns **36 top-level symbols instead of 2** correctly nested ones — including `<unknown>` ×7, `1.8` ×2 and `[t]`.
- **Truncated.** `DiffViewer`'s component span stops at 1130 instead of 1150, so everything after it falls outside the component.
- **Registered at usage positions.** The `lineHtml` symbol had span **709–709** — a call site — while the real declaration is at 683.

The last one manufactures false edges. A `textDocument/references` query anchored on a usage-position phantom harvests the *real* symbol's entire reference set and files it under the phantom's qualified name, so a single call site fans out across every same-named symbol in the repo.

## Measured

Static analysis only, wall clock unchanged (41.71s → 41.96s):

| | before | after |
|---|---|---|
| typescript edges | 1656 | **1923** (+267, +16%) |
| typescript nodes | 3225 | 3243 |
| unnamed / garbage nodes | 6 | **0** |
| caller+name pairs resolving to >1 symbol | 51 | **14** (surplus edges 60 → 15) |

## On the 217 edges that disappear

This is not a strict superset, so I checked what leaves.

**47 are the fan-out being deduplicated** — the caller keeps its edge to the correctly resolved symbol and loses the wrong one:

| dropped | kept |
|---|---|
| `AnalysisViewer.shorten` | `lib.ref.shorten` |
| `RepoWorkflow.primaryBtn` | `lib.ui.primaryBtn` |
| `ReviewPanel.threads.sortThreads` | `lib.review.commentView.sortThreads` |
| `CommentTime.CommentTime.formatAbsoluteTime` | `lib.review.formatTime.formatAbsoluteTime` |

The remaining 170 were hand-sampled rather than exhaustively verified. **Corrected below** — an exhaustive re-measurement shows most are false positives but a substantial minority are not:

- `MergePrDialog → customise` — `customise` is a `useState` **variable**, not a function
- `FileBlock → renderDiffFor` — cited `CodeStream.tsx:377`, a props-list continuation with no call on it
- `DiffViewer → lineHtml` — pointed at the usage-site phantom described above

**I did not verify all 170 individually.** An automated name-match test can't adjudicate these, because every member of a fan-out group cites a line that genuinely contains the call — just to a different symbol.

## Tests

`tests/static_analyzer/test_typescript_adapter.py`, 6 cases, all confirmed to **fail on `main`**: `.tsx`/`.jsx` take the JSX dialect, plain suffixes keep the adapter id, the TypeScript adapter still returns `javascriptreact` for a `.jsx` it owns, unknown suffixes fall back, and other languages are untouched.

Full suite: 2136 passed. The 3 `test_cli_parser` failures are pre-existing on `main`.

## Why this one first

It is the root cause underneath three other planned fixes. ~~Synthetic edge sources drop from 25.5% to ~0.2%, which is most of the containment-attribution problem~~ — **this figure does not reproduce; see the update below**; and it invalidates the accuracy case for switching TypeScript to the definitions edge strategy, since that case was measured entirely on `.tsx` files with this bug present.

---

## Update: exhaustive re-measurement

Every call-site claim on `main` and on this branch was resolved through the TypeScript compiler API — no sampling, no LLM — on **CodeBoarding-webview** and on **excalidraw v0.18.0**. Determinism was verified by repeat runs (identical content hashes), and the comparison was checked against this branch's actual merge-base so #505 could not contaminate it. Full report: https://claude.ai/code/artifact/4542d54a-9dd8-4256-89be-976fc02a218b

**The diagnosis holds, and is understated.** Precision (`confirmed / (confirmed + wrong)`, excluding interface-signature disagreements from both terms):

| | main | this branch |
|---|---|---|
| webview | 88.7% | **99.2%** |
| excalidraw | **25.1%** | **96.3%** |

**Four corrections to the text above.**

1. **The dropped edges are not all false positives.** Classified exhaustively (webview / excalidraw): **159 / 7,499** are provably false, **65 / 780** are the *same call* re-credited to a corrected caller — the call still exists at the same `file:line:column`, the source symbol changed because its span was wrong — and **4 / 21** are genuinely lost. On webview, 46% of the re-credited group had `<unknown>` or a one-line phantom as the old caller.

2. **"Synthetic edge sources 25.5% → ~0.2%" does not reproduce** under any definition I could construct; two metrics look conflated. `<unknown>`-sourced edges: 1.9% → 0.0% (webview), 4.1% → 0.1% (excalidraw). Anonymous-callback-sourced edges: 25.2% → 25.4% (webview), 28.7% → **36.2%** (excalidraw). A quarter of all edges still have an anonymous source after this change.

3. **The `.jsx` half of the title is a no-op.** Opening a real `.jsx` file against tsserver under `javascript` versus `javascriptreact` returns an **identical symbol tree** — the JavaScript dialect already permits JSX. Only `.tsx` was broken, and there it is dramatic (`ReviewPanel.tsx`: 800 symbols / 368 top-level / 108 `<unknown>` under `typescript`, versus 390 / 31 / 7 under `typescriptreact`).

4. **"+16% edges" is repo-specific and inverts.** On excalidraw this change is **12,665 → 5,506 edges (−56%)**, because that repo had far more same-name fan-out: one call to the i18n `t` produced **81 edges**, one per same-named symbol. Both directions are the same correction. The 56% collapse is the single largest consequence of this PR and is worth stating here rather than only in a fixture commit.

**The 25 genuinely lost edges** all hit one pre-existing guard: correct JSX parsing creates new one-line containers (an arrow in a JSX attribute, and the constant in `const x = f()`), so the call lands on its own container's declaration line and `is_reference_in_declaration_body` rejects it — it accepts only `statement_block` and `arrow_expression_clause`, the latter a C# node type behind an adapter opt-in. Fixed in #515, which recovers all of them and adds 1,270 further confirmed edges across three repositories with none dropped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved language detection for source files, including JSX and TSX files.
  * File language handling is now based on each file’s extension for more reliable analysis.
  * Updated cached analysis handling to safely rebuild incompatible older cache entries.

* **Tests**
  * Added coverage for language detection across supported file types and JSX dialects.

## Cache tag

Bumped `_TAG_VERSION` v3 → v4. The freshness gate is the source SHA, so a repo whose tree has not moved keeps loading the pickle the previous engine built — without a bump nobody sees the corrected graph until they happen to edit a file.

This also widens the comment on that constant. It was documented as an on-disk *layout* version, but a materially different graph from the same source needs a bump for the same reason. Worth noting: **#477 changed C# edges (eShop 1252 → 1399) after the v3 bump without one**, so v3 pickles in the wild are already a mixture of two engines. v4 clears them.


Svilen-Stefanov:
Addressed the review.

**Real bug, fixed.** The JSX id never reached three `did_open` sites, so the malformed parse survived on exactly the paths that matter most: the warm-start cross-boundary restore (`incremental_orchestrator.py:176`), the incremental re-open (`:236`), and `notify_file_changed` (`__init__.py:560`). All three route through `language_id_for` now. Good catch — the fix was silently a no-op for incremental users.

**Style, fixed.** `_JSX_LANGUAGE_IDS` annotated `ClassVar` (RUF012), docstring cut to one line per AGENTS.md.

ivanmilevtues:
Heads-up: the private integration suite is still reporting a failure in one test group. Keeping this as a comment rather than a formal review.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -3,13 +3,16 @@
...
 from pathlib import Path
+from typing import ClassVar
 
 from static_analyzer.constants import Language
 from static_analyzer.engine.language_adapter import LanguageAdapter
 
 
 class TypeScriptAdapter(LanguageAdapter):
 
+    _JSX_LANGUAGE_IDS: ClassVar[dict[str, str]] = {".tsx": "typescriptreact", ".jsx": "javascriptreact"}
```
ivanmilevtues: This should probably come from the LANGUAGE_EXTENSIONS in the static_anlayzer/constants.py where we drop Javascript as alanguage and use just the typescript or something like that, as there is also .js, .ts and so on. Can you please take a quick look and unify use only that.
Same goes for the language and language_id we already have them in the enums.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -207,7 +207,7 @@ def _bulk_did_open(self, source_files: list[Path]) -> None:
         for i in range(0, total, DID_OPEN_BATCH_SIZE):
             batch = source_files[i : i + DID_OPEN_BATCH_SIZE]
             for file_path in batch:
-                self._lsp.did_open(file_path, self._adapter.language_id)
+                self._lsp.did_open(file_path, self._adapter.language_id_for(file_path))
```
ivanmilevtues: i think this is probably fine as the tests pass but maybe take a look @Svilen-Stefanov

static_analyzer/__init__.py
```diff
@@ -557,7 +557,7 @@ def notify_file_changed(self, file_path: Path, content: str) -> None:
             adapter = engine_config.adapter
             if suffix in adapter.file_extensions:
                 # Open + change to ensure the server has the latest content
-                client.did_open(file_path, adapter.language_id)
+                client.did_open(file_path, adapter.language_id_for(file_path))
```
ivanmilevtues: this seems like the wrong patchwork, we should make both javascript and typescript resolve to the same typescript adapter which has the typescript language_id but as typescript is a superset over JS this here should be fine. I think the langauge_id_for is not really needed it is a second way of picking an adapter (after we have already picked an adapter).

ivanmilevtues (approved):
Just few stylistic comments, from functional standpoint it is all good. If you can make use of our enums that would be my main comment and i don't fully see the idea behind language_for as the adapters already have been selected for the proper language, and it is fine imp to have the TS adapter also handle JS, as it already handles tsx, ts and so on.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Agree on the end state, can we do it as its own PR though? I had a look and the fold isn't a local edit:

- `source_inspector.py:42` builds `_LANGUAGE_BY_SUFFIX` by filtering `LANGUAGE_EXTENSIONS` through `_LANGUAGE_FACTORY_BY_LANGUAGE`, which has JAVASCRIPT but deliberately not TYPESCRIPT. Fold js into the TS entry and all four js suffixes drop out of the map, `_parser_for` returns None and `find_call_sites` returns `[]` with no error.
- `SOURCE_EXTENSION_TO_LANGUAGE` has no consumers in this repo but the wrapper uses it to pick a tree-sitter grammar (`tree_sitter_method_differ.py:377`). Widening TS is inert there, but dropping the JAVASCRIPT key sends `.jsx` to the typescript grammar, and I checked in the venv, that grammar returns `has_error=True` on JSX while the javascript one parses it. Silent method-hash regression for every `.jsx` file.
- Removing the enum member is a hard import break in 4 core sites plus the wrapper, and `test_registry_coverage.py:316` and `test_vscode_constants.py:215` become mutually unsatisfiable unless `vscode_constants.py:89` moves too.

So it's a cross-repo change plus a cache bump, not a line in this diff. Happy to own it right after the stack.

On the ids specifically: those two strings genuinely aren't in any enum and can't be derived from one, for the reason above.

static_analyzer/engine/call_graph_builder.py
Svilen-Stefanov: Had a look, the code is right, but you're right that "tests pass" isn't evidence here. `_make_adapter()` in `test_call_graph_builder.py:35` is a MagicMock that only sets `language_id`, so `language_id_for` returns an auto-Mock and we only ever assert did_open call counts. Adding a wiring test with a real TypeScriptAdapter over a `.tsx` and a `.ts` asserting the actual didOpen args.

static_analyzer/__init__.py
Svilen-Stefanov: I think it's the other way round? The adapter is already picked, the guard on the line directly above is `if suffix in adapter.file_extensions`, and globally `_adapter_names_for` already removes JavaScript when TypeScript is present. `language_id_for` only fills the per-document `languageId` on the didOpen payload (`lsp_client.py:287`), and LSP makes that a per-document field, so per-file is the shape the protocol asks for and the adapter-wide constant was the actual mismatch. It also never crosses adapter boundaries, the base default returns `self.language_id` and the only override is the TS/JS hierarchy, which already owns all 8 suffixes.

The single-id version can't fix the bug though: that adapter's id is `typescript`, and a `.tsx` opened as `typescript` is exactly the mis-parse this PR measures, 36 flattened top-level symbols with `<unknown>` x7 on one file.

Agree completely on both families resolving to one adapter, that's #507.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Agree, this shouldn't be a dict on the adapter. Moving it to `constants.py` as an enum:

```python
class LanguageId(StrEnum):
    TYPESCRIPT = "typescript"
    TYPESCRIPT_JSX = "typescriptreact"
    JAVASCRIPT = "javascript"
    JAVASCRIPT_JSX = "javascriptreact"
    ...

LANGUAGE_ID_BY_EXTENSION: dict[str, LanguageId] = {
    **{ext: LanguageId(lang.value) for lang, exts in LANGUAGE_EXTENSIONS.items() for ext in exts},
    ".tsx": LanguageId.TYPESCRIPT_JSX,
    ".jsx": LanguageId.JAVASCRIPT_JSX,
}
```

The two JSX entries are the only ones that aren't just the `Language` value, since a `.tsx` is still TypeScript as far as our enum is concerned. Then `_JSX_LANGUAGE_IDS` and `language_id_for` both go away.

static_analyzer/engine/call_graph_builder.py
Svilen-Stefanov: Good instinct to poke at it. The better answer is that the second argument shouldn't be here: it's always derived from the path we're already passing. Letting `did_open` look it up off `LANGUAGE_ID_BY_EXTENSION` instead, so this line is just `self._lsp.did_open(file_path)`.

Also adding a real wiring test, the current ones use a MagicMock adapter so they only assert call counts.

static_analyzer/__init__.py
Svilen-Stefanov: Agree, wrong shape. Dropping `language_id_for` entirely and having `did_open` resolve the id from the suffix, so there's no second lookup and no path-keyed method on the adapter.

Nice side effect: the id then follows the file rather than whichever adapter owns the family, so a `.ts` opened by `JavaScriptAdapter` stops being read as javascript, and #515 doesn't need its adapter-side map either. `language_id` stays, `config_key` still uses it.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -207,7 +207,7 @@ def _bulk_did_open(self, source_files: list[Path]) -> None:
         for i in range(0, total, DID_OPEN_BATCH_SIZE):
             batch = source_files[i : i + DID_OPEN_BATCH_SIZE]
             for file_path in batch:
-                self._lsp.did_open(file_path, self._adapter.language_id)
+                self._lsp.did_open(file_path)
```
Svilen-Stefanov: You were right to poke at it. The argument is gone entirely, this is just `did_open(file_path)` now.

Also dropping the "tests pass" defence, the builder tests use a MagicMock adapter so they only asserted call counts. The new `test_language_id_by_suffix.py` covers the mapping directly.

static_analyzer/__init__.py
```diff
@@ -557,7 +557,7 @@ def notify_file_changed(self, file_path: Path, content: str) -> None:
             adapter = engine_config.adapter
             if suffix in adapter.file_extensions:
                 # Open + change to ensure the server has the latest content
-                client.did_open(file_path, adapter.language_id)
+                client.did_open(file_path)
```
Svilen-Stefanov: Agreed, and this is where it showed worst: the guard above already picked the adapter, then we asked the adapter again. Both the call and `language_id_for` are gone.

static_analyzer/engine/call_graph_builder.py
Svilen-Stefanov, line 210: You were right to point it out. The argument is gone entirely, this is just `did_open(file_path)` now.

Also dropping the "tests pass" defence, the builder tests use a MagicMock adapter so they only asserted call counts. The new `test_language_id_by_suffix.py` covers the mapping directly.

static_analyzer/__init__.py
Svilen-Stefanov, line 560: Agreed, and this is where it showed worst: the guard above already picked the adapter, then we asked the adapter again. Both the call and `language_id_for` are gone.

Svilen-Stefanov (commented):
Both fair, went with your version. Pushed.

`language_id_for(path)` is gone. The id was a pure function of the suffix and `did_open` already had the path, so all five call sites read `did_open(p, adapter.language_id_for(p))`. Now `did_open(file_path)` resolves it itself.

Two enums in `constants.py`. `SourceSuffix` names every suffix, so `LANGUAGE_EXTENSIONS` is written in terms of it and a suffix is never a loose string. `JsxLanguageId` holds only `typescriptreact` and `javascriptreact`, the two ids with no `Language` of their own, and everything else keeps coming from `Language` as you said.

Two consequences worth knowing:

- The adapter files are now untouched vs main, so your comment there is outdated. The id follows the file rather than the adapter, which also makes #515's per-suffix map unnecessary, so I dropped that commit.
- An unopenable suffix now raises instead of quietly taking the adapter's id.

On the internal suite: expected fixture drift, the JSX fix takes excalidraw from 12577 edges to ~6135 and the fixture was pinned pre-fix, CodeBoarding-tests#18 has the refresh. The suite is also gated off for `pull_request`, so the run you saw fell back to engine ref `main`. Re-running it with the stack pinned.

static_analyzer/constants.py
```diff
@@ -26,19 +26,64 @@ class Language(StrEnum):
...
+    CPP = ".cpp"
+    CC = ".cc"
+    CXX = ".cxx"
+    HPP = ".hpp"
+    HH = ".hh"
+    HXX = ".hxx"
+    H = ".h"
+
+
+class JsxLanguageId(StrEnum):
```
Svilen-Stefanov: Here is the enum, in case it's easier to see next to the code. Only these two, since they're the only ids that aren't already a `Language` value.

`SourceSuffix` just below covers the misspelling worry for `.tsx` / `.jsx` and the rest, and `LANGUAGE_EXTENSIONS` is written in terms of it now.

static_analyzer/engine/lsp_client.py
```diff
@@ -266,15 +267,21 @@ def pid(self) -> int | None:
...
+        The languageId follows the file's suffix, not the adapter: one server can serve
+        several dialects, and tsserver mis-parses a .tsx opened as plain "typescript".
+
         Idempotent: silently skips if the file is already open, since the
         LSP spec forbids duplicate didOpen notifications for the same URI.
         """
         uri = file_path.resolve().as_uri()
         if uri in self._opened_uris:
             return
+        language_id = LANGUAGE_ID_BY_SUFFIX.get(file_path.suffix)
```
Svilen-Stefanov: This is the fail-fast bit: a suffix we don't know now raises rather than quietly falling back to the adapter's id. It cost the old `.json` fallback test. Shout if you'd rather it stayed lenient.

ivanmilevtues:
#515 was merged

PR #507, opened 2026-08-20
perf(engine): one engine per language family, not per detected language

Svilen-Stefanov:
Stacked on #506.

The scanner reports TypeScript, TSX, JavaScript and JSX as separate languages, and
`_create_engine_configs` emitted one `EngineConfig` per entry. They all resolve to the same
`typescript-language-server`, so a TypeScript repo started **three** servers over the same files and
clustered the results as separate codebases.

This makes one adapter own the whole family, and gives it the family's whole file set.

## Measured

CodeBoarding-webview (bf351de), static analysis only, one full pass per branch:

| | main | #506 | this branch |
|---|---|---|---|
| engines | 3 | 3 | **1** |
| wall clock | 38.86s | 36.00s | **14.10s** |
| files analysed | 200 | 200 | **215** |
| `typescript` bucket | 3290 nodes / 1688 edges | 3309 / 1956 | **3590 / 2004** |
| `javascript` bucket | 2969 nodes / 1614 edges | 3309 / 1956 | removed |

Two things worth reading off that table.

**On #506 the second bucket is an exact duplicate**, 3309 nodes each, not a degraded copy. Once every
file is opened with the id its suffix implies, the second engine indexes the same files again and
they get clustered as two codebases. On main they differ because that engine read `.ts` under the
wrong dialect.

**The graph grows.** webview's tsconfig sets `allowJs`, but its `include` lists only `**/*.ts` and
`**/*.tsx`, so 14 `.mjs` and 2 `.js` matched no project and were never analysed. `tsc --showConfig`
now runs with `--allowJs`, and a residual sweep adds family files no `include` reaches while leaving
anything the tsconfig excludes alone.

## Two things that had to come along

**A pure-JavaScript repo must keep an engine.** TypeScript only wins the family when TypeScript
exists, so a repo with no `.ts` still gets `JavaScriptAdapter`. Covered by tests for a pure-JS repo,
a JS repo with `jsconfig.json`, and a Python repo.

**The results bucket is keyed by the family.** It used to be keyed by whichever adapter won, so a
repo's graph moved from the `javascript` key to the `typescript` key the moment it gained its first
`.ts`, and the next warm start read the cache under a language it was never written to. The key
cannot move now, and `present_languages()` reads what a bucket actually holds off its files'
suffixes, so a JavaScript-only repo still reports JavaScript.

## Expected downstream churn

Component ids and counts change on **every TypeScript repo**. That is the correction, not a side
effect, but it will read as a large diff in analysis baselines and in the positional tree-diff, so
re-baseline deliberately. The cache tag stays at **v4** for the whole stack, which ships as one
release.

## Do not cherry-pick this without #506

The `javascript` bucket this deletes is almost entirely TypeScript, and on `main` it is the *more*
accurate of the two, because the JavaScript dialect parses JSX correctly while `languageId:
"typescript"` does not. Confirmed edges that exist only in the deleted bucket, unique by call site:

| deleting the bucket… | webview | excalidraw |
|---|---|---|
| on `main` alone | **-397** | **-320** |
| on top of #506 | **0** | -74 |

Real JavaScript is untouched: lodash 4.17.21 (54 `.js`, no TypeScript) produces a byte-identical
graph on `main`, #506 and this branch.

## Tests

`tests/static_analyzer/test_engine_configs.py`, 32 cases: family collapsing, one engine for a TS
repo, pure-JS and jsconfig-JS repos keeping theirs, mixed-family coverage, tsconfig `exclude`
honoured, and the bucket key staying put across a family flip.

Full suite: 2175 passed. The 3 `test_cli_parser` failures are pre-existing on `main` (macOS
`/tmp` vs `/private/tmp`).

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -168,6 +168,13 @@ def _create_engine_configs(
                         f"({len(union)} files across {len(typescript_projects)} tsconfig project(s): "
                         f"{project_dirs})"
                     )
+                    # tsconfig membership is authoritative for what it covers, but it omits
+                    # .js unless ``allowJs`` is set. One adapter owns the whole family, so top
+                    # up with what no project claimed rather than analysing half a mixed repo.
+                    unclaimed = ts_config_scanner.find_unclaimed_family_files(seen)
+                    if unclaimed:
+                        logger.info(f"Adding {len(unclaimed)} family file(s) claimed by no tsconfig")
+                        union.extend(unclaimed)
                     configs.append(EngineConfig(adapter, repository_path, source_files=union))
                 else:
                     logger.info(f"No TypeScript config files found, using repository root for {adapter_name}")
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -96,6 +96,32 @@ def recommended_engine_concurrency(engine_count: int) -> int:
     return max(1, min(engine_count, cpu_bound, memory_bound))
 
 
+def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list[str]:
+    """Deduplicated engine-adapter names for the scanner's detected languages.
+
+    Why: the scanner reports TypeScript, TSX, JavaScript and JSX as separate languages that
+    all share one ``typescript-language-server``. One adapter has to own the whole family, or
+    the same files are indexed once per detected language into separate ``Language`` buckets
+    that are then clustered as if they were different codebases. TypeScript wins the family
+    whenever any TypeScript exists: opening a ``.ts`` under ``languageId: "javascript"`` drops
+    its declarations, while ``"typescript"`` reads a ``.js`` in full.
+    """
+    names: list[str] = []
+    for pl in programming_languages:
+        if not pl.is_supported_lang():
+            logger.warning(f"Unsupported programming language: {pl.language}. Skipping.")
+            continue
+        adapter_name = _lang_to_adapter_name(pl.language)
+        if adapter_name is None:
+            logger.warning(f"No engine adapter for language: {pl.language}. Skipping.")
+            continue
+        if adapter_name not in names:
+            names.append(adapter_name)
+    if "TypeScript" in names and "JavaScript" in names:
+        names.remove("JavaScript")
+    return names
+
+
 def _create_engine_configs(
     programming_languages: list[ProgrammingLanguage],
     repository_path: Path,
@@ -108,19 +134,7 @@ def _create_engine_configs(
     """
     configs: list[EngineConfig] = []
 
-    for pl in programming_languages:
-        if not pl.is_supported_lang():
-            logger.warning(f"Unsupported programming language: {pl.language}. Skipping.")
-            continue
-
-        lang_lower = pl.language.lower()
-
-        # Map CodeBoarding ProgrammingLanguage to engine adapter name
-        adapter_name = _lang_to_adapter_name(pl.language)
-        if adapter_name is None:
-            logger.warning(f"No engine adapter for language: {pl.language}. Skipping.")
-            continue
-
+    for adapter_name in _adapter_names_for(programming_languages):
         try:
             adapter = get_adapter(adapter_name)
         except ValueError:
@@ -128,7 +142,7 @@ def _create_engine_configs(
             continue
 
         try:
-            if lang_lower in (Language.TYPESCRIPT, Language.JAVASCRIPT):
+            if adapter_name in ("TypeScript", "JavaScript"):
                 ts_config_scanner = TypeScriptConfigScanner(repository_path, ignore_manager=ignore_manager)
                 typescript_projects = ts_config_scanner.find_typescript_projects()
 
@@ -159,7 +173,7 @@ def _create_engine_configs(
                     logger.info(f"No TypeScript config files found, using repository root for {adapter_name}")
                     configs.append(EngineConfig(adapter, repository_path))
 
-            elif lang_lower == Language.JAVA:
+            elif adapter_name == "Java":
                 java_config_scanner = JavaConfigScanner(repository_path, ignore_manager=ignore_manager)
                 java_projects = java_config_scanner.scan()
 
@@ -173,7 +187,7 @@ def _create_engine_configs(
                 else:
                     logger.info("No Java projects detected")
 
-            elif lang_lower in (Language.CSHARP, "c#"):
+            elif adapter_name == "CSharp":
                 csharp_scanner = CSharpConfigScanner(repository_path, ignore_manager=ignore_manager)
                 csharp_projects = csharp_scanner.scan()
 
@@ -191,7 +205,7 @@ def _create_engine_configs(
                 configs.append(EngineConfig(adapter, repository_path))
 
         except RuntimeError as e:
-            logger.error(f"Failed to create engine config for {pl.language}: {e}")
+            logger.error(f"Failed to create engine config for {adapter_name}: {e}")
 
     return configs
 
@@ -970,12 +984,22 @@ class StaticAnalyzer:
         self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
     def _loc_for_adapter(self, adapter: LanguageAdapter) -> int:
-        """Return scanner LOC that should have been covered by this adapter."""
+        """Return scanner LOC that should have been covered by this adapter.
+
+        Folds the family the way ``_adapter_names_for`` does, so the JavaScript LOC a
+        TypeScript engine actually reads is credited to it rather than to nobody.
+        """
+        configured = {name.lower() for name in _adapter_names_for(self.programming_langs)}
         adapter_name = adapter.language.lower()
         total = 0
         for pl in self.programming_langs:
             mapped = _lang_to_adapter_name(pl.language)
-            if mapped is not None and mapped.lower() == adapter_name:
+            if mapped is None:
+                continue
+            mapped = mapped.lower()
+            if mapped == "javascript" and "javascript" not in configured:
+                mapped = "typescript"
+            if mapped == adapter_name:
                 total += pl.size
         return total
 
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -213,17 +213,18 @@ def _create_engine_configs(
 def _lang_to_adapter_name(language: str) -> str | None:
     """Map a ProgrammingLanguage name to the engine adapter registry key."""
     mapping: dict[str, str] = {
-        "python": AdapterName.PYTHON,
-        "typescript": AdapterName.TYPESCRIPT,
-        "javascript": AdapterName.JAVASCRIPT,
+        Language.PYTHON: AdapterName.PYTHON,
+        Language.TYPESCRIPT: AdapterName.TYPESCRIPT,
+        Language.JAVASCRIPT: AdapterName.JAVASCRIPT,
+        Language.CSHARP: AdapterName.CSHARP,
+        Language.GO: AdapterName.GO,
+        Language.JAVA: AdapterName.JAVA,
+        Language.PHP: AdapterName.PHP,
+        Language.RUST: AdapterName.RUST,
+        # Scanner spellings with no ``Language`` member of their own.
         "tsx": AdapterName.TYPESCRIPT,
         "jsx": AdapterName.JAVASCRIPT,
         "c#": AdapterName.CSHARP,
-        "csharp": AdapterName.CSHARP,
-        "go": AdapterName.GO,
-        "java": AdapterName.JAVA,
-        "php": AdapterName.PHP,
-        "rust": AdapterName.RUST,
     }
     return mapping.get(language.lower())
 
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -171,7 +171,7 @@ def _create_engine_configs(
                     # tsconfig membership is authoritative for what it covers, but it omits
                     # .js unless ``allowJs`` is set. One adapter owns the whole family, so top
                     # up with what no project claimed rather than analysing half a mixed repo.
-                    unclaimed = ts_config_scanner.find_unclaimed_family_files(seen)
+                    unclaimed = ts_config_scanner.find_unclaimed_family_files(typescript_projects)
                     if unclaimed:
                         logger.info(f"Adding {len(unclaimed)} family file(s) claimed by no tsconfig")
                         union.extend(unclaimed)
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -99,12 +99,9 @@ def recommended_engine_concurrency(engine_count: int) -> int:
 def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list[str]:
     """Deduplicated engine-adapter names for the scanner's detected languages.
 
-    Why: the scanner reports TypeScript, TSX, JavaScript and JSX as separate languages that
-    all share one ``typescript-language-server``. One adapter has to own the whole family, or
-    the same files are indexed once per detected language into separate ``Language`` buckets
-    that are then clustered as if they were different codebases. TypeScript wins the family
-    whenever any TypeScript exists: opening a ``.ts`` under ``languageId: "javascript"`` drops
-    its declarations, while ``"typescript"`` reads a ``.js`` in full.
+    Why: the scanner reports TypeScript, TSX, JavaScript and JSX separately, but one server
+    serves them all, and one adapter per detected name indexes the same files into separate
+    buckets that then cluster as separate codebases.
     """
     names: list[str] = []
     for pl in programming_languages:
@@ -117,6 +114,8 @@ def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list
             continue
         if adapter_name not in names:
             names.append(adapter_name)
+    # TypeScript's adapter already covers the JavaScript suffixes, so a second engine over
+    # the same family would only duplicate it. JavaScript keeps its own when alone.
     if AdapterName.TYPESCRIPT in names and AdapterName.JAVASCRIPT in names:
         names.remove(AdapterName.JAVASCRIPT)
     return names
@@ -1015,9 +1014,8 @@ class StaticAnalyzer:
     def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
         """Whether a cached TypeScript/JavaScript bucket belongs to a language nobody owns now.
 
-        Adding a repository's first ``.ts`` (or deleting its last) flips which adapter owns the
-        family, and the warm start extracts the cached state by the *new* language — finding
-        nothing, and silently rebuilding only the changed files.
+        Why: gaining a first ``.ts`` (or losing the last) flips the owner, and a warm start
+        would then read the cache under a language it was never written to.
         """
         family = {Language.TYPESCRIPT, Language.JAVASCRIPT}
         cached_family = {language for language in cached_results.results if language in family}
@@ -1027,11 +1025,7 @@ class StaticAnalyzer:
         return not (cached_family & live)
 
     def _loc_for_adapter(self, adapter: LanguageAdapter) -> int:
-        """Return scanner LOC that should have been covered by this adapter.
-
-        Folds the family the way ``_adapter_names_for`` does, so the JavaScript LOC a
-        TypeScript engine actually reads is credited to it rather than to nobody.
-        """
+        """Scanner LOC this adapter should have covered, family-folded like ``_adapter_names_for``."""
         configured = {name.lower() for name in _adapter_names_for(self.programming_langs)}
         adapter_name = adapter.language.lower()
         total = 0
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -1001,17 +1001,16 @@ class StaticAnalyzer:
 
     def _loc_for_adapter(self, adapter: LanguageAdapter) -> int:
         """Scanner LOC this adapter should have covered, family-folded like ``_adapter_names_for``."""
-        configured = {name.lower() for name in _adapter_names_for(self.programming_langs)}
-        adapter_name = adapter.language.lower()
+        configured = set(_adapter_names_for(self.programming_langs))
         total = 0
         for pl in self.programming_langs:
             mapped = _lang_to_adapter_name(pl.language)
             if mapped is None:
                 continue
-            mapped = mapped.lower()
-            if mapped == "javascript" and "javascript" not in configured:
-                mapped = "typescript"
-            if mapped == adapter_name:
+            # JavaScript LOC is read by the TypeScript engine whenever that one owns the family.
+            if mapped == AdapterName.JAVASCRIPT and AdapterName.JAVASCRIPT not in configured:
+                mapped = AdapterName.TYPESCRIPT
+            if mapped == adapter.language:
                 total += pl.size
         return total
 
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -11,7 +11,7 @@ from repo_utils.ignore import RepoIgnoreManager
 from static_analyzer.analysis_cache import StaticAnalysisCache
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.cfg import CallGraph
-from static_analyzer.config import Language
+from static_analyzer.config import AdapterName, Language
 from static_analyzer.csharp_config_scanner import CSharpConfigScanner
 from static_analyzer.engine.adapters import get_adapter
 from static_analyzer.engine.call_graph_builder import CallGraphBuilder
@@ -117,8 +117,8 @@ def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list
             continue
         if adapter_name not in names:
             names.append(adapter_name)
-    if "TypeScript" in names and "JavaScript" in names:
-        names.remove("JavaScript")
+    if AdapterName.TYPESCRIPT in names and AdapterName.JAVASCRIPT in names:
+        names.remove(AdapterName.JAVASCRIPT)
     return names
 
 
@@ -142,7 +142,7 @@ def _create_engine_configs(
             continue
 
         try:
-            if adapter_name in ("TypeScript", "JavaScript"):
+            if adapter_name in (AdapterName.TYPESCRIPT, AdapterName.JAVASCRIPT):
                 ts_config_scanner = TypeScriptConfigScanner(repository_path, ignore_manager=ignore_manager)
                 typescript_projects = ts_config_scanner.find_typescript_projects()
 
@@ -173,7 +173,7 @@ def _create_engine_configs(
                     logger.info(f"No TypeScript config files found, using repository root for {adapter_name}")
                     configs.append(EngineConfig(adapter, repository_path))
 
-            elif adapter_name == "Java":
+            elif adapter_name == AdapterName.JAVA:
                 java_config_scanner = JavaConfigScanner(repository_path, ignore_manager=ignore_manager)
                 java_projects = java_config_scanner.scan()
 
@@ -187,7 +187,7 @@ def _create_engine_configs(
                 else:
                     logger.info("No Java projects detected")
 
-            elif adapter_name == "CSharp":
+            elif adapter_name == AdapterName.CSHARP:
                 csharp_scanner = CSharpConfigScanner(repository_path, ignore_manager=ignore_manager)
                 csharp_projects = csharp_scanner.scan()
 
@@ -213,17 +213,17 @@ def _create_engine_configs(
 def _lang_to_adapter_name(language: str) -> str | None:
     """Map a ProgrammingLanguage name to the engine adapter registry key."""
     mapping: dict[str, str] = {
-        "python": "Python",
-        "typescript": "TypeScript",
-        "javascript": "JavaScript",
-        "tsx": "TypeScript",
-        "jsx": "JavaScript",
-        "c#": "CSharp",
-        "csharp": "CSharp",
-        "go": "Go",
-        "java": "Java",
-        "php": "PHP",
-        "rust": "Rust",
+        "python": AdapterName.PYTHON,
+        "typescript": AdapterName.TYPESCRIPT,
+        "javascript": AdapterName.JAVASCRIPT,
+        "tsx": AdapterName.TYPESCRIPT,
+        "jsx": AdapterName.JAVASCRIPT,
+        "c#": AdapterName.CSHARP,
+        "csharp": AdapterName.CSHARP,
+        "go": AdapterName.GO,
+        "java": AdapterName.JAVA,
+        "php": AdapterName.PHP,
+        "rust": AdapterName.RUST,
     }
     return mapping.get(language.lower())
 
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -707,6 +707,15 @@ class StaticAnalyzer:
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
+                    raise StaticAnalysisFatalError(
+                        f"{cache.pkl_path} was written by a different engine version and cannot be "
+                        "reused for an incremental run. Re-run a full analysis to rebuild it."
+                    )
                 logger.info("static_analysis_cache: outcome=miss_absent")
                 results = self._run_full_lsp_pass()
             else:
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -516,7 +516,7 @@ class StaticAnalyzer:
         for engine_config, client in self._live_clients("collect_fresh_diagnostics"):
             diags = client.get_collected_diagnostics()
             if diags:
-                result[engine_config.adapter.language_enum] = diags
+                result[engine_config.adapter.results_language] = diags
         return result
 
     def get_diagnostics_generation(self) -> int:
@@ -727,25 +727,13 @@ class StaticAnalyzer:
                 results = self._run_full_lsp_pass()
             else:
                 cached_results, cached_sha = warm_start
-                if self._family_owner_changed(cached_results):
-                    # A full run rebuilds from scratch and is fine. An incremental one cannot:
-                    # the warm start would keep only the changed files of a bucket nobody owns.
-                    if self.changed_files is not None:
-                        raise StaticAnalysisFatalError(
-                            f"{cache.pkl_path} stores the TypeScript/JavaScript graph under a language "
-                            "no longer configured, so a warm start would rebuild only the changed files "
-                            "and drop the rest. Re-run a full analysis to rebuild it."
-                        )
-                    logger.info("static_analysis_cache: outcome=miss_family_owner_changed")
-                    results = self._run_full_lsp_pass()
-                else:
-                    logger.info(
-                        "static_analysis_cache: outcome=warmstart (cached_sha=%s, current_sha=%s, changes=%s)",
-                        cached_sha,
-                        source_sha or "<none>",
-                        "supplied" if self.changed_files is not None else "git",
-                    )
-                    results = self._update_cached_results(cached_results, cached_sha)
+                logger.info(
+                    "static_analysis_cache: outcome=warmstart (cached_sha=%s, current_sha=%s, changes=%s)",
+                    cached_sha,
+                    source_sha or "<none>",
+                    "supplied" if self.changed_files is not None else "git",
+                )
+                results = self._update_cached_results(cached_results, cached_sha)
 
         self._validate_analysis_results(results)
         results.diagnostics = self.collected_diagnostics
@@ -776,7 +764,7 @@ class StaticAnalyzer:
         def run_one(engine_config: EngineConfig, engine_client: LSPClient | None, order: int = 0) -> None:
             """Analyze one engine. Owns the client's lifetime when given none."""
             adapter, project_path = engine_config.adapter, engine_config.project_path
-            language = adapter.language_enum
+            language = adapter.results_language
             t_lang_start = time.monotonic()
             owned_client: LSPClient | None = None
             try:
@@ -803,7 +791,7 @@ class StaticAnalyzer:
                         status="success",
                         duration_ms=duration_ms,
                         analysis=analysis,
-                        diagnostics=self.collected_diagnostics.get(adapter.language_enum, {}),
+                        diagnostics=self.collected_diagnostics.get(adapter.results_language, {}),
                     )
             except StaticAnalysisFatalError:
                 raise
@@ -896,7 +884,7 @@ class StaticAnalyzer:
         results = StaticAnalysisResults()
         for engine_config, engine_client in self._live_clients("warm-start"):
             adapter, project_path = engine_config.adapter, engine_config.project_path
-            language = adapter.language_enum
+            language = adapter.results_language
             cached_lang_dict = self._extract_language_dict(cached_results, language)
             t_lang_start = time.monotonic()
             changed_files = self._changed_files_for_language(project_path, cached_sha, adapter.language)
@@ -925,7 +913,7 @@ class StaticAnalyzer:
                 status="success",
                 duration_ms=round((time.monotonic() - t_lang_start) * 1000),
                 analysis=analysis,
-                diagnostics=self.collected_diagnostics.get(adapter.language_enum, {}),
+                diagnostics=self.collected_diagnostics.get(adapter.results_language, {}),
             )
         results.incremental_base_results = cached_results
         return results
@@ -1009,20 +997,7 @@ class StaticAnalyzer:
         # Merge, not replace: a monorepo yields several configs per language, and
         # replacing kept only whichever finished last -- which under the concurrency
         # bound is a different one run to run.
-        self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
-
-    def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
-        """Whether a cached TypeScript/JavaScript bucket belongs to a language nobody owns now.
-
-        Why: gaining a first ``.ts`` (or losing the last) flips the owner, and a warm start
-        would then read the cache under a language it was never written to.
-        """
-        family = {Language.TYPESCRIPT, Language.JAVASCRIPT}
-        cached_family = {language for language in cached_results.results if language in family}
-        if not cached_family:
-            return False
-        live = {config.adapter.language_enum for config in self._engine_configs}
-        return not (cached_family & live)
+        self.collected_diagnostics.setdefault(adapter.results_language, {}).update(merged_diags)
 
     def _loc_for_adapter(self, adapter: LanguageAdapter) -> int:
         """Scanner LOC this adapter should have covered, family-folded like ``_adapter_names_for``."""
@@ -1096,7 +1071,7 @@ class StaticAnalyzer:
             adapter = engine_config.adapter
             if adapter.fail_on_empty_symbols is not True:
                 continue
-            language = adapter.language_enum
+            language = adapter.results_language
             source_files = results.get_source_files(language)
             if not source_files:
                 continue
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -720,6 +720,12 @@ class StaticAnalyzer:
                 results = self._run_full_lsp_pass()
             else:
                 cached_results, cached_sha = warm_start
+                if self._family_owner_changed(cached_results):
+                    raise StaticAnalysisFatalError(
+                        f"{cache.pkl_path} stores the TypeScript/JavaScript graph under a language no "
+                        "longer configured, so a warm start would rebuild only the changed files and "
+                        "drop the rest. Re-run a full analysis to rebuild it."
+                    )
                 logger.info(
                     "static_analysis_cache: outcome=warmstart (cached_sha=%s, current_sha=%s, changes=%s)",
                     cached_sha,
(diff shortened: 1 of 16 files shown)


Svilen-Stefanov:
Addressed the review.

**P2 (per-file ids for the expanded family)** is resolved by the parent PR: `notify_file_changed` now calls `language_id_for(file_path)`, so a `.ts`/`.tsx` claimed by the JavaScript adapter is no longer opened as `javascript`.

**Style, fixed.** Adapter import moved to module scope; test module docstring cut to one line.

Svilen-Stefanov:
Second round addressed. Both were real.

**Incompatible cache before incremental (P1).** Correct, and it is an AGENTS.md violation my tag bump introduced: `load_with_sha()` returning `None` sent `analyze()` down `_run_full_lsp_pass()` even for an incremental request, then saved that over the artifact a later run could have reused — so the caller pays for a full analysis *and* hits `IncrementalCacheMissingError` afterwards. An artifact that exists but cannot be read now raises `StaticAnalysisFatalError` naming the remedy, matching the refusal pattern `diagram_generator` already uses.

**JavaScript extensions after the fold (P2).** Also correct. `_try_qualified_name_as_path` derived suffixes from `LANGUAGE_EXTENSIONS[Language(lang)]`, so a `.js` endpoint now stored in the TypeScript bucket was looked up with TypeScript suffixes only. Added `file_extensions_for()` in constants, used by both the resolver and the adapter, so the family question is answered in one place rather than duplicated.

**Style, fixed.** Private fixture moved below the test methods.

On CI: the earlier `review` failure was `HTTP 401` on the OpenRouter key, not this change — #506 passed the same workflow minutes earlier. Worth checking the key or its credit balance.

Svilen-Stefanov:
Third round addressed.

**Family owner flip (P1) — real, and a direct consequence of this PR.** Adding a repo's first `.ts` (or deleting its last) flips which adapter owns the family while the cache tag stays valid, so the warm start accepts it and `_update_cached_results` extracts by the *new* language, finds nothing, and rebuilds only changed files — silently dropping every unchanged file of the other family. Now refused with the remedy named, same pattern as the incompatible-tag case. Three tests: the flip is caught, a matching cache is accepted, a Python-only cache is never a false positive.

**`.ts`/`.js` qualified-name collision (P2) — real, not fixed here, and I want to be straight about why.** `build_qualified_name` strips the suffix, so `src/foo.ts` and `src/foo.js` mint the same name and `register_symbols` overwrites one with the other. This PR does increase exposure: a configless repo now walks both families through one adapter where before they landed in separate buckets.

It is a pre-existing defect rather than a new one, and I measured it before: over CodeBoarding-webview there are **zero** actual instances of a module prefix backed by more than one file. The fix is suffix-disambiguated qualified names, which changes every qname in the graph and needs its own re-baseline — too large to smuggle into a perf fix. Tracked as a separate item rather than left silent.

Svilen-Stefanov:
Fourth round addressed. One was a real bug I introduced in the previous round.

**Full runs rejected by the owner-flip guard (P2, but the most important of the three).** Correct — the guard was unconditional, so `codeboarding full` without `--force`, which warm-starts with `changed_files=None`, was told to "run a full analysis": the operation already in progress. A full run can simply rebuild; only an incremental one has to refuse, because only it would keep just the changed files of a bucket nobody owns. Now scoped to `changed_files is not None`, with the full path falling through to a normal rebuild.

**`unittest.main()` mid-file (P2).** Also real, and I caused it by appending classes after an existing guard. Running the module directly exited before four classes were defined and silently skipped them. Guard moved to the end — direct execution now runs all 15 rather than 11.

**Method-local imports (P1).** Hoisted.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -5,14 +5,24 @@
...
+from static_analyzer.constants import Language, file_extensions_for
 from static_analyzer.engine.language_adapter import LanguageAdapter
 
 
 class TypeScriptAdapter(LanguageAdapter):
 
     _JSX_LANGUAGE_IDS: ClassVar[dict[str, str]] = {".tsx": "typescriptreact", ".jsx": "javascriptreact"}
 
+    @property
+    def file_extensions(self) -> tuple[str, ...]:
```
ivanmilevtues: we have an enum in the constants with these, as per my request in the other PR, we should unify both the js and ts under one entry I think. If you want to make them explicit we can have JS as separate and just append it to the TS (within the constants).

static_analyzer/constants.py
```diff
@@ -41,6 +41,18 @@ class Language(StrEnum):
...
 }
 
+
+def file_extensions_for(language: Language) -> tuple[str, ...]:
+    """Every extension whose symbols may be stored under *language*.
+
+    Why wider than ``LANGUAGE_EXTENSIONS`` for TypeScript: one server owns the whole TS/JS
+    family, so a ``.js`` file's symbols live in the TypeScript bucket.
+    """
+    if language is Language.TYPESCRIPT:
```
ivanmilevtues: i don't like this, seems like patchwork, rather make the LANGAUEG_EXETENSION for TS to be TS + JS

static_analyzer/reference_resolver.py
```diff
@@ -451,7 +451,7 @@ def _try_qualified_name_as_path(
         file_path = qname.replace(".", os.sep)
         full_path = os.path.join(self.repo_dir, file_path)
         file_ref = ".".join(full_path.rsplit(os.sep, 1))
-        language_extensions = LANGUAGE_EXTENSIONS[Language(lang)]
+        language_extensions = file_extensions_for(Language(lang))
```
ivanmilevtues: i prefer the other thing, not the proxy method. pelase update the extensions enums.

static_analyzer/__init__.py
```diff
@@ -693,17 +707,38 @@ def analyze(
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
```
ivanmilevtues, line 721: hmm unsure if we should raise as we have a clean fallback which is also "free" - just run the anlaysis again.

So I suppose I agree with the comment. I think this is a blocker actually as this will fail for all the github action runs I think.

static_analyzer/__init__.py
```diff
@@ -159,7 +173,7 @@ def _create_engine_configs(
                     logger.info(f"No TypeScript config files found, using repository root for {adapter_name}")
                     configs.append(EngineConfig(adapter, repository_path))
 
-            elif lang_lower == Language.JAVA:
+            elif adapter_name == "Java":
```
ivanmilevtues: can we revert to using our constants/enums here

static_analyzer/__init__.py
```diff
@@ -969,13 +1004,37 @@ def _collect_diagnostics_for(self, adapter: LanguageAdapter, engine_client: LSPC
         # bound is a different one run to run.
         self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
+    def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
```
ivanmilevtues: this seems to handle an invalid/old cache which we will anyway regenerate. I think this method is not really needed and it is a weird backward compat thing which will never run? Unsure though you might have to check on this one.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -5,14 +5,24 @@
...
+        """Both families: one tsserver serves them, and ``allowJs`` puts .js in a TS project.
+
+        Why not just the TypeScript set: this adapter owns the whole family, so a narrower
+        tuple makes the incremental pass invalidate a changed ``.js`` and then filter it back
+        out — its nodes are dropped and never rebuilt.
```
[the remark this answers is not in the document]
ivanmilevtues: one line seems better ;d

ivanmilevtues (approved):
So my main concern is the raising when running incremental, that sounds like everything has to explicitly rerun full (which means all github actions will fail).
Apart from that we should continue using our enums/constants and not introduce new inline constant strings and i am suspicuous there is unused code which was intended as backward compat but we won't reach it because of static analysis versioning.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Same thread as the #506 one, agree on the end state, it just isn't a local edit. Doing the fold as its own PR so the `source_inspector` and wrapper bits land in the same commit.

static_analyzer/constants.py
Svilen-Stefanov: Fair, it is a proxy. One thing to carry into the fold: `file_extensions_for` branches on `is Language.TYPESCRIPT` and `Language` is a StrEnum, so a plain str key doesn't raise, it silently returns the narrow tuple. I measured it, `file_extensions_for("typescript")` gives back only the 4 TS suffixes. So the fold should flip that to `==`, otherwise we swap a raise for a wrong answer.

static_analyzer/reference_resolver.py
Svilen-Stefanov: Will do in the fold PR. nit: let's not drop the `Language(lang)` coercion in the same move, see the other comment, `file_extensions_for` uses `is` on a StrEnum so a str slips straight through. Either keep the coercion or switch that branch to `==` in the same edit.

static_analyzer/__init__.py
Svilen-Stefanov: Dug into this one properly. Three things:

1. The failure is real but the tag bump causes it, not this line. `main` is on v3, this stack is on v6, so `load_with_sha()` returns None on any repo with a committed baseline regardless of what we do here. `main`'s own committed `static_analysis.sha` says `v2` against `main`'s `v3` today, so the repo is already a permanent cache miss.
2. The fallback isn't free. On `main` today `_run_full_lsp_pass()` produces results with an empty ClusterCache, so `generate_analysis_incremental` raises `IncrementalCacheMissingError` at `diagram_generator.py:1474` anyway, after paying for the full LSP re-index. And by then `stop_clients()` has flushed a current-tag, cluster-cache-less pkl over the committed one, so every later incremental fails too. That's the wedge AGENTS.md section 5 is about.
3. On the action: `github_action.py:119` builds the generator with no `changes=`, so `changed_files` is None and this line can't fire on the in-repo entry point. The CLI catches broadly and already emits `requiresFullAnalysis: true`. Can you check the published action though? I don't have it checked out here, and if it calls `run_incremental` in-process without a catch it would get an uncaught `StaticAnalysisFatalError`.

Where you're right, and I think this is what your instinct was pointing at: we're leaking a new exception type that no surface maps to "needs full". The wrapper catches only `(BaselineUnavailableError, IncrementalCacheMissingError)` at `analysis_controller.py:436`, so the IDE gets `CoreFailed` instead of `CoreRequiresFull`. Fixing that in core, translating at `run_incremental_workflow` so it comes out as `IncrementalCacheMissingError` with an explicit reason, no wrapper release needed. Plus a test for the stale-tag case.

Two smaller ones I'll fix while in there: the message claims "different engine version" for any unreadable pkl including a corrupt current-tag one, and the guard only covers pkl-and-sha-both-present, so sha-present/pkl-absent still full-passes and then clobbers.

Does that work as the resolution?

static_analyzer/__init__.py
Svilen-Stefanov: Yeah, that reads worse. It changed because we iterate adapter names now rather than languages, so `Language.JAVA` was the wrong type there, but a bare `"Java"` literal isn't better. Adding named constants for the adapter registry keys so `_adapter_names_for`, `_lang_to_adapter_name` and this branch all come from one place. There are already hardcoded `"TypeScript"` / `"JavaScript"` literals at `:121` that should use them too.

static_analyzer/__init__.py
Svilen-Stefanov: Checked, and it does run. It's a same-version flip, which the tag bump can't cover because a bump invalidates old caches and this one was written by this build.

Concretely: repo has only `.js` at tag v6, so the cache is written with a `Language.JAVASCRIPT` bucket, because we only drop JavaScript when TypeScript is also present (`__init__.py:120`). Someone adds the first `.ts`. Next run the tag still matches so `load_with_sha` succeeds, but the live config is TypeScript alone, `_extract_language_dict(cached, TYPESCRIPT)` hits its `except ValueError` arms and hands back an empty CFG, and we rebuild only the changed files. The rest of the graph is silently dropped and re-saved. `_validate_analysis_results` doesn't catch it, it only raises at zero nodes. Reverse case, last `.ts` deleted, is symmetric.

Two things I'll fix while defending it, both fair hits:

- `test_a_full_run_rebuilds_instead_of_refusing` only asserts `analyzer.changed_files is None`, which is a constructor default, not the branch its name claims. No test drives `analyze()` through that path at all. Replacing it with one that does.
- The predicate also fires when a polyglot repo deletes all its TS/JS, which hard-fails an incremental whose only sin was deleting the JS tree. Narrowing it.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Agreed, trimming it.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -4,12 +4,22 @@
...
 
-from static_analyzer.constants import Language
+from static_analyzer.constants import Language, file_extensions_for
 from static_analyzer.engine.language_adapter import LanguageAdapter
 
 
 class TypeScriptAdapter(LanguageAdapter):
 
+    @property
+    def file_extensions(self) -> tuple[str, ...]:
```
Svilen-Stefanov: Agree on the end state, doing the fold as its own PR. It isn't a local edit: `source_inspector.py` builds its parser map by filtering `LANGUAGE_EXTENSIONS` through a factory dict that has JAVASCRIPT and not TYPESCRIPT, so folding js into the TS entry silently drops all four js suffixes and `find_call_sites` starts returning `[]`. The wrapper also picks its tree-sitter grammar off `SOURCE_EXTENSION_TO_LANGUAGE`, and the typescript grammar errors on JSX (checked).

static_analyzer/constants.py
```diff
@@ -86,6 +99,18 @@ class JsxLanguageId(StrEnum):
...
```
Svilen-Stefanov: Fair, it is a proxy. One thing for the fold: it branches on `is Language.TYPESCRIPT` and `Language` is a StrEnum, so a plain str key doesn't raise, it silently returns the narrow tuple. Measured. Worth flipping to `==` in the same edit so we don't swap a raise for a wrong answer.

static_analyzer/reference_resolver.py
Svilen-Stefanov: Will do in the fold PR. nit: let's keep the `Language(lang)` coercion when we do, see the other comment.

static_analyzer/__init__.py
```diff
@@ -159,7 +173,7 @@ def _create_engine_configs(
                     logger.info(f"No TypeScript config files found, using repository root for {adapter_name}")
                     configs.append(EngineConfig(adapter, repository_path))
 
-            elif lang_lower == Language.JAVA:
+            elif adapter_name == AdapterName.JAVA:
```
Svilen-Stefanov: Done. Added `AdapterName` to `constants.py` and used it here, in the registry itself and in the scanner mapping. It can't be `Language` since one adapter owns several.

static_analyzer/__init__.py
Svilen-Stefanov: Dug into this one properly since you called it a blocker.

1. The failure is real but the tag bump causes it, not this line. Main is v3 and this stack is v6, so `load_with_sha()` returns None on any repo with a committed baseline either way. Main's own committed sha says v2 against main's v3, so we're already a permanent cache miss.
2. The fallback isn't free. On main today the full pass yields an empty ClusterCache, so `generate_analysis_incremental` raises `IncrementalCacheMissingError` anyway, after paying for a full LSP re-index, and `stop_clients()` has already flushed a cluster-cache-less pkl over the committed one so every later run fails too. That's the wedge AGENTS.md section 5 is about.
3. `github_action.py:119` builds the generator with no `changes=`, so `changed_files` is None and this can't fire there. Can you check the published action though? I don't have it here.

Where you're right: we leak an exception type nothing maps to "needs full", so the IDE shows `CoreFailed` instead of `CoreRequiresFull`. Fixing that in core so no wrapper release is needed. Does that work as the resolution?

static_analyzer/__init__.py
Svilen-Stefanov: Checked, it does run. It's a same-version flip, which a tag bump can't catch because this cache was written by this build.

A repo with only `.js` writes a JAVASCRIPT bucket, since we drop JavaScript only when TypeScript is present. Add the first `.ts` and the tag still matches, so the cache loads, but the live config is TypeScript alone and `_extract_language_dict` hands back an empty CFG. We then rebuild only the changed files and write the truncated graph back. `_validate_analysis_results` misses it, it only raises at zero nodes.

Two fair hits though: `test_a_full_run_rebuilds_instead_of_refusing` only asserts a constructor default, not the branch its name claims, and the predicate over-fires when a polyglot repo deletes all its TS/JS. Fixing both.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -4,12 +4,22 @@
...
+        """Both families: one tsserver serves them, and ``allowJs`` puts .js in a TS project.
+
+        Why not just the TypeScript set: this adapter owns the whole family, so a narrower
+        tuple makes the incremental pass invalidate a changed ``.js`` and then filter it back
+        out — its nodes are dropped and never rebuilt.
```
Svilen-Stefanov: Agreed, trimmed.

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -4,12 +4,17 @@
...
 
-from static_analyzer.constants import Language
+from static_analyzer.constants import Language, file_extensions_for
 from static_analyzer.engine.language_adapter import LanguageAdapter
 
 
 class TypeScriptAdapter(LanguageAdapter):
 
+    @property
+    def file_extensions(self) -> tuple[str, ...]:
```
Svilen-Stefanov: Agree on the end state, doing the fold as its own PR. It isn't a local edit: `source_inspector.py` builds its parser map by filtering `LANGUAGE_EXTENSIONS` through a factory dict that has JAVASCRIPT and not TYPESCRIPT, so folding js into the TS entry silently drops all four js suffixes and `find_call_sites` starts returning `[]`. The wrapper also picks its tree-sitter grammar off `SOURCE_EXTENSION_TO_LANGUAGE`, and the typescript grammar errors on JSX (checked in the venv).

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -4,12 +4,17 @@
...
+        """Both families: one tsserver serves them, and ``allowJs`` puts .js in a TS project."""
```
Svilen-Stefanov: Trimmed, good call.

static_analyzer/constants.py
```diff
@@ -86,6 +99,18 @@ class JsxLanguageId(StrEnum):
     ),
 }
 
+
+def file_extensions_for(language: Language) -> tuple[str, ...]:
```
Svilen-Stefanov: Fair, it is a proxy. One thing for the fold: it branches on `is Language.TYPESCRIPT` and `Language` is a StrEnum, so a plain str key doesn't raise, it silently returns the narrow tuple. Measured. Worth flipping to `==` in the same edit so we don't swap a raise for a wrong answer.

static_analyzer/reference_resolver.py
Svilen-Stefanov: Will do in the fold PR. nit: let's keep the `Language(lang)` coercion when we do, see the other comment.

static_analyzer/__init__.py
Svilen-Stefanov: Done. `AdapterName` in `constants.py` now backs this, the registry itself and the scanner mapping. It can't be `Language` since one adapter owns several.

static_analyzer/__init__.py
Svilen-Stefanov: Dug into this properly since you called it a blocker.

1. The failure is real but the tag bump causes it, not this line. Main is v3 and this stack is v6, so `load_with_sha()` returns None on any repo with a committed baseline either way. Main's own committed sha says v2 against main's v3, so we're already a permanent cache miss.
2. The fallback isn't free. On main today the full pass yields an empty ClusterCache, so `generate_analysis_incremental` raises `IncrementalCacheMissingError` anyway, after paying for a full LSP re-index, and `stop_clients()` has already flushed a cluster-cache-less pkl over the committed one, so every later run fails too. That's the wedge AGENTS.md section 5 is about.
3. `github_action.py:119` builds the generator with no `changes=`, so `changed_files` is None and this can't fire there. Can you check the published action though? I don't have it checked out here.

Where you're right: we leak an exception type nothing maps to "needs full", so the IDE shows `CoreFailed` instead of `CoreRequiresFull`. Fixing that in core so no wrapper release is needed. Does that work as the resolution?

static_analyzer/__init__.py
Svilen-Stefanov: Checked, it does run. It's a same-version flip, which a tag bump can't catch because this cache was written by this build.

A repo with only `.js` writes a JAVASCRIPT bucket, since we drop JavaScript only when TypeScript is present. Add the first `.ts` and the tag still matches, so the cache loads, but the live config is TypeScript alone and `_extract_language_dict` hands back an empty CFG. We rebuild only the changed files and write the truncated graph back. `_validate_analysis_results` misses it, it only raises at zero nodes.

Two fair hits though: `test_a_full_run_rebuilds_instead_of_refusing` only asserts a constructor default rather than the branch its name claims, and the predicate over-fires when a polyglot repo deletes all its TS/JS. Fixing both.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Agree on the end state, but I want to do the fold in a separate PR rather than here, and it doesn't exist yet, so nothing to look at for now.

Reason it isn't a one-liner: `source_inspector.py` builds its parser map by filtering `LANGUAGE_EXTENSIONS` through a factory dict that has JAVASCRIPT and not TYPESCRIPT, so folding js into the TS entry silently drops all four js suffixes and `find_call_sites` starts returning `[]`. The wrapper also picks its tree-sitter grammar off `SOURCE_EXTENSION_TO_LANGUAGE`, and the typescript grammar errors on JSX (checked in the venv). So it's a cross-repo change plus a cache bump.

Want me to raise a ticket for it so it doesn't get lost?

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Trimmed, good call.

static_analyzer/constants.py
Svilen-Stefanov: Fair, it is a proxy, and it goes when the fold lands.

One thing to carry into that change: it branches on `is Language.TYPESCRIPT` and `Language` is a StrEnum, so a plain str key doesn't raise, it silently returns the narrow tuple. Measured. Worth flipping to `==` in the same edit so we don't swap a raise for a wrong answer.

static_analyzer/reference_resolver.py
Svilen-Stefanov: Same, goes with the fold. nit: let's keep the `Language(lang)` coercion when we do, see the other comment.

static_analyzer/__init__.py
```diff
@@ -191,25 +205,26 @@ def _create_engine_configs(
...
-        "javascript": "JavaScript",
-        "tsx": "TypeScript",
-        "jsx": "JavaScript",
-        "c#": "CSharp",
-        "csharp": "CSharp",
-        "go": "Go",
-        "java": "Java",
-        "php": "PHP",
-        "rust": "Rust",
+        Language.PYTHON: AdapterName.PYTHON,
```
Svilen-Stefanov: Done, both halves. `AdapterName` in `constants.py` now backs the registry, the family fold and this mapping, and the keys here are `Language` wherever one exists so a typo can't silently miss.

The three that stay literals are the ones with no `Language` member: tokei spells them `tsx`, `jsx` and `c#`. Keeping them as strings also makes the point that this dict translates the scanner's vocabulary into ours.

static_analyzer/__init__.py
```diff
@@ -693,17 +708,38 @@ def analyze(
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
```
Svilen-Stefanov: Dug into this properly since you called it a blocker.

1. The failure is real but the tag bump causes it, not this line. Main is v3 and this stack is v6, so `load_with_sha()` returns None on any repo with a committed baseline either way. Main's own committed sha still says v2, so we're already a permanent cache miss.
2. The fallback isn't free, and it's worse than a slow path. The pkl carries the ClusterCache, and that's written by the clustering stage, not the LSP pass, which is why the save is deferred to `stop_clients`. So a fallback full LSP pass produces results with an empty partition, `generate_analysis_incremental` raises at `diagram_generator.py:1474`, and then `__exit__` flushes that partition-less pkl over the committed one, stamped with the new tag. The next incremental now loads it cleanly and raises again, forever, until someone runs a full analysis. Falling back doesn't rescue the run, it destroys the artifact that would have been reusable.
3. `github_action.py:119` builds the generator with no `changes=`, so `changed_files` is None and this can't fire there. Can you check the published action though? I don't have it checked out here.

Where you're right: we leak an exception type nothing maps to "needs full", so the IDE shows `CoreFailed` instead of `CoreRequiresFull`. Fixing that in core so no wrapper release is needed. Does that work as the resolution?

static_analyzer/__init__.py
```diff
@@ -969,13 +1005,37 @@ def _collect_diagnostics_for(self, adapter: LanguageAdapter, engine_client: LSPC
         # bound is a different one run to run.
         self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
+    def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
```
Svilen-Stefanov: Checked, it does run. It's a same-version flip, which a tag bump can't catch because this cache was written by this build.

`_adapter_names_for` drops JavaScript only when TypeScript is also present, so a repo with only `.js` stores its graph under `Language.JAVASCRIPT`. Add the first `.ts` and TypeScript takes the whole family, so the live set is `{TYPESCRIPT}` while the cache still says JAVASCRIPT, and the intersection is empty.

Without the guard the warm start asks for the cached state by the new language, `_extract_language_dict` hits its `except ValueError` arms and hands back an empty CFG, and `update_cfg_for_changed_files` rebuilds only the changed files. The rest of the graph is dropped and saved back. `_validate_analysis_results` doesn't catch it, it only raises at 0 nodes.

Deleting the last `.ts` is the mirror image. A full run doesn't raise here, it just does a full pass.

Two fair hits though: `test_a_full_run_rebuilds_instead_of_refusing` only asserts a constructor default rather than the branch its name claims, and the predicate over-fires when a polyglot repo deletes all its TS/JS, where refusing is wrong because there's no TS/JS engine left to truncate anything. Fixing both.

static_analyzer/__init__.py
```diff
@@ -154,12 +168,19 @@ def _create_engine_configs(
                         f"({len(union)} files across {len(typescript_projects)} tsconfig project(s): "
                         f"{project_dirs})"
                     )
+                    # tsconfig membership is authoritative for what it covers, but it omits
+                    # .js unless ``allowJs`` is set. One adapter owns the whole family, so top
+                    # up with what no project claimed rather than analysing half a mixed repo.
+                    unclaimed = ts_config_scanner.find_unclaimed_family_files(seen)
```
Svilen-Stefanov: This is the fix for the thing your "one engine per family" question exposed. tsconfig stays authoritative for what it claims; we just stop dropping what it structurally cannot claim.

Checked that tsserver actually serves the added files: it puts them in an inferred project, so no `allowJs` injection is needed. `.tsx` and `.jsx` declarations produce nodes and cross-family edges. Separately, `export const X = () => ...` produces no node in any file type, which is a pre-existing shape gap rather than a family one.

static_analyzer/typescript_config_scanner.py
```diff
@@ -85,6 +85,19 @@ def find_typescript_projects(self) -> list[TypeScriptProject]:
             )
         return projects
 
+    def find_unclaimed_family_files(self, claimed: set[Path]) -> list[Path]:
```
Svilen-Stefanov: The walk is ignore-filtered, so `node_modules` stays out. Worth knowing it inherits the `.codeboardingignore`-replaces-defaults behaviour, so a repo with a custom ignore file that drops `node_modules` would pull it in here as it already would in `_fallback_walk`.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Agree on the end state, but I want the fold in its own PR rather than here, and it doesn't exist yet, so there's nothing to look at for now.

It isn't a one-liner: `source_inspector.py` builds its parser map by filtering `LANGUAGE_EXTENSIONS` through a factory dict that has JAVASCRIPT and not TYPESCRIPT, so folding js into the TS entry silently drops all four js suffixes and `find_call_sites` starts returning `[]`. The wrapper also picks its tree-sitter grammar off `SOURCE_EXTENSION_TO_LANGUAGE`, and the typescript grammar errors on JSX. Cross-repo plus a cache bump.

Want me to raise a ticket so it doesn't get lost?

static_analyzer/constants.py
Svilen-Stefanov: Fair, it is a proxy, and it goes when the fold lands. One thing to carry into that change: it branches on `is Language.TYPESCRIPT` and `Language` is a StrEnum, so a plain str key doesn't raise, it silently returns the narrow tuple. Measured.

static_analyzer/reference_resolver.py
Svilen-Stefanov: Same, goes with the fold. nit: let's keep the `Language(lang)` coercion when we do, see the other comment.

static_analyzer/__init__.py
```diff
@@ -191,25 +212,26 @@ def _create_engine_configs(
...
-        "javascript": "JavaScript",
-        "tsx": "TypeScript",
-        "jsx": "JavaScript",
-        "c#": "CSharp",
-        "csharp": "CSharp",
-        "go": "Go",
-        "java": "Java",
-        "php": "PHP",
-        "rust": "Rust",
+        Language.PYTHON: AdapterName.PYTHON,
```
Svilen-Stefanov: Done, both halves. `AdapterName` now backs the registry, the family fold and this mapping, and the keys are `Language` wherever one exists so a typo can't silently miss.

Only `tsx`, `jsx` and `c#` stay literals, since those have no `Language` member.

static_analyzer/__init__.py
```diff
@@ -693,17 +715,38 @@ def analyze(
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
```
Svilen-Stefanov: Rewriting this one, I argued the weaker position first time.

You were right that the fallback is where the problem is, but the defect isn't that we fall back, it's that we **persist a half-finished artifact**. `get_static_analysis` wraps only `analyze()`, so the `with` block exits normally and flushes the pkl before the incremental step ever runs. The pkl carries the ClusterCache, and clustering hasn't happened yet, so what lands on disk is a partition-less pkl stamped with the current tag. The next run then loads it cleanly and fails the cluster check, forever. Verified: save a result with no clustering, `load_with_sha()` succeeds, `all_cluster_ids()` is empty, `IncrementalCacheMissingError` again.

So one bad write turns a one-time cache miss into a permanent one, which is most likely what's behind the volume of these errors in the logs.

What I think we should actually do, rather than this raise:
1. Stop writing a partition-less artifact over a good one. That alone removes the permanence.
2. Make the fallback complete the full pipeline, clustering included, and say loudly that component ids get renumbered.
3. Keep a hard failure only for when that fails, naming the step to take.

AGENTS.md section 5 is about a *silent* full run, so an announced one doesn't conflict with it.

I've left the raise in for now because it's still better than the silent wedge, but it's a backstop, not the fix. Two open questions before I do the above: is a fallback full run acceptable in CI given the cost, and is renumbering components tolerable there? Also, `github_action.py:119` passes no `changes=` so this line can't fire on the in-repo entry point, but I can't see the published action from here, could you check whether it calls `run_incremental` in-process without a catch?

static_analyzer/__init__.py
```diff
@@ -969,13 +1012,37 @@ def _collect_diagnostics_for(self, adapter: LanguageAdapter, engine_client: LSPC
         # bound is a different one run to run.
         self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
+    def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
```
Svilen-Stefanov: Checked, it does run. It's a same-version flip, which a tag bump can't catch because the cache was written by this build.

`_adapter_names_for` drops JavaScript only when TypeScript is present, so a repo with only `.js` stores its graph under `Language.JAVASCRIPT`. Add the first `.ts` and TypeScript takes the family, so the live set is `{TYPESCRIPT}` while the cache still says JAVASCRIPT, and the intersection is empty.

Without the guard the warm start asks for the cached state by the new language, `_extract_language_dict` hits its `except ValueError` arms and returns an empty CFG, and only the changed files get rebuilt. The rest is dropped and saved back. `_validate_analysis_results` misses it, it only raises at 0 nodes.

Two fair hits: the test named `test_a_full_run_rebuilds_instead_of_refusing` only asserts a constructor default, and the predicate over-fires when a polyglot repo deletes all its TS/JS, where refusing is wrong. Fixing both.

static_analyzer/constants.py
```diff
@@ -68,7 +81,18 @@ class JsxLanguageId(StrEnum):
 # the same edit. ``mypy`` enforces total coverage via the assertion below.
 LANGUAGE_EXTENSIONS: dict[Language, tuple[SourceSuffix, ...]] = {
     Language.PYTHON: (SourceSuffix.PY,),
-    Language.TYPESCRIPT: (SourceSuffix.TS, SourceSuffix.TSX, SourceSuffix.MTS, SourceSuffix.CTS),
+    # TypeScript carries the JavaScript suffixes too: one tsserver serves the whole family,
+    # and ``allowJs`` puts .js inside a TypeScript project.
+    Language.TYPESCRIPT: (
```
Svilen-Stefanov: Done as you asked. The suffixes live on the TypeScript entry now, so `file_extensions_for` and the adapter's `file_extensions` override are both gone and `reference_resolver` reads `LANGUAGE_EXTENSIONS` directly again.

JavaScript keeps its own entry so a repo with no TypeScript still gets its adapter. Everything derived from the table is unchanged: `SOURCE_EXTENSION_TO_LANGUAGE`, `LANGUAGE_ID_BY_SUFFIX` and the tree-sitter parser map all come out byte-identical.

static_analyzer/__init__.py
```diff
@@ -96,6 +96,31 @@ def recommended_engine_concurrency(engine_count: int) -> int:
...
+        adapter_name = _lang_to_adapter_name(pl.language)
+        if adapter_name is None:
+            logger.warning(f"No engine adapter for language: {pl.language}. Skipping.")
+            continue
+        if adapter_name not in names:
+            names.append(adapter_name)
+    # TypeScript's adapter already covers the JavaScript suffixes, so a second engine over
+    # the same family would only duplicate it. JavaScript keeps its own when alone.
+    if AdapterName.TYPESCRIPT in names and AdapterName.JAVASCRIPT in names:
+        names.remove(AdapterName.JAVASCRIPT)
```
Svilen-Stefanov: Added a comment. It's needed: the TypeScript adapter covers the JavaScript suffixes, so a second engine over the same family only duplicates it, which is what main does. JavaScript keeps its own adapter when there's no TypeScript.

static_analyzer/__init__.py
```diff
@@ -191,25 +211,26 @@ def _create_engine_configs(
...
-        "javascript": "JavaScript",
-        "tsx": "TypeScript",
-        "jsx": "JavaScript",
-        "c#": "CSharp",
-        "csharp": "CSharp",
-        "go": "Go",
-        "java": "Java",
-        "php": "PHP",
-        "rust": "Rust",
+        Language.PYTHON: AdapterName.PYTHON,
```
Svilen-Stefanov: Keys are `Language` wherever one exists, so a typo can't silently miss. Only `tsx`, `jsx` and `c#` stay literals, since those have no `Language` member.

static_analyzer/typescript_config_scanner.py
```diff
@@ -122,13 +183,20 @@ def _resolve_project_files(
...
+                kept = [f for f in files if f.suffix in _ALL_EXTENSIONS and not self.ignore_manager.should_ignore(f)]
+                return kept, excluded
             # tsc available but failed for this project — try the FS walk.
-        return self._fallback_walk(project_dir, all_candidates)
+        return self._fallback_walk(project_dir, all_candidates), []
 
     def _showconfig(self, project_dir: Path, tsc_cmd_prefix: list[str]) -> dict | None:
-        """Invoke ``tsc --showConfig -p <dir>`` and return parsed JSON."""
-        cmd = [*tsc_cmd_prefix, "-p", str(project_dir)]
+        """Invoke ``tsc --showConfig -p <dir> --allowJs`` and return parsed JSON.
```
Svilen-Stefanov: This is what makes a mixed repo work. `--showConfig` reports only the TypeScript unless the flag is set, so on main a repo with both families had its JavaScript analysed by a second, degraded engine, and on a repo with a tsconfig covering only `src` the JS was simply absent.

The flag overrides the file's own `allowJs` while `include`, `exclude`, `files` and `references` are all still honoured, so tsc stays the authority on membership rather than us re-implementing its globs.

static_analyzer/typescript_config_scanner.py
```diff
@@ -85,6 +124,27 @@ def find_typescript_projects(self) -> list[TypeScriptProject]:
             )
         return projects
 
+    def find_unclaimed_family_files(self, projects: list[TypeScriptProject]) -> list[Path]:
```
Svilen-Stefanov: For files no tsconfig `include` reaches, which main picked up only by accident and without edges. A path the tsconfig actively excludes stays out.

Measured on a repo whose tsconfig covers `src` only, with a `.js` and a `.ts` in `scripts/` calling into it: main analysed 4 of 5 files, missed the out-of-include `.ts` entirely and produced 1 edge; this produces all 5 files and 4 edges.

static_analyzer/__init__.py
```diff
@@ -693,17 +714,38 @@ def analyze(
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
```
Svilen-Stefanov: This raise is a backstop, not the fix, and I'd like your call on the fix.

The pkl carries the ClusterCache, and that comes from clustering rather than the LSP pass. `get_static_analysis` wraps only `analyze()`, so on a tag change the `with` block exits normally and writes a partition-less pkl stamped with the current tag before the incremental step runs. The next run loads it cleanly and fails the cluster check, and so does every run after that. Verified: save a result with no clustering, `load_with_sha()` succeeds, `all_cluster_ids()` is empty.

So one bad write turns a one-time cache miss into a permanent one, which is likely most of the `IncrementalCacheMissingError` volume in the logs.

What I think we should do instead of relying on this raise:
1. Stop writing a partition-less artifact over a good one.
2. Let the fallback finish the full pipeline, clustering included, and say loudly that component ids get renumbered.
3. Keep a hard failure only for when that fails, naming the step to take.

Should we do this here, in a separate PR or something else? Maybe I can just remove the error here and address in a separate PR?

static_analyzer/__init__.py
```diff
@@ -969,13 +1011,32 @@ def _collect_diagnostics_for(self, adapter: LanguageAdapter, engine_client: LSPC
         # bound is a different one run to run.
         self.collected_diagnostics.setdefault(adapter.language_enum, {}).update(merged_diags)
 
+    def _family_owner_changed(self, cached_results: StaticAnalysisResults) -> bool:
```
Svilen-Stefanov: Worth keeping, and it's independent of the cache-save problem in the other thread.

A repo with only `.js` stores its graph under `Language.JAVASCRIPT`. Add the first `.ts` and TypeScript takes the family, so the live set is `{TYPESCRIPT}` while the cache still says JAVASCRIPT. Same tag, same build, so no version check catches it. Without the guard the warm start reads the cache under the new language, gets an empty CFG, rebuilds only the changed files and writes that truncated graph back. `_validate_analysis_results` misses it because it only raises at 0 nodes.

One thing I still owe: it over-fires when a polyglot repo deletes all its TS/JS, where refusing is wrong because there's no TS/JS engine left to truncate anything.

static_analyzer/config.py
```diff
@@ -68,7 +81,18 @@ class JsxLanguageId(StrEnum):
 # the same edit. ``mypy`` enforces total coverage via the assertion below.
 LANGUAGE_EXTENSIONS: dict[Language, tuple[SourceSuffix, ...]] = {
     Language.PYTHON: (SourceSuffix.PY,),
-    Language.TYPESCRIPT: (SourceSuffix.TS, SourceSuffix.TSX, SourceSuffix.MTS, SourceSuffix.CTS),
+    # TypeScript carries the JavaScript suffixes too: one tsserver serves the whole family,
+    # and ``allowJs`` puts .js inside a TypeScript project.
+    Language.TYPESCRIPT: (
```
Svilen-Stefanov, line 87: Done as you asked. The suffixes live on the TypeScript entry, so `file_extensions_for` and the adapter's `file_extensions` override are both gone and `reference_resolver` reads `LANGUAGE_EXTENSIONS` directly again.

JavaScript keeps its own entry so a repo with no TypeScript still gets its adapter. Everything derived from the table is byte-identical: `SOURCE_EXTENSION_TO_LANGUAGE`, `LANGUAGE_ID_BY_SUFFIX` and the tree-sitter parser map.

static_analyzer/__init__.py
Svilen-Stefanov, line 120: Commented. It is needed: the TypeScript adapter covers the JavaScript suffixes, so a second engine over the same family only duplicates it, and on #506 that duplicate is exact, 3309 nodes each. JavaScript keeps its own adapter when there is no TypeScript.

static_analyzer/__init__.py
Svilen-Stefanov, line 222: Keys are `Language` wherever one exists, so a typo cannot silently miss. Only `tsx`, `jsx` and `c#` stay literals, since those are scanner spellings with no `Language` member. Those three are the last language strings in the engine.

static_analyzer/typescript_config_scanner.py
Svilen-Stefanov, line 162: This is what makes a mixed repo work. `--showConfig` reports only the TypeScript unless the flag is set, so a repo with both families had its JavaScript analysed by a second engine on main, and simply absent once that engine goes.

The flag overrides the file's own `allowJs` while `include`, `exclude`, `files` and `references` are all still honoured, so tsc stays the authority on membership rather than us re-implementing its globs.

static_analyzer/typescript_config_scanner.py
Svilen-Stefanov, line 102: For files no tsconfig `include` reaches, which main picked up only by accident and without edges. A path the tsconfig actively excludes stays out.

Measured on a fixture whose tsconfig covers `src` only, with a `.js` and a `.ts` in `scripts/` calling into it: main analysed 4 of 5 files, missed the out-of-include `.ts` entirely and produced 1 edge; this produces all 5 and 4 edges. On webview it is the 16 `.mjs`/`.js` its `include` never matched.

static_analyzer/engine/language_adapter.py
```diff
@@ -41,6 +41,16 @@ def language_enum(self) -> Language:
         per-language extension set lives in exactly one place.
         """
 
+    @property
+    def results_language(self) -> Language:
```
Svilen-Stefanov, line 45: This is the answer to the warm-start problem rather than the guard that used to detect it.

The bucket used to be keyed by whichever adapter won the family, so a repo's graph moved from the javascript key to the typescript key the moment it gained its first `.ts`, and the next warm start read the cache under a language it was never written to. The key is the family now, so it cannot move.

static_analyzer/analysis_result.py
```diff
@@ -297,6 +311,34 @@ def get_languages(self) -> list[Language]:
         """Return the list of languages for which any data has been recorded."""
         return list(self.results)
 
+    def present_languages(self) -> set[Language]:
```
Svilen-Stefanov, line 313: And this is how the cache still says what it holds. `source_languages()` reads the languages off the suffixes of the files stored in a bucket, `present_languages()` is the union across buckets, so a JavaScript-only repo reports JavaScript and one holding both reports both. Gaining a first `.ts` shows up as a new language instead of a mismatch.

Reads resolve through the family, so callers keep asking for the language they mean.

static_analyzer/__init__.py
```diff
@@ -693,6 +714,15 @@ def analyze(
         else:
             warm_start = cache.load_with_sha()
             if warm_start is None:
+                # An artifact that is present but unreadable means an engine version whose
+                # graph this build would not reproduce. Refuse rather than quietly running a
+                # full pass: the caller asked for incremental, and the full result would
+                # overwrite the very artifact a later run could have reused.
+                if self.changed_files is not None and cache.pkl_path.exists() and cache.sha_path.exists():
```
Svilen-Stefanov, line 721: This raise is a backstop, not the fix, and I would like your call on the fix.

The pkl carries the ClusterCache, and that comes from clustering rather than the LSP pass. `get_static_analysis` wraps only `analyze()`, so on a tag change the `with` block exits normally and writes a partition-less pkl stamped with the current tag before the incremental step runs. The next run loads it cleanly and fails the cluster check, and so does every run after that. Verified: save a result with no clustering, `load_with_sha()` succeeds, `all_cluster_ids()` is empty.

So one bad write turns a one-time cache miss into a permanent one, which is likely most of the `IncrementalCacheMissingError` volume in the logs.

What I think we should do instead of relying on this raise:
1. Stop writing a partition-less artifact over a good one.
2. Let the fallback finish the full pipeline, clustering included, and say loudly that component ids get renumbered.
3. Keep a hard failure only for when that fails, naming the step to take.

Should we remove the raised error here and open a separate PR to address the above?

Svilen-Stefanov (commented):
Took the enums. One thing I want your call on, at the bottom.

Against main: one engine per TS/JS family instead of one per detected language, the family's whole file set instead of whatever a tsconfig happened to claim, and a results bucket keyed by the family so it cannot move when a repo gains its first `.ts`.

Re-measured on webview (bf351de) after the review changes: 3 engines to 1, 38.86s to 14.10s, and 200 files to 215 because its tsconfig `include` lists only `**/*.ts` and `**/*.tsx`, so 14 `.mjs` and 2 `.js` were never analysed. Worth noting the second bucket on #506 is an exact duplicate of the first, 3309 nodes each, not a degraded copy.

Cache tag stays at v4 for the whole stack, since it ships as one release.

ivanmilevtues:
#515 was merged

PR #508, opened 2026-08-20
fix(engine): credit a call to the declaration a reader would name

Svilen-Stefanov:
Stacked on #507.

An LSP server reports inline callbacks and anonymous functions as symbols, and `find_containing_symbol` returns the innermost one. So a call written inside an arrow callback is credited to `listReviewComments.raw.ghPaginate() callback` rather than to `listReviewComments`.

On CodeBoarding-webview that is **499 of 1923 TypeScript edges — 25.9%** — whose source is a name nobody uses. It shows up in the graph, in the diagram labels, and in the CFG evidence handed to the model (`forEach() callback -> highlightCode`).

Worth noting: this did **not** shrink after the JSX language-id fix, contrary to what I expected when planning. #506 removed synthetic *targets* (46 → 6); synthetic *sources* stayed at 25.9%. I re-measured before writing this rather than assuming.

## The obvious fix is wrong, and I measured it first

Teaching `lift_to_callable` to walk past synthetic callables looks fine on aggregate counts but **silently destroys grounding**. `container` isn't only the credit line — it's also the input to the declaration-line guard. Once it becomes `listReviewComments.raw`, whose `start_line` equals the reference line (`const raw = await ghPaginate(...)`), the reference is dropped. That version loses 30 grounded call sites and takes `ghPaginate` from 12 callers down to 3.

So: **every guard keeps seeing the innermost container, and only the recorded source name rolls up**, via a new `SymbolTable.attribution_symbol`.

## Measured — it is a pure relabel

| | before | after |
|---|---|---|
| synthetic edge sources | 499 (25.9%) | **17 (1.0%)** |
| call sites | 2324 | **2324** (unchanged) |
| grounding `(target, line, col)` | 2298 | **2298** (3 lost, 3 gained) |
| nodes | 3243 | 3242 |
| edges | 1923 | 1758 |
| `ghPaginate` callers | 12, of which 2 named | **10, all 10 named** |

Edges drop because callbacks merge into their parent — two callbacks in one function calling the same target become one edge carrying both call sites, which is why the call-site count is unchanged.

## Why nodes are untouched

Nodes are clustering input. Deleting 1198 of 3243 would move the partition for reasons that have nothing to do with the partition being wrong, and the CONTAINS reference edges tying a callback to its parent would all need rewriting. The symbol table, `methods_index` and those reference edges all keep the callback; only the credit line on a call edge moves.

Applied to **both** edge strategies, so Java and C# get it too.

## Tidying

`_CALLBACK_PATTERNS` in `node.py` had no production caller and duplicated the marker list. Both now read one tuple in `constants.py`.

## Tests

`tests/static_analyzer/test_attribution_symbol.py`, 6 cases: a named symbol credits itself, a callback credits its enclosing function, a promoted `const` wrapper is stepped past, the innermost *named* enclosure wins, a callback nested in a callback still reaches the function, and a module-level anonymous function credits itself.

Full suite: 2158 passed. The 3 `test_cli_parser` failures are pre-existing on `main`. Tag bumped to v6.

### DIFF
--- a/static_analyzer/engine/edge_builder.py
+++ b/static_analyzer/engine/edge_builder.py
@@ -521,14 +521,18 @@ def _resolve_definitions(
                         if adapter.is_class_like(parent_kind):
                             parent_qname = parent_qualified_name(target.qualified_name)
                             parent_sym = st.symbols.get(parent_qname)
-                            if parent_sym is not None and _is_valid_edge(caller, parent_sym):
-                                _add_edge_call_site(edge_set, caller.qualified_name, parent_qname, call_site)
+                            if (
+                                parent_sym is not None
+                                and _is_valid_edge(caller, parent_sym)
+                                and _is_valid_edge(attributed, parent_sym)
+                            ):
+                                _add_edge_call_site(edge_set, attributed.qualified_name, parent_qname, call_site)
 
                     # Queue implementation query for polymorphic dispatch
                     if adapter.is_callable(target.kind):
                         impl_queries_pending.append(
                             ImplementationQuery(
-                                caller_qname=caller.qualified_name,
+                                caller_qname=attributed.qualified_name,
                                 target_file=target.file_path,
                                 target_line=target.start_line,
                                 target_char=target.start_char,
--- a/static_analyzer/engine/edge_builder.py
+++ b/static_analyzer/engine/edge_builder.py
@@ -290,7 +290,12 @@ def _process_references_for_position(
                     continue
             if sym.qualified_name.startswith(container.qualified_name + "."):
                 continue
-            _add_edge_site(edge_set, container.qualified_name, sym.qualified_name, ref_file, ref_line, ref_char)
+            # Every guard above sees the innermost container, which is what decides whether
+            # this reference is an edge at all. Only the credit line is rolled up.
+            attributed = st.attribution_symbol(container).qualified_name
+            if attributed == sym.qualified_name or sym.qualified_name.startswith(attributed + "."):
+                continue
+            _add_edge_site(edge_set, attributed, sym.qualified_name, ref_file, ref_line, ref_char)
 
     return refs_total, refs_call_sites
 
@@ -380,12 +385,15 @@ def _resolve_iterated_types(
                     if target is None or not _is_valid_edge(caller, target):
                         continue
                     resolved += 1
-                    _add_edge_call_site(edge_set, caller.qualified_name, target.qualified_name, site)
+                    attributed = st.attribution_symbol(caller)
+                    if not _is_valid_edge(attributed, target):
+                        continue
+                    _add_edge_call_site(edge_set, attributed.qualified_name, target.qualified_name, site)
                     # The loop calls the enumerator, so name it too when the
                     # type declares one rather than inheriting it.
                     for enumerator in _members_named(target, st, "GetEnumerator"):
-                        if _is_valid_edge(caller, enumerator):
-                            _add_edge_call_site(edge_set, caller.qualified_name, enumerator.qualified_name, site)
+                        if _is_valid_edge(caller, enumerator) and _is_valid_edge(attributed, enumerator):
+                            _add_edge_call_site(edge_set, attributed.qualified_name, enumerator.qualified_name, site)
     return resolved
 
 
@@ -489,16 +497,23 @@ def _resolve_definitions(
                     if not _is_valid_edge(caller, target):
                         continue
 
-                    _add_edge_call_site(edge_set, caller.qualified_name, target.qualified_name, call_site)
+                    # Guards keep the innermost caller; only the credit line rolls up.
+                    attributed = st.attribution_symbol(caller)
+                    if not _is_valid_edge(attributed, target):
+                        continue
+
+                    _add_edge_call_site(edge_set, attributed.qualified_name, target.qualified_name, call_site)
 
                     for override in _override_targets(target, st, dispatch):
-                        if _is_valid_edge(caller, override):
-                            _add_edge_call_site(edge_set, caller.qualified_name, override.qualified_name, call_site)
+                        if _is_valid_edge(caller, override) and _is_valid_edge(attributed, override):
+                            _add_edge_call_site(edge_set, attributed.qualified_name, override.qualified_name, call_site)
 
                     if (call_site.lsp_line, call_site.lsp_column) in collection_positions:
                         for adder in _members_named(target, st, "Add"):
-                            if _is_valid_edge(caller, adder):
-                                _add_edge_call_site(edge_set, caller.qualified_name, adder.qualified_name, call_site)
+                            if _is_valid_edge(caller, adder) and _is_valid_edge(attributed, adder):
+                                _add_edge_call_site(
+                                    edge_set, attributed.qualified_name, adder.qualified_name, call_site
+                                )
 
                     # If target is a callable with a class-like parent, also add edge to the parent class
                     if adapter.is_callable(target.kind) and target.parent_chain:
--- a/static_analyzer/engine/models.py
+++ b/static_analyzer/engine/models.py
@@ -23,6 +23,9 @@ class SymbolInfo:
     end_line: int
     end_char: int
     parent_chain: list[tuple[str, int]] = field(default_factory=list)
+    # Arrived as VARIABLE/CONSTANT and was promoted to CLASS for having callable children.
+    # A `const raw = await fn(() => ...)` wrapper is a scope, not a name a reader would use.
+    promoted_from_variable: bool = False
 
     @property
     def definition_location(self) -> tuple[str, int, int]:
--- a/static_analyzer/engine/models.py
+++ b/static_analyzer/engine/models.py
@@ -23,8 +23,8 @@ class SymbolInfo:
     end_line: int
     end_char: int
     parent_chain: list[tuple[str, int]] = field(default_factory=list)
-    # Arrived as VARIABLE/CONSTANT and was promoted to CLASS for having callable children.
-    # A `const raw = await fn(() => ...)` wrapper is a scope, not a name a reader would use.
+    # Promoted to CLASS for having callable children: a namespace object, or a
+    # `const raw = await fn(() => ...)` wrapper that is a scope rather than a name.
     promoted_from_variable: bool = False
 
     @property
--- a/static_analyzer/engine/symbol_table.py
+++ b/static_analyzer/engine/symbol_table.py
@@ -280,18 +280,27 @@ class SymbolTable:
 
         best: SymbolInfo | None = None
         best_size = float("inf")
+        # A promoted `const x = ...` wrapper is a name a reader recognises, unlike an anonymous
+        # callback, so it is a usable credit — but only when nothing named encloses it. Module
+        # level is exactly that case: there is no function above it to prefer.
+        fallback: SymbolInfo | None = None
+        fallback_size = float("inf")
         for other in self._file_symbols.get(str(sym.file_path), []):
-            if other.qualified_name == sym.qualified_name or self._is_unnameable(other):
+            if other.qualified_name == sym.qualified_name or self._is_anonymous(other):
                 continue
             if not (self._naming.is_callable(other.kind) or self._naming.is_class_like(other.kind)):
                 continue
             if not self._encloses(other, sym):
                 continue
             size = (other.end_line - other.start_line) * 10000 + (other.end_char - other.start_char)
-            if size < best_size:
+            if other.promoted_from_variable:
+                if size < fallback_size:
+                    fallback = other
+                    fallback_size = size
+            elif size < best_size:
                 best = other
                 best_size = size
-        return best or sym
+        return best or fallback or sym
 
     def get_equivalent_names(self, qualified_name: str) -> list[str]:
         """Get equivalent symbol names for edge expansion using pre-built index."""
@@ -357,7 +366,12 @@ class SymbolTable:
         return False
 
     def _is_unnameable(self, sym: SymbolInfo) -> bool:
-        return sym.promoted_from_variable or any(marker in sym.name for marker in ANONYMOUS_SYMBOL_MARKERS)
+        """Whether this symbol's own name is not one a reader would use as a caller."""
+        return sym.promoted_from_variable or self._is_anonymous(sym)
+
+    @staticmethod
+    def _is_anonymous(sym: SymbolInfo) -> bool:
+        return any(marker in sym.name for marker in ANONYMOUS_SYMBOL_MARKERS)
 
     @staticmethod
     def _encloses(outer: SymbolInfo, inner: SymbolInfo) -> bool:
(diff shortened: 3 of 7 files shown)


Svilen-Stefanov:
Addressed the review. Two were real.

**Character bounds.** `attribution_symbol` compared only line numbers, so several declarations sharing a line let an unrelated neighbour win. It now applies the same bounds `find_containing_symbol` does, via a shared `_encloses` helper. No change to webview's output (no same-line collisions there), but it was wrong.

**Attributed caller not carried all the way.** The direct, override and collection edges used `attributed`, but the parent-class edge and the queued `ImplementationQuery` still carried the callback name — so constructor and polymorphic calls from inside a callback kept producing edges from the anonymous node. Both fixed.

**Tag enforcement on un-gated reads.** Correct: `_get_unlocked` skipped `_read_tag_sha_unlocked` entirely when `expected_sha is None`, so health and status consumers kept unpickling a v5 artifact after the v6 bump. Added `_tag_version_is_stale` — only a tag that exists and disagrees rejects, so the deliberate untagged legacy path still reads.

**Style, fixed.** Private helpers moved below the public API, docstring condensed.

Svilen-Stefanov:
Ordering fixed in both places — `_tag_version_is_stale` moved below the public cache API, `_load` moved below the test methods.

Svilen-Stefanov:
Third round addressed.

**Promoted module declarations as fallback (P2) — real gap in my own logic.** `_is_unnameable` excluded every promoted symbol from being *chosen*, so a module-level `const raw = await ghPaginate(() => ...)` had no enclosing candidate at all and the callback credited itself — precisely the synthetic source this PR removes everywhere else.

Split the predicate: a promoted wrapper carries a real name, so it is a usable credit when nothing better encloses, while a named function still wins. Synthetic sources on webview **17 → 14**. Two tests: module-level wrapper used, named function still preferred over a wrapper.

static_analyzer/engine/edge_builder.py
```diff
@@ -290,7 +290,12 @@ def _process_references_for_position(
                     continue
             if sym.qualified_name.startswith(container.qualified_name + "."):
                 continue
-            _add_edge_site(edge_set, container.qualified_name, sym.qualified_name, ref_file, ref_line, ref_char)
+            # Every guard above sees the innermost container, which is what decides whether
+            # this reference is an edge at all. Only the credit line is rolled up.
+            attributed = st.attribution_symbol(container).qualified_name
+            if attributed == sym.qualified_name or sym.qualified_name.startswith(attributed + "."):
```
ivanmilevtues, line 296: this here seems interesting :D, but if it works it works. YOu can double check the P1 comment from codex on this one

static_analyzer/engine/models.py
```diff
@@ -23,6 +23,9 @@ class SymbolInfo:
     end_line: int
     end_char: int
     parent_chain: list[tuple[str, int]] = field(default_factory=list)
+    # Arrived as VARIABLE/CONSTANT and was promoted to CLASS for having callable children.
+    # A `const raw = await fn(() => ...)` wrapper is a scope, not a name a reader would use.
+    promoted_from_variable: bool = False
```
ivanmilevtues, line 28: hmm interesting if it is prompted from variable true then the kind should be whatever the promotion is, can we make some sort of an assert for this case so wehnever this gets set to true that kind is w.e. we are promoting it to. It just feels a bit like a patch that we have to monitor later on and validate (hense i am a bit opposed to it).

If it is too much you can leave it I just dislike it a bit as it makes the SymbolInfo (which in my head is what the LSP says) into a class which is somewhat not exactly what the lsp says, but also not our own type, as we transform it for the analysis.

static_analyzer/engine/symbol_table.py
```diff
@@ -76,10 +76,12 @@ def register_symbols(
 
             # Promote variables/constants with method children to class
             children = sym.get("children", [])
+            promoted = False
             if kind in (NodeType.VARIABLE, NodeType.CONSTANT) and children:
                 child_kinds = {c.get("kind", 0) for c in children}
                 if child_kinds & CALLABLE_KINDS:
                     kind = NodeType.CLASS
+                    promoted = True
```
ivanmilevtues, line 84: so this here would be a callback, interesting that we are assigning it into a class. It seems a bit odd child_kinds & callable (i'd expect a metho or a func here).

ivanmilevtues (approved):
Again few minor comments, would prefer to have callables be promoted to METHOD/FUNCTIONS and not classes the others are nits.
Overall good job on this!

static_analyzer/engine/edge_builder.py
Svilen-Stefanov: Checked both of them.

The P2 about definition-side edges is stale, `82a6476e` already switched the parent-class and queued implementation edges to `attributed` and that commit is on this head. The only raw `caller.qualified_name` left is inside `_is_valid_edge`, which is the guard, not the credit.

The P1 about the incremental rebuild is real and unfixed. The warm-start path in `incremental_orchestrator.py` never goes through `_process_references_for_position` or even the SymbolTable, it resolves call sites against CallGraph nodes and writes `src_node.fully_qualified_name` raw, and `_containing_source_node` picks the innermost enclosing callable, which for a call in a callback is the callback. It's actually broader than codex says: `_add_outbound_edges_from_changed_files` re-resolves every call site in a changed file, not only cross-file ones, so it both reintroduces `X.map() callback -> target` and duplicates same-file edges the fresh build already added under the attributed name.

It's a pre-existing path this PR doesn't touch, but shipping without it means the attribution only holds on cold runs and quietly unwinds for every file you edit. Doing it as a follow-up commit on this branch, cleanest shape is the roll-up inside `_containing_source_node` so both call sites inherit it. Shout if you'd rather it were its own PR.

static_analyzer/engine/models.py
Svilen-Stefanov: Good call, adding it as a `__post_init__` on `SymbolInfo`. The invariant does hold: `promoted = True` is only ever set in the same two-line block that sets `kind = NodeType.CLASS`, and `kind` isn't reassigned before any of the three construction sites.

On `SymbolInfo` being what the LSP says, I think that's the bit worth naming: it already isn't. `qualified_name` is adapter-computed, `parent_chain` is walker-computed, `definition_location` is derived, and the stored span is a hybrid of `selectionRange` start and `range` end, which no LSP payload contains. LSP's own `detail`, `containerName`, `tags` and `deprecated` are all dropped, and `kind` has been a transformed field since #224. So the flag isn't adding a transformation to a mirror, it's recording one we were already doing silently, and the assert is what makes it checkable rather than something to monitor later.

Rewording the docstring too, it documents only the callback case, which is what made line 84 read odd.

static_analyzer/engine/symbol_table.py
Svilen-Stefanov: Worth saying first that this block isn't new here, it's verbatim on `main` since #224, this PR only added the boolean next to it, and attribution works under either kind.

Your instinct is right for a different reason though: the branch is kind-blind and fires on two unrelated shapes. `const plugin = { register() {} }` is a namespace object where CLASS is fair, and `const raw = await fn(() => ...)` where the only callable child is an anonymous callback, where CLASS is a lie.

On METHOD/FUNCTION I'd rather not, for two reasons. Neither shape is callable, you can't call `plugin()` or `raw()`, so METHOD asserts something falser than CLASS, which at least means container. And it breaks four things concretely:

- CONTAINS edges only get built to parents in `CLASS_TYPES` (`result_converter.py:174`), so every object-literal method loses its container and goes isolated, which is the exact isolation the CFG completion work fixed.
- The reference filter at `edge_builder.py:255` gates class-like symbols on `is_invocation` but lets callables through unfiltered, so `import { plugin }`, `export default plugin` and `plugin.register(...)` would all start producing call edges.
- The parent roll-up at `:518` only fires for class-like parents, so the `api.get` aggregation edge disappears.
- `symbol_table.py:358` marks a PROPERTY local when its parent is callable, so the non-method members of every object literal get dropped from references.

The real cost of the current behaviour is the wrapper case leaking into the class hierarchy and being flaggable as a god class. If that's what bothers you, the narrower fix is to promote only when a child is a *named* callable, which leaves `const raw = await fn(() => ...)` a VARIABLE and keeps namespace objects as CLASS. That's a behaviour change with its own eval so I'd file it separately rather than fold it in here. wdyt?

static_analyzer/engine/edge_builder.py
Svilen-Stefanov: Checked both. The P2 is stale, `82a6476e` already moved the definition-side edges to `attributed` and that's on this head.

The P1 is real. The warm-start path in `incremental_orchestrator.py` never touches the SymbolTable at all, it resolves against CallGraph nodes and writes `src_node.fully_qualified_name` raw, so attribution only holds on cold runs and unwinds for every file you edit. Broader than codex says too, it re-resolves every call site in a changed file so same-file edges get duplicated under both names.

Pre-existing path this PR doesn't touch, doing it as a follow-up commit here. Shout if you'd rather it were separate.

static_analyzer/engine/models.py
Svilen-Stefanov: Good call, adding it as a `__post_init__`. The invariant holds, `promoted = True` is only set in the same two lines that set `kind = NodeType.CLASS`.

On `SymbolInfo` being what the LSP says: it already isn't. `qualified_name` is adapter-computed, `parent_chain` walker-computed, and the span is a hybrid of `selectionRange` start and `range` end that no payload contains. So the flag records a transformation we were already doing silently, and the assert makes it checkable rather than something to watch.

static_analyzer/engine/symbol_table.py
Svilen-Stefanov: This block isn't new here, it's on main since #224, this PR only added the boolean beside it.

Your instinct is right for a different reason though: it's kind-blind and fires on two shapes. `const plugin = { register() {} }` where CLASS is fair, and `const raw = await fn(() => ...)` where CLASS is a lie.

METHOD/FUNCTION I'd rather not. Neither shape is callable, so it asserts something falser than CLASS. And CONTAINS edges only build to `CLASS_TYPES` parents, so every object-literal method would go isolated, and the reference filter lets callables through ungated so `import { plugin }` would start producing call edges.

If the wrapper case is what bothers you, the narrower fix is promoting only on a named callable child. That needs its own eval so I'd file it separately. wdyt?

static_analyzer/engine/edge_builder.py
```diff
@@ -290,7 +290,12 @@ def _process_references_for_position(
                     continue
             if sym.qualified_name.startswith(container.qualified_name + "."):
                 continue
-            _add_edge_site(edge_set, container.qualified_name, sym.qualified_name, ref_file, ref_line, ref_char)
+            # Every guard above sees the innermost container, which is what decides whether
+            # this reference is an edge at all. Only the credit line is rolled up.
+            attributed = st.attribution_symbol(container).qualified_name
```
Svilen-Stefanov: Checked both. The P2 is stale, `82a6476e` already moved the definition-side edges to `attributed` and that's on this head.

The P1 is real. The warm-start path in `incremental_orchestrator.py` never touches the SymbolTable at all, it resolves against CallGraph nodes and writes `src_node.fully_qualified_name` raw, so attribution only holds on cold runs and unwinds for every file you edit. Broader than codex says too: it re-resolves every call site in a changed file, not just cross-file ones, so same-file edges get duplicated under both names.

Pre-existing path this PR doesn't touch, so doing it as a follow-up commit here. Shout if you'd rather it were separate.

static_analyzer/engine/models.py
Svilen-Stefanov: Good call, adding it as a `__post_init__`. The invariant holds, `promoted = True` is only set in the same two lines that set `kind = NodeType.CLASS`.

On `SymbolInfo` being what the LSP says: it already isn't. `qualified_name` is adapter-computed, `parent_chain` walker-computed, and the span is a hybrid of `selectionRange` start and `range` end that no payload contains. So the flag records a transformation we were already doing silently, and the assert makes it checkable rather than something to watch.

static_analyzer/engine/symbol_table.py
Svilen-Stefanov: This block isn't new here, it's on main since #224, this PR only added the boolean beside it.

Your instinct is right for a different reason though: it's kind-blind and fires on two shapes. `const plugin = { register() {} }` where CLASS is fair, and `const raw = await fn(() => ...)` where CLASS is a lie.

METHOD/FUNCTION I'd rather not. Neither shape is callable, so it asserts something falser than CLASS. And CONTAINS edges only build to `CLASS_TYPES` parents, so every object-literal method would go isolated, and the reference filter lets callables through ungated so `import { plugin }` would start producing call edges.

If the wrapper case is what bothers you, the narrower fix is promoting only on a named callable child. That needs its own eval so I'd file it separately. wdyt?

static_analyzer/engine/edge_builder.py
Svilen-Stefanov: Checked both codex ones. The P2 about definition-side edges is stale, `82a6476e` already moved them to `attributed`.

The P1 is real. The warm-start path in `incremental_orchestrator.py` never touches the SymbolTable, it resolves against CallGraph nodes and writes `src_node.fully_qualified_name` raw, so attribution holds on cold runs and unwinds for every file you edit. Broader than codex says too: it re-resolves every call site in a changed file, not just cross-file ones, so same-file edges get duplicated under both names.

It is a pre-existing path this PR does not touch, so I would rather do it as its own PR than widen this one. Flagging it because shipping without it means the 24.8% to 0.6% above only holds on a cold run.

static_analyzer/engine/models.py
```diff
@@ -23,6 +23,9 @@ class SymbolInfo:
     end_line: int
     end_char: int
     parent_chain: list[tuple[str, int]] = field(default_factory=list)
+    # Promoted to CLASS for having callable children: a namespace object, or a
+    # `const raw = await fn(() => ...)` wrapper that is a scope rather than a name.
+    promoted_from_variable: bool = False
```
Svilen-Stefanov: Good call, adding it as a `__post_init__`. The invariant holds: `promoted = True` is only set in the same two lines that set `kind = NodeType.CLASS`.

On `SymbolInfo` being what the LSP says, that is the bit worth naming: it already isn't. `qualified_name` is adapter-computed, `parent_chain` walker-computed, and the span is a hybrid of `selectionRange` start and `range` end that no payload contains. The flag records a transformation we were already doing silently, and the assert makes it checkable.

static_analyzer/engine/symbol_table.py
Svilen-Stefanov: This block isn't new here, it's on main since #224, this PR only added the boolean beside it.

Your instinct is right for a different reason though: it is kind-blind and fires on two shapes. `const plugin = { register() {} }` where CLASS is fair, and `const raw = await fn(() => ...)` where CLASS is a lie.

METHOD/FUNCTION I'd rather not. Neither shape is callable, so it asserts something falser than CLASS. And CONTAINS edges only build to `CLASS_TYPES` parents, so every object-literal method would go isolated, and the reference filter lets callables through ungated so `import { plugin }` would start producing call edges.

If the wrapper case is what bothers you, the narrower fix is promoting only on a named callable child. That needs its own eval so I'd file it separately. wdyt?

ivanmilevtues:
#515 was merged

PR #509, opened 2026-08-20
feat(incremental): show the model the calls the commit touched

Svilen-Stefanov:
Stacked on #508.

`build_scope_cfg_string` has no notion of `changed_members`, so an incremental run hands the model **every** cross-component edge in a scope — the same helper the full-run agents use — across **two** prompts per changed scope (`step_api_surfaces` and `step_relation_analysis`). On a real webview step, 30 of 366 root-scope edges touch a changed method, so 92% of the evidence is unrelated to the change.

## The gate

A pair carrying no touched edge renders as one line stating its weight:

```
Code Review & Stream Interface -> Shared UI Utilities (61 edges, none touched by this change)
```

instead of up to ten call lines. It **stays visible** — dropping it entirely would hide that A talks to B at all, degrading the api-surface answer and inviting the model to invent the relation back.

**Component-pair gating was measured and rejected first.** Changed-ness propagates to ancestors, so 47 of 47 root pairs stay hot and a pair gate cuts 5% of lines. Only edge-level hotness has leverage.

## The trap, and why the restore path widens in the same commit

The model's output **replaces** the scope's relations wholesale. So evidence hidden from it is a relation it cannot describe — and today's restore path is far too narrow to catch that: `touches_change` is true for every root pair, and `_relation_edges_unmoved` fails whenever an edge count drifted. **Gating alone would trade edge churn for wording churn.**

So the prompt gate and the restore path share one predicate. `pair_untouched_by_change` becomes a third disjunct in the wording-restore condition, which makes the invariant hold:

> a pair the prompt rendered as a bare count never takes a model-authored label.

No plumbing is needed to keep the two sides in sync — both are pure functions of `(the relation's fresh edges, changed_members)`, and both sides already hold both.

`edge_touches_change` is deliberately **symmetric**, unlike the existing `_edge_touches_changed_method`. That one asks whether the commit deleted a call, which only a changed *source* can do. This one asks whether the edge is evidence for re-describing the pair, which either end moving supplies.

## Measured

One real webview step, **357 changed methods** — a large change, so this is a conservative figure. A small PR gates far harder.

| scope | now | gated | cut |
|---|---|---|---|
| root | 324 | 178 | **45%** |
| scope 1 | 170 | 154 | 9% |
| scope 3 | 64 | 26 | 59% |
| **all scopes** | **657** | **439** | **33%** |

Rendered twice per changed scope, so the token saving doubles.

A **full run is untouched**: `changed_members` is empty and the output is byte-identical to today's.

## Ordering

⚠️ This touches `preserve_unchanged_relations`, which **#505 also rewrites**. #505 is open against `main`; this sits on the engine stack. They conflict in one place — the wording-restore condition. Resolution is additive: keep #505's `_restore_baseline_wording(relation, previous)` call and add `or pair_untouched_by_change(relation, changed_members)` to the same condition. **Merge #505 first.**

## Tests

`tests/agents/test_cfg_evidence_gating.py`, 8 cases covering both predicates, including the symmetry that distinguishes them from the deletion predicate and the full-run degradation.

Full suite: 2158 passed. The 3 `test_cli_parser` failures are pre-existing on `main`.

## Summary by CodeRabbit

* **Improvements**
  * Improved incremental relation updates by focusing analysis on changed or affected code.
  * Preserved existing descriptions for unchanged relationships while identifying removed or modified connections.
  * Accounted for changes in files without attributed members, including module-level and configuration updates.
  * Highlighted affected connections and prioritized them in generated relationship views.

* **Tests**
  * Added coverage for evidence filtering, changed and deleted connections, relationship preservation, and connection ordering.

### DIFF
--- a/agents/incremental_agent.py
+++ b/agents/incremental_agent.py
@@ -300,7 +300,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         self,
         scope: AnalysisInsights,
         scope_name: str,
-        clustering: ClusterScopeResult,
+        static_call_evidence: str,
     ) -> ComponentApiSurfaces:
         """Analyze API surfaces for one updated scope."""
         logger.info("[IncrementalAgent] Analyzing API surfaces for scope: %s", scope_name)
@@ -308,10 +308,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
             component_summaries=ComponentArchitecture(
                 description=scope.description, components=scope.components
             ).llm_str(),
-            static_call_evidence=render_scope_connections(
-                clustering,
-                {component.component_id: component.name for component in scope.components},
-            ),
+            static_call_evidence=static_call_evidence,
         )
         return self._parse_invoke(prompt, ComponentApiSurfaces)
 
@@ -322,6 +319,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         scope_name: str,
         api_surfaces: ComponentApiSurfaces,
         clustering: ClusterScopeResult,
+        static_call_evidence: str,
     ) -> list[Relation]:
         """Discover evidence-backed relations and attach deterministic CFG edges."""
         logger.info("[IncrementalAgent] Discovering component relations for scope: %s", scope_name)
@@ -338,10 +336,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
                 description=scope.description, components=scope.components
             ).llm_str(),
             api_surfaces=api_surfaces.llm_str(),
-            static_call_evidence=render_scope_connections(
-                clustering,
-                {component.component_id: component.name for component in scope.components},
-            ),
+            static_call_evidence=static_call_evidence,
         )
         relation_result: ComponentRelations = self._invoke_validate(
             prompt,
@@ -409,12 +404,19 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         }
 
         if context.changed_ids:
-            api_surfaces = self.step_api_surfaces(scope, scope_name, context.clustering)
+            # Rendered once and shared: both prompts must agree on what the commit touched.
+            static_call_evidence = render_scope_connections(
+                context.clustering,
+                {component.component_id: component.name for component in scope.components},
+                changed_members,
+            )
+            api_surfaces = self.step_api_surfaces(scope, scope_name, static_call_evidence)
             rels = self.step_relation_analysis(
                 scope,
                 scope_name,
                 api_surfaces,
                 context.clustering,
+                static_call_evidence,
             )
         else:
             # Nothing in this scope changed, so preserve_unchanged_relations below would discard any
--- a/agents/incremental_agent.py
+++ b/agents/incremental_agent.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import threading
+from collections.abc import Collection
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
@@ -377,6 +378,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         scope_name: str,
         context: ScopeRelationContext,
         changed_members: set[str] | None = None,
+        unattributed_files: Collection[str] = (),
     ) -> list[Relation]:
         """Run the API-surface and relation stages for one updated scope, or skip both when nothing in it changed."""
         if len(scope.components) < 2:
@@ -403,12 +405,15 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
             if relation.src_name and relation.dst_name
         }
 
+        gated_members: set[str] = set()
         if context.changed_ids:
             # Rendered once and shared: both prompts must agree on what the commit touched.
+            gated_members = set() if unattributed_files else changed_members or set()
             static_call_evidence = render_scope_connections(
                 context.clustering,
                 {component.component_id: component.name for component in scope.components},
-                changed_members,
+                gated_members,
+                baseline_by_pair,
             )
             api_surfaces = self.step_api_surfaces(scope, scope_name, static_call_evidence)
             rels = self.step_relation_analysis(
@@ -463,6 +468,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
                 live_ids,
                 live_qnames,
                 changed_members,
+                gated_members,
             )
             # Preservation runs after the grounding filters and re-injects baseline edges, so
             # the assembled list is filtered once more — otherwise a baseline's invented or
@@ -482,6 +488,7 @@ class IncrementalAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         sub_analyses: dict[str, AnalysisInsights],
         relation_contexts: dict[str, ScopeRelationContext],
         changed_members: set[str] | None = None,
+        unattributed_files: Collection[str] = (),
     ) -> None:
         """Regenerate relations for every touched scope with at least two components.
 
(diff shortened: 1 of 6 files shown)


Svilen-Stefanov:
Addressed the review.

**Deleted calls — real, and it undercut the whole gate.** A call the commit removed is absent from the rebuilt edges, so hotness judged on those alone reported the pair untouched and pinned a label still describing the removed interaction. `pair_untouched_by_change` now reads the baseline's edges too, and `preserve_unchanged_relations` passes `previous`. Two tests cover it.

**Double indentation — real, and it falsified a claim in this PR body.** The cold marker was two spaces on top of the format string's own two, so every rendered call moved from two-space to four-space indentation. The "full run is byte-identical" claim was therefore wrong as written. Marker is empty for cold edges now, and it is byte-identical.

**Style, fixed.** Public predicates moved above the private helpers.

**Not done: unattributed file changes.** You are right that a module-level edit (a registry, a route table) leaves `changed_members` non-empty while touching no method in a pair, so the pair is summarised and its wording pinned. I have not wired `unattributed_files` through, because the signal lives on `ChangedMembers` at a different layer and threading it is a bigger change than this PR should carry. Worth a follow-up — flagging rather than quietly leaving it.

agents/cluster_methods_mixin.py
```diff
@@ -743,17 +750,32 @@ def build_scope_cfg_string(self, analysis: AnalysisInsights) -> str:
...
-            dst_label = id_to_name.get(dst_id, dst_id)
-            edge_count = len(relation.all_edges)
-            lines.append(f"\n{src_label} -> {dst_label} ({edge_count} edge{'s' if edge_count != 1 else ''}):")
-            for edge in relation.all_edges[:10]:
+            src_label = id_to_name.get(relation.src_cluster_id, relation.src_cluster_id)
+            dst_label = id_to_name.get(relation.dst_cluster_id, relation.dst_cluster_id)
+            edges = relation.all_edges
+            edge_count = len(edges)
+            plural = "s" if edge_count != 1 else ""
+            hot = [e for e in edges if edge_touches_change(e, changed_members)] if changed_members else []
```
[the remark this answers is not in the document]
ivanmilevtues: I think we should give only the real things, I have made changes in PR for these so this can be a nasty merge conflict :D

agents/cluster_methods_mixin.py
```diff
@@ -724,7 +725,7 @@ def _prefix_local_cluster_ids(self, analysis: AnalysisInsights, prefix: str) ->
                 component.source_cluster_ids, prefix
             )
 
-    def build_scope_cfg_string(self, analysis: AnalysisInsights) -> str:
+    def build_scope_cfg_string(self, analysis: AnalysisInsights, changed_members: set[str] | None = None) -> str:
```
ivanmilevtues: How does this impact our things on evals? As this for sure impacst the incremental analysis so curios to see the full eval results here if we are not introducing a regression.

ivanmilevtues (approved):
Can you send me over the eval results for this change if you have run anything to see if we have not introduced a regression ehre?

agents/cluster_methods_mixin.py
Svilen-Stefanov: No eval numbers yet, you're right to ask. The "Measured" table in the description is rendered prompt line counts, not benchmark scores, and the only other number is the unit suite.

Running the `incremental` benchmark, this head vs #508 as base, same fixed baseline, and I'll post `drift.edge_churn.{cosmetic_relabelled_count, relabelled_count, added_count, removed_count}` plus the graded op set, same shape as the existing `relation-stability-{main,fix}` pair. Including e2e since it exercises the incremental relation path.

Skipping the `full` benchmark on purpose, and worth saying why: with `changed_members` falsy, `hot` is empty, the header falls back to the old `({edge_count} edges):`, `marker` is `""` and `pair_untouched_by_change` returns False, so full-run output is byte identical. Verified in code, not assumed.

Caveat up front: `docs/benchmarks.md` records that counts vary between two runs of the same engine while structural operations don't, so one pair of runs on `small` is directional, not decisive.

agents/cluster_methods_mixin.py
Svilen-Stefanov: On the conflict, it's worse than a conflict: #499 deletes `cluster_methods_mixin.py` outright and re-homes this as `llm_renderers/clustering.py::render_scope_connections`.

The bit to watch is that this PR's two halves live in different files. The prompt gate is in the file you delete, and the restore half at `relation_edges.py:319` sits in a region that's byte identical between main and your head, so it survives the merge. A "take mine, the file's gone" resolution silently keeps the wording pin and drops the evidence gate that justifies it, and then the model gets shown every edge and is still forbidden from re-wording cold pairs. Can we agree the rebase either ports both halves or removes both? I'll add a regression test for the invariant so it can't drift quietly.

On "only the real things", just checking I've got you right: your version in #499 looks like a verbatim port of today's renderer, every pair with its count and up to ten call lines, no gate and no markers. So I don't think your diff resolves this one either way, the disagreement is whether an incremental prompt should ever summarise a pair rather than list it? The counts aren't fabricated, they're true, so it's really a prompt design call. Happy to go your way on it, I'd just want the gate to stay so cold pairs keep their labels.

agents/cluster_methods_mixin.py
Svilen-Stefanov: No numbers yet, fair hit. The "Measured" table is rendered prompt line counts, not benchmark scores.

Running the incremental benchmark, this head vs #508, same baseline, and I'll post `drift.edge_churn` plus the graded op set. Skipping the full benchmark on purpose: with `changed_members` falsy the marker is empty and the header falls back to the old one, so full-run output is byte identical. Verified in code.

Caveat up front, `docs/benchmarks.md` says counts vary between two runs of the same engine while structural ops don't, so one pair on `small` is directional not decisive.

agents/cluster_methods_mixin.py
Svilen-Stefanov: It's worse than a conflict, #499 deletes this file and re-homes the renderer.

The trap is that this PR's two halves live in different files: the gate is in the file you delete, and the wording pin at `relation_edges.py:319` is byte identical between main and your head so it survives. Resolving by taking yours keeps the pin and drops the gate that justifies it, and the model then gets shown every edge and is still forbidden from re-wording cold pairs. Can we agree the rebase ports both or neither? I'll add a regression test so it can't drift quietly.

On "only the real things", just checking I've got you: your #499 version looks like a verbatim port of today's renderer, count plus ten call lines, no gate. So I don't think it resolves this either way? The counts aren't fabricated, so it's really a prompt design call. Happy to go your way, I'd just want the gate kept.

agents/cluster_methods_mixin.py
Svilen-Stefanov: No numbers yet, fair hit. The "Measured" table is rendered prompt line counts, not benchmark scores.

Running the incremental benchmark, this head vs #508, same baseline, and I'll post `drift.edge_churn` plus the graded op set. Skipping the full benchmark on purpose: with `changed_members` falsy the marker is empty and the header falls back to the old one, so full-run output is byte identical. Verified in code, not assumed.

Caveat up front, `docs/benchmarks.md` says counts vary between two runs of the same engine while structural ops don't, so one pair on `small` is directional not decisive.

agents/cluster_methods_mixin.py
Svilen-Stefanov: It's worse than a conflict, #499 deletes this file and re-homes the renderer.

The trap is that this PR's two halves live in different files: the gate is in the file you delete, and the wording pin at `relation_edges.py:319` is byte identical between main and your head so it survives. Resolving by taking yours keeps the pin and drops the gate that justifies it, and the model then gets shown every edge and is still forbidden from re-wording cold pairs. Can we agree the rebase ports both or neither? I'll add a regression test so it can't drift quietly.

On "only the real things", just checking I've got you: your #499 version looks like a verbatim port of today's renderer, count plus ten call lines, no gate. So I don't think it resolves this either way? The counts aren't fabricated, so it's really a prompt design call. Happy to go your way, I'd just want the gate kept.

agents/cluster_methods_mixin.py
Svilen-Stefanov: No eval numbers yet. The "Measured" table in the description is rendered prompt line counts, not benchmark scores, so it does not answer your question.

Blocked rather than skipped: the OpenRouter key is out of credits (402 on `analyze_project_metadata`), which is the same thing failing `review` here. Once it is topped up the run is the `incremental` benchmark, this head vs #508 on a fixed baseline, reporting `drift.edge_churn` plus the graded op set.

One thing I can state without the run: the `full` benchmark cannot move. With `changed_members` falsy the marker is empty and the header falls back to the old one, so full-run output is byte identical. Verified in code.

agents/cluster_methods_mixin.py
Svilen-Stefanov: It is worse than a conflict, #499 deletes this file and re-homes the renderer.

The trap is that this PR's two halves live in different files: the gate is in the file you delete, and the wording pin at `relation_edges.py` is byte identical between main and your head so it survives. Resolving by taking yours keeps the pin and drops the gate that justifies it, and the model then gets shown every edge and is still forbidden from re-wording cold pairs. Can we agree the rebase ports both or neither?

On "only the real things", just checking I have you right: your #499 version looks like a verbatim port of today's renderer, count plus ten call lines, no gate. So I do not think it resolves this either way? The counts are true, so it is a prompt design call. Happy to go your way, I would just want the gate kept.

PR #510, opened 2026-08-21
refactor(analysis): remove details LLM cache

ivanmilevtues:
## Goal

Remove the persistent details-agent LLM cache before #498 starts making clustered scopes authoritative. Cache entries are not reusable safely across scopes, and our primary GitHub Actions workflow rarely preserves the repository-local database long enough to benefit from a hit.

## What changes

- Deletes the SQLite-backed `FinalAnalysisCache` and its public exports.
- Makes details analysis invoke the LLM for every requested component.
- Removes the cache-only `run_id` argument from `DetailsAgent`.
- Removes details-cache-backed run-ID reuse and cache-pruning lifecycle hooks.
- Leaves the meta-agent cache and static-analysis cache unchanged.

## Explicit behavior changes

- Details analysis no longer returns a cached architecture; every component analysis invokes the LLM.
- Full, partial, and incremental commands now receive a fresh run ID instead of reusing the run ID stored by the details cache.
- Existing `final_analysis_llm.sqlite` files are ignored and are not deleted from user repositories.

## Why this is separate

This is intentionally a small predecessor to #498. It resolves the stale scope/component-ID cache concern structurally without adding reconciliation logic to the clustering migration, and keeps #498 focused on consuming `ClusterScopeResult`.

## Validation

- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` — 2,152 passed, 28 skipped, 83.12% coverage
- `uv run mypy .`
- `uv run black ...` on all changed Python files

## Stack

#497 → **this PR** → #498 → #499 → #500

Amp-Thread-ID: https://ampcode.com/threads/T-01a024ad-20dd-707e-b96a-70cb73fdaa78


ivanmilevtues (commented):
I remvoed this cache as it was causing lot's of issues and our cache hits are close to 0. The github action could never hit the cache as peopel do not commit the db so it is pointless to have it, but it messes up with the ids.

diagram_analysis/run_context.py
```diff
@@ -1,13 +1,9 @@
-import logging
 from dataclasses import dataclass
```
Svilen-Stefanov, line 1: Looking at this file, I'm not sure we really need it if we have another place where to put these data classes.

diagram_analysis/run_context.py
```diff
@@ -38,34 +34,10 @@ def resolve(
         cls,
         repo_dir: Path,
         project_name: str,
```
Svilen-Stefanov, line 36: **Overall on this PR:** Good removal, the cache was doing very little for us and the run_id plumbing it dragged along was the annoying part. Two things before this goes in: it breaks CodeBoarding-wrapper (details in the inline on run_context.py), and the run_id machinery in caching/cache.py is now dead but still there.

Two smaller notes that didn't land on a changed line:
- tests/test_main.py:387,404,424 still stub `finalize=lambda: None` on the RunContext mock.
- dropping `reuse_latest_run_id` also changes the telemetry run_id dimension: incremental and partial used to reuse the previous run's id (telemetry reads `getattr(instance, "run_id")` off the generator), so a full plus its follow-up runs used to group under one id and now each gets its own. Probably more correct, just flagging it as a dashboard change nobody asked for.

---

blocker-ish: I think this breaks the wrapper. We need a wrapper PR before this merges, right?

diagram_analysis/run_context.py
```diff
@@ -38,34 +34,10 @@ def resolve(
...
         """Resolve the run metadata needed to construct a DiagramGenerator."""
-        run_id = _load_existing_run_id(repo_dir) if reuse_latest_run_id else None
-        if run_id is None:
-            run_id = generate_run_id()
-
         return cls(
-            run_id=run_id,
+            run_id=generate_run_id(),
             log_path=generate_log_path(project_name),
             repo_dir=repo_dir,
```
Svilen-Stefanov, line 42: Building on my other comment about whether we need this file: `repo_dir` is now write-only. Nothing in production reads `run_context.repo_dir` any more (github_action.py and codeboarding_workflows/analysis.py only touch `run_id` and `log_path`), it only existed for `finalize()` and `_load_existing_run_id`. Can we drop the field and the param too, or is something outside core reading it?

caching/__init__.py
```diff
@@ -1,11 +1,8 @@
 from caching.cache import BaseCache
-from caching.details_cache import DetailsCacheKey, FinalAnalysisCache
 from caching.meta_cache import MetaCache, MetaCacheKey
 
 __all__ = [
```
Svilen-Stefanov, line 4: The details cache was the only user of the run_id in `caching/cache.py`, so a fair bit of it is dead now: `load_most_recent_run()` has no production caller left (only tests/agents/test_meta_agent.py:165), `clear(keep_run_ids=...)`'s `keep_run_ids` is never passed by anything, and `_CLEAR_BEFORE_STORE = False` no longer has a subclass using it. MetaCache still writes the `run_id` column plus `idx_cache_entries_namespace_run_id` that nothing reads back. Should we rip that out here, or ticket it so this PR stays small?

codeboarding_workflows/orchestration.py
```diff
@@ -25,24 +24,17 @@
...
 
     Returns ``None`` when *source* yields ``None`` (cache hit). Otherwise
-    returns the scope's return value. The RunContext is finalized in a
-    ``finally`` block so it's released even if the scope raises.
+    returns the scope's return value.
     """
     with source as src:
         if src is None:
             return None
         run_context = RunContext.resolve(
```
Svilen-Stefanov, line 36: nit: with the try/finally gone this is basically `with source as src: resolve(); return scope(...)`, but the module docstring still frames it as owning a lifecycle so callers can't drift. Is the indirection still earning its keep? `incremental_analysis.py` already resolves the RunContext directly and doesn't use this at all.

codeboarding_cli/commands/incremental_analysis.py
```diff
@@ -49,7 +49,6 @@ def run_from_args(args: argparse.Namespace, parser: argparse.ArgumentParser) ->
         run_context = RunContext.resolve(
```
Svilen-Stefanov, line 49: nit: this try/except was here because `resolve(reuse_latest_run_id=True)` used to stat `.git` and open SQLite. Now `resolve` is two pure calls, so catching bare `Exception` and emitting `requiresFullAnalysis: true` can only ever hide a programming error. Can we drop it?

codeboarding_cli/commands/partial_analysis.py
```diff
@@ -62,7 +62,6 @@ def scope(src: SourceContext, run_context: RunContext) -> None:
             artifact_dir=run_paths.output_dir,
         ),
         scope=scope,
```
Svilen-Stefanov, line 64: Something I hadn't thought about until I traced it: this cache was what made expand idempotent. `step_final_analysis` keyed on prompt + model settings and returned the cached `AnalysisInsights` on a hit, so hitting expand twice on unchanged code gave you the same sub-components. Without it every expand is a fresh LLM call and the subtree can come back differently each time. Given expand is a button in the IDE, is that a regression we're happy with, or do we want a cheaper idempotency key (source_tree_hash + component_id) somewhere?

Svilen-Stefanov (approved):
LGTM, left mostly nits. The biggest part to be careful about is breaking other callers - before merging, let's make sure the wrapper is ready.

codeboarding_cli/commands/partial_analysis.py
ivanmilevtues: I think it is fine to have it like that for now for simplicity.

diagram_analysis/run_context.py
ivanmilevtues: Keeping `RunContext` as the small shared run-metadata boundary used by workflow orchestration, the incremental CLI, and the GitHub Action. Commit 29c71ecd narrows the class to the two values that boundary actually owns (`run_id` and `log_path`) and removes the overstated lifecycle wording.

diagram_analysis/run_context.py
ivanmilevtues: Addressed in 29c71ecd: removed the dead cache run-ID schema and APIs, removed the stale `finalize` test doubles, and simplified `RunContext` to generate a fresh ID per execution. The fresh telemetry ID is intentional. The required wrapper API migrations are now explicitly tracked in CodeBoarding/CodeBoarding-wrapper#93.

diagram_analysis/run_context.py
ivanmilevtues: Fixed in 29c71ecd. `repo_dir` is removed from both `RunContext` and `RunContext.resolve()` because no core consumer reads it. Wrapper call-site migration is tracked in CodeBoarding/CodeBoarding-wrapper#93.

caching/__init__.py
ivanmilevtues: Removed in 29c71ecd. The cache no longer stores `run_id` or `updated_at`, the unused run-ID index and lookup/retention APIs are gone, and the sole remaining cache implementation keeps only its current key/value entry.

codeboarding_workflows/orchestration.py
ivanmilevtues: Kept the helper because it still owns the shared source materialization and dispatch used by three full/partial CLI paths. Commit 29c71ecd replaces the overstated lifecycle documentation with a terse description of that actual responsibility.

codeboarding_cli/commands/incremental_analysis.py
ivanmilevtues: Fixed in 29c71ecd. The broad catch around `RunContext.resolve()` is removed so programming errors are no longer reported as a full-analysis requirement.

codeboarding_cli/commands/partial_analysis.py
ivanmilevtues: Accepted for now: repeated partial expansion intentionally makes a fresh LLM call. Reintroducing cache semantics solely for idempotency would work against the simplification in this stack, so no replacement idempotency layer is being added.

ivanmilevtues:

PR #515, opened 2026-08-24
fix(engine): recover calls a declaration-line guard was suppressing

Svilen-Stefanov:
Stacked on #508. Two follow-ups from the measurement write-up on #506/#507.

Originally opened on #507; **rebased onto #508** because the two interact — see "Why this sits on #508" below.

## 1. Every family suffix gets its own language id

#507 makes one adapter own the whole TypeScript/JavaScript family and widens `file_extensions` to match, but `language_id_for` still fell back to the *adapter's* id for anything that wasn't `.tsx`/`.jsx`. A JavaScript-owned family would open a `.ts` as `languageId: "javascript"` — the exact degradation #507 measured in the other direction (2903 nodes against 3225 for identical source).

Every suffix now maps explicitly, so the id follows the file rather than whichever adapter won the family. Measured byte-identical on CodeBoarding-webview: hardening, not a graph change.

## 2. A call in a declaration's body or initialiser is an edge

These shapes put the call on the same line as the declaration containing it:

```ts
const device = useDevice();
const files  = await fetchPrFiles(...).then(r => r.files);
const shown  = spent(role) ? ROLES.find(c => !spent(c)) : role;
onClick={() => track('brand_link_clicked', { surface })}
```

Correct JSX parsing (#506) makes tsserver report those one-line declarations as real symbols. They become the innermost container, the reference lands on its own container's declaration line, and the guard added by #427 discards it.

That guard came in to narrow **C#** incremental edges. `is_reference_in_declaration_body` accepts only `statement_block` and `arrow_expression_clause` — the latter a C# tree-sitter node type behind `include_references_on_declaration_line`, which only the C# adapter sets. A TypeScript concise arrow body or a `const x = f()` initialiser can never satisfy it.

The rule: once the walk has passed a call node, an enclosing `body` or `value` field counts as body. Scoped to the TypeScript/JavaScript grammars.

### Measured

Static analysis only, every call site adjudicated against the TypeScript compiler API, on top of #508:

| repo | edges | dropped | added | added confirmed | precision |
|---|---|---|---|---|---|
| CodeBoarding-webview | 1826 → **1945** | **0** | 119 | **100.0%** | — |
| excalidraw v0.18.0 | 5211 → **6135** | **0** | 924 | **99.6%** | 96.3% → **96.7%** |
| lodash 4.17.21 | 1376 → **1388** | **0** | 12 | **100.0%** | 95.2% → **95.3%** |

Nothing is dropped on any of the three. It recovers all four edges #506 was measured to lose, and with #508 underneath they are credited to the real component (`DiscordLink`, `BrandFooter`, `SetupResult`, `CommentsTab`) rather than to an anonymous arrow.

### Why this sits on #508

The two are not independent. On #507 alone, `lift_to_callable` happened to lift the one-line `const` container up to its enclosing function, so some of these references escaped the guard by accident. #508 stops that, which exposed a gap in the first version of this rule: it only accepted a call **directly** occupying the body field, so `await f().then(...)`, `a && f(b)` and ternaries still fell through. Measured on the combination, that cost 51 call sites; generalising the rule brings it to 8.

Those remaining 8 are **recursion** (`walk -> walk`, `scrollToComment -> scrollToComment`) plus one intra-component call. After #508 rolls the source up to the named enclosure they become self- or containment edges, which the engine already excludes by policy. Correct, not lost.

### Scope

- Python's `def g(x=f())` occupies the same `value` field but is evaluated by the enclosing scope, so every other grammar keeps the #427 default and its opt-in. Covered by a test.
- A bare reference in an object literal (`{ run: target }`) never passes a call node, so `test_object_literal_is_not_a_declaration_body` still holds unchanged.

### Tests

Two tests asserted the old behaviour for shapes that are real calls, both introduced by #427:

- `test_reference_later_on_declaration_line_is_ignored_by_default` — `const caller = () => target()`
- `test_reference_in_template_interpolation_is_not_a_block_body_edge` — ``const caller = () => `${target()}` ``

Both now assert the edge, and a new C# sibling proves the #427 default is untouched there. New: `tests/static_analyzer/test_declaration_line_body.py` (4 cases), one case in `test_typescript_adapter.py`.

Full suite: 2171 passed. The 3 `test_cli_parser` failures are pre-existing on `main`.

Tag bumped **v6 → v7**: the graph is materially different from the same source, so without it an unchanged tree keeps serving the previous engine's graph.

### Downstream

Fixtures: CodeBoarding-tests#18.


static_analyzer/analysis_cache.py
```diff
@@ -65,8 +65,10 @@
 # second, degraded ``javascript`` bucket that was clustered as a separate codebase.
 # v6: a call inside an inline callback is credited to the enclosing declaration, so edge
 # sources are names a reader recognises rather than ``map() callback``.
+# v7: a call in a declaration's body or initialiser is an edge, so `const x = f()` and
+# `onClick={() => f()}` stop being dropped — every TypeScript/JavaScript graph gains edges.
 # Older pickles are treated as cache misses and re-run.
-_TAG_VERSION = "v6"
+_TAG_VERSION = "v7"
```
ivanmilevtues: In this stack we have bumped the version too many times please keep it at v5. (which is the initial bump ;d)_

static_analyzer/engine/adapters/typescript_adapter.py
```diff
@@ -11,7 +11,18 @@
 
 class TypeScriptAdapter(LanguageAdapter):
 
-    _JSX_LANGUAGE_IDS: ClassVar[dict[str, str]] = {".tsx": "typescriptreact", ".jsx": "javascriptreact"}
+    # Every suffix the family serves, so the id follows the file rather than whichever adapter
+    # happens to own the family. A ``.ts`` opened as "javascript" loses its declarations.
+    _LANGUAGE_IDS: ClassVar[dict[str, str]] = {
```
ivanmilevtues: here as in previous comment all these have to go in the constants.py and we should reuse them from there.

static_analyzer/engine/source_inspector.py
```diff
@@ -232,9 +242,25 @@ def is_reference_in_declaration_body(
...
+            if node.type in _CALL_NODE_TYPES or node.type in _CONSTRUCTOR_NODE_TYPES:
+                # The reference is part of a call; an enclosing body or initialiser now counts.
+                passed_a_call = True
+            if body_starts_in_declaration and body_fields and passed_a_call and self._occupies_body_field(node):
+                return True
             node = node.parent
         return False
 
+    @staticmethod
+    def _occupies_body_field(node: TreeSitterNode) -> bool:
```
ivanmilevtues: can you put the method below the public ones.

static_analyzer/engine/source_inspector.py
```diff
@@ -109,6 +109,14 @@
...
 _DECLARATION_BLOCK_NODE_TYPES = frozenset({"block", "compound_statement", "statement_block"})
 _EXPRESSION_BODY_NODE_TYPES = frozenset({"arrow_expression_clause"})
+# Fields whose occupant is the declaration's body or initialiser rather than its signature.
+# Only counted once the walk has passed a call, so a bare reference in an object literal
+# (`{ run: target }`) stays out of the body while `const x = await f().then(...)` does not.
+# Why only TS/JS: a one-line `onClick={() => f()}` or `const x = f()` puts the call on its own
+# declaration's line, where the block-node test cannot reach it. Python's `def g(x=f())` occupies
+# `value` too but is evaluated by the enclosing scope, so the rule stays off the other grammars.
+_BODY_FIELD_NAMES = ("body", "value")
+_BODY_FIELD_SUFFIXES = frozenset({".ts", ".mts", ".cts", ".tsx", ".js", ".mjs", ".cjs", ".jsx"})
```
ivanmilevtues: I'd rather have this as something like BODY_DECLARATION_LANGUAGES and then we reuse our enums to get the suffices. the field names you can stay here i suppose.

ivanmilevtues (approved):
Just few minor comments on the versioning + usage of constants.

static_analyzer/analysis_cache.py
Svilen-Stefanov: Pushing back on this one. Two corrections first: the initial bump in the stack is v4 on #506, v5 is already #507's, so "keep it at v5" is two bumps rather than one. And the bumps are free, `analysis_cache.py:150` and `:161` compare with `!=`, never `<`, so a user coming from main's v3 to a build carrying v7 misses exactly once and rebuilds exactly once, identical to what a single v4 would cost. There's no user who rebuilds four times.

They're load-bearing because these land on main one at a time, each becoming a state main sits at. With one shared tag, a cache written by the #506 build is accepted by the #508 build, so the fixes never reach an unchanged tree. Sharpest case: a 506-build on a mixed `.ts`/`.js` repo writes both a TYPESCRIPT and a JAVASCRIPT bucket, then `_family_owner_changed` sees a non-empty intersection with the 507-build's live `{TYPESCRIPT}` and returns False, so the degraded javascript bucket that #507 exists to delete survives the upgrade.

We already ran this policy once and it broke: #476 bumped v2 to v3 for the whole #474 to #483 stack, then #480 moved CallGraph/Edge to a new module and #483 turned reference edges into ReferenceEdge objects under that same v3. Four pickle layouts in the wild sharing one tag.

If the annoyance is the comment block churn and the rebase conflict on that one line every PR, happy to collapse the history into a single "v4 to v7: TypeScript/JavaScript graph changes across #506 to #515" note while keeping the distinct values. And if you do want one bump, it needs one squashed merge of the whole stack, which is a merge strategy call rather than a code one, fine by me either way.

nit on my own side: the v7 comment names only the declaration-line fix, but the PR also carries the per-suffix language ids, which change the graph for a JS repo with one `.ts`. Naming both.

static_analyzer/engine/adapters/typescript_adapter.py
Svilen-Stefanov: Doing this. One correction to carry over though: the two JSX ids aren't in any enum today and can't be derived from one, because `LANGUAGE_EXTENSIONS` already maps `.tsx` to `Language.TYPESCRIPT`, so an enum lookup returns `typescript`, which is the bug. So it wants to be a derived map with those two as literals:

```python
JSX_LANGUAGE_IDS = {".tsx": "typescriptreact", ".jsx": "javascriptreact"}
LSP_LANGUAGE_ID_BY_EXTENSION = {
    ext: JSX_LANGUAGE_IDS.get(ext, language.value)
    for language, exts in LANGUAGE_EXTENSIONS.items()
    for ext in exts
}
```

Then the adapter drops `_LANGUAGE_IDS` and `language_id_for` is a one-line lookup keeping the existing fallback, which `test_unknown_suffix_falls_back_to_the_adapter_id` covers. That gets the 6 non-JSX entries out of the hardcoded list, which is your point in its strong form.

While there: the scalar `language_id` is genuinely redundant, I instantiated all 8 adapters and it equals `language_enum.value` every time, so I can make the base return that and delete all 8 overrides in the same pass. Want that in here or separately?

static_analyzer/engine/source_inspector.py
Svilen-Stefanov: Will do. Moving it to just after `find_member_modifiers`, which is the only spot actually below all the publics, since `find_type_bases` and `find_member_modifiers` sit under the earlier private block. Pure move, it's a staticmethod called through `self` and its only free name is module level, so no behaviour change and no test edits. Codex asked for the same thing under the method's old name, so this closes both.

Heads up that the class already interleaves on `main`, `_method_group_candidates` and the block below it sit above both of those publics. This fixes what I introduced, not the file. Want a ticket for the rest, or leave it?

static_analyzer/engine/source_inspector.py
Svilen-Stefanov: Agreed, and it comes out as exactly the same 8 suffixes, checked element by element against `LANGUAGE_EXTENSIONS`:

```python
BODY_DECLARATION_LANGUAGES = (Language.TYPESCRIPT, Language.JAVASCRIPT)
_BODY_FIELD_SUFFIXES = frozenset(
    suffix for language in BODY_DECLARATION_LANGUAGES for suffix in LANGUAGE_EXTENSIONS[language]
)
```

`source_inspector.py:16` already imports both and `constants.py` imports nothing but `enum`, so no new import and no cycle. Keeping `_BODY_FIELD_NAMES` local as you said.

static_analyzer/analysis_cache.py
Svilen-Stefanov: Pushing back on this one.

Two corrections first: the initial bump in the stack is v4 on #506, so "keep it at v5" is already two. And they're free, `analysis_cache.py` compares with `!=` and never `<`, so someone coming from main's v3 to v7 misses once and rebuilds once, same as a single v4 would cost.

They're load-bearing because these land on main one at a time. With one shared tag a cache written by the #506 build is accepted by the #508 build, so the fixes never reach an unchanged tree. Sharpest case: a 506-build on a mixed repo writes both buckets, `_family_owner_changed` then sees a non-empty intersection and returns False, and the degraded javascript bucket #507 exists to delete survives the upgrade.

We ran this policy once already: #476 bumped once for the whole #474 to #483 stack, then #480 and #483 changed the pickled shape under that same v3. Four layouts sharing one tag.

Happy to collapse the comment block into one line if the churn is the annoyance. And if you do want a single bump it needs one squashed merge of the stack, which is a merge strategy call, fine either way.

static_analyzer/engine/source_inspector.py
```diff
@@ -109,6 +109,15 @@
...
 _DECLARATION_BLOCK_NODE_TYPES = frozenset({"block", "compound_statement", "statement_block"})
 _EXPRESSION_BODY_NODE_TYPES = frozenset({"arrow_expression_clause"})
+# Fields holding a declaration's body or initialiser rather than its signature.
+# Why only TS/JS: a one-line `onClick={() => f()}` puts the call on its own declaration's
+# line, out of the block-node test's reach. Python's `def g(x=f())` occupies `value` too but
+# the enclosing scope evaluates it, so the rule stays off the other grammars.
+_BODY_FIELD_NAMES = ("body", "value")
+BODY_DECLARATION_LANGUAGES = (Language.TYPESCRIPT, Language.JAVASCRIPT)
+_BODY_FIELD_SUFFIXES = frozenset(
+    suffix for language in BODY_DECLARATION_LANGUAGES for suffix in LANGUAGE_EXTENSIONS[language]
```
Svilen-Stefanov: Done, and it comes out as exactly the same 8 suffixes, checked element by element:

```python
BODY_DECLARATION_LANGUAGES = (Language.TYPESCRIPT, Language.JAVASCRIPT)
_BODY_FIELD_SUFFIXES = frozenset(
    suffix for language in BODY_DECLARATION_LANGUAGES for suffix in LANGUAGE_EXTENSIONS[language]
)
```

The file already imported both so no new import. Kept `_BODY_FIELD_NAMES` local as you said, and trimmed the comment above it, it was longer than the code.

static_analyzer/analysis_cache.py
Svilen-Stefanov: Pushing back on this one.

Two corrections first: the initial bump in the stack is v4 on #506, so "keep it at v5" is already two. And they're free, `analysis_cache.py` compares with `!=` and never `<`, so someone coming from main's v3 to v7 misses once and rebuilds once, same as a single v4 would cost.

They're load-bearing because these land on main one at a time. With one shared tag a cache written by the #506 build is accepted by the #508 build, so the fixes never reach an unchanged tree. Sharpest case: a 506-build on a mixed repo writes both buckets, `_family_owner_changed` then sees a non-empty intersection and returns False, and the degraded javascript bucket that #507 exists to delete survives the upgrade.

We ran this policy once already: #476 bumped once for the whole #474 to #483 stack, then #480 and #483 changed the pickled shape under that same v3. Four layouts sharing one tag.

Happy to collapse the comment block into one line if the churn is the annoyance. And if you do want a single bump it needs one squashed merge of the stack, which is a merge strategy call, fine either way.

static_analyzer/engine/source_inspector.py
```diff
@@ -109,6 +109,15 @@
 _PREPROCESSOR_SUFFIXES = frozenset({".cs"})
 _DECLARATION_BLOCK_NODE_TYPES = frozenset({"block", "compound_statement", "statement_block"})
 _EXPRESSION_BODY_NODE_TYPES = frozenset({"arrow_expression_clause"})
+# Fields holding a declaration's body or initialiser rather than its signature.
+# Why only TS/JS: a one-line `onClick={() => f()}` puts the call on its own declaration's
+# line, out of the block-node test's reach. Python's `def g(x=f())` occupies `value` too but
+# the enclosing scope evaluates it, so the rule stays off the other grammars.
+_BODY_FIELD_NAMES = ("body", "value")
+BODY_DECLARATION_LANGUAGES = (Language.TYPESCRIPT, Language.JAVASCRIPT)
```
Svilen-Stefanov: Done, named exactly as you suggested, and it comes out as the same 8 suffixes, checked element by element. The file already imported both so no new import. Kept `_BODY_FIELD_NAMES` local as you said, and trimmed the comment above it, it was longer than the code.

static_analyzer/engine/source_inspector.py
```diff
@@ -596,6 +612,17 @@ def text(node: TreeSitterNode) -> str:
                 modifiers[(type_name, text(member_name_node))] = frozenset(found)
         return modifiers
 
+    @staticmethod
+    def _occupies_body_field(node: TreeSitterNode) -> bool:
```
Svilen-Stefanov: Moved down here, just after `find_member_modifiers`. That's the only spot actually below all the publics, since `find_type_bases` and `find_member_modifiers` sit under the earlier private block. Same request codex made under the old name, so this closes both.

Heads up that the class already interleaves on main, so this fixes what I introduced rather than the file. Want a ticket for the rest?

static_analyzer/engine/source_inspector.py
Svilen-Stefanov: Done, named as you suggested, and it comes out as the same 8 suffixes, checked element by element. The file already imported both so no new import, and `_BODY_FIELD_NAMES` stays local as you said.

static_analyzer/engine/source_inspector.py
Svilen-Stefanov: Moved down here, just after `find_member_modifiers`. That is the only spot actually below all the publics, since `find_type_bases` and `find_member_modifiers` sit under the earlier private block. Pure move, and it closes the same request codex made under the old name.

Heads up that the class already interleaves on main, so this fixes what I introduced rather than the file. Want a ticket for the rest?

PR #516, opened 2026-08-24
Unify scoped clustering lineage and ownership

### LINKED ISSUE #514: Index persisted symbol ownership once for incremental clustering
## Problem

Incremental clustering currently reconstructs persisted symbol ownership through two incomplete and overlapping paths:

- `_previous_ownership` limits persisted members to files present in `ClusterResult.cluster_to_files`. That map describes structural Leiden clusters, so it excludes unchanged callable/class symbols assigned through the orphan fallback when their file produced no leaf cluster. An unrelated regroup can then move those symbols.
- `_persisted_members` rebuilds the complete normalized live-file set from every graph node for every persisted `FileMethodGroup`, causing repeated path normalization at roughly `O(file groups × graph nodes)`.

Review context:
- https://github.com/CodeBoarding/CodeBoarding/pull/499#discussion_r3836082525
- https://github.com/CodeBoarding/CodeBoarding/pull/499#discussion_r3836082526

## Intended design

Build one authoritative persisted symbol-ownership index before incremental hierarchy reconstruction:

1. Normalize each language graph once into `normalized file → live qualified names`.
2. Walk persisted component `file_methods` once and join them against that live index by normalized file and qualified name.
3. Produce `language → qualified name → component ID` ownership maps.
4. Reuse those maps for both persisted-lineage validation and previous member-ownership recovery.
5. Keep `cluster_to_files` limited to structural cluster claims; it must not define symbol ownership.

The index remains function-level. File paths only disambiguate language membership and connect persisted methods to their current language graph.

## Acceptance criteria

- An unchanged callable/class symbol assigned outside Leiden leaf clusters keeps its persisted component through an unrelated incremental regroup.
- Identical qualified names in different languages do not share ownership.
- Live graph paths are normalized once per language, not once per persisted file group.
- Structural cluster ownership continues to use `cluster_to_files` without broadening architectural membership to data-only symbols.

This should be implemented as a focused follow-up PR stacked after #500 rather than folded into the clustering cleanup.

ivanmilevtues:
## Summary

- replace the root partition, flattened method paths, and per-scope omitted-member map with one complete lineage record per clustering scope
- reroot complete scope records during single-child absorption so the partition and omitted members remain aligned, including at the architecture root
- index persisted callable/class ownership once from normalized live graphs and reuse it for lineage validation, cluster anchoring, and member ownership repair
- remove incremental root reseeding now that hierarchy preparation records the authoritative root and nested partitions together

This is stacked on #500. The cache schema changes, so existing static-analysis caches require one full analysis to reindex; no migration path is added.

Closes #514

## Verification

- `uv run black --check .`
- `uv run mypy .`
- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` (2141 passed, 28 skipped; 83.40% coverage)


static_analyzer/analysis_cache.py
```diff
@@ -56,8 +56,9 @@
 # v3: MethodClusterPaths moved to static_analyzer.clustering, then CallGraph and the edge
 # types to static_analyzer.cfg, and reference edges became ReferenceEdge objects.
 # v4: ClusterCache records intentionally unclustered members for each hierarchy scope.
+# v5: structural and unclustered lineage share one authoritative record per scope.
 # Older pickles are treated as cache misses and re-run.
-_TAG_VERSION = "v4"
+_TAG_VERSION = "v5"
```
ivanmilevtues: we can keep at v4, it is yet to be released.

static_analyzer/clustering/service.py
```diff
@@ -347,9 +354,7 @@ def seeded_input(scope_id: ScopeId, graphs: Mapping[str, CallGraph]) -> ClusterS
 
         return self._cluster_hierarchy(static_analysis.available_cfgs(), max_depth, seeded_input)
 
-    def _build_leaf_clusters(
-        self, static_analysis: Any, cluster_caches: Mapping[str, ClusterCache] | None = None
-    ) -> dict[str, ClusterResult]:
+    def _build_leaf_clusters(self, static_analysis: Any) -> dict[str, ClusterResult]:
```
ivanmilevtues: can we have the static_analysis typed i..e I think it has to be a static analysis Result or?

static_analyzer/analysis_cache.py
ivanmilevtues: Done in commit 47c13a23. The cache tag remains v4 because this schema is unreleased; the v4 description now reflects the complete scoped-lineage representation.

static_analyzer/clustering/service.py
ivanmilevtues: Done in commit 47c13a23. The clustering service now types its full, incremental, partial-scope, helper, and cache-recording boundaries as `StaticAnalysisResults`; the incremental baseline is typed the same way.

PR #517, opened 2026-08-25
feat: support third-party Anthropic providers

ivanmilevtues:
## Summary
- support custom Anthropic-compatible Messages API endpoints through `ANTHROPIC_BASE_URL` / `anthropic_base_url`
- require an Anthropic API key when selecting Anthropic through a custom base URL
- update the default Anthropic agent model to Claude Sonnet 5 and omit unsupported sampling parameters
- document Azure Foundry configuration and add provider/config regression coverage

## Testing
- `uv run pytest tests/agents/test_llm_config.py tests/test_user_config.py` (148 passed)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring custom Anthropic-compatible endpoints, including Azure AI/Foundry deployments.
  * Added `anthropic_base_url` to configuration files, environment settings, and TOML examples.
  * Updated Anthropic configuration guidance to use canonical model IDs, such as `claude-sonnet-5`.

* **Bug Fixes**
  * Improved endpoint validation and API key handling.
  * Improved compatibility for Anthropic models that do not support sampling parameters.

### DIFF
--- a/PYPI.md
+++ b/PYPI.md
@@ -116,9 +116,11 @@ LLM provider keys and model overrides are stored in `~/.codeboarding/config.toml
 # Uncomment exactly one provider key
 # openai_api_key    = "sk-..."
 # anthropic_api_key = "sk-ant-..."
-# anthropic_base_url = "https://resource.services.ai.azure.com/anthropic"
 # google_api_key    = "AIza..."
 # ollama_base_url   = "http://localhost:11434"
+
+# Optional Anthropic endpoint override; requires anthropic_api_key above
+# anthropic_base_url = "https://resource.services.ai.azure.com/anthropic"
 
 [llm]
 # Optional: override the default model for your active provider
--- a/PYPI.md
+++ b/PYPI.md
@@ -116,6 +116,7 @@ LLM provider keys and model overrides are stored in `~/.codeboarding/config.toml
 # Uncomment exactly one provider key
 # openai_api_key    = "sk-..."
 # anthropic_api_key = "sk-ant-..."
+# anthropic_base_url = "https://resource.services.ai.azure.com/anthropic"
 # google_api_key    = "AIza..."
 # ollama_base_url   = "http://localhost:11434"
 
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
 # openai_api_key            = "sk-..."
 # openai_base_url           = "https://api.example.com/v1"  # any OpenAI-compatible gateway
 # anthropic_api_key         = "sk-ant-..."
+# anthropic_base_url        = "https://resource.services.ai.azure.com/anthropic"  # Azure Foundry
 # google_api_key            = "AIza..."
 # vercel_api_key            = "vck_..."
 # aws_bearer_token_bedrock  = "..."
@@ -142,6 +143,8 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
 ```
 
 `openai_base_url` points CodeBoarding at any OpenAI-compatible gateway. Set the gateway API key as `openai_api_key`, then choose its model IDs with `agent_model` and `parsing_model`. The equivalent shell variables are `OPENAI_BASE_URL`, `OPENAI_API_KEY`, `AGENT_MODEL`, and `PARSING_MODEL`.
+
+`anthropic_base_url` points the Anthropic client at a compatible Messages API, including Azure Foundry Claude deployments. Set the deployment key as `anthropic_api_key` and use deployment names for `agent_model` and `parsing_model`.
 
 Shell environment variables such as `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, and `OLLAMA_BASE_URL` take precedence over the config file. For private repositories, set `GITHUB_TOKEN` in your environment.
 
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
 
 `openai_base_url` points CodeBoarding at any OpenAI-compatible gateway. Set the gateway API key as `openai_api_key`, then choose its model IDs with `agent_model` and `parsing_model`. The equivalent shell variables are `OPENAI_BASE_URL`, `OPENAI_API_KEY`, `AGENT_MODEL`, and `PARSING_MODEL`.
 
-`anthropic_base_url` points the Anthropic client at a compatible Messages API, including Azure Foundry Claude deployments. Set the deployment key as `anthropic_api_key` and use deployment names for `agent_model` and `parsing_model`.
+`anthropic_base_url` points the Anthropic client at a compatible Messages API, including Azure Foundry Claude deployments. Set the deployment key as `anthropic_api_key` and use canonical Anthropic model IDs such as `claude-sonnet-5` for `agent_model` and `parsing_model` so CodeBoarding selects the correct model capabilities and prompts.
 
 Shell environment variables such as `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, and `OLLAMA_BASE_URL` take precedence over the config file. For private repositories, set `GITHUB_TOKEN` in your environment.
 
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -31,6 +31,7 @@ _OPENROUTER_FALLBACK_CONTEXT_WINDOW = ContextWindow(1_048_576, 65_536, is_fallba
 _SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS = (
     "claude-opus-4-7",
     "claude-opus-4-8",
+    "claude-opus-5",
     "claude-sonnet-5",
     "claude-fable-5",
     "claude-mythos-5",
@@ -174,13 +175,14 @@ LLM_PROVIDERS = {
     ),
     "anthropic": LLMConfig(
         chat_class=ChatAnthropic,
-        selection_envs=["ANTHROPIC_API_KEY", "ANTHROPIC_BASE_URL"],
+        selection_envs=["ANTHROPIC_API_KEY"],
         api_key_env="ANTHROPIC_API_KEY",
         agent_model="claude-sonnet-5",
         parsing_model="claude-haiku-4-5",
         llm_type=LLMType.CLAUDE,
         extra_args={
             "base_url": lambda: os.getenv("ANTHROPIC_BASE_URL"),
+            "thinking": {"type": "disabled"},
             "max_tokens": 8192,
             "timeout": None,
             "max_retries": 0,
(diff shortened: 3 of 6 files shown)


agents/llm_config.py
```diff
@@ -173,12 +174,13 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
     ),
     "anthropic": LLMConfig(
         chat_class=ChatAnthropic,
-        selection_envs=["ANTHROPIC_API_KEY"],
+        selection_envs=["ANTHROPIC_API_KEY", "ANTHROPIC_BASE_URL"],
         api_key_env="ANTHROPIC_API_KEY",
-        agent_model="claude-sonnet-4-6",
+        agent_model="claude-sonnet-5",
```
Svilen-Stefanov, line 180: blocker-ish, and sorry because it's about the bit that isn't really the point of the PR :D

On `claude-sonnet-4-6`, omitting `thinking` means no thinking. On `claude-sonnet-5`, omitting it runs adaptive thinking, and langchain never sends the param (the field defaults to None), so we get thinking turned on implicitly just from the default bump.

Two things follow from that. Thinking tokens bill against `max_tokens`, which stays at 8192 right below, so the visible answer budget shrinks and with `max_retries: 0` a truncation reaches the extractor as malformed output rather than an error. And once thinking is on, the AIMessage content is a list of blocks rather than a string, so `agents/agent.py`'s `"".join(str(m) ...)` splices `{'type': 'thinking', 'thinking': '', 'signature': '<long base64>'}` straight into the text we hand to the parsing model.

Could we either pass `thinking={"type": "disabled"}` for this provider, or raise `max_tokens` and make that join thinking-aware?

agents/llm_config.py
```diff
@@ -26,11 +26,12 @@
 _OPENROUTER_FALLBACK_CONTEXT_WINDOW = ContextWindow(1_048_576, 65_536, is_fallback=True)
 
 # Model families that reject sampling params (temperature/top_p/top_k) with HTTP 400.
-# Why: Anthropic removed them starting with Opus 4.7; sending temperature (even 0) 400s.
+# Why: newer Anthropic models deprecate them; sending temperature (even 0) 400s.
 # Substrings match bare IDs and Bedrock-prefixed forms (e.g. us.anthropic.claude-opus-4-8-v1:0).
 _SAMPLING_PARAM_FREE_MODEL_SUBSTRINGS = (
     "claude-opus-4-7",
     "claude-opus-4-8",
+    "claude-sonnet-5",
```
Svilen-Stefanov, line 35: nit: while we're in here, `claude-opus-5` is missing and it isn't a substring of the 4-7 / 4-8 entries, so `_model_accepts_temperature("claude-opus-5")` returns True and anyone setting `AGENT_MODEL=claude-opus-5` gets temperature sent and a hard 400 with `max_retries: 0`. Can we add it here and to the parametrized list in the test?

Separately, this tuple grows every release. I wonder if we should flip it around and list the families that still accept sampling params instead, since that set is shrinking rather than growing. Not for this PR, but maybe worth a ticket?

agents/llm_config.py
```diff
@@ -173,12 +174,13 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
     ),
     "anthropic": LLMConfig(
         chat_class=ChatAnthropic,
-        selection_envs=["ANTHROPIC_API_KEY"],
+        selection_envs=["ANTHROPIC_API_KEY", "ANTHROPIC_BASE_URL"],
```
Svilen-Stefanov: I'm a bit worried about this one and I think it needs an explicit decision rather than just going in.

`ANTHROPIC_BASE_URL` is exported on a lot of dev machines already (Claude Code behind a corporate proxy or a LiteLLM gateway, often straight in the shell profile). With this change, someone whose actual key is `OPENAI_API_KEY` now gets `Multiple LLM providers selected (openai, anthropic); please set only one.` at bootstrap, which is pretty confusing when they only have one key set and never touched Anthropic.

And on the action / wrapper path there's no ambiguity check at all: `validate_api_key_provided()` is only called from `codeboarding_cli/bootstrap.py`, everything else goes straight to `_resolve_selected_provider`, which just takes the first selected entry, and anthropic sits above google/aws/cerebras/deepseek/... in the dict. So with `ANTHROPIC_BASE_URL` + `GOOGLE_API_KEY` set we'd silently switch provider and send `no-key-required` to the proxy.

Given the new check below says `ANTHROPIC_BASE_URL` alone is never a valid config anyway, can't we leave `selection_envs` as `["ANTHROPIC_API_KEY"]` and keep the base URL purely as an `extra_args` pass-through? The Foundry case still works since the key is required there.

agents/llm_config.py
```diff
@@ -434,6 +436,8 @@ def validate_api_key_provided() -> None:
             name,
             config.api_key_env,
         )
+    elif config.api_key_env and not config.has_real_api_key():
```
Svilen-Stefanov: just checking, was the wider blast radius intended here? This fires for any selected provider that has an `api_key_env` and isn't `keyless_capable`, which is vercel, deepseek, glm and kimi as well as anthropic.

So today setting only `DEEPSEEK_BASE_URL` (pointing at an unauthenticated local gateway) passes validation and falls through to the `no-key-required` placeholder, and after this it exits at bootstrap with `Provider 'deepseek' requires DEEPSEEK_API_KEY.` That might well be what we want, but it isn't in the PR description and there's no test for those four.

Also the docstring just above still says the sole-selected-provider case only logs a warning, could you update it while you're here?

agents/llm_config.py
```diff
@@ -173,12 +174,13 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
...
         chat_class=ChatAnthropic,
-        selection_envs=["ANTHROPIC_API_KEY"],
+        selection_envs=["ANTHROPIC_API_KEY", "ANTHROPIC_BASE_URL"],
         api_key_env="ANTHROPIC_API_KEY",
-        agent_model="claude-sonnet-4-6",
+        agent_model="claude-sonnet-5",
         parsing_model="claude-haiku-4-5",
         llm_type=LLMType.CLAUDE,
         extra_args={
+            "base_url": lambda: os.getenv("ANTHROPIC_BASE_URL"),
```
Svilen-Stefanov, line 184: just wanted to check whether this was tried against a real Foundry deployment, because the two new tests only assert the kwarg is forwarded.

`ChatAnthropic` builds a plain `anthropic.Client`, whose `_api_key_auth` sends `X-Api-Key`. The SDK's own `AnthropicFoundry` client overrides `_prepare_options` to send `api-key: <key>` or `Authorization: Bearer <entra token>` instead, which reads to me like the Azure gateway doesn't authenticate on `x-api-key`. If that's right we'd get a 401 on the first call and nothing here would catch it.

`ChatAnthropic` does take `default_headers`, so plumbing `api-key` through would be cheap if it turns out we need it. Either way I'd rather we confirm against a real deployment before we put the Azure Foundry claim in the README.

README.md
```diff
@@ -143,6 +144,8 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
 
 `openai_base_url` points CodeBoarding at any OpenAI-compatible gateway. Set the gateway API key as `openai_api_key`, then choose its model IDs with `agent_model` and `parsing_model`. The equivalent shell variables are `OPENAI_BASE_URL`, `OPENAI_API_KEY`, `AGENT_MODEL`, and `PARSING_MODEL`.
 
+`anthropic_base_url` points the Anthropic client at a compatible Messages API, including Azure Foundry Claude deployments. Set the deployment key as `anthropic_api_key` and use deployment names for `agent_model` and `parsing_model`.
```
Svilen-Stefanov: nit: the "use deployment names for `agent_model` and `parsing_model`" part worries me a little, because both of our Anthropic behaviours key off the model id string.

`_model_accepts_temperature` matches substrings, and `LLMType.from_model_name` only returns `CLAUDE` when the name contains claude/opus/sonnet/haiku. So someone who follows this literally and sets `agent_model = "prod-messages-v2"` gets temperature sent to a sonnet-5 deployment (exactly the 400 the substring change in this PR is meant to prevent) and silently falls through to the Gemini Flash prompt factory.

Doesn't the `/anthropic` route serve the canonical Anthropic ids? If it does, maybe we just tell people to use those.

PYPI.md
```diff
@@ -116,6 +116,7 @@ LLM provider keys and model overrides are stored in `~/.codeboarding/config.toml
 # Uncomment exactly one provider key
 # openai_api_key    = "sk-..."
 # anthropic_api_key = "sk-ant-..."
+# anthropic_base_url = "https://resource.services.ai.azure.com/anthropic"
```
Svilen-Stefanov: nit: this sits under "Uncomment exactly one provider key", but a base URL isn't a key, and with the new validation it now needs `anthropic_api_key` uncommented alongside it or the run fails. Can we word it the way README does for `openai_base_url`?

Svilen-Stefanov (commented):
Thanks for this, having a base URL for Anthropic makes sense and the config plumbing looks clean.

Most of my comments are actually about the default model bump rather than the base URL itself: sonnet-5 flips thinking on by default and claude-opus-5 is missing from the sampling list, so I'd like those sorted before this goes in. The selection_envs change also needs an explicit decision from us, I think it can bite people who already have ANTHROPIC_BASE_URL exported for Claude Code.

Also just checking, did anyone actually run this against a real Foundry deployment, or is it unit tests only so far?

ivanmilevtues:
I ran it against a foundry deployment and it works.

agents/llm_config.py
ivanmilevtues: It works against Azure's deployemtn yes, haven't tested against anthropic themselves as i don't have a key, but should be fine.

agents/llm_config.py
ivanmilevtues: thinking disabled.

agents/llm_config.py
ivanmilevtues: Good call — opened #518 to track replacing the release-by-release denylist with capability-based detection / an allowlist of families that still accept sampling parameters. This PR will stay scoped to adding `claude-opus-5` and its parametrized test case.

agents/llm_config.py
ivanmilevtues: Confirmed against a real Azure Foundry deployment: a full CodeBoarding run completed successfully using this configuration.

agents/llm_config.py
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): the Anthropic client now explicitly passes `thinking={"type": "disabled"}`, preserving the existing non-thinking response shape and token budget.

agents/llm_config.py
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): `claude-opus-5` is included in the sampling-parameter exclusion tuple and the parametrized test. #518 tracks replacing the release-by-release denylist with capability-based detection.

agents/llm_config.py
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): provider selection reads the full process environment after config values are injected, so `ANTHROPIC_BASE_URL` is now pass-through only and `ANTHROPIC_API_KEY` remains the sole Anthropic selector. OpenAI remains URL-selectable because it intentionally supports keyless compatible endpoints. Regression coverage verifies that a stray Anthropic URL does not conflict with another provider.

agents/llm_config.py
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): the generic API-key requirement was removed. Only an Anthropic URL with no selected provider gets the Anthropic-specific missing-key error; Vercel, DeepSeek, GLM, and Kimi keep their existing behavior. The validation docstring and regression tests now cover that scope.

README.md
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): the README now instructs users to provide canonical Anthropic model IDs such as `claude-sonnet-5`, ensuring temperature gating and prompt-family detection continue to work.

PYPI.md
ivanmilevtues: Addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b): `anthropic_base_url` is now documented separately from the exactly-one-provider-key list as an optional endpoint override that requires `anthropic_api_key`.

ivanmilevtues:
All review feedback from @Svilen-Stefanov has been addressed in [47d142a9793f4b48e254676fba6042f32459f49b](https://github.com/CodeBoarding/CodeBoarding/commit/47d142a9793f4b48e254676fba6042f32459f49b). Adaptive thinking is explicitly disabled, `claude-opus-5` is covered, the Anthropic base URL is pass-through-only, validation remains scoped, the real Azure Foundry run is confirmed, and the README/PyPI guidance is corrected. Follow-up #518 tracks the longer-term sampling-capability cleanup.

PR #521, opened 2026-08-26
Update DeepSeek defaults to V4 Flash

### LINKED ISSUE #520: Update discontinued DeepSeek default model IDs
## Context

The strict live-catalog integration check fails for both configured DeepSeek roles because `LLM_PROVIDERS` still uses `deepseek-chat`, which is absent from the current public catalog.

The Java/JDTLS failure originally reported here has been split into #523 and assigned to @Svilen-Stefanov.

## Finding

DeepSeek's official April 24, 2026 changelog states that `deepseek-chat` mapped to the non-thinking mode of `deepseek-v4-flash` and would be discontinued on July 24, 2026. The current official API documentation lists `deepseek-v4-flash`, `deepseek-v4-pro`, and `deepseek-v4-flash-vision-exp` as the supported production IDs.

Both the agent and parsing roles still configure the discontinued alias.

## Recommended direction

Replace both DeepSeek defaults with the supported production ID that preserves the old alias behavior: `deepseek-v4-flash`.

## Acceptance criteria

- Both configured DeepSeek roles use a current production model ID verified against DeepSeek's official documentation.
- Both roles pass the strict live-catalog/default validation.
- Existing DeepSeek model-family detection remains green for user-supplied model IDs.

## Pull request

- #521

## Reproduction environment

- CodeBoarding `81fc8941967f19ff488b204f139106785bebf352`
- Current keyless models.dev catalog

ivanmilevtues:
## Summary

- replace the discontinued `deepseek-chat` default with `deepseek-v4-flash`
- update both agent and parsing roles to preserve the old alias's non-thinking behavior
- keep legacy DeepSeek model-family detection intact for user-supplied IDs

The JDTLS readiness work has been split into #523 and a separate pull request.

## Root cause

DeepSeek's official April 24, 2026 changelog states that `deepseek-chat` mapped to `deepseek-v4-flash` and would be discontinued on July 24, 2026. Both CodeBoarding roles still configured that discontinued alias, so the strict current-catalog validation failed twice.

## Verification

- private live provider catalog suite — 17 passed
- private live model-capability suite — 64 passed
- `tests/agents/test_llm_config.py` — passed as part of the full suite
- full suite — 2,227 passed, 28 skipped, 83.80% coverage
- `mypy .` — no issues in 339 source files
- `black .` — 339 files unchanged

Closes #520

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -247,8 +247,8 @@ LLM_PROVIDERS = {
         chat_class=ChatOpenAI,
         selection_envs=["DEEPSEEK_API_KEY", "DEEPSEEK_BASE_URL"],
         api_key_env="DEEPSEEK_API_KEY",
-        agent_model="deepseek-chat",
-        parsing_model="deepseek-chat",
+        agent_model="deepseek-v4-flash",
+        parsing_model="deepseek-v4-flash",
         llm_type=LLMType.DEEPSEEK,
         extra_args={
             "base_url": lambda: os.getenv("DEEPSEEK_BASE_URL", "https://api.deepseek.com/v1"),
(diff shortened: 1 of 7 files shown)


Svilen-Stefanov (approved):
LGTM

PR #528, opened 2026-08-26
fix(relations): give the scope gate the whole hierarchy's ownership

Svilen-Stefanov:
**Alternative to #526. Same defect, same 17 edges removed, fixed at the source instead of at finalize. Only one of the two should merge.**

## The defect

An edge stored under a relation is evidence only if it runs between the components the relation declares. `edge_crosses_components` (`agents/relation_edges.py:128-141`) is exactly that predicate, and it already runs inside `merge_scope_relations`. The bug is the **map** it is given.

`merge_scope_relations` built `ComponentOwnershipIndex.from_analysis(analysis)` — indexed from that one scope's own components. `owner_of` returns `""` for anything it cannot uniquely resolve, and the guard reads `if owner and component_id and not is_self_or_descendant(...)`, so **a blank owner keeps the edge**. An endpoint owned anywhere else in the tree was therefore never judged.

Measured: the per-scope map rejects **2** of the 17 edges the eval corpus flags. The other **15 are structurally invisible to it**.

What ships today, from `click-before-get-strerror-removal` scope 5 — relation `5.3 → 5.1`, *"Command metadata feeds parser configuration"*, with four code references attached as backing:

| cited edge | caller owned by | callee owned by |
|---|---|---|
| `decorators._make_command -> core.Command` | 5.3 ✓ | **2.2.2** ✗ |
| `decorators.option -> core.Option` | **1.3.3** ✗ | **1.3.2** ✗ |
| `decorators.argument -> core.Argument` | **1.3.3** ✗ | **1.1** ✗ |
| `core.Command -> parser.OptionParser` | **2.2.2** ✗ | 5.1 ✓ |

None connects the pair it is attached to; two have neither endpoint in either component.

## The map already exists, before any LLM call

- `ClusteringService._cluster_hierarchy` ends `root.index_hierarchy(); return root`, and **both** `build_full_hierarchy` and `build_incremental_hierarchy` route through it.
- `index_hierarchy` recurses `group.children` and fills `clustering_groups` at **every depth**. On a real serilog pickle: 41 groups across depths 1-3, 823 distinct `qualified_names` against 840 CFG nodes.
- `static_analysis_enricher_mixin.py:65` — `component.component_id = group.group_id`. **Component ids are group ids.** The LLM labels fixed groups; it never creates or renumbers them.
- `build_full_hierarchy` is called at `diagram_generator.py:629`; the abstraction agent does not run until `:1231`.

So a complete, full-depth symbol → component map is available before the first scope agent runs. There is no need to repair the artifact afterwards.

## The change

**7 source files, +114/−33** (plus tests).

- **`agents/component_ownership.py`** — `from_clustering_hierarchy(hierarchy)` iterates `clustering_groups` sorted by `group_id.count(".")` ascending so the **deepest** group wins, giving one owner per symbol. A flat union over all depths would leave nearly every symbol with several owners and blank `owner_of` out entirely — strictly worse than today; there is a test pinning that. `from_analysis` is **deleted** along with its last caller. `from_node_owners` stays; `rebuild_global_relations` and `build_component_relations` still use it, and the new constructor delegates to it.
- **`agents/static_analysis_enricher_mixin.py`** — `merge_scope_relations` and `prune_relations` become instance methods reading `self.component_ownership`. The per-call `from_analysis` at `:107` and `:193` is gone. The predicate call itself is unchanged.
- **`diagram_analysis/diagram_generator.py`** — `_initialize_agents` builds the index **once per run** and injects it into `DetailsAgent`, `AbstractionAgent`, `IncrementalAgent` as a **required** kwarg, so a missed construction site is a `TypeError` rather than a silently dead gate. `IncrementalAgent._clone_for_worker` forwards it.

The three `merge_scope_relations` call sites did not change textually — the root hierarchy reaches the agents through construction, not through the scope argument. No global, no threading through `run()`.

`edge_crosses_components` and `is_self_or_descendant` are untouched, and no downstream filter is added.

## Measured impact

Re-scored with the evals harness's own check code over the 16 committed full-run `analysis.json` in CodeBoarding-evals:

| | baseline | this PR | #526 |
|---|---:|---:|---:|
| `relations.edges_cross_declared_components` | **17** | **0** | 0 |
| `relations.edge_endpoints_are_known_symbols` | 12 | 12 | 12 |
| edges removed | — | **17** | 17 |
| non-flagged edges lost | — | **0** | 0 |
| relations deleted | — | 0 | 0 |
| edgeless inferred relations (of 162) | 75 | 82 | 82 |
| relations with neither edges nor prose | 0 | **0** | 0 |

**It removes the same 17 edges as #526 — set identity, verified in both directions**, not merely the same count. Every removal lands on an `is_static:false` relation; no CFG-derived edge is touched. All other registered checks stay at 0.

The map strictly dominates the one it replaces. Over all 13,834 `all_edges` endpoints:

| | count |
|---|---:|
| blank under old map | 61 |
| blank under new map | 42 |
| old blank → new resolved | 19 |
| **old resolved → new blank** | **0** |
| **both resolve, different branch** | **0** |
| keep→reject flips at sub-scopes | 19 endpoints → 17 edges |
| **reject→keep flips** | **0** |

It never contradicts the old map — it only refines or fills in. That is the strongest available evidence that moving to it cannot regress a judgement the current code gets right.

## Tests

`3 failed, 2226 passed, 28 skipped` against baseline's `3 failed, 2220 passed` — the same three pre-existing `test_cli_parser` failures (a macOS `/private/tmp` symlink artifact; they should pass on Linux CI). mypy clean on 340 files, black clean.

New tests cover the deepest-owner rule, the ancestor-ambiguity trap, injection into all three agents, and the gate rejecting an out-of-branch endpoint.

## Review points

1. **`--target-component` builds a partial map.** `deterministic_analysis` (`:633-635`) constructs a bare `ClusterScopeResult` and registers only the target subtree, so `clustering_groups` covers that subtree alone. Harmless here — out-of-subtree endpoints blank and are kept, which is today's behaviour — but the "every analysed symbol has an owner" premise is false on that path. Worth either building a full-repo map there or having the constructor refuse a hierarchy that does not cover the CFG. **I would take this as a follow-up, not a blocker, but it should be a conscious decision.**
2. **`prune_relations` gets a behaviour change beyond the stated defect.** It now uses the full map, so `find_relation_pair_for_edge` resolves endpoints it previously blanked and an edge can be classified internal and dropped. Bounded to the same 17 edges; on this corpus it is **one extra deletion** (2 → 3 dropped as internal), further gated by `changed_members`.
3. **Absorption is safe, analysed rather than assumed.** `absorb_single_child_components` collapses only a single-component scope and `reroot_indexes` raises on id collision, so ids are rewritten along one ancestor chain and sibling branches never merge — a different-branch verdict is preserved. Both consumers run during agent execution, before `finalize_for_save`.
4. **`assert self.clustering_hierarchy is not None`** at `:1049` is control flow via assert, stripped under `python -O`.
5. **Scope.** This fixes **misattribution**, not invention. It cannot reach `edge_endpoints_are_known_symbols` by construction — an invented endpoint resolves to `""` and the guard keeps it. See below.

## What this deliberately does not include

I built and measured a reject-unknown rule (drop an edge whose endpoint matches no indexed symbol) and **removed it before opening this PR**. It drove criterion 2 to 0, but **4 of its 12 drops are real Serilog methods**: `candidates_by_leaf` keys on `rsplit(".", 1)[-1]`, and 707 of serilog's 840 CFG keys carry a parameter list, so the key is `Equals(object x, object y)` while a signature-less reference keys on `Equals` and misses. `ByReferenceStringComparer.Equals` exists twice in the CFG and the rule would have deleted it.

The useful finding stands and is worth its own issue: **the baseline 12 is 8 real inventions plus 4 measurement artifacts**, and the cheaper first move on criterion 2 is fixing the suffix matcher (`_qualified_names_match` / `candidates_by_leaf`) to roll up class↔member and tolerate signatures — which would also recover a real CFG edge currently flagged as ungrounded.

## Summary by CodeRabbit

* **Improvements**
  * Improved component ownership tracking across nested project hierarchies.
  * More accurately filters and preserves relationship data based on ownership and scope.
  * Ensured ownership information remains consistent during parallel and incremental processing.

* **Tests**
  * Added coverage for nested ownership resolution, ambiguous memberships, and scope-based relationship filtering.
  * Updated agent and relationship tests to reflect the enhanced ownership handling.

### DIFF
--- a/agents/abstraction_agent.py
+++ b/agents/abstraction_agent.py
@@ -13,7 +13,7 @@ from agents.agent_responses import (
     MetaAnalysisInsights,
     assign_relation_ids,
 )
-from agents.component_ownership import group_ids_by_name
+from agents.component_ownership import ComponentOwnershipIndex, group_ids_by_name
 from agents.llm_renderers import (
     cluster_group_descriptions,
     cluster_group_ids,
@@ -58,10 +58,12 @@ class AbstractionAgent(StaticAnalysisEnricherMixin, CodeBoardingAgent):
         meta_context: MetaAnalysisInsights,
         agent_llm: BaseChatModel,
         parsing_llm: BaseChatModel,
+        component_ownership: ComponentOwnershipIndex,
     ):
         system_message = format_project_system_message(get_system_message(), project_name, meta_context)
         super().__init__(repo_dir, static_analysis, system_message, agent_llm, parsing_llm)
 
+        self.component_ownership = component_ownership
         self.project_name = project_name
         self.meta_context = meta_context
 
--- a/agents/component_ownership.py
+++ b/agents/component_ownership.py
@@ -5,7 +5,8 @@ from __future__ import annotations
 from collections.abc import Collection
 from dataclasses import dataclass
 
-from agents.agent_responses import AnalysisInsights, Component, SourceCodeReference
+from agents.agent_responses import Component, SourceCodeReference
+from static_analyzer.clustering import ClusterScopeResult
 
 
 def group_ids_by_name(components: list[Component], valid_group_ids: Collection[str]) -> dict[str, str]:
@@ -26,15 +27,21 @@ class ComponentOwnershipIndex:
     candidates_by_leaf: dict[str, list[tuple[str, str]]]
 
     @classmethod
-    def from_analysis(cls, analysis: AnalysisInsights) -> ComponentOwnershipIndex:
-        candidates_by_leaf: dict[str, list[tuple[str, str]]] = {}
-        for component in analysis.components:
-            for file_methods in component.file_methods:
-                for method in file_methods.methods:
-                    candidates_by_leaf.setdefault(method.qualified_name.rsplit(".", 1)[-1], []).append(
-                        (method.qualified_name, component.component_id)
-                    )
-        return cls(candidates_by_leaf)
+    def from_clustering_hierarchy(cls, hierarchy: ClusterScopeResult) -> ComponentOwnershipIndex:
+        """Index every symbol at every depth against the deepest group that owns it.
+
+        Why: component IDs are group IDs, so the hierarchy already knows the owner of every
+        symbol in the repo, not only the ones in the scope an agent happens to be naming. A
+        symbol belongs to its leaf group and to every ancestor of it, so a flat union over
+        all depths would leave almost every symbol with several owners and blank out
+        ``owner_of``; the deepest group is the single answer, the same one
+        ``build_global_node_to_component_map`` gives for the expanded frontier.
+        """
+        owner_by_symbol: dict[str, str] = {}
+        for group_id, group in sorted(hierarchy.clustering_groups.items(), key=lambda item: item[0].count(".")):
+            for qualified_name in group.qualified_names:
+                owner_by_symbol[qualified_name] = group_id
+        return cls.from_node_owners(owner_by_symbol)
 
     @classmethod
     def from_node_owners(cls, node_owners: dict[str, str]) -> ComponentOwnershipIndex:
--- a/agents/details_agent.py
+++ b/agents/details_agent.py
@@ -14,7 +14,7 @@ from agents.agent_responses import (
     MetaAnalysisInsights,
     assign_relation_ids,
 )
-from agents.component_ownership import group_ids_by_name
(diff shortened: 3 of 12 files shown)


ivanmilevtues:
#526 merged this one goes.

PR #536, opened 2026-08-31
Export Core LLM environment variable contract

ivanmilevtues:
## Summary
- model each provider's endpoint environment variable and built-in default directly on `LLMConfig`
- resolve `base_url` from that provider metadata while preserving SDK-default behavior when no URL is configured
- derive `LLM_PROVIDER_ENV_VARS` entirely from provider selectors, API-key variables, and `base_url_env`
- export the complete provider and model subprocess contract as `LLM_ENV_VARS`

## Verification
- `uv run pytest -q tests/agents/test_llm_config.py tests/test_user_config.py` — 162 passed
- `uv run black --check agents/llm_config.py tests/agents/test_llm_config.py`
- `uv run mypy agents/llm_config.py`
- Anthropic constructor without a base URL resolves to `https://api.anthropic.com`

## Summary by CodeRabbit

* **New Features**
  * Added centralized configuration for LLM provider base URLs, including environment-variable overrides and provider-specific defaults.
  * Expanded supported base URL configuration for all available providers, including Anthropic, OpenRouter, and OrcaRouter.

* **Tests**
  * Added coverage confirming provider configuration variables are accurately recognized.
  * Added validation for default and environment-provided base URL resolution.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -333,6 +333,24 @@ LLM_PROVIDERS = {
 }
 
 
+# Endpoint overrides that affect an already-selected provider rather than
+# selecting one themselves.
+_PROVIDER_ENDPOINT_OVERRIDE_ENV_VARS = frozenset(
+    {
+        "ANTHROPIC_BASE_URL",
+        "OPENROUTER_BASE_URL",
+        "ORCAROUTER_BASE_URL",
+    }
+)
+
+LLM_PROVIDER_ENV_VARS: frozenset[str] = (
+    frozenset(var for config in LLM_PROVIDERS.values() for var in config.selection_envs)
+    | frozenset(config.api_key_env for config in LLM_PROVIDERS.values() if config.api_key_env)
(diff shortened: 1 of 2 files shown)


agents/llm_config.py
```diff
@@ -139,6 +143,10 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
             value = v() if callable(v) else v
             if value is not None:
                 resolved[k] = value
+        if self.base_url_env:
+            base_url = os.getenv(self.base_url_env, self.default_base_url or None)
```
ivanmilevtues, line 147: I double checked if self.default_base_url is set to an empty string we will resolve this to None:
```python
>>> n = os.getenv("non_exist", "" or None)
>>> print(n)
None
>>> n is None
True
>>> n == ""
False
>>> ```

Svilen-Stefanov (approved):
Thanks for fixing this!

PR #539, opened 2026-08-31
feat(clustering): partition by qualified name

Svilen-Stefanov:
Top-level components now come from **what files are called**, not from who calls whom.

Stacked: #540 (qualified names) → **#539** → #542 (delete the graph partitioner).

## How it works

1. **One leaf per file.** Leiden's communities straddle component boundaries — on Beacon 7 of 21 held symbols from two different components, capping *any* grouping over them at 0.34 where the same names over files reach 0.94.
2. **The LLM reads a naming model, once per full analysis** — component names, the vocabulary each owns, and a machinery list. Stored in `analysis.json`.
3. **A gate picks which half of the name carries the architecture.** If the model's components name the directories, the **structural** half leads; if they contradict them, the **lexical** half does. eShop matched 8 of 20 scopes → structural. A layered repo (`Application`, `Infrastructure`, `Api`) matches 0 → lexical, because grouping *that* by directory just reproduces the layering, which scores below not partitioning at all.
4. **Assignment is arithmetic over strings — no edge is read.** The call graph still supplies every arrow, and still shapes everything below depth 1.

The two halves read **different inputs**: structural reads the *file path* and only the outermost directory (`src/A/B/C/x.cs` → `A`); lexical reads the *qualified name*, where every segment votes, weighted `2^(pos − last)` within its segment so the head noun wins — `IncidentResolvedMetricsHandler` → Analytics, not Incidents.

## How depth is decided — and why the formula no longer fits

Depth is not a fixed number. A component expands if `load ≥ 1.0` (>12 files or >120 methods) **or** it splits into ≥2 groups **and** has ≥30 methods **and** modularity ≥ `0.15 × (1 − load)`. The bar *falls as a component grows*. `--depth-level` is a ceiling, never a target.

On CodeBoarding this stops at depth 2 for 7 of 13: three forced open by size, three cleared a low bar; `Caching` (23 methods) and `Plugin Registry` (16) are under the 30-method floor and can never expand, while `Health Checks`, `Output Generators`, `Telemetry` and `Tool Registry` passed it but were too cohesive to split.

**This gate was not updated for the rewrite, and it asks a graph question.** `modularity(name_groups, call_graph)` was self-consistent while Leiden produced the groups — Leiden maximises exactly that quantity. Now the groups come from names, so modularity is an external judge that was never asked to agree: a correct name split in a chatty codebase scores low and gets suppressed.

The name-native replacement asks what the partition actually answers — does the next level of names divide this component, and does the division carry mass? Both are already computed before the gate runs:

> expand if `load ≥ 1.0`, or if `keys ≥ 2` **and** `(files − largest_child) / files ≥ 0.25 × (1 − load)`

Same shape, measuring name dispersion instead of edge density.

## Depth as recursion

Sub-scopes are Leiden at every depth today, **on full runs too** — names shape only the root. The intended design is that depth *N* is depth 1 applied to a sub-tree: re-run the gate for that scope, partition by the winning half, expand if the split carries mass.

Applying the depth-1 structural rule one level down on CodeBoarding shows why the gate must be **per scope**, not once per repo:

| component | files | keys at depth 2 | outside largest | shipped |
|---|---:|---:|---:|---|
| Static Analysis Engine | 55 | 5 (`engine`, `clustering`, `cfg`, `lsp_client`) | 56% | expanded |
| LLM Agents | 46 | 4 (`tools`, `prompts`, `llm_renderers`) | 52% | expanded |
| Analysis Orchestration | 10 | 3 (`commands`, `sources`) | 50% | expanded |
| Health Checks | 6 | 2 (`checks`) | 50% | **leaf** |
| Diagram Analysis | 12 | **1** — all files sit directly in the dir | 0% | expanded |
| Application Configuration | 10 | **1** | 0% | expanded |
| Repository Utilities | 7 | **1** | 0% | expanded |
| Execution Monitoring | 6 | **1** | 0% | expanded |

Four of the seven currently-expanded components have **nothing structural to split on** — Leiden expanded them on call-graph communities alone. That is not a reason to stop: `diagram_analysis/` has no sub-directories, but `diagram_generator`, `file_index`, `file_coverage`, `scope_plan` carry plenty of vocabulary. One structural key means *go lexical here*, not *stop*.

## Results

**eShop — pair-F1 = 1.0000** vs `codeboarding_evals/reference/eshop-depth-1.mmd` (P 1.0000, R 1.0000, 390 files, all-in-one floor 0.3311). Each of the 9 reference boxes maps to exactly one component. The 3 extra components are directories the reference doesn't model. *Caveat: 77 of 467 files sit outside the reference; it judges depth 1 only.*

```mermaid
graph LR
    Basket_API["Basket API"]
    Catalog_API["Catalog API"]
    Mobile_Client_App["Mobile Client App"]
    Aspire_AppHost_and_Service_Defaults["Aspire AppHost and Service Defaults"]
    Event_Bus["Event Bus"]
    Hybrid_App["Hybrid App"]
    Identity_API["Identity API"]
    Integration_Event_Log["Integration Event Log"]
    Ordering_Service["Ordering Service"]
    Payment_Processor["Payment Processor"]
    Web_Application["Web Application"]
    Webhook_Client["Webhook Client"]
    Basket_API -- "calls" --> Aspire_AppHost_and_Service_Defaults
    Payment_Processor -- "Publishes payment result events" --> Event_Bus
    Web_Application -- "Calls basket gRPC/REST endpoints" --> Basket_API
    Web_Application -- "Queries catalog REST API" --> Catalog_API
    Web_Application -- "Authenticates users via OIDC/OAuth" --> Identity_API
    Web_Application -- "Submits orders via REST API" --> Ordering_Service
    Web_Application -- "Invokes shared catalog utilities" --> Hybrid_App
    Web_Application -- "calls" --> Aspire_AppHost_and_Service_Defaults
    Webhook_Client -- "calls" --> Aspire_AppHost_and_Service_Defaults
    Catalog_API -- "Persists events via transactional outbox" --> Integration_Event_Log
    Mobile_Client_App -- "Manages basket via gRPC/REST" --> Basket_API
    Mobile_Client_App -- "Browses catalog via REST" --> Catalog_API
    Mobile_Client_App -- "Authenticates via Identity API" --> Identity_API
    Mobile_Client_App -- "Places orders via REST" --> Ordering_Service
    Event_Bus -- "Dispatches order-started events" --> Basket_API
    Event_Bus -- "Dispatches stock-confirmed events" --> Payment_Processor
    Event_Bus -- "Dispatches payment confirmation events" --> Ordering_Service
    Hybrid_App -- "Calls basket gRPC/REST endpoints" --> Basket_API
    Hybrid_App -- "Queries catalog REST API" --> Catalog_API
    Hybrid_App -- "Authenticates via Identity API" --> Identity_API
    Hybrid_App -- "Submits orders via REST" --> Ordering_Service
    Integration_Event_Log -- "Publishes integration events" --> Event_Bus
    Ordering_Service -- "Persists events via transactional outbox" --> Integration_Event_Log
    Ordering_Service -- "calls" --> Aspire_AppHost_and_Service_Defaults
    click Catalog_API href "./Catalog_API.md" "Details"
    click Mobile_Client_App href "./Mobile_Client_App.md" "Details"
    click Hybrid_App href "./Hybrid_App.md" "Details"
    click Identity_API href "./Identity_API.md" "Details"
    click Ordering_Service href "./Ordering_Service.md" "Details"
    click Web_Application href "./Web_Application.md" "Details"
    click Webhook_Client href "./Webhook_Client.md" "Details"
```

**CodeBoarding** — 13 components, each resolving to exactly one source directory, 94% coverage.

```mermaid
graph LR
    Application_Configuration["Application Configuration"]
    LLM_Agents["LLM Agents"]
    Caching_Layer["Caching Layer"]
    Analysis_Orchestration["Analysis Orchestration"]
    Plugin_Registry["Plugin Registry"]
    Diagram_Analysis["Diagram Analysis"]
    Health_Checks["Health Checks"]
    Execution_Monitoring["Execution Monitoring"]
    Output_Generators["Output Generators"]
    Repository_Utilities["Repository Utilities"]
    Static_Analysis_Engine["Static Analysis Engine"]
    Telemetry["Telemetry"]
    Tool_Registry["Tool Registry"]
    Application_Configuration -- "Invokes static health checks" --> Static_Analysis_Engine
    Application_Configuration -- "Bootstraps external tools" --> Tool_Registry
    Application_Configuration -- "Manages remote repository lifecycle" --> Repository_Utilities
    Application_Configuration -- "Dispatches analysis pipelines" --> Analysis_Orchestration
    Application_Configuration -- "Resolves analysis metadata" --> Diagram_Analysis
    Application_Configuration -- "Renders documentation outputs" --> Output_Generators
    Application_Configuration -- "Triggers health diagnostics" --> Health_Checks
    Repository_Utilities -- "Provides source tree hashes" --> LLM_Agents
    Repository_Utilities -- "Supplies change fingerprints" --> Diagram_Analysis
    Static_Analysis_Engine -- "Reads analysis configuration" --> Application_Configuration
    Static_Analysis_Engine -- "Accesses repository file system" --> Repository_Utilities
    Static_Analysis_Engine -- "Emits analysis telemetry" --> Telemetry
    Static_Analysis_Engine -- "Provisions language servers and tools" --> Tool_Registry
    Static_Analysis_Engine -- "Builds component relations via agent models" --> LLM_Agents
    Telemetry -- "Forwards token usage stats" --> Execution_Monitoring
    Tool_Registry -- "Resolves executable paths" --> Application_Configuration
    LLM_Agents -- "calls" --> Application_Configuration
    LLM_Agents -- "Traces agent execution steps" --> Execution_Monitoring
    LLM_Agents -- "calls" --> Repository_Utilities
    LLM_Agents -- "Consumes static analysis artifacts" --> Static_Analysis_Engine
    LLM_Agents -- "Populates diagram indexes" --> Diagram_Analysis
    LLM_Agents -- "Caches metadata and model settings" --> Caching_Layer
    LLM_Agents -- "calls" --> Plugin_Registry
    Caching_Layer -- "Resolves cache directories" --> Application_Configuration
    Caching_Layer -- "calls" --> Repository_Utilities
    Caching_Layer -- "calls" --> LLM_Agents
    Analysis_Orchestration -- "Consumes global configuration" --> Application_Configuration
    Analysis_Orchestration -- "Introspects repository state" --> Repository_Utilities
    Analysis_Orchestration -- "Configures probabilistic analysis layer" --> LLM_Agents
    Analysis_Orchestration -- "Discovers and loads plugins" --> Plugin_Registry
    Analysis_Orchestration -- "Manages analysis artifact lifecycle" --> Diagram_Analysis
    Analysis_Orchestration -- "Instruments pipeline execution" --> Execution_Monitoring
    Analysis_Orchestration -- "Triggers documentation generation" --> Output_Generators
    Diagram_Analysis -- "calls" --> Application_Configuration
    Diagram_Analysis -- "Normalizes paths and change sets" --> Repository_Utilities
    Diagram_Analysis -- "Orchestrates deterministic analysis" --> Static_Analysis_Engine
    Diagram_Analysis -- "Reports analysis progress" --> Telemetry
    Diagram_Analysis -- "Enriches artifacts with semantic insights" --> LLM_Agents
    Diagram_Analysis -- "Triggers repository health diagnostics" --> Health_Checks
    Diagram_Analysis -- "Initializes monitoring streams" --> Execution_Monitoring
    Health_Checks -- "Applies repository ignore rules" --> Repository_Utilities
    Health_Checks -- "Inspects code structure" --> Static_Analysis_Engine
    Health_Checks -- "Executes plugin health checks" --> Plugin_Registry
    Execution_Monitoring -- "calls" --> Application_Configuration
    Output_Generators -- "Sanitizes output content" --> Application_Configuration
    Output_Generators -- "Resolves language labels" --> Static_Analysis_Engine
    Output_Generators -- "Incorporates LLM-generated narratives" --> LLM_Agents
    Output_Generators -- "Consumes unified analysis JSON" --> Diagram_Analysis
    click Application_Configuration href "./Application_Configuration.md" "Details"
    click LLM_Agents href "./LLM_Agents.md" "Details"
    click Analysis_Orchestration href "./Analysis_Orchestration.md" "Details"
    click Diagram_Analysis href "./Diagram_Analysis.md" "Details"
    click Execution_Monitoring href "./Execution_Monitoring.md" "Details"
    click Repository_Utilities href "./Repository_Utilities.md" "Details"
    click Static_Analysis_Engine href "./Static_Analysis_Engine.md" "Details"
```

## The two halves fail in opposite directions

They look like the same idea. They are not: they read different inputs, and each breaks where the other works. Every file assigned both ways on three real repos:

| repo | files | scopes | gate | disagree | why |
|---|---:|---:|---|---:|---|
| Modulify | 28 | 4 | structural | **0%** | each feature has its own directory *and* its own vocabulary — either half works |
| eShop | 527 | 21 | structural | **21%** | domain words leak across service boundaries |
| Beacon | 111 | 6 | lexical | **70%** | every feature is spread across four layer directories |

Modulify is the case where they agree completely — which is why they look interchangeable.

**eShop, where lexical is wrong.** `src/WebApp/Services/BasketService.cs` → structural `Web App`, lexical `Basket`. It is the web app's *client* for the basket service; the word `Basket` is there because of what it talks to. Lexical also swept `Program.cs` / `Extensions.cs` / `GlobalUsings.cs` from every service into one Infrastructure bucket, collapsing 10 structural boxes into 1 (70% coverage).

**Beacon, where structural is wrong.** `Beacon.Application/Analytics/IncidentResolvedMetricsHandler.cs` → structural `Beacon.Application`, lexical `Analytics`. The outermost directory is a layer, so structural puts Analytics, Escalation, Alerts and Incidents in one box.

> **Rule of thumb.** Structural fails when *one directory holds many domains* (a layer). Lexical fails when *one domain word appears in many directories* (a client, UI, test, or shared contract naming what it talks to).

This is the sharpest argument for a **per-scope** gate: eShop is structural at depth 1 but would need lexical at depth 2 inside `Ordering.API`, whose sub-directories are `Application` and `Infrastructure` — Beacon's problem one level down, in a repo whose top level had the opposite shape.

## Where a per-layer decision must live

**Not implemented.** The recursion is designed, not built: the gate runs once over top-level scopes, and sub-scopes never use names.

When built, the decision cannot be recomputed per run — it depends on the scope set, which changes as files are added, so a repo could flip structural→lexical mid-life and reshuffle every box. It goes in `analysis.json` metadata beside the model, keyed by scope id (already the component id):

```json
"naming_model": {
  "components": [...],          // global — a repo has one vocabulary
  "machinery":  [...],          // global
  "decisions": {                // per scope, replayed verbatim
    "root": {"half": "structural", "matched": 0.40},
    "11":   {"half": "lexical",    "matched": 0.00},
    "11.2": {"half": "structural", "matched": 0.67}
  }
}
```

Keeping `matched` makes a surprising box explainable after the fact. The static-analysis pickle is the wrong home: it caches derived facts about code, and this is a presentation decision that must outlive cache invalidation.

## Incremental — read this before approving

The naming model is loaded on an incremental run, is **required**, and is then **never consulted**. `_anchored_group` carries ownership forward and places new files by the call graph; a new component appears only if Leiden isolates ≥2 new leaf clusters, so **a single new file can never create one**. A drift budget can re-derive the whole scope from the modularity peak, silently discarding the name-derived shape.

Sub-scopes are Leiden at every depth, **on full runs too** — names shape only the root scope today.

**Making incremental add a component is mostly deterministic.** A component key *is* a directory's distinctive word, so the component set is a pure function of the directory set: a new directory whose word matches no existing key **is** a new component; one that loses its last file **is** retired. No model call. Only the lexical half needs the LLM, and only for files that drew no votes. Generalising to any depth needs `scope_of` to return the *d*-th path segment rather than `parts[0]`, plus the gate evaluated **per scope** — inside `Ordering.API` the second level is `Application`/`Infrastructure`, layers again, so depth 2 there must go lexical.

Also missing: `NamingModel.to_dict` persists components and machinery but **not** the structural/lexical decision, so it would be recomputed and could flip a repo mid-life.

## Known gaps

- `NamePartition.coverage` reaches only a `logger.info`, never a caller — and that log is gated off exactly when everything falls into `Infrastructure`.
- A scope whose every word is machinery falls back to its raw mixed-case name, which can never collide with a lowercase stem — a guaranteed singleton component that still counts as "placed".

## Verified

`pytest` 2278 passed; mypy and black clean. Private suite green on all 12 jobs (8 languages × 3 platforms) via CodeBoarding-tests#25, which pairs with the #542 branch and therefore covers this PR plus the deletion.

### DIFF
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -661,6 +661,28 @@ class MetaAnalysisInsights(LLMBaseModel):
         return title + content
 
 
+class ComponentNaming(LLMBaseModel):
+    """One component the naming model proposes, and the vocabulary it owns."""
+
+    name: str = Field(description="One responsibility, never two joined by 'and'")
+    owns: list[str] = Field(description="Domain words naming this component, as they appear in identifiers")
+
+    def llm_str(self):
+        return f"{self.name}: {', '.join(self.owns)}"
+
+
+class NamingModelInsights(LLMBaseModel):
+    """How a repo's own vocabulary divides it, decided once per full analysis."""
+
+    components: list[ComponentNaming] = Field(description="The top-level components the identifiers name")
+    machinery: list[str] = Field(description="Words naming how software is built rather than what this system is about")
+    notes: str = Field(default="", description="What the vocabulary says about the repo's shape")
+
+    def llm_str(self):
+        lines = [f"- {component.llm_str()}" for component in self.components]
+        return "# Naming model\n" + "\n".join(lines) + f"\n\nMachinery: {', '.join(self.machinery)}\n{self.notes}"
+
+
 class FileClassification(LLMBaseModel):
     """Classification of a file to a component."""
 
--- a/agents/naming_model_agent.py
+++ b/agents/naming_model_agent.py
@@ -1,0 +1,88 @@
+"""Read a repo's components out of its own identifiers, once per full analysis."""
+
+import logging
+from collections import Counter
+from pathlib import Path
+
+from langchain_core.language_models import BaseChatModel
+from langchain_core.prompts import PromptTemplate
+
+from agents.agent import CodeBoardingAgent
+from agents.agent_responses import NamingModelInsights
+from agents.prompts import get_naming_model_message, get_system_message
+from monitoring import trace
+from static_analyzer.analysis_result import StaticAnalysisResults
+from static_analyzer.clustering.naming import ComponentVocabulary, NamingModel, scope_of, tokenize
+
+logger = logging.getLogger(__name__)
+
+VOCABULARY_SAMPLE = 120
+IDENTIFIER_SAMPLE = 120
+
+
+class NamingModelAgent(CodeBoardingAgent):
+    """Decides which components a repo's vocabulary names, and which words are machinery.
+
+    Run on a full analysis only. An incremental run reuses the stored answer, so the
+    partition cannot move underneath unchanged code.
+    """
+
+    def __init__(
+        self,
+        repo_dir: Path,
+        static_analysis: StaticAnalysisResults,
+        agent_llm: BaseChatModel,
+        parsing_llm: BaseChatModel,
+    ):
+        super().__init__(repo_dir, static_analysis, get_system_message(), agent_llm, parsing_llm)
+        self.prompt = PromptTemplate(
+            template=get_naming_model_message(),
+            input_variables=["scopes", "vocabulary", "identifiers"],
+        )
+
+    @trace
+    def read_naming_model(self) -> NamingModel:
+        scopes, vocabulary, identifiers = self._evidence()
+        if not identifiers:
+            logger.warning("[NamingModelAgent] No identifiers to read; returning an empty model")
+            return NamingModel(components=(), machinery=frozenset())
+
+        insights = self._parse_invoke(
+            self.prompt.format(
+                scopes="\n".join(scopes),
+                vocabulary=", ".join(f"{word}({count})" for word, count in vocabulary),
+                identifiers="\n".join(identifiers),
+            ),
+            NamingModelInsights,
+        )
+        model = NamingModel(
+            components=tuple(
+                ComponentVocabulary(component.name, tuple(component.owns)) for component in insights.components
+            ),
+            machinery=frozenset(insights.machinery),
+        )
+        logger.info(
+            "[NamingModelAgent] %d components, %d machinery words, scopes lead: %s",
+            len(model.components),
+            len(model.machinery),
+            model.scopes_are_components(set(scopes)),
+        )
+        return model
+
+    def _evidence(self) -> tuple[list[str], list[tuple[str, int]], list[str]]:
+        """The scopes, the commonest words, and a sample of identifiers. Never the answer."""
+        scopes: set[str] = set()
+        words: Counter[str] = Counter()
+        identifiers: set[str] = set()
+        for node in self.static_analysis.iter_reference_nodes():
+            scope = scope_of(node.file_path)
+            if scope:
+                scopes.add(scope)
+            name = node.fully_qualified_name.rsplit(".", 1)[-1]
+            identifiers.add(name)
+            words.update(tokenize(name))
+        return (
+            sorted(scopes),
+            words.most_common(VOCABULARY_SAMPLE),
+            sorted(identifiers)[:IDENTIFIER_SAMPLE],
+        )
--- a/agents/naming_model_agent.py
+++ b/agents/naming_model_agent.py
@@ -23,8 +23,8 @@ IDENTIFIER_SAMPLE = 120
 class NamingModelAgent(CodeBoardingAgent):
     """Decides which components a repo's vocabulary names, and which words are machinery.
 
-    Run on a full analysis only. An incremental run reuses the stored answer, so the
-    partition cannot move underneath unchanged code.
+    Run on a full analysis only; an incremental run reads the stored answer back rather than
+    re-asking, so the vocabulary cannot move underneath unchanged code.
     """
 
     def __init__(
@@ -70,12 +70,16 @@ class NamingModelAgent(CodeBoardingAgent):
         return model
 
     def _evidence(self) -> tuple[list[str], list[tuple[str, int]], list[str]]:
-        """The scopes, the commonest words, and a sample of identifiers. Never the answer."""
+        """The scopes, the commonest words, and a sample of identifiers. Never the answer.
+
+        Why the root: nodes carry absolute paths, so without it every file's outermost part is
+        the filesystem root and the model is shown one scope for the whole repo.
+        """
         scopes: set[str] = set()
         words: Counter[str] = Counter()
         identifiers: set[str] = set()
         for node in self.static_analysis.iter_reference_nodes():
-            scope = scope_of(node.file_path)
+            scope = scope_of(node.file_path, str(self.repo_dir))
             if scope:
                 scopes.add(scope)
             name = node.fully_qualified_name.rsplit(".", 1)[-1]
--- a/agents/naming_model_agent.py
+++ b/agents/naming_model_agent.py
@@ -75,18 +75,34 @@ class NamingModelAgent(CodeBoardingAgent):
         Why the root: nodes carry absolute paths, so without it every file's outermost part is
         the filesystem root and the model is shown one scope for the whole repo.
         """
-        scopes: set[str] = set()
         words: Counter[str] = Counter()
-        identifiers: set[str] = set()
+        by_scope: dict[str, set[str]] = {}
         for node in self.static_analysis.iter_reference_nodes():
             scope = scope_of(node.file_path, str(self.repo_dir))
-            if scope:
-                scopes.add(scope)
             name = node.fully_qualified_name.rsplit(".", 1)[-1]
-            identifiers.add(name)
+            by_scope.setdefault(scope, set()).add(name)
             words.update(tokenize(name))
         return (
-            sorted(scopes),
+            sorted(scope for scope in by_scope if scope),
             words.most_common(VOCABULARY_SAMPLE),
-            sorted(identifiers)[:IDENTIFIER_SAMPLE],
+            self._identifier_sample(by_scope),
         )
+
+    @staticmethod
+    def _identifier_sample(by_scope: dict[str, set[str]]) -> list[str]:
+        """A sample that reaches every scope, taken round-robin.
+
+        Why not the first N of a sorted set: that is an alphabetical slice. On eShop it showed
+        the model identifiers from A to C only -- nothing named Order, Payment or Webhook --
+        so the sample argued for components the repo does not have and stayed silent on the
+        ones it does.
+        """
+        queues = [sorted(names) for _, names in sorted(by_scope.items())]
+        sample: list[str] = []
+        for index in range(max((len(q) for q in queues), default=0)):
+            for queue in queues:
+                if index < len(queue):
+                    sample.append(queue[index])
+                    if len(sample) == IDENTIFIER_SAMPLE:
+                        return sample
+        return sample
--- a/agents/prompts/__init__.py
+++ b/agents/prompts/__init__.py
@@ -29,6 +29,7 @@ from .prompt_factory import (
     get_scope_relations_message,
     get_api_surfaces_message,
     get_relation_analysis_message,
+    get_naming_model_message,
 )
 
 
--- a/agents/prompts/abstract_prompt_factory.py
+++ b/agents/prompts/abstract_prompt_factory.py
@@ -6,6 +6,33 @@ Defines the abstract base class for prompt factories with all prompt methods.
 
 from abc import ABC, abstractmethod
 
+NAMING_MODEL_MESSAGE = """You are reading a codebase through its identifiers only.
+
+Name the top-level architectural components a maintainer of this repo would name, and the
+vocabulary each one owns.
+
+Rules:
+- `machinery` is vocabulary naming how software is built rather than what this system is
+  about: Handler, Repository, Controller, View, Context, Dto, Options, Factory, Entry, Item,
+  Event, Service. A machinery word must never own a component, or every handler in the system
+  lands in one box -- which is a layer, not a component.
+- A component's `owns` is domain vocabulary: the words naming the problem this system solves.
+  Use words that actually appear in the identifiers below.
+- Every component must be one responsibility. Never name a component by joining two things
+  with "and" or "&".
+- Where a scope below is named after part of the problem, name a component after it too, so
+  that the scope and the vocabulary agree.
+
+Top-level scopes:
+{scopes}
+
+Most frequent identifier words, with counts:
+{vocabulary}
+
+A sample of identifiers:
+{identifiers}
+"""
+
 
 class AbstractPromptFactory(ABC):
     """Abstract base class for prompt factories."""
@@ -61,3 +88,11 @@ class AbstractPromptFactory(ABC):
     @abstractmethod
     def get_relation_analysis_message(self) -> str:
         pass
+
+    def get_naming_model_message(self) -> str:
+        """Read a repo's architecture out of its identifiers.
+
+        Concrete rather than abstract: this asks for a structured reading of a vocabulary,
+        not for prose, so no provider needs its own wording.
+        """
+        return NAMING_MODEL_MESSAGE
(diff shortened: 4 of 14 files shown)


Svilen-Stefanov:
## Downstream impact, measured

**No first-party repo is affected.** The committed baselines in CodeBoarding, CodeBoarding-wrapper, CodeBoarding-vscode and CodeBoarding-action are Python or TypeScript only — I checked the language mix of every `key_entities` entry in each. The rename touches C# and Java, so none of them moves.

**The exposed case is a user with a C#/Java repo and a committed baseline.** I measured what their first incremental run after upgrading actually does: baseline the C# edge-case project with the old naming, change nothing on disk, re-analyse with the new one.

```
baseline names : 97
after upgrade  : 98
look deleted   : 42
look added     : 43
unchanged      : 55
```

**It does not raise and it does not lose anything — it churns.** About 43% of nodes appear to move on a project where no line changed, because a renamed node is indistinguishable from a deleted one plus an added one.

Two ways to handle it, your call:

1. **A release note.** Tell C#/Java users the first run after upgrading shows a one-time large diff, and to run a full analysis to reset the baseline. Zero extra code; the churn is visible and explainable.
2. **A baseline migration.** The rename is mechanical — drop the file-stem segment — so an upgrade step could rewrite the stored names in place and the first incremental would be clean. More code, and it has to run before the fingerprint comparison.

I'd take (1): the affected population is C#/Java users on an incremental cadence, the failure mode is a noisy diff rather than a wrong answer, and a migration that silently rewrites stored identities is a worse thing to get wrong than a noisy diff is to explain.

Svilen-Stefanov:
## Result

Full pipeline, both rulers, scored over the symbols clustering is answerable for. (Namespaces, fields and properties are never tracked — 113 of Beacon's 369 — so counting them would penalise the pipeline for a set it is not asked to place.)

| ruler | shipped | this branch | precision | recall | boxes |
|---|---:|---:|---:|---:|---|
| beacon | 0.3511 | **0.9439** | 0.9497 | 0.9382 | 5 → 9 |
| eShop | 0.6672 | **0.9993** | 1.0000 | 0.9987 | 5 → 10 |

Both truths have nine boxes; the pipeline emits five for each today. On eShop every pair the new partition groups is correct.

The change that carries this is the **unit**: leaves are now files rather than Leiden communities, which are drawn from the call graph and cross the boundaries a name partition wants to keep. On Beacon 7 of 21 held symbols from two components, capping *any* grouping over them at 0.3427.

One caveat on my own result: the file ceiling is 1.0 partly by construction, since both rulers label at file granularity or coarser. Partitioning at *symbol* granularity — where a prediction can disagree with its own file and the ruler gives it no help — still reaches 0.8610, which is what says the method reads the vocabulary rather than inheriting the ruler.

## What this makes redundant

Nothing is dead today: every path below still runs when there is no naming model. But two things are now superseded in principle, and one is worth deleting outright.

**`Level.FILE` in `clustering/engine.py` should go.** It reaches for file granularity by chopping the last two segments off a qualified name:

```python
if level is Level.FILE and len(parts) > 2:
    return delimiter.join(parts[:-2])
```

That is the "qualified names are paths" assumption, and it is measurably false — only 7% of eShop's node keys start with their file's declared namespace. `file_leaf_clusters` reads `node.file_path` and gets the real answer. The string-surgery version is a worse implementation of the same idea and has no reason to survive.

**The rest becomes fallback-only, and that is a decision rather than a cleanup.** If the naming model becomes the default, these run only when it is absent:

- `cluster_graph`'s level-up search and `_score_clustering`
- `_group_by_modularity_peak`'s 13-resolution sweep
- the `[5,8]` bracket, `_seeds_from_partition` padding and `_absorb_leftovers`

That is a large block to keep alive as a fallback. Worth deciding deliberately once you have a view on whether the model is on by default — I have not assumed it.

ivanmilevtues:
The collapse is an implementation/model-contract failure, not evidence that MarkItDown is architecturally one component.

### Lexical model returned

Proposed component
`owns` entries
Effective matching vocabulary
`markitdown`
85
12 tokens
`markitdown-ocr`
19
**0**
`markitdown-sample-plugin`
7
**0**
`markitdown-mcp`
2
**0**

The prompt requested domain **words**, but Gemini mostly returned complete identifiers and constants:

- `DocxConverterWithOCR`
- `_extract_and_ocr_images`
- `ACCEPTED_FILE_EXTENSIONS`
- `__plugin_interface_version__`

The implementation builds the ownership map by stemming each entry directly—it does not tokenize it ([`owner_by_word`](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-pr539-run/static_analyzer/clustering/naming.py#L137-L145)). Thus:

- Model vocabulary: `DocxConverterWithOCR` → `docxconverterwithocr`
- Code identifier tokens: `Docx`, `Converter`, `With`, `OCR`

They never match. The same happens for underscores: `_extract_and_ocr_images` remains one ownership key, while code identifiers become `extract`, `and`, `ocr`, `images`.

### Why everything went to core

`MarkItDown` was one of the core component’s owned entries, producing the key `markitdown`. That word appears in the package prefix of practically every qualified name—including OCR and plugin files.

Measured from the saved static analysis:

- `markitdown` appeared **538 times**
- Every one of the **39 file leaves** received a core vote
- OCR received **zero votes**
- Sample plugin received **zero votes**
- MCP received **zero votes**

Sample plugin did own `accepts` and `convert`, but core listed them first. Ownership uses `setdefault`, so the first component wins collisions. Those words therefore also voted for core.

For example, an OCR file effectively scored:

```text
_docx_converter_with_ocr.py
  markitdown namespace votes → core
  accepts/convert votes       → core
  OCR-specific votes          → none
```

### Why it chose lexical instead of structural

The scopes were essentially:

```text
markitdown
markitdown-ocr
markitdown-sample-plugin
markitdown-mcp
```

This is actually a feature/package-shaped repository, so **structural probably should have won**.

However, the gate does not match scope words against component names. It only checks them against `owns` vocabulary ([gate logic](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-pr539-run/static_analyzer/clustering/naming.py#L147-L163)). After removing ubiquitous `markitdown`, the distinctive words were `ocr`, `sample`, `plugin`, and `mcp`. None existed as atomic ownership keys, so the match ratio was 0% and the gate selected lexical.

### Why “100% placement” is misleading

Coverage only means “did this file receive at least one recognized vote?” ([voting logic](file:///Users/imilev/StartUp/CodeBoarding-core/CodeBoarding-pr539-run/static_analyzer/clustering/naming.py#L300-L313)). Since every qualified name contained `markitdown`, every file was considered placed—even though all placements went to one component.

A robust correction would require:

1. Validating that `owns` contains atomic words, not identifiers.
2. Including component names when evaluating the structural gate.
3. Filtering ubiquitous repository/package words from lexical votes.
4. Treating a one-owner partition as degenerate rather than successful 100% coverage.

PR #551, opened 2026-09-02
feat(clustering): draw every component from the names partition

Svilen-Stefanov:
Second of three, on top of #550 (design + partition core). Next: `feat/name-tree-cleanup` deletes Leiden, anchoring, the cluster lineage and their dependencies.

## What changes

`ClusteringService` now draws every component from the names partition, on all three paths:

- **Full run.** Units from every language's graph → `draft_tree` one level deeper than `--depth-level` → `TreeSpec` persisted in `analysis.json` metadata as `tree_spec` → per scope, `replay` gives the `ClusterGroup`s the agents already consume. File leaves are the cluster ids; members are the callables and classes, as before. `expandable` is now a recorded decision (the child scope has rules), not a modularity gate.
- **Incremental run.** No seeded Leiden, no anchoring, no drift budget. The stored spec is replayed over the live names. A component whose rule still claims units keeps its id, so the scope plan emits UPDATE only where membership or bodies changed. Units no prefix or word claims are grouped where they leave the known tree; a group of two or more all-new files becomes a new rule with a fresh id, and the created component keeps that id (`scope_plan` hands it through the CREATE operation). A retired directory's component is deleted by the existing plan. Whether anything changed is read off the replayed tree against the persisted one; agents initialise only then.
- **Partial run.** Replays the stored spec for the component, or drafts its scope if the cap never reached it. A component the spec calls a leaf is reported as not expandable with the reason.

One `Grouper` interface, two implementations, switchable by `CODEBOARDING_GROUPER` or `[clustering] grouper` in `config.toml`:

- `KinshipGrouper` (default): merges scopes sharing a word. Deterministic, no model call.
- `TreePlannerAgent`: kinship first; only a scope left with more than nine groups is shown to the model (groups, sizes, a few identifiers) to fold across words toward the budget. The answer is folded deterministically: every label lands in exactly one component, a forgotten label keeps its own group. No silent fallback: asking for the planner without an LLM is an error. Kinship stays the default until the planner's variance across draws is measured on the rulers.

A baseline without a `tree_spec` (written before this PR) cannot support an incremental or partial run and raises `IncrementalCacheMissingError` with the reason, as a missing cluster baseline did.

## Tuning found on real graphs

Running the service over CodeBoarding's own pickled graph opened both of its two large packages at the probe's 25% share and scattered their sub-packages and loose files across the root (19 boxes). The probe's own sweep shows 50% identical on seven rulers and better on kubernetes (0.24 → 0.43); at 50% CodeBoarding is its 13 top-level packages and eShop is the maintainers' 14 boxes with Ordering, Webhooks and EventBus merged with their namesakes. Adopted, documented in the design doc.

## What stays, for now

`engine.py`, `grouping.py`'s modularity path, `delta.py`, `snapshot.py`, the cluster lineage in the pickle, `expansion.py` and `leiden_utils.py` are no longer called by the service but are still imported by their own tests; the next PR deletes them together with the dependencies. The lineage is still written (file leaves), so the pickle stays consistent until then. The tests of the old service internals (scope inputs, anchoring, method-level fallback) go with the internals here.

## Tests

`pytest` 2,347 passed (the three `test_cli_parser` failures also fail on `origin/main`), `mypy` and `black` clean. New: service tests for full, incremental (unchanged, new file, new directory, lone new file, retired directory, missing spec) and partial runs; planner folding; config switch; `tree_spec` persistence; generator tests rewritten off the cluster delta.

## Summary by CodeRabbit

* **New Features**
  * Added configurable component grouping with deterministic and LLM-assisted planning options.
  * Added support for persisting and replaying tree specifications across full and incremental analyses.
  * Component identifiers now remain stable when analyses are updated.
  * Added configuration for selecting the grouping strategy.

* **Bug Fixes**
  * Improved incremental analysis handling for new, changed, and removed scopes.
  * Added clearer errors when planning or required cached analysis data is unavailable.

* **Documentation**
  * Updated guidance on grouping thresholds, tree planning, and incremental analysis behavior.


Svilen-Stefanov:
Review round in 9c7079eb (+ fixtures in 6232ea93), rebased on the core fixes: `TreeSpec.reroot` follows save-time single-child absorption so the next incremental run replays the ids the analysis carries; `_stored_tree_spec` refuses a specification of another version (`IncrementalCacheMissingError`); the incremental baseline inventory is the baseline graph's files, so a data-only file a layer falls back to is not admitted as a new scope (test with two constants-only files under a layer); the planner's `owns` words are stemmed the way replay looks them up; the planner is registered in `_monitoring_agents`; class docstring cut to one line.

Svilen-Stefanov:

PR #553, opened 2026-09-02
feat(clustering): fold toward nine components along the graph, nest from forty units, and a planner that draws three times

Svilen-Stefanov:
Follow-up to #552, kept separate so the stack's original shape and this one can be compared side by side.

> **Read this first.** Three groupers exist: `kinship` (deterministic, the #552 default: directory boxes merged by a shared word), `affinity` (deterministic, **the new default here**: kinship plus a fold along the call graph) and `planner` (the opt-in LLM grouper). **Every real run in this PR and in CodeBoarding-tests#33 is deterministic; the planner was not used in any of them.** Part 2 below stabilises the planner, measured partition-only (60 model calls), so the escape hatch is sane if it is kept. On the evidence, affinity is the one to ship: better on the only maintainer-drawn ruler, agrees with the planner's own grouping at 0.96 on django, bit-identical across runs, no model call. No measured repository needs the planner; it remains for a case no formula over names and links can group (kubernetes-scale roots, 44–53 candidates, are the only unmeasured ones). Deleting it would remove ~230 lines and the run's only LLM dependency before description time; that is a product call this PR leaves open.

Two changes, each measured with no LLM on the six static analyses of the grouper study (CodeBoarding-tests#32) plus real depth-3 runs.

## 1. Fewer root boxes, real nesting, no grab bags (deterministic)

The study's kinship arm drew 14/18/12 root boxes on eShop/django/mermaid with 1–2 expandable components, and nothing below depth 2. Three causes, three fixes in `draft.py`/`frontier.py`:

- **`AffinityGrouper`, the new default.** Kinship first, then fold along the call graph: a candidate below the scope's floor, and the smallest candidate while the scope holds more than nine, joins the sibling with the highest *observed over expected* link count (call + `INHERITS`/`TYPEREF` edges, cross-file; `links × total / (degree × degree)`, at least two links, never past 60% of the scope). The ratio is what defeats the `utils`/`core` hub. The links reach the grouper as context only; the persisted rules are still prefixes and words and `replay` never sees an edge. On django it recovers the planner's own merges (`http` → `views`, `templatetags` → `template`, `conf`/`apps` → `core`) at 0.96 agreement with the LLM planner's root grouping, byte-identical across runs and unit order.
- **Nesting from 40 units.** A component of ≥ 40 units reads the frontier of its own sub-trie; transposition and the vocabulary rung stay above 135, because a transposition below the cap over-splits every maintainer-drawn leaf on the eShop depth-2 ruler (0.59–0.67) while a directory split costs it 0.04. The un-merge rung gives nesting for free: a folded root box opens into its candidates at depth 2, each of which opens its sub-trie at depth 3.
- **The guard no longer absorbs into the largest rule** (django's `contrib` had become a `gis` grab bag of twelve packages), and **a layered node without a grid draws its layers** instead of one box, which is what sent markitdown and serilog to the vocabulary rung (88% / 69% in one box).

| repo (units) | root boxes | expandable components at depth 1 / at depth 2 | largest root box | F1 d1 | F1 d2 |
|---|---|---|---|---|---|
| eShop (475) | 14 → **9** | 2/0 → **7/1** | 28% → 28% | 0.986 → 0.985 | 0.974 → 0.934 |
| django (635) | 18 → **9** | 1/1 → **7/8** | 46% → 47% | 0.423 → 0.410 | 0.545 → **0.709** |
| mermaid (386) | 12 → **9** | 2/1 → **4/2** | 44% → 47% | 0.086 → 0.075 | 0.099 → 0.087 |
| markitdown (40) | 3 → 3 | 0/0 | **88% → 57%** | | |
| serilog (109) | 7 → **9** | 1/0 → **3/1** | **69% → 43%** | | |
| fzf (42) | 5 → 5 | 0/0 | 57% | | |

eShop's nine root boxes are the maintainers' nine at 0.985; the one cost is `PaymentProcessor` + AppHost folded into Basket, and depth 2 drops to 0.934 because `Identity.API` now opens into its layers. Depth-1 F1 is unchanged on all eight synthesised rulers of the design doc.

## 2. The LLM planner, close to deterministic

The study's variance is the model itself: Kimi-K2.6 at temperature 0 gives three different answers to a byte-identical prompt (its hidden reasoning is sampled; `seed` is ignored, thinking cannot be disabled through the proxy), and markitdown's 2-box draw was entirely an `owns` word (`markitdown`) that replay's vote turned into 35 of 40 files. So `TreePlannerAgent` now asks in one JSON call with no tools, lists groups largest first with the scope's floor, draws three answers concurrently and folds the medoid, and keeps an `owns` word only when it is a lowercase stem in the component's own identifiers and nobody else's.

| repo | before (study, single draw) | after (3 draws, medoid) |
|---|---|---|
| markitdown root agreement, boxes | 0.27–0.66, 2–7 | 0.62, 6–7 |
| serilog root agreement | 0.53–0.95 | 0.69–0.79 |
| django scope 1 (label level) | 0.67 | 1.00 |

Three times the planner's tokens, the wall clock of one call. The planner stays opt-in (`CODEBOARDING_GROUPER=planner`); the default is the affinity fold.

## Real runs at depth 3 (affinity grouper, no planner; Kimi-K2.6 names and describes)

| repo | root | expandable components at depth 1 / at depth 2 | boxes at depth 2 / depth 3 | calls | tokens |
|---|---|---|---|---|---|
| mem0 (499 files, Python + TypeScript) | 7 | 3 / 5 | 14 / 25 | 72 | 708k |
| serilog (109, C#) | 9 | 3 / 1 | 8 / 4 | 54 | 531k |

The runs, their nested diagrams and a renderer are in CodeBoarding-tests (`shape-study/`, branch `study/name-tree-shape`).

## Incremental behaviour

Measured in CodeBoarding-tests#33 (`shape-study/incremental/`): one add+delete+modify change set per repository on both arms. Collateral was 0 in all six runs and unchanged components were byte-identical. Three defects found there (a deleted project resurrected by a second engine config, a new directory lost to a word vote, dead rules left in the spec) are fixed in #555 on top of this PR.

## Known limits

- Every expanded scope has at least two real children (the guard requires it), but three splits in the real runs are lopsided, one child holding 85–89% (eShop's Catalog, mem0's ingestion layer, serilog's Core). A share cap on the split is the obvious next rule.

- Small root boxes with no neighbour in the graph stay (eShop's two TypeScript files, mermaid's `scripts`/`cypress`); the budget cannot fold what nothing links.
- A unit the parent placed by a word has no home among the un-merged parts and lands in "Unassigned" (two or three files on eShop, serilog, mermaid).
- The fold is measured up to 635 units; kubernetes-scale roots need links the synthesised ruler harness does not carry.
- k=3 draws leave the model's ambiguity over one-file groups; more draws buy it down at linear cost.

## Summary by CodeRabbit

* **New Features**
  * Added affinity-based clustering that groups related components using call and reference relationships.
  * Added configurable multi-draw planning with JSON responses, retries, timeouts, and consensus selection.
  * Added support for choosing among affinity, kinship, and planner-based grouping strategies.

* **Improvements**
  * Improved diagram organization with clearer component size limits and nesting behavior.
  * Refined ownership labels to use concise, component-specific terms.
  * Updated layered diagrams to preserve meaningful boxes instead of over-collapsing them.

* **Documentation**
  * Updated design documentation to describe the new grouping strategies, layout rules, and measurements.

### DIFF
--- a/agents/llm_config.py
+++ b/agents/llm_config.py
@@ -534,3 +534,8 @@ def supports_prompt_caching(llm: BaseChatModel) -> bool:
     today; other providers either cache transparently or not at all.
     """
     return llm.__class__.__module__.startswith("langchain_anthropic")
+
+
+def supports_json_mode(llm: BaseChatModel) -> bool:
+    """True when *llm* takes OpenAI's ``response_format={"type": "json_object"}``."""
+    return isinstance(llm, ChatOpenAI)
--- a/agents/prompts/abstract_prompt_factory.py
+++ b/agents/prompts/abstract_prompt_factory.py
@@ -6,23 +6,30 @@ Defines the abstract base class for prompt factories with all prompt methods.
 
 from abc import ABC, abstractmethod
 
-TREE_PLAN_MESSAGE = """You are grouping the scopes of a codebase into the components a maintainer would draw.
+TREE_PLAN_SYSTEM_MESSAGE = (
+    "You group the candidate groups of one scope of a codebase into the components a maintainer "
+    "would draw in an architecture diagram. Answer with JSON only."
+)
 
-Scope: {scope}
+TREE_PLAN_MESSAGE = """Scope: {scope} ({units} files in {count} candidate groups, listed largest first)
+{groups}
 
-Each candidate group below has already merged the scopes that share a name. Fold them into at most
-{budget} components where they serve one purpose together, for example the web app, the mobile app
-and the hybrid app into "Customer experiences", and leave apart what does not belong together.
+Each candidate group has already merged the scopes that share a name. Fold them into at most
+{budget} components, each for one responsibility, for example the web app, the mobile app and the
+hybrid app into "Customer experiences".
 
 Rules:
-- Never split a candidate group; a component gathers whole groups.
-- Every candidate group label must appear in exactly one component's members.
+- Every label appears in exactly one component's members. Never split a candidate group.
+- A component holds at least {floor} files. A group with fewer files joins the component whose
+  purpose it serves; it never stands alone.
+- Keep apart what does not belong together: use the budget before lumping unrelated groups.
 - Name each component for its one responsibility. Never join two things with "and" or "&".
-- `owns` is domain vocabulary a component claims beyond its groups' names: words that appear in the
-  identifiers below, never words naming how software is built (Handler, Service, Repository).
+- owns: at most 5 lowercase single words this component claims beyond its group names, taken from
+  the identifiers listed under its own groups. Never words naming how software is built (handler,
+  service, repository, converter).
 
-Candidate groups:
-{groups}
+Answer with JSON only, no prose, in this shape:
+{{"groups": [{{"name": "Customer experiences", "members": ["G1", "G4"], "owns": ["customer"]}}]}}
 """
 
 
--- a/agents/tree_planner_agent.py
+++ b/agents/tree_planner_agent.py
@@ -1,20 +1,29 @@
 """The planner: an LLM folds a scope's candidate groups into components, once per scope."""
 
+import json
 import logging
+import re
+from collections import Counter
 from collections.abc import Sequence
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 from langchain_core.language_models import BaseChatModel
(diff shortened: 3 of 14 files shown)


Svilen-Stefanov:

PR #555, opened 2026-09-02
fix(incremental): one graph per language across engine configs, new directories over word votes, retired rules leave the spec

Svilen-Stefanov:
Follow-up to #553, from the incremental study in CodeBoarding-tests#33 (`shape-study/incremental/`): one change set per repository with an addition, a deletion and a modification at once (eShop: a new `ShippingProcessor` project, `WebhookClient` deleted, Basket publishing to the event bus; serilog: a new `Sampling` directory, `Settings` deleted, the logger calling into formatting; mem0: a new plugin cloned from `openclaw`, `vercel-ai-sdk` deleted, a new helper called from `memory/main.py`), run on both engine arms (**both deterministic**: kinship = #552, affinity = #553; the LLM planner was not used) and diffed component by component. Collateral (components touched without holding a changed file) was **0 in all six runs**; unchanged components were byte-identical. Three defects were found and are fixed here.

## 1. Deleted and modified files survived when a language has more than one engine config

eShop's C# has two engine configs (the solution and `src/ClientApp`). `_update_cached_results` handed each config its own copy of the language's cached graph, invalidated only the changed files under that config's root, and merged the results: the second config's untouched copy put every node the first had deleted or re-analysed back. On eShop the deleted `WebhookClient` project reappeared in `analysis.json` with all 11 files while `fingerprint.json` no longer listed them, and DELETE, single-child absorption and the spec reroot never ran. Now one dict per language is threaded through its configs and absorbed once. Re-run on the same change set: 11 deleted files dropped, `WebhookClient` retired, scope 6 absorbs its remaining child, the spec is rerooted, and the diff shows exactly the expected set.

## 2. A new directory whose names carried a word an existing rule owned was never a new scope

Replay tries prefixes, then a weighted word vote, then fallbacks; only fallback-caught and unplaced units were reported as new-scope candidates. serilog's kinship root is drawn from words (`Sink`, `Logger`, …), so three of the four new `Sampling` files were voted into `Core` by `IsEnabled(LogEvent)` and the fourth stayed unplaced: no component. eShop was one unit from the same failure (4 of 5 ShippingProcessor units vote `order`). Now every unit no prefix claims is a candidate, and the service keys the new units by where they leave the tree of the units the baseline knew rather than the tree of owned prefixes (a scope drawn from words owns no prefix at all, so everything diverged at `Serilog`). A new directory of two or more new files is a new component whatever its names say.

## 3. A rule that claimed nothing stayed in the spec after its component was deleted

serilog deleted component `6 Pair` (its files went with `Settings`) and `8 Settings`, but the rules stayed in `tree_spec`, ready to re-claim units later while the tree no longer had the component. A rule with no members is now retired from the spec with the scopes below it; fallback-only rules stay, they are legitimately empty; a directory that comes back is a new scope with a fresh id.

Also: `[ScopePlan]` now logs every operation it emits (action, id, rationale); the study could not attribute the extra root operations from the count alone.

## Re-run of the two failing cases on this branch (`shape-study/incremental/runs-fixed/`)

| repo × arm | result | collateral |
|---|---|---|
| eShop × affinity | 11 deleted files dropped; `ShippingProcessor` created with 6 units; `WebhookClient` retired, scope 6 absorbs its remaining child, spec rerooted; 35 of 40 components unchanged | 0 |
| serilog × kinship | `Sampling` created with all 4 files; `Settings` deleted and `Pair` retired from the spec; 10 of 13 unchanged | 0 |

## Not fixed, recorded in the study

- A modification can move a word-placed file (serilog: dropping the `KeyValuePairs` overloads moved `LoggerSettingsConfiguration` from `Pair` to `Logger`). Correct by the rules; membership of word-placed files depends on method naming.
- The intended Basket → EventBus edge was not extracted (interface-member call, C# adapter), so the kinship arm's LLM dropped the baseline Basket → EventBus relation. Relation quality after a modification is bounded by edge extraction.
- The affinity root stores the global relation list and re-resolves it by name on every incremental run (~48 "Dropping relation" lines, harmless).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved incremental analysis when multiple configurations share a language, preventing outdated or deleted results from being restored.
  - New directories are now correctly recognized as new components, including when names match existing terms.
  - Empty component rules are retired while fallback-only rules remain available.
  - Scope planning now records operation actions, affected components, and rationales.

- **Documentation**
  - Clarified incremental-run behavior for new and retired components, including fallback-only rules.

- **Tests**
  - Added coverage for new-component detection, retired components, and multi-configuration warm starts.

### DIFF
--- a/diagram_analysis/scope_plan.py
+++ b/diagram_analysis/scope_plan.py
@@ -145,6 +145,8 @@ def _plan_scope_operations(
         f"[ScopePlan] {scope_id}: {len(groups)} groups, {untouched} unchanged (no operation), "
         f"{len(operations)} operation(s)"
     )
+    for operation in operations:
+        logger.info(f"[ScopePlan] {scope_id}: {operation.action.value} {operation.component_id}: {operation.rationale}")
     return ScopeUpdateDecision(operations=operations)
 
 
--- a/docs/design/name-tree.md
+++ b/docs/design/name-tree.md
@@ -188,8 +188,8 @@ clustering. Per scope, walking the tree top-down:
    membership; `previous_component_id` is the rule id, so the existing scope plan emits
    UPDATE only where members or bodies changed, and nothing else moves. Measured: 0.00%
    of unchanged files move across 27 repositories and four edit types.
-2. **New components.** Every unit that no prefix and no word claims, whether a fallback
-   rule caught it or nothing did, is reported in `Partition.new_scopes` under the prefix at
+2. **New components.** Every unit that no prefix claims, whether a word vote, a fallback
+   rule or nothing placed it, is reported in `Partition.new_scopes` under the prefix at
    which its position first leaves every prefix a rule owns: a new top-level directory
    diverges at that directory; a new file deep inside a known one is claimed by its
    ancestor's rule and never gets here. A group of ≥ 2 units whose members are all new to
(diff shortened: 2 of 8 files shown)


Svilen-Stefanov:

PR #558, opened 2026-09-03
Remove legacy architecture analysis agents

ivanmilevtues:
## Summary
- remove the abstraction, detailing, and incremental LLM agents and their provider-specific prompts
- keep deterministic clustering authoritative for component membership, hierarchy, IDs, and static relations
- move deterministic scope assembly and incremental update behavior out of the agents layer
- remove obsolete response models, tools, validators, and tests

This is stack 1 of 2. It intentionally leaves deterministic placeholder component metadata; stack 2 adds the single scope-analysis agent that names components and enriches relations.

## Verification
- `uv run black .`
- `uv run mypy .` — 324 source files, no issues
- targeted tests — 81 passed
- `uv run python -m compileall -q agents diagram_analysis`
- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` — 2092 passed, 28 skipped, 1 environment-dependent failure; coverage 83.44%. The failure is the registry-coverage check for the unavailable `dotnet` executable.

Benchmark instrumentation is intentionally excluded from this stack.

## Summary by CodeRabbit

* **New Features**
  * Added deterministic component assembly for more consistent architecture diagrams.
  * Added incremental updates that preserve unchanged components and metadata while refreshing affected relationships.
  * Added automatic fallback descriptions and component naming based on analyzed code structure.

* **Changes**
  * Component relationships are now rebuilt from static code analysis for affected scopes.
  * Removed advanced AI-driven architecture, detail, validation, repair, and bridge-edge analysis capabilities.
  * Simplified prompt handling to shared planning and metadata prompts.


diagram_analysis/scope_assembly.py
```diff
@@ -16,62 +15,51 @@
...
-            )
-            if component is None or id(component) in used:
-                symbols = sorted(group.qualified_names)
-                component = Component(
-                    name=symbols[0].split(".")[-1] if symbols else group_name,
-                    description=descriptions[group_name],
+        for group in scope.groups:
+            components.append(
+                Component(
+                    name=f"Component {group.group_id}",
```
Svilen-Stefanov: This is fine as a stack step, but it does mean that if #558 lands on main and we cut a release before #559, every user gets `Component 1` through `Component 8` in their diagrams. Can we merge the two together, or at least agree that no release goes out in between?

Better still: `ComponentRule` already carries a `name` from the name-tree work and `ClusteringService._materialize` throws it away. If we plumbed that through to here instead, this branch would be safe standalone and #559 would degrade to a real name whenever the model fails or returns junk. Left the longer version of this on #559.

agents/validation.py
```diff
@@ -62,420 +16,10 @@ class ValidationResult:
...
-        logger.warning("[Validation] Relations without static backing or evidence: %s", "; ".join(unsupported))
-    if invalid_key_edges:
-        logger.warning("[Validation] Relations with degenerate key edges: %s", "; ".join(invalid_key_edges))
-    if unresolved_key_edges:
-        logger.warning("[Validation] Relations with unresolved key edges: %s", "; ".join(unresolved_key_edges))
-    return ValidationResult(
-        is_valid=False,
-        feedback_messages=feedback_messages,
-        score=(valid_relation_count / total_relations) if total_relations else 0.0,
+def score_validation_results(validator_results: list[tuple[Callable, ValidationResult]]) -> float:
```
Svilen-Stefanov: nit: `VALIDATOR_WEIGHTS` is an empty dict now and nothing outside `tests/` imports this module any more, and #559 deletes the file anyway. Can we just drop it here rather than carrying a 20 line stub plus its tests through the stack? Or is something in between still leaning on `score_validation_results`?

Svilen-Stefanov, line 19: nit: not sure if I see this correctly but it seems to me like `VALIDATOR_WEIGHTS` is only used here and in the `tests/` now. Can we just drop it as a global constant?

agents/llm_config.py
```diff
@@ -317,7 +303,6 @@ def get_resolved_extra_args(self) -> dict[str, Any]:
         api_key_env="ORCAROUTER_API_KEY",
         agent_model="openai/gpt-5.4-mini",
         parsing_model="openai/gpt-5.4-mini",
-        llm_type=LLMType.GPT4,
```
Svilen-Stefanov, line 320: Glad to see this is gone now.

Svilen-Stefanov (approved):
Thanks for doing this. It is a clean deletion, I'm mostly going to look at #559.
One question - I saw you deleted all the claude, gemini etc prompts and factory - does this mean we will have a single prompt for all models for now and overwrite it later if needed?

PR #559, opened 2026-09-03
Add one bounded agent for deterministic scopes

ivanmilevtues:
## Summary
- add one level-agnostic `ScopeAnalysisAgent` for both root and nested scopes
- supply complete deterministic groups, per-file placement reasons, boundary files, and directed boundary method pairs in the prompt
- expose only scope-restricted `readFile` and targeted `getMethodCalls` tools, with hard limits of 6 tool calls and 8 model calls per scope
- keep LangGraph's recursion ceiling above those middleware budgets so multi-turn tool use terminates through the intended limits rather than an early `GraphRecursionError`
- apply names, descriptions, key entities, and relation labels without allowing the model to change deterministic membership, hierarchy, IDs, or static edges
- accept non-static REST/queue/plugin/config relations only when exact source symbols resolve on both component sides
- rerun only affected scopes incrementally, lock names when membership is stable, preserve untouched metadata, and avoid re-analyzing newly generated child scopes twice

Boundary data is passed directly instead of spending a tool call on `getBorderingFiles`; the two tools are reserved for targeted source and CFG questions.

This is stack 2 of 2 and depends on #558.

## Verification
- `uv run black .`
- `uv run mypy .` — 327 source files, no issues
- focused scope/tool/incremental tests — 71 passed, 12 skipped
- recursion regression — 7 attempted reads, 6 executed, bounded completion with valid JSON
- `uv run python -m compileall -q agents diagram_analysis`
- `uv run pytest --cov=. --cov-report=term --cov-fail-under=80` — 2104 passed, 28 skipped, 1 environment-dependent failure; coverage 83.74%. The failure is the registry-coverage check for the unavailable `dotnet` executable.

Benchmark instrumentation is intentionally excluded from this stack.

## Evaluation against Gemini main

Reference: CodeBoarding `a01b20eb` using `google/gemini-3.8-flash`, on the same subject commits at depth 3 with no judge.

### Complete full suite (12 tests)

One worker per side, compared with PR head `3d6f0424`:

| Measurement | Gemini main | PR #559 | Saving |
|---|---:|---:|---:|
| Total tokens | 87,125,309 | 4,652,511 | **82,472,798 fewer (94.66%, 18.73× less)** |
| Summed wall time | 4,689.04s (1h 18m 09s) | 1,556.68s (25m 57s) | **3,132.36s faster (52m 12s; 66.80%, 3.01×)** |
| Deterministic full criteria | 60 met / 0 violated | 60 met / 0 violated | unchanged |

### Current-head smoke confirmation (`4b0ca45b`)

| Measurement (two full tests) | Gemini main | PR #559 | Saving |
|---|---:|---:|---:|
| Total tokens | 17,899,253 | 481,486 | **17,417,767 fewer (97.31%, 37.18× less)** |
| Summed wall time | 933.95s | 173.58s | **760.37s faster (81.41%, 5.38×)** |
| Deterministic full criteria | 10 met / 0 violated | 10 met / 0 violated | unchanged |

The two matched current-head incremental tests retained the same outcomes as main while summed wall time fell from 1,310.76s to 150.11s: **88.55% lower (8.73× faster)**. Incremental token totals were not retained by the historical runs, so no incremental token reduction is claimed.

Token totals come from the engine's raw `llm_usage.json`. Wall-time results are directional rather than hardware-controlled: main ran on macOS/ARM and the PR on Linux/x86. The complete full comparison used one worker per side; the current-head smoke used two workers and reports summed per-test time. [Detailed per-test results](https://github.com/CodeBoarding/CodeBoarding/pull/559#issuecomment-5534194238).

## Summary by CodeRabbit

* **New Features**
  * Added scope-aware semantic analysis that enriches diagrams with component names, descriptions, key entities, and relationships.
  * Added model-specific prompt profiles, including stricter instructions for supported model families.
  * Added clearer explanations for why files are assigned to diagram groups.
  * Added scoped method-call exploration and safer file access within analyzed repositories.

* **Changes**
  * Simplified LLM configuration to use a single agent model; parsing-model settings and related configuration options were removed.
  * Retired legacy metadata analysis, dependency discovery, and several unused analysis tools.


ivanmilevtues:
## Eval results: token and wall-time delta vs Gemini main

Reference: CodeBoarding `a01b20eb` using `google/gemini-3.8-flash`, compared on the same subject commits at depth 3 with no judge.

### Complete full suite (12 tests)

This comparison was run with one worker on each side against PR head `3d6f0424`:

| Measurement | Gemini main | PR #559 | Saving |
|---|---:|---:|---:|
| Total tokens | 87,125,309 | 4,652,511 | **82,472,798 fewer (94.66%, 18.73× less)** |
| Summed wall time | 4,689.04s (1h 18m 09s) | 1,556.68s (25m 57s) | **3,132.36s faster (52m 12s; 66.80%, 3.01×)** |
| Deterministic full criteria | 60 met / 0 violated | 60 met / 0 violated | unchanged |

### Current-head smoke confirmation (`4b0ca45b`)

Two representative full→incremental pairs were rerun after the latest commits:

| Full test | Tokens: main → PR | Token reduction | Wall time: main → PR | Time reduction |
|---|---:|---:|---:|---:|
| `click-before-get-strerror-removal` | 558,223 → 46,934 | **91.59%** | 188.32s → 54.68s | **70.96%** |
| `codeboarding-before-health-checks` | 17,341,030 → 434,552 | **97.49%** | 745.63s → 118.90s | **84.05%** |
| **Combined** | **17,899,253 → 481,486** | **97.31% (37.18× less)** | **933.95s → 173.58s** | **81.41% (5.38×)** |

The matched incrementals also retained the same outcomes while reducing summed wall time:

| Incremental test | Main | PR | Reduction |
|---|---:|---:|---:|
| `referenced-symbol-deleted` | 531.18s | 36.27s | **93.17%** |
| `new-subsystem-added` | 779.58s | 113.84s | **85.40%** |
| **Combined** | **1,310.76s** | **150.11s** | **88.55% (8.73×)** |

Current-head quality was unchanged: full **10 met / 0 violated**; incremental **16 met / 3 violated**, with the same three known `new-subsystem-added` violations as the earlier PR head and Gemini main.

Token totals come from the engine's raw `llm_usage.json`. Incremental token totals were not retained by either historical run, so no unsupported incremental token claim is made. Wall-time comparisons are directional rather than hardware-controlled: main ran on macOS/ARM and the PR on Linux/x86; the complete full comparison used one worker per side, while the current-head smoke used two workers and reports summed per-test time rather than batch elapsed time.

agents/llm_config.py
```diff
@@ -67,15 +65,12 @@ def configure_models(
...
-      1. Shell environment variables (set before the process starts)
-      2. ``api_keys`` passed here  /  values from ~/.codeboarding/config.toml
-      3. AGENT_MODEL / PARSING_MODEL environment variables (for model names)
-      4. Provider defaults defined in LLM_PROVIDERS
+    Existing shell environment variables take precedence over values loaded
+    from ``~/.codeboarding/config.toml``. Explicit ``model_override`` arguments
+    to ``initialize_agent_llm`` remain the highest-priority library API.
     """
-    global _agent_model_override, _parsing_model_override
+    global _agent_model_override
```
Svilen-Stefanov, line 72: blocker-ish: this one takes the wrapper down with it. `CodeBoarding-wrapper/codeboarding_pro/bootstrap.py:56` still calls `configure_models(agent_model=..., parsing_model=...)` and then reads `user_cfg.llm.parsing_model`, so that's a TypeError followed by an AttributeError the moment this ships. And the wrapper floors on `codeboarding>=0.13.11` rather than pinning, so it doesn't break on merge, it breaks on the next PyPI release, which is the nastier version of this.

Can we land a companion wrapper PR first? The action also still exposes a documented `parsing_model` input that silently becomes a no-op (`PARSING_MODEL_INPUT` is set in three places in action.yml), so that probably wants a deprecation note in the same pass.

static_analyzer/clustering/service.py
```diff
@@ -209,6 +209,11 @@ def _materialize(
                 nodes = graphs[unit.language].nodes
                 owned = {name for name in unit.names if nodes[name].type in CALLABLE_TYPES | CLASS_TYPES}
                 group.symbol_members_by_language.setdefault(unit.language, set()).update(owned)
+                group.file_reasons[unit.unit_id] = self._placement_reason(
```
Svilen-Stefanov, line 212: Really glad the placement reasons finally make it into the prompt, that's the bit I always wanted the model to see.

While we're in here though: `rule.name` is sitting right there and we drop it on the floor, and then `ScopeAssembler.build` names everything `Component <id>` and waits for the LLM to fix it. Can't we carry `rule.name` onto `ClusterGroup` as well and use it as the default component name?

It's the single change that would make every failure path in this PR degrade to a real name instead of `Component 3`, and it would make #558 safe to sit on main on its own. We put a fair amount of work into computing those names in the name-tree stack and they still never actually ship anywhere.

diagram_analysis/scope_assembly.py
```diff
@@ -87,6 +89,102 @@ def populate_file_methods(
...
+                group_id not in locked_name_ids
+                and proposed_name
+                and (proposed_name == component.name or proposed_name not in used_names)
+            ):
+                used_names.discard(component.name)
+                component.name = proposed_name
+                used_names.add(proposed_name)
+            if component_semantics.description.strip():
+                component.description = component_semantics.description.strip()
+            component.key_entities = [reference.model_copy(deep=True) for reference in component_semantics.key_entities]
```
Svilen-Stefanov, line 122: nit-ish: this assignment is unconditional, while `name` and `description` just above are both guarded by a `.strip()` check. So a valid response that simply omits `key_entities` for a changed group wipes the entities that group already had.

Shouldn't this follow the same rule and only replace when the list is non-empty?

diagram_analysis/scope_assembly.py
```diff
@@ -87,6 +89,102 @@ def populate_file_methods(
...
+        if updated_ids:
+            reference_resolver.fix_key_entities_refs(analysis, updated_ids)
+        if result.description.strip():
+            analysis.description = result.description.strip()
+
+        ownership = ComponentOwnershipIndex.from_analysis(analysis)
+        preserved = [
+            relation.model_copy(deep=True)
+            for relation in analysis.components_relations
+            if relation.src_id not in editable_group_ids and relation.dst_id not in editable_group_ids
```
Svilen-Stefanov, line 134: just checking my reading here: `preserved` only keeps relations where neither side is editable. So every relation touching a changed group gets dropped, and then re-added by `merge_scope_relations` with `DEFAULT_STATIC_RELATION_LABEL` unless the model happened to repeat it. The prompt says relations "may" be returned for pairs touching a changed group, so it's optional from the model's side.

Doesn't that mean one changed group resets the labels on all of its neighbours? Feels like the same relation churn family we've been chasing for a while, and the eval does still report `ungrounded_edge_churn`. Cheap guard would be falling back to the previous label when the model omits a pair that already had one.

agents/scope_analysis_agent.py
```diff
@@ -0,0 +1,177 @@
...
+            if isinstance(block, str):
+                parts.append(block)
+            elif isinstance(block, dict):
+                text = block.get("text") or block.get("output_text")
+                if isinstance(text, str):
+                    parts.append(text)
+        return "".join(parts)
+
+    @staticmethod
+    def _json_object(text: str) -> str:
```
Svilen-Stefanov, line 171: Now that the parsing LLM is gone, this brace slice is the only thing between us and losing a whole scope's names. First `{` to last `}` will happily swallow a sentence of prose containing a brace, or two objects glued together.

And the failure is completely silent from the user's side: `analyze` logs a warning and returns None, `_enrich_scope` swallows the None, and the scope ships as `Component 3`. Can we at least count these into monitoring so we can see how often it actually happens in the wild? Right now we'd have no idea.

agents/prompts/profiles.py
```diff
@@ -0,0 +1,27 @@
...
+
+class PromptProfile(StrEnum):
+    """Instruction style used by tree planning and semantic scope analysis."""
+
+    STANDARD = "standard"
+    STRICT = "strict"
+
+
+_STRICT_MODEL_MARKERS = (
+    "deepseek",
```
Svilen-Stefanov, line 14: I'm a bit nervous about this one. We collapsed six provider prompt files into two profiles, but the evals only ran gemini flash, which resolves to standard. So strict is completely unvalidated, and it's the profile every deepseek, glm, kimi and qwen user gets.

Can we do at least one full run on glm or deepseek before this merges? Also, substring matching on the model name feels fragile, `gpt-oss` on cerebras lands in standard, is that deliberate?

agents/llm_renderers/scope.py
```diff
@@ -0,0 +1,194 @@
...
+) -> str:
+    """Return complete group files, boundary candidates, and known calls as JSON."""
+    boundary_reasons = _boundary_reasons(scope, repo_dir)
+    components = {component.component_id: component for component in analysis.components}
+    groups = []
+    for group in scope.groups:
+        component = components.get(group.group_id)
+        file_reasons = _group_file_reasons(group, scope, repo_dir)
+        for file_path in file_reasons.keys() & changed_files:
+            boundary_reasons[group.group_id][file_path].add("changed in incremental analysis")
```
Svilen-Stefanov, line 32: nit: we're pushing changed files into `bordering_files`, but the same path already appears in `files` with `changed: true`, and the system prompt tells the model bordering files are boundary candidates.

Isn't this just muddying what bordering means? The `changed` flag on its own seems like enough.

Svilen-Stefanov (approved):
Nice, amazing to see the reduction in time and tokens! I saw you capped the tool and model calls - is there a reason for the numbers or did these just seem reasonable?

One thing I noticed is that it breaks the wrapper. I opened a PR to fix it.

A few smaller things:
- `llms.txt` is still the old one and embeds the full source of `abstraction_agent`, `details_agent`, `validation` and friends. Should we regenerate it in this stack?
- I agree with the codex P1s. Two worth calling out: the child-scope auth one is a genuine regression, main had an explicit `except LLMAuthError: raise` in `_process_component` and #558 removed it because there was no LLM left in there, so putting an LLM call back in this PR quietly undid it. And the wall-clock timeout one is real, all our clients are constructed with `timeout=None` and we lost the 300/600s wrapper along with the old agent path.

@ivanmilevtues nice work on the eval writeup by the way.

PR #565, opened 2026-09-05
fix(naming): keep the scope agent inside the model's context and never ship a rule name or a disk path

Svilen-Stefanov:
## Why

The 0.14.0 release eval ([CodeBoarding-evals#23](https://github.com/CodeBoarding/CodeBoarding-evals/pull/23)) read `origin/main` b61e5abf against 0.13.11 on ten repositories with Kimi-K2.6 (262k context). The clustering was a large improvement everywhere, but on Gson and jsoup the top level shipped **raw rule names** (`internal`, `Loose files in gson.com.google.gson`, `helper 7`) and **every description was a template printing absolute paths of the analysed checkout** (`Owns 631 symbols across 37 files: /private/var/folders/…/repo/…`). Root cause, measured: `render_scope_context` rendered every cross-group edge as a nine-field indented object — Gson root = **1,676,924 chars ≈ 420k tokens, 93% of it `known_connections` (2,980 edges)** — the model answered 400, `_enrich_scope` caught it, and the scope kept its deterministic output. Polly / JUnit 4 / mermaid / Hono kept template descriptions one level down for a second reason (the bounded agent ended in prose, "response contains no JSON object"). CodeBoarding shipped `1` and `1.1` both named "Static Analysis Engine" (`doc_filenames_unique` violated), and Hono drew two components owning zero files.

## What changes

- **`known_connections` is one entry per directed group pair**: the number of distinct calls and up to five of them in full (`MAX_EXAMPLE_EDGES`). Gson root prompt: 1,676,924 → **106,841 chars**; the scope names itself.
- **JSON recovery**: a bounded run that ends without a JSON object gets one tool-free completion over its transcript asking for the object alone.
- **Fallback descriptions are repository-relative** (`normalize_repo_path`).
- **No child named after its parent**: the renderer lists `enclosing_components`, the prompt forbids reusing them, `apply_semantics` reserves them, and `finalize_for_save` runs `distinguish_expanded_component_names` — any two expanded components that still sanitise to one document name get an id suffix, relations following.
- **A rule whose units own no callable or class is not drawn** (`ClusteringService`).

## Verified

Re-ran the five affected rulers through the evals harness with this branch (`local:` engine, tag `release-0.14.0-fix`, Kimi-K2.6), same commits as the eval:

| ruler | template descriptions | path leaks | empty components | numbered names | duplicate document names | tokens |
|---|---:|---:|---:|---:|---:|---:|
| gson | 8 → **0** | 8 → 0 | 0 | 0 | 0 | 222k → 130k |
| jsoup | 8 → **0** | 8 → 0 | 0 | 1 → 0 | 0 | 305k → 127k |
| codeboarding | 2 → **0** | 2 → 0 | 0 | 0 | 1 → **0** | 306k → 207k |
| hono | 9 → **0** | 7 → 0 | 2 → **0** | 0 | 0 | 212k → 188k |
| polly | 10 → **0** | 10 → 0 | 0 | 1 → 0 | 0 | 686k → 455k |

25 of 25 deterministic criteria met (the candidate had 24/25). Gson root now reads "Gson Core API / Gson Internal Bindings and Utilities / JSON Streaming Layer / Configuration and Annotations / Extras and Extensions / Shrinker Test Samples / Performance Benchmarks / Build Configuration"; jsoup "DOM Node Model / HTML/XML Parser / Selector Engine / Helper Utilities / Internal Utilities / HTML Sanitizer / Facade and Exceptions / Java 11 HTTP Client".

`pytest tests/` 2041 passed; mypy and black clean.

## Summary by CodeRabbit

* **New Features**
  * Improved diagram generation with unique names for expanded components, preventing naming conflicts.
  * Added clearer connection summaries with call counts and representative examples.
  * Added enclosing-component context to improve analysis results.
  * Standardized default document naming across generated documentation.

* **Bug Fixes**
  * Improved recovery when analysis responses are incomplete or lack structured results.
  * Prevented child components from reusing enclosing component names.
  * Updated fallback descriptions to use repository-relative file paths.
  * Improved reliability when recording component names after analysis issues.

### DIFF
--- a/agents/scope_analysis_agent.py
+++ b/agents/scope_analysis_agent.py
@@ -31,10 +31,7 @@ MAX_SCOPE_TOOL_CALLS = 6
 MAX_SCOPE_MODEL_CALLS = 8
 SCOPE_RECURSION_LIMIT = 40
 
-#: Asked once, without tools, when the bounded run ended in prose instead of the object.
-#: Measured on Hono with Kimi-K2.6: four of eleven scopes ended that way in one run and the
-#: scope then shipped its deterministic names; one plain completion over the same transcript
-#: is far cheaper than losing the scope.
+#: Sent once, without tools, when the bounded run did not end in a usable JSON object.
 JSON_RECOVERY_MESSAGE = (
     "Your previous message did not contain the JSON object. Reply now with exactly one JSON object "
     "in the requested shape and nothing else — no prose, no markdown fences."
@@ -155,15 +152,11 @@ class ScopeAnalysisAgent(MonitoringMixin):
             raise_if_auth_error(error)
             raise
         messages = response.get("messages", [])
-        text = self._last_response_text(messages)
-        if "{" not in text:
-            logger.info("Scope %s answered without a JSON object; asking once more for it", scope.scope_id)
-            text = self._recover_json(messages)
-        try:
-            return ScopeAnalysisResult.model_validate_json(self._json_object(text))
-        except (json.JSONDecodeError, ValidationError, ValueError) as error:
-            logger.warning("Scope %s returned unusable semantic output: %s", scope.scope_id, error)
-            return None
+        result = self._parse(self._last_response_text(messages), scope.scope_id)
+        if result is None:
+            logger.info("Scope %s: asking once more for the JSON object", scope.scope_id)
+            result = self._parse(self._recover_json(messages), scope.scope_id)
+        return result
 
     def _recover_json(self, messages: list) -> str:
         """One tool-free completion over the transcript, asking for the object alone."""
@@ -177,6 +170,14 @@ class ScopeAnalysisAgent(MonitoringMixin):
             logger.warning("JSON recovery call failed: %s", error)
             return ""
         return self._last_response_text([answer])
+
+    @classmethod
+    def _parse(cls, text: str, scope_id: str) -> ScopeAnalysisResult | None:
+        try:
+            return ScopeAnalysisResult.model_validate_json(cls._json_object(text))
+        except (json.JSONDecodeError, ValidationError, ValueError) as error:
+            logger.warning("Scope %s returned unusable semantic output: %s", scope_id, error)
+            return None
 
     @staticmethod
     def _last_response_text(messages: list) -> str:
--- a/agents/scope_analysis_agent.py
+++ b/agents/scope_analysis_agent.py
@@ -2,17 +2,17 @@
 
 from __future__ import annotations
 
-import json
 import logging
 from collections.abc import Sequence
 from pathlib import Path
 
 from langchain.agents import create_agent
+from langchain.agents.structured_output import ToolStrategy
 from langchain.agents.middleware import ModelCallLimitMiddleware, ToolCallLimitMiddleware
 from langchain_core.language_models import BaseChatModel
-from langchain_core.messages import AIMessage, HumanMessage
+from langchain_core.messages import HumanMessage
 from langgraph.graph.state import CompiledStateGraph
-from pydantic import Field, ValidationError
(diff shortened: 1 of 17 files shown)


Svilen-Stefanov:
Addressed in 18cfaae1:
- recovery runs after a failed parse (prose with braces, truncated JSON), one call at most; test covers prose-with-braces and a malformed recovery answer
- expanded document names are compared sanitised **and case-folded**, matching `render_docs`; ids ordered with `CodeBoardingClusterIds.sort` (legacy ids included, test added)
- `readFile` backtick closed
- docstrings and comments cut to a summary plus a `Why:` line; history moved to the evals doc
- suffix-collision case covered (`Core`, `Core`, `Core (2)` → `Core`, `Core (2)`, `Core (2) (3)`)

Full suite 2044 passed, mypy and black clean.

Svilen-Stefanov:
Round two, in 6859ca26: the root document names (`overview`, `on_boarding`, now constants in `constants.py` used by the renderer and the CLI) are taken before any expanded component is keyed; a partial run records the persisted tree's names and the target component before its children are named; the test module docstring is back. Full suite 2046 passed, mypy and black clean.

Svilen-Stefanov:
P2 addressed in 722a4d11: the partial flow passes `{root, **sub_analyses}` as `persisted_scopes`, so `_enclosing_names` sees every ancestor of the expanded component. Full-run output is untouched by this commit; the evals re-run (CodeBoarding-evals#23, tag `release-0.14.0-fix`) verified 88d1e4cc, the last commit that changed the full path.

Svilen-Stefanov:
b5e21f1a: the JSON recovery prompt is gone. The scope agent now runs with `response_format=ToolStrategy(ScopeAnalysisResult)`, so the model finishes only by calling the schema tool and a validation failure is retried inside the run by langchain; `_json_object`/`_last_response_text`/`_recover_json` are deleted. Smoke on Gson root with Kimi over Azure: 8 named components, 27 of 27 relations labelled, no warnings. The example-edge cap stays: Hono root with 5 vs 50 examples labels the same 18 relations, and the document's `all_edges` come from the call graph regardless (identical counts before/after).

Svilen-Stefanov:
4141c31d: found while re-running the rulers on the ToolStrategy commit — `ToolCallLimitMiddleware` counts every tool call, including the structured answer, so a scope that spent its six reads had its answer blocked (Polly scope 3 kept rule names). `RepositoryToolBudget` is the same shared six-call budget with `_matches_tool_filter` ignoring `ScopeAnalysisResult`; per-tool limiters were tried first and loop inside langchain when a message's only call is the blocked one (recursion limit), which the end-to-end fake-model test caught.

agents/scope_analysis_agent.py
```diff
@@ -78,6 +79,17 @@ def llm_str(self) -> str:
         return "\n".join(part for part in (self.description, components, relations) if part)
 
 
+class RepositoryToolBudget(ToolCallLimitMiddleware):
```
ivanmilevtues, line 82: blckerish: This seems rather like a patchwork fix, we can keep it for now. This is an indication of badly designed prompts or tools. We should rather design the tools and the agents in a way where we shouldn't restrict the usage of tools either remove the tool whatsoever or redesign it to be used more efficiently.

We can merge it in, but we should open a new task to take a closer look and design the tools better or have some prompt optimisation as this fallback will make our analysis's inconsistent as one time we might hit it another not really.

ivanmilevtues (approved):
Only a single blocker-ish comment, I have added it but I think we should probably address it after we release. I would rather have this in and then we can take a closer look and analyze the tooling as it has not been adapted to the new llm agent workflow.

PR #568, opened 2026-09-07
feat(cfg): type-reference edges from source, drawn as relations; fix the callable-usage walk; two save-path quadratics

Svilen-Stefanov:
Answers "where do abp's disconnected components come from?" and closes the C# half of it: a type-reference (`TYPEREF`) layer read from source with tree-sitter, reference edges drawn as relations, the callable-usage walk fixed, and the two save-path quadratics that a type-complete graph exposes. Measured on abp (6,615 files, 29 .NET solutions) across three full runs.

## Where the floating boxes came from (0.14.0 run, 504 components)

| cause | scale | evidence |
|---|---|---|
| **C#: missing *type* edges, not missing calls** | 43 of 456 components | every one is a type-wired shape — `*Module`/`*Installer` (`[DependsOn(typeof(X))]`), Mapperly mappers (generic base args), DbContexts (`DbSet<T>`), `templates/*` (framework only via NuGet), consts/enums; 885 hard-isolated C# nodes, 777 of them classes |
| **TS/JS: a hub fan-in bug, then genuinely independent scripts** | 40 of 44 | 29 copies of `var abp = abp \|\| {}` absorbed **95.9% of all TS edges** (11,368 of 11,859); of the 29,725 call sites behind them, 58 genuinely passed or returned the object |
| **root/L1: attribution** | 9/9 → 3/9 rolled up | relations credit the deepest owner, so parents look floating |

## The commits

1. **`fix(edges)`** — the callable-usage walk climbed through function boundaries and member chains, so `$(function () { abp.x(); })`, `f(abp.x.y)` and `return { k: abp.x }` all counted as "abp passed as a callable". Stop at a function boundary; a member-chain receiver is never the value passed or returned (both walks, every grammar's receiver field). abp TS edges 11,859 → 1,263; the 783 hub edges left are the genuine `f(abp)` uses.
2. **`feat(cfg)`** — `SourceInspector.find_type_reference_sites` reads type positions (C# type slots, `type_argument_list`, `base_list`, `typeof`, `as`/`is`, catch, member-access receivers; TS `type_identifier`, `extends`, decorator arrays). `type_reference_builder.TypeIndex` resolves names against the **merged** graph's class-like nodes — C# by declared-namespace visibility (`find_namespace_context`; a file-scoped `namespace X;` spans the file, its declarations are siblings in this grammar), then usings, then directory proximity; TS through relative imports (`find_import_bindings`), package imports stay external. `StaticAnalyzer.analyze` rebuilds the layer after the per-project graphs merge, which is what lets a template's `AbpModule` resolve into `framework/`. abp: 163,509 sites → 29,887 edges in 10 s; hard-isolated C# nodes 885 → 224; isolated files 625 → 53. Receivers are C#-only: in a script a promoted variable like `data` resolves to the wrong file.
3. **`feat(relations)`** — `RELATION_REFERENCE_KINDS = (INHERITS, TYPEREF)` count as relations in `build_component_relations`, the clustering service's group connections (`ClusterConnectionEdge.kind`) and the per-scope merge. A pair with calls keeps "calls"; reference-only pairs get "uses" / "inherits from" (`static_relation_label`), and `RelationEdge.from_reference` tags the edge's `description` — no schema change. The scope prompt gains a per-pair `type_references` count; reference edges stay out of `bordering_files` (one line per file per reference bordered nearly every abp file and grew the root prompt 472k → 598k tokens in the partial run).
4. **`perf(save)`** — `FileEntry.merge_from` deep-copied every owned method on every merge, and every progress save re-merges every file of every scope: 4.4M deepcopies per save on abp (81.5 s → 5.7 s). `build_global_relations` scanned every semantic relation per static pair (28M `is_self_or_descendant` calls, 11.3 s → 0.3 s).

11. **`feat(clustering)`** — `TYPEREF` was already in `AFFINE_REFERENCE_KINDS` on main, where nothing emitted it and it was inert; this branch makes it live, which would have re-partitioned every repository as a side effect. It is now excluded, so the layer draws an edge and leaves the boxes alone. Measured both ways: on eShop counting it costs two of the nine boxes its maintainers publish (7/9 → 5/9, because every service names the event bus's types and the fold reads that density as affinity); on abp it would help (largest box 53.8% → 24.9%). Helping where names cannot separate and hurting where they already do is not one switch, so drawing and partitioning stay separate decisions.

Three more from the review round, after the bot comments (9 findings: 4 real, 5 declined with evidence — see [the reply](https://github.com/CodeBoarding/CodeBoarding/pull/568#issuecomment-5577581381)):

5. **`fix(cfg)`** — the C# reader threw away part of what the source wrote and the resolver guessed at the rest. A `using` was flattened to its target, losing the alias name (so `using BasketItem = ...Models.Basket.BasketItem;` was decided by directory proximity, and on eShop picked the generated gRPC type), the `static` kind, and the lexical scope (a using inside one namespace block was in force in its siblings). A written type name reached the index unnormalised: `A.Box<T>` looked up as `Box<T>` where the index is keyed `Box`, `global::A.Foo` carrying a qualifier no namespace can end with. And a TS relative import that writes its extension matched nothing — `./svc.js` never string-prefixes `svc.ts`, which is most of what a NodeNext project writes.
6. **`chore(cache)`** — `_TAG_VERSION` v7 → v8. Commit 1 changes which call edges come out of identical source, and the freshness gate is the source SHA, so without a bump a warm start keeps serving the old graph.
7. **`perf(cfg)`** — the pass decided the language *after* parsing the file, so every Python, Java, Go, PHP and Rust file paid a full tree-sitter parse to have the tree thrown away.
8. **`refactor(relations)`** — one vocabulary for edge kinds. `EdgeKind` gains `CALL` and a `relation_label` per kind, `RelationEdge` carries the kind of the static edge behind it, and `Relation.default_label` records that a label was computed rather than written — the carry filter used to decide that by comparing the text, so a model that described a pair as "uses" had its own wording and evidence thrown away on the next incremental run. `ReferenceEdge` now carries the sites that produced it, so a type reference reaches the prompt and the relation with the positions where the type is named, the way a call does with its call sites.
9. **`feat(cfg)`** — resolution now walks the scopes C# and TS actually have. One `ImportDirective`/`NameScope` describes a using and an import alike; lookup goes enclosing types, then each namespace with the usings written in *its* declaration, then the compilation unit with its `global using`s, stopping at the first scope holding a candidate. Consequences: a qualifier matches whole segments (`Domain` no longer accepts `FooDomain`), `Outer.Inner` resolves through the enclosing type instead of comparing it with a namespace, a `global using` governs its whole compilation rather than its own file, and a base written in a base list is no longer doubled as a type reference on top of the inheritance edge the graph already has. Replayed offline on eShop: 20 edges gained, one lost; serilog and nest identical.
10. **`fix(cfg)`** — a bare name never means a type nested in another type. A namespace can be opened by an import this layer cannot see, so a name unique in the graph resolves from anywhere; a nested type is not like that, and abp's `TranslateCommand.Options.File` was answering every `File.Create(...)` in the repository. 440 edges over 10 targets and 47 component pairs removed, 4 added where the wrong candidate had made a name ambiguous. See "Where the edge was simply wrong" below.

## What kind of edge this actually discovers

A call is a **runtime** dependency: this path executes that one. A type reference is a
**compile-time** dependency: this code does not build without that type. Both are real and they
break differently, which is why these carry the label `uses` rather than `calls` and stay off
`CallGraph.edges`.

Everything below is measured on abp itself rather than on a fixture, by resolving each type
position against the real index and grouping the resulting edges by the syntax that produced them
(`runs/abp-type-references/tools/edge_positions.py` in the evals PR). 163,509 C# positions read,
37,367 resolving to a type of this repository, **19,123 crossing a component boundary**. They
touch 2,687 component pairs, and **1,659 of those pairs have no call edge in either direction** —
those relations could not be drawn at all before this branch.

| position | edges | on a pair no call connects | as written in abp |
|---|---:|---:|---|
| static member receiver | 5,872 | 2,524 | `request.Headers.Add(TimeZoneConsts.DefaultTimeZoneKey, …)` |
| generic argument | 4,100 | 1,611 | `Configure<AbpLocalizationOptions>(…)` |
| parameter type | 3,565 | 1,478 | `MauiBlazorContentFileProvider(IVirtualFileProvider provider)` |
| base type or interface | 2,743 | 1,804 | `public class AbpAIAbstractionsModule : AbpModule` |
| field or property type | 1,490 | 573 | `protected IClock Clock { get; }` |
| `typeof(…)` | 627 | 142 | `typeof(IApplicationService).IsAssignableFrom(t)` |
| return type | 280 | 72 | `public static NumberBag BuildBag()` |
| generic constraint | 172 | 61 | `where TDestination : IHasExtraProperties` |
| pattern, `as`, `is`, cast | 220 | 39 | `if (o is Container c)` |
| `catch` type | 24 | 16 | `catch (AbpDbConcurrencyException ex)` |
| array creation | 12 | 12 | `.AddRange(new EntityAction[] { … })` |
| local variable type | 17 | 7 | rare in C#, which writes `var` |

The tool splits the generic row further by what the generic sits on — 1,331 on a static member
receiver, 802 on a return type, 717 on a call's arguments, 339 on a parameter — which is where to
look if one of these groups deserves to be dropped.

Of the 2,743 base-list references, 1,353 are inheritance the language server already recorded:
those stay `INHERITS` and read "inherits from". The other **1,390 are inheritance no server ever
saw**, because the base type lives in a solution that only references it as a NuGet package.

### Groups that no call graph could have found

**The module graph.** abp wires itself with `[DependsOn(typeof(X))]` and `Configure<TOptions>`;
not one of those is a call. `AbpIoLocalizationModule.cs:37` names `AbpValidationResource`, line 22
names `AbpVirtualFileSystemOptions`, line 18 declares `: AbpModule`. Before this branch that whole
module, and every `*Module` like it, was a box with no edges. This is the clearest case where the
edge is not merely defensible but is the architecture.

**Constructor injection.** `MauiBlazorContentFileProvider(IVirtualFileProvider virtualFileProvider)`
is the strongest dependency in a DI codebase and produces no call site anywhere: the container does
the constructing. 3,565 edges, 1,478 of them the only thing joining their two components.

**Implementing an interface from another module.**
`public class ChatClientAccessor : IChatClientAccessor, ITransientDependency` — the implementer
never calls the interface, so a call graph sees nothing. 1,804 of these join components that were
otherwise unconnected, and they are the reason abp's `*Installer` classes and Mapperly mappers
stopped floating.

**Constants, enums and options classes** — the case you named. `TimeZoneConsts.DefaultTimeZoneKey`
is a field read, not a call, so the class holding the constants was unreachable by any call edge;
same for an enum's member and for a `static class` of option names. This is what the "static member
receiver" row is mostly made of, and it is the only way such a class is reachable at all.

**Exception contracts.** `catch (AbpDbConcurrencyException ex)` in `IdentityRoleStore` is a real
compile-time dependency on `Volo.Abp.Data`, and one nobody would draw by hand. Small (24), and
honestly closer to incidental than to architecture.

### Where the edge is real but the signal is weaker

- **A static member receiver double counts.** `Formatter.FormatLabel(name)` is both a call and a
  type reference: the pair usually already has a call edge, and the reference adds the *class*
  where the call named the method. It stays because dropping it would lose every consts class and
  enum, which are never called — but it inflates a pair that was already connected. 3,348 of the
  5,872 are on pairs a call already joins.
- **A local variable type** is inside a method body rather than on an interface, so it says less
  about the architecture than a parameter does. In C# this is nearly nothing (17) because the
  language writes `var`.
- **`as` / `is` / a cast** is genuine and usually incidental.

### Where the edge was simply wrong

Grouping abp's real edges this way is what turned up a defect, and it is the largest single
false-positive family I found:

**A nested type answering a bare name.** abp declares `class TranslateCommand { … static class
Options { static class File } }`. `File` is then the only type of that name in the repository, so
every `File.Create(…)`, `Directory.Exists(…)` and `Url` written anywhere in abp — all meaning
`System.IO` — resolved to the CLI's nested option classes. 440 edges over 10 targets, and **47
component pairs that existed for no other reason**, nearly all of them an arrow from some component
into abp's CLI. A nested type is reachable unqualified only from inside its enclosing type or
through `using static`, never through an import the reader failed to see, so the last commit stops
treating it like a namespace. 4 edges appear in exchange, where the wrong candidate had been making
a name ambiguous.

**A receiver that is a property, not a type.** `if (Clock.SupportsMultipleTimezone)` reads the
enclosing class's own injected `IClock` property, but the reader sees an identifier in a receiver
position and the index answers with abp's concrete `Clock` class. The C# graph keeps no property
symbols (27 in 24k nodes), so there is nothing to check the name against. Counting receiver names
that are also bound as a member, parameter or local within reach puts an **upper bound of 1,158 of
the 5,872 receiver edges (19.7%)** on this — an upper bound because one name can be both, so some
of those are genuine. The names are abp's injected base-class properties: `CurrentTenant` 143,
`Clock` 63, `UnitOfWorkManager` 58, `CurrentUser` 33. Not fixed here; the fix is to let the reader
skip a receiver the file itself binds, and I would rather measure that on more than one repository
before shipping it.

### What is deliberately not drawn

- **`new Other()`** — the constructor call already carries that edge; counting the type as well
  would double the pair.
- **A declaration's own name** — `class Service` is not a reference to `Service`.
- **A Go method's receiver** — `func (e *Entity) SetType(…)` is containment, already `CONTAINS`
  (fixed in #574 after this same exercise; it was 42% of Go's references).
- **Anything outside the repository.** 126,142 of the 163,509 positions resolve to nothing —
  `Guid`, `Task`, `DateTime`, `CancellationToken`, lambda parameters. That is load-bearing rather
  than wasted work: the layer only ever draws between two components of *this* repository, so a
  framework type cannot manufacture an edge.

## Where the type references come from, and what they cost

Nothing queries the language server for them. `SourceInspector.find_type_reference_sites` walks the tree-sitter parse a file already needs and yields the positions where a type is *named* rather than called; `TypeIndex` then resolves each name against the class-like nodes of the **merged** graph, which is what lets a template resolve into `framework/` although its own solution only sees a package reference. The C# rule is namespace visibility — an alias in scope decides outright, then plain `using` imports, then directory proximity for a genuine tie. The TypeScript rule is the import table, with package specifiers left external on purpose.

**Two of the eight wired languages produce them.** C# reads type slots, `type_argument_list`, `base_list`, `typeof`, `as`/`is`, `catch` and member-access receivers; TypeScript reads `type_identifier`, `extends` and the class names inside decorator arrays. JavaScript shares the TS selector but has almost nothing for it to find — it has no type annotations, so only `class X extends Y` and decorators qualify. Python, Java, Go, PHP and Rust have no selector and return immediately.

Measured over exactly the files a real run analyses (ignore rules applied), after commit 7:

| language | repo | files | MB | pass | sites found |
|---|---|---:|---:|---:|---:|
| C# | abp | 6,284 | 14.4 | **9.4 s** | 163,509 |
| TypeScript | mermaid | 388 | 1.5 | **1.3 s** | 3,281 |
| JavaScript | abp | 210 | 0.6 | **0.5 s** | 0 |
| Python | django | 894 | 5.6 | **0.00 s** | 0 |
| Java | mockito | 990 | 3.5 | **0.00 s** | 0 |
| Go | fzf | 53 | 0.6 | **0.00 s** | 0 |
| Rust | regex | 181 | 5.3 | **0.00 s** | 0 |

So the cost is ~0.7 s per MB of C# or TypeScript source and nothing at all elsewhere. In the abp run it is 10.1 s of a 55-minute analysis. Before commit 7 the languages in the bottom half of that table paid for a parse they could not use — 1.9 s on django, 0.4 s on regex, 0.3 s on mockito, always for zero sites.

The private integration suite was dispatched against this branch per language (`workflow_dispatch` → `CodeBoarding-tests`). Python: `integration-tests (windows-latest, python)` **passed**; the run is red only on `live-provider-checks`, an unrelated model-capability coverage assertion about ids like `o1-mini` and `claude-3-5-sonnet-20241022` falling back to generic capabilities. C# dispatched at `7474129`.

## Three full abp runs, Kimi-K2.6 over Azure

| | main `e38c07e3` | branch, commits 1-3 partial | branch, all four |
|---|---:|---:|---:|
| wall clock | 55.5 min | 2h09m | **55.5 min** |
| static / LLM phase | 25.5 / 30.1 | 25.3 / 104 | 25.8 / 29.8 |
| TYPEREF pass | — | 10 s | 10 s |
| disconnected components (rolled up) | 85 / 504 | 50 / 482 | **50 / 485** |
| disconnected leaves | 73 / 417 | 42 / 397 | 42 / 399 |
| disconnected C# | 43 / 456 | 0 / 426 | **0 / 429** |
| relations (global) | 1,955 | 3,585 | 3,612 (678 "uses", 14 "inherits from") |
| reference-backed edges in relations | 0 | 24,647 | 24,730 |
| scopes falling back to deterministic names | 2 / 88 | 2 / 86 (scope 1 = 34 children) | **1 / 87** (root only) |
| largest non-root prompt | 207k | 236k | **145k** |
| TS call edges / into `abp` hubs | 11,859 / 11,368 | 8,515 / 8,033 | 1,263 / 783 |
| tokens in / out | 1.80M / 0.86M | 1.96M / 0.95M | 1.77M / 0.91M |

> [!IMPORTANT]
> **These three runs predate the last commit, and the columns that depend on the grouping no longer
> describe what this PR does.** They were made while `TYPEREF` still counted as a link for the fold
> — the line `AFFINE_REFERENCE_KINDS` has since been narrowed to inheritance alone, so the partition
> is now left exactly as it is on `main`.
>
> Unaffected: everything about the *edges* — the TYPEREF pass and its cost, the C# isolation figures,
> the callable-usage fix and its TS edge counts, and the two save-path quadratics.
> Affected: `components`, `disconnected components`, `disconnected leaves`, `scopes falling back to
> deterministic names` and `largest non-root prompt`, all of which move with the partition. On the
> shipped setting abp keeps main's 504 components and gains relations rather than re-partitioning.
>
> Measured like for like instead, same graph and same fold, no model: abp gains **4,695 concrete
> edges** across 11 new component pairs at depth 1, and eShop 74 across 5. Both are browsable, with
> the file and line behind every edge, in [CodeBoarding-evals#25](https://github.com/CodeBoarding/CodeBoarding-evals/pull/25)
> (`ui/edges.html`). A fresh depth-3 abp run on the shipped setting has not been made.

The partial run's 2h09m was the save path, not the analysis: 85 progress saves at 50-80 s each. Commit 4 makes the tail faster than main's (265+405 s of CPU-bound gaps vs 580 s).

All 50 remaining floating boxes are JS/TS/Go — per-module page scripts, Angular apps depending on gitignored `@abp/ng.*`, build scripts, service workers. They cannot connect statically today: the `abp.*` API is built by property assignment (`abp.ui.setBusy = function`), which `documentSymbol` never reports, and the C# side is reached over HTTP.

## What it looks like

The same scope of abp on each side, drawn by the graph viewer from the two analyses recorded in
[CodeBoarding-evals#25](https://github.com/CodeBoarding/CodeBoarding-evals/pull/25). Each side is
rendered as its own document so both sit at a comparable zoom; the heights differ because the
layouts do — a scope with no relations is one flat rank of boxes, and edges are what give it depth.

**abp's `Host Applications`, the whole scope.** Seven boxes and not one edge, against six boxes and
three, the module template's hosts wired to the HttpApi host they share. The C# was always there to
read — `[DependsOn(typeof(...))]`, constructor parameter types — but none of it was a call, so none
of it was an edge.

The Host Applications scope before the change: seven component boxes in a single row with no edges between any of them.

*Before — `main` e38c07e3: 7 components, 0 relations.*

The Host Applications scope after the change: six component boxes, three with arrows converging on the Module Template HttpApi Host and Shared component.

*After — this branch: 6 components, 3 relations.*

**`CmsKit`, a whole top-level component**, at the scale a reader actually opens: 30 boxes with 13
relations before, 18 with 31 after. The four boxes still floating along the top right are the
JavaScript and TypeScript ones, which is the residue described under "Not in this PR".

The CmsKit component before the change: thirty component boxes in flat rows, with a long right-hand tail of boxes carrying no edge at all.

*Before — `main` e38c07e3: 30 components, 13 relations, and a long tail of boxes with no edge.*

The CmsKit component after the change: eighteen boxes in a layered graph with thirty-one edges, and four unconnected boxes at the top right.

*After — 18 components, 31 relations, and the four JS/TS boxes that still float.*

## Not in this PR

- The root scope still overflows (492k tokens; `groups[].files` at 6,428 entries is 81% of it) — the #565 gap, unchanged by this work.
- An LLM relation labelled literally "uses" is treated as a static default by the carry filter in `scope_assembly` (it is not carried across runs). Cosmetic; a distinct label would avoid it.
- Incremental runs rebuild the TYPEREF layer from scratch (10 s on abp) rather than re-deriving per changed file; correct, not optimal.
- `IMPORT` edges still have no producer.
- A member-access receiver that is a property rather than a type still resolves to the class of that name — `Clock.SupportsMultipleTimezone` on an injected `IClock` property finds abp's concrete `Clock`. Upper bound 1,158 of 5,872 receiver edges; measured, not fixed (see "Where the edge was simply wrong").

## Tests

Type sites for C#/TS/JS, name scopes and import directives, every `TypeIndex` resolution rule (including the three new nested-type cases), the builder end to end, reference-edge relations and labels, group connections, scope merge, renderer counts, the walk rules. `uv run mypy .` and `black --check` clean; **2,153 passed** — the 3 failures in `tests/test_cli_parser.py::test_main_renders_after_any_successful_analysis` fail identically on pristine `e38c07e3`.

The runs, the zero-LLM harnesses (offline TYPEREF replay against a committed pkl, hub call-site classifier, run comparison) and the write-up are in [CodeBoarding-evals#25](https://github.com/CodeBoarding/CodeBoarding-evals/pull/25); the `analysis.json` artifacts are on the CodeBoarding/abp fork — run 1 on `codeboarding-analysis-20260907` ([abp#2](https://github.com/CodeBoarding/abp/pull/2)), run 3 on `codeboarding-analysis-20260907-typeref`, one commit on top of it so the two diff directly.

## Summary by CodeRabbit

* **New Features**
  * Diagrams now include type-reference and inheritance relationships alongside call relationships.
  * Static relationships are labeled more clearly, including “inherits from” and “uses.”
  * Type references resolve across merged projects, namespaces, aliases, global imports, and module imports.
  * Scope views distinguish calls from type references and provide more accurate connection and boundary details.
  * Analysis now reports unreadable source files and preserves reference locations.

* **Bug Fixes**
  * Improved detection of callable usage and type names across supported languages.
  * Older analysis results are safely rejected and reprocessed when necessary.

### DIFF
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -11,6 +11,7 @@ from pydantic import BaseModel, Field
 from clustering_ids import ComponentId
 from agents.file_index_models import FileEntry, FileMethodGroup, MethodIndexEntry
 from agents.scope_ids import ROOT_SCOPE_ID
+from static_analyzer.cfg.edge import EdgeKind
 
 logger = logging.getLogger(__name__)
 
@@ -77,6 +78,13 @@ class RelationCallSite(BaseModel):
     column: int = Field(description="One-based column number of the call site in the source file.")
 
 
+# What a static reference edge says about itself; ``static_relation_label`` reads it back.
+REFERENCE_EDGE_DESCRIPTIONS: dict[EdgeKind, str] = {
+    EdgeKind.INHERITS: "inherits from",
+    EdgeKind.TYPEREF: "references type",
+}
+
+
 class RelationEdge(LLMBaseModel):
     """A source-to-target code reference that supports a component relation."""
 
@@ -119,6 +127,25 @@ class RelationEdge(LLMBaseModel):
                 reference_end_line=edge.dst_node.line_end,
             ),
             call_sites=[RelationCallSite.model_validate(call_site) for call_site in edge.call_sites],
+        )
+
+    @classmethod
+    def from_reference(cls, source, target, kind: EdgeKind) -> RelationEdge:
+        """An edge backed by a reference (inheritance, a type mention) rather than a call site."""
+        return cls(
+            source=SourceCodeReference(
+                qualified_name=source.fully_qualified_name,
+                reference_file=source.file_path,
+                reference_start_line=source.line_start,
+                reference_end_line=source.line_end,
+            ),
+            target=SourceCodeReference(
+                qualified_name=target.fully_qualified_name,
+                reference_file=target.file_path,
+                reference_start_line=target.line_start,
+                reference_end_line=target.line_end,
+            ),
+            description=REFERENCE_EDGE_DESCRIPTIONS[kind],
         )
 
     def llm_str(self) -> str:
--- a/agents/agent_responses.py
+++ b/agents/agent_responses.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import abc
 import logging
 from abc import abstractmethod
-from collections.abc import Hashable
+from collections.abc import Hashable, Sequence
 from enum import StrEnum
 
 from pydantic import BaseModel, Field
@@ -11,7 +11,8 @@ from pydantic import BaseModel, Field
 from clustering_ids import ComponentId
 from agents.file_index_models import FileEntry, FileMethodGroup, MethodIndexEntry
 from agents.scope_ids import ROOT_SCOPE_ID
-from static_analyzer.cfg.edge import EdgeKind
+from static_analyzer.cfg.edge import CallSiteLocation, EdgeKind
+from static_analyzer.node import Node
 
 logger = logging.getLogger(__name__)
 
@@ -78,13 +79,6 @@ class RelationCallSite(BaseModel):
     column: int = Field(description="One-based column number of the call site in the source file.")
 
 
-# What a static reference edge says about itself; ``static_relation_label`` reads it back.
-REFERENCE_EDGE_DESCRIPTIONS: dict[EdgeKind, str] = {
-    EdgeKind.INHERITS: "inherits from",
-    EdgeKind.TYPEREF: "references type",
-}
-
-
 class RelationEdge(LLMBaseModel):
     """A source-to-target code reference that supports a component relation."""
 
@@ -93,7 +87,12 @@ class RelationEdge(LLMBaseModel):
     description: str = Field(default="", description="Short explanation of how source reaches or configures target.")
     call_sites: list[RelationCallSite] = Field(
         default_factory=list,
-        description="Call-site line and column pairs for this edge.",
+        description="Line and column pairs where this edge occurs.",
+        exclude=True,
+    )
+    kind: EdgeKind = Field(
+        default=EdgeKind.CALL,
+        description="The static edge behind this one: a call unless it came from a reference edge.",
         exclude=True,
     )
 
@@ -130,8 +129,10 @@ class RelationEdge(LLMBaseModel):
         )
 
     @classmethod
-    def from_reference(cls, source, target, kind: EdgeKind) -> RelationEdge:
-        """An edge backed by a reference (inheritance, a type mention) rather than a call site."""
+    def from_reference(
+        cls, source: Node, target: Node, kind: EdgeKind, sites: Sequence[CallSiteLocation] = ()
+    ) -> RelationEdge:
+        """An edge backed by a reference (inheritance, a type mention) rather than a call."""
         return cls(
             source=SourceCodeReference(
                 qualified_name=source.fully_qualified_name,
@@ -145,7 +146,9 @@ class RelationEdge(LLMBaseModel):
                 reference_start_line=target.line_start,
                 reference_end_line=target.line_end,
             ),
-            description=REFERENCE_EDGE_DESCRIPTIONS[kind],
+            description=kind.relation_label,
+            call_sites=[RelationCallSite.model_validate(site) for site in sites],
+            kind=kind,
         )
 
     def llm_str(self) -> str:
@@ -187,6 +190,20 @@ def _relation_endpoint_from_key(
     )
 
 
+def static_relation_label(edges: Sequence[RelationEdge]) -> str:
+    """The label a relation gets from its static edges alone.
+
+    A single call outranks any reference; one kind of reference reads as that kind; mixed
+    references read as the umbrella ``uses``.
+    """
+    kinds = {edge.kind for edge in edges}
+    if not kinds or EdgeKind.CALL in kinds:
+        return EdgeKind.CALL.relation_label
+    if len(kinds) == 1:
+        return next(iter(kinds)).relation_label
+    return EdgeKind.TYPEREF.relation_label
+
+
 class Relation(LLMBaseModel):
     """A relationship between two components."""
 
@@ -210,6 +227,11 @@ class Relation(LLMBaseModel):
     src_id: str = Field(default="", description="Component ID of the source.", exclude=True)
     dst_id: str = Field(default="", description="Component ID of the destination.", exclude=True)
     is_static: bool = Field(default=False, description="True if derived from static CFG analysis.", exclude=True)
+    default_label: bool = Field(
+        default=False,
+        description="True when the label is what the static edges give the pair, not a description someone wrote.",
+        exclude=True,
+    )
     all_edges: list[RelationEdge] = Field(
         default_factory=list,
         description="All known source-to-target edges for this relation, populated deterministically when available.",
@@ -217,26 +239,17 @@ class Relation(LLMBaseModel):
     )
 
     @classmethod
-    def from_edges(
-        cls,
-        relation: str,
-        src_name: str,
-        dst_name: str,
-        src_id: str,
-        dst_id: str,
-        edges: list[RelationEdge],
-        is_static: bool,
-        evidence: str = "",
-    ) -> Relation:
+    def from_edges(cls, src_name: str, dst_name: str, src_id: str, dst_id: str, edges: list[RelationEdge]) -> Relation:
+        """The relation the static edges alone make between two components, labelled by their kinds."""
         return cls(
-            relation=relation,
+            relation=static_relation_label(edges),
             src_name=src_name,
             dst_name=dst_name,
-            evidence=evidence,
             key_edges=[],
             src_id=src_id,
             dst_id=dst_id,
-            is_static=is_static,
+            is_static=True,
+            default_label=True,
             all_edges=cls.unique_edges(edges),
         )
 
@@ -261,6 +274,7 @@ class Relation(LLMBaseModel):
             src_id=self.src_id,
             dst_id=self.dst_id,
             is_static=self.is_static,
+            default_label=self.default_label,
             all_edges=all_edges,
         )
 
--- a/agents/llm_renderers/scope.py
+++ b/agents/llm_renderers/scope.py
@@ -10,7 +10,7 @@ from typing import Any
 
 from agents.agent_responses import AnalysisInsights
 from repo_utils.path_utils import normalize_repo_path
-from static_analyzer.cfg.edge import EdgeKind
+from static_analyzer.cfg.edge import CALL_EDGE_KIND, EdgeKind
 from static_analyzer.clustering import ClusterConnectionEdge, ClusterGroup, ClusterScopeResult
 from static_analyzer.node import Node
 
@@ -146,12 +146,15 @@ def _known_connections(scope: ClusterScopeResult, repo_dir: Path) -> list[dict[s
             seen.add(key)
             distinct.append(edge)
         distinct.sort(key=lambda edge: (edge.source_qualified_name, edge.target_qualified_name))
+        calls = [edge for edge in distinct if edge.kind == CALL_EDGE_KIND]
+        references = [edge for edge in distinct if edge.kind != CALL_EDGE_KIND]
         connections.append(
             {
                 "source_group_id": source_group_id,
                 "target_group_id": target_group_id,
-                "calls": len(distinct),
-                "examples": [_example(edge, scope, repo_dir) for edge in distinct[:MAX_EXAMPLE_EDGES]],
+                "calls": len(calls),
+                "type_references": len(references),
+                "examples": [_example(edge, scope, repo_dir) for edge in (calls + references)[:MAX_EXAMPLE_EDGES]],
             }
         )
     return connections
@@ -186,6 +189,10 @@ def _boundary_reasons(
     }
     for connection in scope.connections:
         for edge in connection.edges:
+            # Only calls border a file here; a type reference is counted per pair in
+            # ``known_connections`` instead, or it would border nearly every file.
+            if edge.kind != CALL_EDGE_KIND:
+                continue
             graph = scope.graphs_by_language.get(edge.language)
             if graph is None:
                 continue
@@ -200,7 +207,9 @@ def _boundary_reasons(
 
     for language, graph in scope.graphs_by_language.items():
         for reference in graph.reference_edges:
-            if reference.kind is EdgeKind.CONTAINS:
+            # TYPEREF is dense enough to border nearly every file; its count rides on
+            # ``known_connections`` instead, where a pair costs one line.
+            if reference.kind is not EdgeKind.INHERITS:
                 continue
             source_group = owners.get((language, reference.src))
             target_group = owners.get((language, reference.dst))
--- a/agents/llm_renderers/scope.py
+++ b/agents/llm_renderers/scope.py
@@ -10,7 +10,7 @@ from typing import Any
 
 from agents.agent_responses import AnalysisInsights
 from repo_utils.path_utils import normalize_repo_path
-from static_analyzer.cfg.edge import CALL_EDGE_KIND, EdgeKind
+from static_analyzer.cfg.edge import EdgeKind
 from static_analyzer.clustering import ClusterConnectionEdge, ClusterGroup, ClusterScopeResult
 from static_analyzer.node import Node
 
@@ -146,8 +146,8 @@ def _known_connections(scope: ClusterScopeResult, repo_dir: Path) -> list[dict[s
             seen.add(key)
             distinct.append(edge)
         distinct.sort(key=lambda edge: (edge.source_qualified_name, edge.target_qualified_name))
-        calls = [edge for edge in distinct if edge.kind == CALL_EDGE_KIND]
-        references = [edge for edge in distinct if edge.kind != CALL_EDGE_KIND]
+        calls = [edge for edge in distinct if edge.kind is EdgeKind.CALL]
+        references = [edge for edge in distinct if edge.kind is not EdgeKind.CALL]
         connections.append(
             {
                 "source_group_id": source_group_id,
@@ -191,7 +191,7 @@ def _boundary_reasons(
         for edge in connection.edges:
             # Only calls border a file here; a type reference is counted per pair in
             # ``known_connections`` instead, or it would border nearly every file.
-            if edge.kind != CALL_EDGE_KIND:
+            if edge.kind is not EdgeKind.CALL:
                 continue
             graph = scope.graphs_by_language.get(edge.language)
             if graph is None:
--- a/agents/relation_edges.py
+++ b/agents/relation_edges.py
@@ -1,11 +1,29 @@
 """Merge component relations and index their source endpoints."""
 
-from collections.abc import Callable, Collection
+from collections.abc import Callable, Collection, Sequence
 from pathlib import Path
 
-from agents.agent_responses import AnalysisInsights, Relation, RelationEdge, SourceCodeReference
+from agents.agent_responses import (
+    REFERENCE_EDGE_DESCRIPTIONS,
+    AnalysisInsights,
+    Relation,
+    RelationEdge,
+    SourceCodeReference,
+)
 from clustering_ids import is_self_or_descendant
 from repo_utils.path_utils import normalize_repo_path
+from constants import DEFAULT_STATIC_RELATION_LABEL, INHERITANCE_RELATION_LABEL, TYPE_REFERENCE_RELATION_LABEL
+from static_analyzer.cfg.edge import EdgeKind
+
+
+def static_relation_label(edges: Sequence[RelationEdge]) -> str:
+    """The label a relation gets from its static edges alone: a single call outranks any reference."""
+    descriptions = {edge.description for edge in edges}
+    if not descriptions or descriptions - set(REFERENCE_EDGE_DESCRIPTIONS.values()):
+        return DEFAULT_STATIC_RELATION_LABEL
+    if descriptions == {REFERENCE_EDGE_DESCRIPTIONS[EdgeKind.INHERITS]}:
+        return INHERITANCE_RELATION_LABEL
+    return TYPE_REFERENCE_RELATION_LABEL
 
 
 def append_or_merge_relation(
--- a/agents/relation_edges.py
+++ b/agents/relation_edges.py
@@ -1,29 +1,11 @@
 """Merge component relations and index their source endpoints."""
 
-from collections.abc import Callable, Collection, Sequence
+from collections.abc import Callable, Collection
 from pathlib import Path
 
-from agents.agent_responses import (
-    REFERENCE_EDGE_DESCRIPTIONS,
-    AnalysisInsights,
-    Relation,
-    RelationEdge,
-    SourceCodeReference,
-)
+from agents.agent_responses import AnalysisInsights, Relation, RelationEdge, SourceCodeReference
 from clustering_ids import is_self_or_descendant
 from repo_utils.path_utils import normalize_repo_path
-from constants import DEFAULT_STATIC_RELATION_LABEL, INHERITANCE_RELATION_LABEL, TYPE_REFERENCE_RELATION_LABEL
-from static_analyzer.cfg.edge import EdgeKind
-
-
-def static_relation_label(edges: Sequence[RelationEdge]) -> str:
-    """The label a relation gets from its static edges alone: a single call outranks any reference."""
-    descriptions = {edge.description for edge in edges}
-    if not descriptions or descriptions - set(REFERENCE_EDGE_DESCRIPTIONS.values()):
-        return DEFAULT_STATIC_RELATION_LABEL
-    if descriptions == {REFERENCE_EDGE_DESCRIPTIONS[EdgeKind.INHERITS]}:
-        return INHERITANCE_RELATION_LABEL
-    return TYPE_REFERENCE_RELATION_LABEL
 
 
 def append_or_merge_relation(
@@ -364,7 +346,11 @@ def _restore_baseline_orientation(relation: Relation, baseline_by_pair: dict) ->
 
 def _restore_baseline_wording(fresh: Relation, previous: Relation) -> Relation:
     """Restore baseline wording and highlighting onto fresh edge metadata."""
-    wording = {"relation": previous.relation, "evidence": previous.evidence}
+    wording = {
+        "relation": previous.relation,
+        "evidence": previous.evidence,
+        "default_label": previous.default_label,
+    }
     if not previous.key_edges:
         return fresh.model_copy(update=wording)
     highlighted = {
--- a/diagram_analysis/scope_assembly.py
+++ b/diagram_analysis/scope_assembly.py
@@ -14,15 +14,13 @@ from agents.relation_edges import (
     edge_crosses_components,
     ground_relation_edges,
     prune_ungrounded_edges,
-    static_relation_label,
 )
 from agents.scope_analysis_agent import ScopeAnalysisResult
 from agents.scope_ids import ROOT_SCOPE_ID
 from clustering_ids import CodeBoardingClusterIds
-from constants import STATIC_RELATION_LABELS
 from diagram_analysis.file_index import build_file_methods_from_nodes, build_files_index
 from static_analyzer import StaticAnalysisFatalError
-from static_analyzer.cfg import CALL_EDGE_KIND, Edge, EdgeKind
+from static_analyzer.cfg import Edge, EdgeKind
 from repo_utils.path_utils import normalize_repo_path
 from static_analyzer.clustering import ClusterGroup, ClusterScopeResult, GroupConnection
 from static_analyzer.reference_resolver import StaticReferenceResolver
@@ -218,14 +216,15 @@ class ScopeAssembler:
 
         # A group the model was asked about drops every relation it touches. Where the model did
         # not label one back, keep the label the previous run gave it rather than resetting a
-        # still-connected edge to the generic default.
+        # still-connected edge to the generic default. A label that was itself the static default
+        # is not carried: the merge recomputes it from today's edges.
         carried = [
             relation.model_copy(deep=True)
             for pair, relation in existing_by_pair.items()
             if pair not in seen_pairs
             and pair not in preserved_pairs
             and relation.relation.strip()
-            and relation.relation not in STATIC_RELATION_LABELS
+            and not relation.default_label
             and scope.connection_between(*pair) is not None
         ]
         for relation in carried:
@@ -290,6 +289,7 @@ class ScopeAssembler:
                     src_id=src_id,
                     dst_id=dst_id,
                     is_static=bool(static_edges),
+                    default_label=llm_relation.default_label,
                     all_edges=all_edges,
                 ),
             )
@@ -306,13 +306,11 @@ class ScopeAssembler:
             append_or_merge_relation(
                 merged,
                 Relation.from_edges(
-                    static_relation_label(edges),
                     source.name if source is not None else connection.source_group_id,
                     target.name if target is not None else connection.target_group_id,
                     connection.source_group_id,
                     connection.target_group_id,
                     edges,
-                    True,
                 ),
             )
 
@@ -405,8 +403,10 @@ class ScopeAssembler:
             target = graph.nodes.get(connection_edge.target_qualified_name)
             if source is None or target is None:
                 continue
-            if connection_edge.kind == CALL_EDGE_KIND:
+            if connection_edge.kind is EdgeKind.CALL:
                 edges.append(RelationEdge.from_edge(Edge(source, target, connection_edge.call_sites)))
             else:
-                edges.append(RelationEdge.from_reference(source, target, EdgeKind(connection_edge.kind)))
+                edges.append(
+                    RelationEdge.from_reference(source, target, connection_edge.kind, connection_edge.call_sites)
+                )
         return edges
--- a/diagram_analysis/scope_assembly.py
+++ b/diagram_analysis/scope_assembly.py
@@ -14,14 +14,15 @@ from agents.relation_edges import (
     edge_crosses_components,
     ground_relation_edges,
     prune_ungrounded_edges,
+    static_relation_label,
 )
 from agents.scope_analysis_agent import ScopeAnalysisResult
 from agents.scope_ids import ROOT_SCOPE_ID
 from clustering_ids import CodeBoardingClusterIds
-from constants import DEFAULT_STATIC_RELATION_LABEL
+from constants import STATIC_RELATION_LABELS
 from diagram_analysis.file_index import build_file_methods_from_nodes, build_files_index
 from static_analyzer import StaticAnalysisFatalError
-from static_analyzer.cfg import Edge
+from static_analyzer.cfg import CALL_EDGE_KIND, Edge, EdgeKind
 from repo_utils.path_utils import normalize_repo_path
 from static_analyzer.clustering import ClusterGroup, ClusterScopeResult, GroupConnection
 from static_analyzer.reference_resolver import StaticReferenceResolver
@@ -224,7 +225,7 @@ class ScopeAssembler:
             if pair not in seen_pairs
             and pair not in preserved_pairs
             and relation.relation.strip()
-            and relation.relation != DEFAULT_STATIC_RELATION_LABEL
+            and relation.relation not in STATIC_RELATION_LABELS
             and scope.connection_between(*pair) is not None
         ]
         for relation in carried:
@@ -305,7 +306,7 @@ class ScopeAssembler:
             append_or_merge_relation(
                 merged,
                 Relation.from_edges(
-                    DEFAULT_STATIC_RELATION_LABEL,
+                    static_relation_label(edges),
                     source.name if source is not None else connection.source_group_id,
                     target.name if target is not None else connection.target_group_id,
                     connection.source_group_id,
@@ -404,5 +405,8 @@ class ScopeAssembler:
             target = graph.nodes.get(connection_edge.target_qualified_name)
             if source is None or target is None:
                 continue
-            edges.append(RelationEdge.from_edge(Edge(source, target, connection_edge.call_sites)))
+            if connection_edge.kind == CALL_EDGE_KIND:
+                edges.append(RelationEdge.from_edge(Edge(source, target, connection_edge.call_sites)))
+            else:
+                edges.append(RelationEdge.from_reference(source, target, EdgeKind(connection_edge.kind)))
         return edges
--- a/static_analyzer/cfg/edge.py
+++ b/static_analyzer/cfg/edge.py
@@ -26,10 +26,14 @@ class EdgeKind(StrEnum):
 
 # What structural consumers fold into ``to_networkx`` on top of call edges. The call graph
 # leaves ~a fifth of symbols isolated (constructors, dunders, DI/interface methods), so
-# completing it with these avoids grab-bag components. TYPEREF and IMPORT are plumbed through
-# ``LanguageAnalysisResult`` but no engine emits them yet, and IMPORT is expected to over-merge
-# (coarse, dense, file-level) when one does.
+# completing it with these avoids grab-bag components. TYPEREF is derived from source once
+# the per-project graphs merge (``engine.type_reference_builder``); IMPORT has no producer
+# yet and is expected to over-merge (coarse, dense, file-level) when one does.
 DEFAULT_REFERENCE_KINDS: tuple[EdgeKind, ...] = (EdgeKind.CONTAINS, EdgeKind.INHERITS)
+# The reference kinds that, like a call, make one component depend on another.
+RELATION_REFERENCE_KINDS: tuple[EdgeKind, ...] = (EdgeKind.INHERITS, EdgeKind.TYPEREF)
+# ``ClusterConnectionEdge.kind`` of a plain call; reference edges carry their ``EdgeKind`` value.
+CALL_EDGE_KIND = "call"
 
 
 @dataclass(frozen=True)
--- a/static_analyzer/cfg/edge.py
+++ b/static_analyzer/cfg/edge.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Hashable, Mapping, Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import StrEnum
 from typing import NotRequired, TypedDict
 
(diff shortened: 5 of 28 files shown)


Svilen-Stefanov:
## Review round: 4 of 9 findings fixed, 5 declined with reasons

Every comment was verified against the code before anything was changed — each one reproduced (or refuted) by running the real functions on real repositories, not read off the diff. Two commits: `5efe373` for the three resolution defects, `5e3b1f1` for the cache tag, which is unrelated to the rest.

### Fixed

**`3951618874` — C# `using` loses the alias, the kind and the scope.** Confirmed on real eShop code: `src/ClientApp/Services/Basket/BasketService.cs` aliases `using BasketItem = eShop.ClientApp.Models.Basket.BasketItem;` precisely because two namespaces in scope both declare `BasketItem`; the resolver ignored the alias and picked the generated gRPC type by directory proximity. A `UsingDirective` now records target, alias, `static` and the range it governs. An alias in scope decides the name outright (C# §7.8) and never falls through to the proximity guess; only a plain namespace import makes a namespace nameable; a using written inside a namespace block no longer reaches its siblings.

This **supersedes `3951605173`**, which asked for an alias-to-target map. A map alone would have left the other two halves broken — the scoping and the `static` kind — so the record carries all three rather than bolting a second field alongside `usings`.

**`3951618881` — a written type name reached the index unnormalised.** `A.Box<T>` was looked up as `Box<T>` where the index is keyed `Box`, and `global::A.Foo` carried a qualifier no declared namespace can end with. The selector's four structural branches collapse into one string rule. The split is bracket-aware, because the two callers sit on opposite sides of the same ambiguity: a site can write a dotted type argument (`Box<a.b.T>`) while a qualified name can carry arity on a non-final segment (`A.Repo<T>.Entry`). Splitting on the first `<` or the last `.` alone is wrong for one of them — verified identical to the previous index key on all 30,285 indexed names of abp, eShop and mermaid.

**`3951618892` — a relative TS import that writes its extension matched nothing.** `./svc.js` never string-prefixes `svc.ts`, so the reference was dropped; this is most of what a NodeNext project writes (24 of the 30 bindings in mermaid's `mermaidAPI.ts`). Both sides now reduce to a module stem, which also covers the `./lib/index.js` barrel. The external-package guard is untouched, and a test pins that dropping the extension did not degrade the filter into basename matching.

**`3951618896` — `_TAG_VERSION` still `v7`.** Correct and the most consequential of the nine: an earlier commit here changes which call edges come out of identical source, so without a bump a warm start keeps serving the old graph. Bumped to v8. The changelog entry names only the call-edge predicate — the type-reference layer is rebuilt from source on every run and cannot go stale in a pickle. Both new tests use the literal `v7`, since the existing ones write an unreadable pickle and would pass either way.

### Declined

- **`3951605172`** (keep `kind` in the dedup key) — the collision is reachable, but it only shifts one integer inside a prompt payload: the pair is still reported, `calls` is still correct, no drawn relation is lost, and the collapse matches the precedence this PR states elsewhere. The proposed fix would regress the `examples` list into byte-identical duplicates.
- **`3951605177`** and **`3951618870`** (protect the TYPEREF wipe against unreadable or partial parses) — the premise does not hold. The site walk deliberately traverses ERROR subtrees, so an error tree still yields a superset of a clean parse's names (measured on serilog's `ILogger.cs`: 217 sites, none missing). TYPEREF is not incrementally cached — every run wipes and rebuilds it from the language's complete source list — so an unreadable file gives byte-identical output warm or cold. Preserving a previous run's edges for a file that cannot be read is the plausible-but-wrong result `AGENTS.md` §5 forbids.
- **`3951618886`** (renamed default imports) — the control case is invalid: with no import statement at all the same name also resolves to nothing, so the `default` clause is not what loses the edge. Only reading `export default` declarations would, which this layer does not claim to do for any language. An absent edge here is the resolver's documented contract, not a wrong one.

### One thing worth flagging

An internal review of the fix caught a regression in the fix itself before it was pushed: the first version of the normalised split reordered the operations, which was right for the site but wrong for the index key, and cost 38 correct C# edges on abp — including `class GenerateProxyCommand : ProxyCommandBase<GenerateProxyCommand>`. That is what the bracket-aware split above exists for, and the discriminating cases are pinned in `test_simple_type_name_strips_path_and_arity`.

abp is unchanged in aggregate — 29,887 → 29,891 type references, the same 2,360 undecidable, isolation still 0.9% — which is the intended outcome: a handful of proximity guesses traded for the same number of correctly resolved names. Full suite 2,131 passed; the 3 `test_cli_parser` failures are pre-existing and fail identically on `e38c07e3`.

static_analyzer/engine/models.py
```diff
@@ -65,6 +65,50 @@ def lsp_column(self) -> int:
...
+    line: int  # one-based, like ``CallSite``
+    column: int
+    name: str
+    # The dotted prefix written before the name (``Volo.Abp`` in ``Volo.Abp.Foo``), else empty.
+    qualifier: str = ""
+
+
+@dataclass(frozen=True)
+class UsingDirective:
+    """One C# ``using`` and the lines it governs."""
```
Svilen-Stefanov: I'm not particularly fond that these things are C# specific, I think they should ideally be generally applicable.

static_analyzer/engine/models.py
```diff
@@ -65,6 +65,50 @@ def lsp_column(self) -> int:
         return self.column - 1
 
 
+@dataclass(frozen=True)
+class TypeReferenceSite:
+    """A position naming a type without calling it: a parameter, base, generic argument, ``typeof``."""
+
+    file: str
```
Svilen-Stefanov: I see file: str, is this a name (if so, the var should be named so), if not, should the type be Path or something else?

agents/agent_responses.py
```diff
@@ -77,6 +78,13 @@ class RelationCallSite(BaseModel):
     column: int = Field(description="One-based column number of the call site in the source file.")
 
 
+# What a static reference edge says about itself; ``static_relation_label`` reads it back.
+REFERENCE_EDGE_DESCRIPTIONS: dict[EdgeKind, str] = {
```
Svilen-Stefanov: Why is this declared here, can't we have this declared once, e.g in edge.py and reused in different places?

Svilen-Stefanov: Why is this declared here, can't we have this declared once, e.g in edge.py and reused in different places?

static_analyzer/cfg/edge.py
```diff
@@ -26,10 +26,14 @@ class EdgeKind(StrEnum):
...
-# ``LanguageAnalysisResult`` but no engine emits them yet, and IMPORT is expected to over-merge
-# (coarse, dense, file-level) when one does.
+# completing it with these avoids grab-bag components. TYPEREF is derived from source once
+# the per-project graphs merge (``engine.type_reference_builder``); IMPORT has no producer
+# yet and is expected to over-merge (coarse, dense, file-level) when one does.
 DEFAULT_REFERENCE_KINDS: tuple[EdgeKind, ...] = (EdgeKind.CONTAINS, EdgeKind.INHERITS)
+# The reference kinds that, like a call, make one component depend on another.
+RELATION_REFERENCE_KINDS: tuple[EdgeKind, ...] = (EdgeKind.INHERITS, EdgeKind.TYPEREF)
+# ``ClusterConnectionEdge.kind`` of a plain call; reference edges carry their ``EdgeKind`` value.
+CALL_EDGE_KIND = "call"
```
Svilen-Stefanov: I find it strange that the CALL_EDGE_KIND is not part of EdgeKind and is hardcoded

static_analyzer/cluster_relations.py
```diff
@@ -73,6 +73,15 @@ def build_component_relations(
...
                 edge_pairs[key].append(RelationEdge.from_edge(edge))
+        for ref in cfg.reference_edges:
+            if ref.kind not in RELATION_REFERENCE_KINDS:
+                continue
+            src_comp = node_to_component.get(ref.src)
+            dst_comp = node_to_component.get(ref.dst)
+            source, target = cfg.nodes.get(ref.src), cfg.nodes.get(ref.dst)
+            if not (src_comp and dst_comp) or src_comp == dst_comp or source is None or target is None:
+                continue
+            edge_pairs[(src_comp, dst_comp)].append(RelationEdge.from_reference(source, target, ref.kind))
```
Svilen-Stefanov: For typerefs, do we store the actual sites of the reference, it seems like we don't and we just add the edge type?

Svilen-Stefanov: For typerefs, do we store the actual sites of the reference, it seems like we don't and we just add the edge type?

static_analyzer/engine/models.py
```diff
@@ -65,6 +65,50 @@ def lsp_column(self) -> int:
...
+    first_line: int
+    last_line: int
+    # The name an alias binds (``Item`` in ``using Item = A.B.Item``), else empty.
+    alias: str = ""
+    # ``using static N.T``: T's members become nameable unqualified, T's own name does not.
+    static: bool = False
+
+
+@dataclass(frozen=True)
+class NamespaceContext:
```
Svilen-Stefanov: This might be a bit confusing even thought it generally makes sense. For now, we use the dir structure as the reference for qualified names, it seems like here we diverge from this and use the namespace instead? 
Should we use one and the same construct everywhere?

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,255 @@
+"""Type-reference edges from source: what names a type without calling it.
```
Svilen-Stefanov, line 1: Doesn't this file have a lot of common with the call sites builder? If so and if we want to keep them separate, should they have a common parent and reuse common code or an interface?

Svilen-Stefanov (commented):
Adding some initial comments

static_analyzer/engine/models.py
Svilen-Stefanov: Agreed. `UsingDirective`, `NamespaceContext` and the TS-only `ImportBinding` are gone. d9d531d replaces them with one `ImportDirective` (the container a directive names, the one member it binds if any, the local alias, the lines it governs, and whether it spans the compilation) and one `NameScope` per file, read by `SourceInspector.find_name_scope` for C# and TS/JS alike. `TypeIndex` resolves against containers for both languages, so the only language-specific code left is the two tree-sitter readers.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/models.py
Svilen-Stefanov: It is a path, not a name: renamed to `file_path` in d9d531d. It stays a `str` because it is matched against `Node.file_path`, which is a `str` throughout the graph; a `Path` here would only add a conversion at every comparison.

---
_Generated by [Claude Code](https://claude.ai/code)_

agents/agent_responses.py
Svilen-Stefanov: Removed in 1505401. `EdgeKind` in edge.py now carries `CALL` and a `relation_label` per kind, which is the single vocabulary; `REFERENCE_EDGE_DESCRIPTIONS` and the three label constants in `constants.py` are gone. `RelationEdge` carries the `kind` of the static edge behind it, so `static_relation_label` reads kinds instead of parsing description text.

---
_Generated by [Claude Code](https://claude.ai/code)_

Svilen-Stefanov: Same fix as the thread above: 1505401 moves the vocabulary onto `EdgeKind.relation_label` in edge.py and gives `RelationEdge` a `kind`, so nothing declares or parses the label text elsewhere.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/cfg/edge.py
Svilen-Stefanov: Fixed in 1505401: `EdgeKind.CALL` exists, `CALL_EDGE_KIND` is gone, `ClusterConnectionEdge.kind` is an `EdgeKind`, and `ReferenceEdge` rejects `CALL` so a call can only ever be an `Edge`.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/cluster_relations.py
Svilen-Stefanov: They were not; only the pair survived. In 1505401 `ReferenceEdge.sites` records every position that produced the edge (d9d531d fills it from the type-reference pass), and it flows through `ClusterConnectionEdge.call_sites` into `RelationEdge.call_sites`, the same way a call's sites do. Sites never take part in edge identity.

---
_Generated by [Claude Code](https://claude.ai/code)_

Svilen-Stefanov: Same as the thread above: as of 1505401 and d9d531d a type-reference edge carries the positions where the type is named, end to end into the relation edge.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/models.py
Svilen-Stefanov: The qualified names stay path-based in every language; nothing here renames a node. The namespace is only used to match what the source wrote: C# writes namespaces (`using`, `Volo.Abp.Foo`) and never directories, while TS/JS writes module paths, which are the directory structure. d9d531d makes that one construct explicit: `TypeIndex.container_of(node)` is the path source writes to reach a declaration, spelled as namespace plus enclosing types in C# and as the module in TS/JS, and both resolvers match written names against it.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
Svilen-Stefanov: They share less than it looks. The call-site builder runs per engine over `SymbolInfo` and asks the language server to resolve each site; this pass runs once over the merged graph's `Node`s and resolves from source alone, because a name has to resolve across sub-projects the server never sees together. A common parent would have to abstract over both models and both resolvers for a loop of five lines. What is genuinely shared already is: `SourceInspector` finds the sites for both, `is_self_or_container_edge` gates both, and after 1505401 both produce edges whose sites use the same `CallSiteLocation` shape. If you would still like a `SiteResolver` protocol over the two, I can extract one, but it would be an interface whose two implementations differ in every method.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/source_inspector.py
```diff
@@ -570,6 +640,115 @@ def _node_is_call_target(self, target: TreeSitterNode) -> bool:
...
+
+        usings: list[str] = []
+        namespaces: list[tuple[str, int, int]] = []
+
+        def visit(scope: TreeSitterNode, prefix: str) -> None:
+            for child in scope.named_children:
+                if child.type in _CSHARP_USING_NODE_TYPES:
+                    target = self._using_target(child)
+                    if target is not None:
+                        usings.append(text(target))
```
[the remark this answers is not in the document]
Svilen-Stefanov: Addressed in 5efe373 (the alias, the `static` kind and the lexical scope are recorded) and superseded by d9d531d, where the directive is the general `ImportDirective` and lookup runs scope by scope.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/source_inspector.py
```diff
@@ -759,6 +948,85 @@ def _last_named_child_of_type(self, node: TreeSitterNode, node_types: frozenset[
...
+        in_list = parent.type in _CSHARP_TYPE_LIST_NODE_TYPES
+        as_operand = parent.type in _CSHARP_TYPE_OPERAND_NODE_TYPES and field == "right"
+        receiver = (
+            node.type == "identifier" and parent.type in _CSHARP_MEMBER_ACCESS_NODE_TYPES and field == "expression"
+        )
+        if not (in_slot or in_list or as_operand or receiver):
+            return None
+        if node.type == "qualified_name":
+            name = node.child_by_field_name("name")
+            return (name, node.child_by_field_name("qualifier")) if name is not None else None
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 5efe373: `split_type_name` is bracket-aware, `global::` is stripped, and tests pin `A.Box<T>` and `global::A.Foo`.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,212 @@
...
+        if binding is None:
+            candidates = self._by_name.get(site.name, [])
+            same_file = [node for node in candidates if node.file_path == site.file]
+            if len(same_file) == 1:
+                return same_file[0]
+            return candidates[0] if len(candidates) == 1 else None
+        if not binding.source.startswith("."):
+            return None
+        wanted = site.name if binding.imported_name in (_NAMESPACE_IMPORT, "default") else binding.imported_name
+        module = os.path.normpath(os.path.join(os.path.dirname(site.file), binding.source))
```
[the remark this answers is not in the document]
Svilen-Stefanov: The extension half is fixed in 5efe373: `./svc.js` resolves to `svc.ts` and `./lib/index.js` to the barrel. A `paths` alias such as `@app/models` deliberately stays external: resolving it needs the project's tsconfig and module rules, which this layer does not read, and a wrong guess would draw an edge into a package that only shares a name. That is a documented limit of the layer rather than a bug.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/source_inspector.py
```diff
@@ -671,16 +851,25 @@ def _node_is_return_value(target: TreeSitterNode) -> bool:
...
     def _node_is_call_argument(self, target: TreeSitterNode) -> bool:
+        if self._is_receiver(target):
+            return False
         node = target
         while node.parent is not None:
             parent = node.parent
             if parent.type in _CALLABLE_USAGE_ANCESTORS and self._parent_is_call_like(parent):
                 return True
-            if self._call_target_node(parent) == target:
+            if parent.type in _FUNCTION_BOUNDARY_NODE_TYPES or self._call_target_node(parent) == target:
```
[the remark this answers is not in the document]
Svilen-Stefanov: Bumped to v8 in 5e3b1f1, with a changelog entry naming the call-edge predicate as the reason.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/source_inspector.py
```diff
@@ -570,6 +683,127 @@ def _node_is_call_target(self, target: TreeSitterNode) -> bool:
             node = parent
         return False
 
+    def find_type_reference_sites(self, file_path: Path) -> list[TypeReferenceSite]:
+        """Positions where the file names a type without calling it.
+
+        Why: a parameter type, a base class, a generic argument or a ``typeof`` operand never
+        reaches the call graph, and in module-oriented code that is where the wiring is written.
+        """
+        parsed = self._parse(file_path)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 7474129: the suffix decides before the file is read, and a test pins that a language without a selector is never parsed.

---
_Generated by [Claude Code](https://claude.ai/code)_

agents/relation_edges.py
```diff
@@ -1,11 +1,29 @@
...
+
+
+def static_relation_label(edges: Sequence[RelationEdge]) -> str:
+    """The label a relation gets from its static edges alone: a single call outranks any reference."""
+    descriptions = {edge.description for edge in edges}
+    if not descriptions or descriptions - set(REFERENCE_EDGE_DESCRIPTIONS.values()):
+        return DEFAULT_STATIC_RELATION_LABEL
+    if descriptions == {REFERENCE_EDGE_DESCRIPTIONS[EdgeKind.INHERITS]}:
+        return INHERITANCE_RELATION_LABEL
+    return TYPE_REFERENCE_RELATION_LABEL
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed. `complete_type_references` (d9d531d) skips a TYPEREF pair the graph already records as INHERITS, so a base list no longer doubles the inheritance, and `static_relation_label` (1505401) reads edge kinds: a pair joined by inheritance alone is "inherits from", mixed references are "uses", and a call still outranks both.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,255 @@
...
+            directive
+            for directive in self._namespace_context(file_path).usings
+            if directive.first_line <= line <= directive.last_line
+        ]
+
+    def _visible_namespaces(self, file_path: str, line: int) -> set[str]:
+        visible = {""}
+        # Only a plain ``using N;`` makes a namespace nameable: an alias binds one name, and
+        # ``using static N.T`` brings in T's members rather than T itself.
+        visible.update(d.target for d in self._usings_at(file_path, line) if not d.alias and not d.static)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d: a `global using` is recorded as compilation-wide, the compilation is the nearest ancestor directory holding a `*.csproj`, and every file's scope is read before any site is resolved so a directive in `GlobalUsings.cs` governs its siblings. Global usings sit at the compilation-unit level of the lookup, below a namespace's own usings, which is what C# does; eShop's Ordering.API needs exactly that order.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/source_inspector.py
```diff
@@ -570,6 +683,127 @@ def _node_is_call_target(self, target: TreeSitterNode) -> bool:
...
+                    if target is not None:
+                        alias = child.child_by_field_name("name")
+                        usings.append(
+                            UsingDirective(
+                                target=_without_alias_qualifier(text(target)),
+                                first_line=first_line,
+                                last_line=last_line,
+                                alias=text(alias) if alias is not None else "",
+                                # ``static`` is an anonymous token, so it is not a named child.
+                                static=any(part.type == "static" for part in child.children),
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d, with same-compilation and other-project tests: `global using` is compilation-wide, the compilation is the nearest `*.csproj` above the file, and the directive applies at the compilation-unit level of the scope-ordered lookup.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def _by_qualifier(self, written: str) -> list[Node]:
+        """Indexed types whose declared namespace matches the dotted prefix written before the name."""
+        qualifier, name = split_type_name(written)
+        candidates = self._by_name.get(name, [])
+        if not qualifier:
+            return candidates
+        # ``endswith``, not equality: a name written from inside an ancestor namespace is only
+        # partially qualified (``Domain.Order`` in ``namespace Lib`` against ``Lib.Domain.Order``).
+        return [node for node in candidates if self._namespace_of(node).endswith(qualifier)]
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d: a qualifier matches the whole container or a dot-bounded suffix of it, and a test pins that `Domain` no longer matches `FooDomain`.

---
_Generated by [Claude Code](https://claude.ai/code)_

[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d: a candidate's container is its namespace plus the types enclosing it, so `Outer.Inner` and `N.Outer.Inner` resolve, and a nested type is nameable unqualified inside its enclosing type and through `using static`.

---
_Generated by [Claude Code](https://claude.ai/code)_

diagram_analysis/scope_assembly.py
```diff
@@ -224,7 +225,7 @@ def apply_semantics(
             if pair not in seen_pairs
             and pair not in preserved_pairs
             and relation.relation.strip()
-            and relation.relation != DEFAULT_STATIC_RELATION_LABEL
+            and relation.relation not in STATIC_RELATION_LABELS
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 1505401: `Relation.default_label` records that the wording was computed from the edges, it is persisted with the relation, and the carry filter checks it instead of the text, so an authored "uses" is carried with its evidence. A file written before the field existed is judged by its wording once, at load.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d: a using's target is resolved from the namespace it is written in outward, against the containers the index knows, so inside `namespace Root` a `using Models;` opens `Root.Models` when that exists.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,212 @@
...
+        binding = bindings.get(site.qualifier or site.name)
+        if binding is None:
+            candidates = self._by_name.get(site.name, [])
+            same_file = [node for node in candidates if node.file_path == site.file]
+            if len(same_file) == 1:
+                return same_file[0]
+            return candidates[0] if len(candidates) == 1 else None
+        if not binding.source.startswith("."):
+            return None
+        wanted = site.name if binding.imported_name in (_NAMESPACE_IMPORT, "default") else binding.imported_name
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in d9d531d: every module's `export default` declaration is read along with its scope, and a default binding resolves to that name rather than to the importer's local alias, so `import Client from './svc'` finds `export default class Service`.

---
_Generated by [Claude Code](https://claude.ai/code)_

agents/llm_renderers/scope.py
```diff
@@ -146,12 +146,15 @@ def _known_connections(scope: ClusterScopeResult, repo_dir: Path) -> list[dict[s
             seen.add(key)
             distinct.append(edge)
         distinct.sort(key=lambda edge: (edge.source_qualified_name, edge.target_qualified_name))
+        calls = [edge for edge in distinct if edge.kind == CALL_EDGE_KIND]
+        references = [edge for edge in distinct if edge.kind != CALL_EDGE_KIND]
```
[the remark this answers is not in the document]
Svilen-Stefanov: Kept as is. The pair list counts distinct symbol pairs, and a pair that is both called and named as a type is one connection to the model, not two; adding `kind` to the key would list the same source and target example twice with identical text. The `calls` count is exact either way.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/engine/type_reference_builder.py
```diff
@@ -0,0 +1,212 @@
...
+                continue
+            pairs.setdefault((container.fully_qualified_name, target.fully_qualified_name), None)
+    return list(pairs)
+
+
+def complete_type_references(
+    graph: CallGraph, source_files: Iterable[Path], inspector: SourceInspector
+) -> TypeReferenceStats:
+    """Replace the graph's TYPEREF edges with what its source files name today."""
+    graph.reference_edges = [ref for ref in graph.reference_edges if ref.kind is not EdgeKind.TYPEREF]
```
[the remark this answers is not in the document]
Svilen-Stefanov: The premise about preexisting edges does not hold: the TYPEREF layer is not carried between runs. It is wiped and rebuilt from the language's full source list on every run, warm or cold, so a partial parse gives the same graph either way and there is no cached edge to protect. A tree with ERROR nodes is walked as a best effort, the same contract the call-site pass has; refusing the run on any grammar gap would fail most real C# repositories. What was wrong is that an unreadable file was silent: as of d9d531d `find_type_reference_sites` returns `None` for a file it cannot read, the pass counts those files in its stats, and the analyzer logs them.

---
_Generated by [Claude Code](https://claude.ai/code)_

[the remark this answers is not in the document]
Svilen-Stefanov: There is no cached TYPEREF edge to protect: the layer is wiped and rebuilt from the language's full source list on every run, warm or cold, so an unreadable file gives the same graph either way. Keeping a previous run's edges for a file that cannot be read would be the plausible-but-wrong result AGENTS.md rules out. What was wrong is that it happened silently: as of d9d531d `find_type_reference_sites` returns `None` for a file it cannot read, the pass counts those files in its stats, and the analyzer logs them.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #573, opened 2026-09-08
feat(cfg): type references for Java and Python

Svilen-Stefanov:
Stacked on [#570](https://github.com/CodeBoarding/CodeBoarding/pull/570) — **base is `feat/relation-edge-kind`**, which is itself stacked on [#568](https://github.com/CodeBoarding/CodeBoarding/pull/568).

## Why

#568 gave the type-reference layer to C# and TypeScript and returned nothing for the other six wired languages. Nothing about the defect was C#-specific — `_process_references_for_position` drops a reference to a class-like symbol that is not an invocation whatever the language, so `void m(Widget w)` and `def m(w: Widget)` were exactly as invisible as `[DependsOn(typeof(X))]`. A Java or Python repository kept the floating components the C# work exists to remove, and since `AFFINE_REFERENCE_KINDS` feeds TYPEREF into clustering, those repositories were also grouped worse — with nothing in the UI to explain why.

The original mandate for this work said all eight languages. This is two of the six that were missing.

## There are only two resolution rules

Adding a language needs a reader and a rule for what a bare name means. It turns out there are two families, and both already existed:

- **A declared scope plus imports** — C#, **Java**. Java reuses `_resolve_by_namespace` unchanged: a single-type import decides a name exactly as a C# alias does, a wildcard import offers a package as a plain `using` does, and a static import offers members rather than the type itself. The only new code is the reader that turns `package`/`import` into the `UsingDirective` records C# already produces.
- **A name bound to a module** — TypeScript, **Python**. Python reuses the import-binding shape, with a dotted module path in place of a relative specifier: relative imports anchor against the importing file, absolute ones match as a path suffix, and a package resolves through its `__init__`.

`_resolve_csharp` is renamed `_resolve_by_namespace` because it is no longer C#-specific, and the suffix→selector chain became a lookup so a language is one row rather than a branch.

## The readers are small because the grammars are explicit

**Java:** every `type_identifier` is a type position — a declaration names itself with a plain `identifier` — so the selector only excludes the segments of a dotted name and the type of a creation, which is already a constructor call. Annotations are included on purpose: in a Spring or Mockito codebase they *are* the wiring, and the ones that matter resolve in-repo while `@Override` and `@Test` resolve to nothing and cost nothing.

**Python:** every annotation is wrapped in a `type` node and base classes hang off the class's `superclasses` list. That is the whole rule; nested generics fall out of it, since `Optional[Deep]` puts `Deep` under another `type`.

## What these two languages actually discover

Real edges from the integration fixtures, absent before this branch. See [#568](https://github.com/CodeBoarding/CodeBoarding/pull/568) for the full framing; the short version is that a call is a runtime dependency and a type reference is a compile-time one, which is why the label is `uses`.

**Java** — 23 references on the fixture project:

| position | real example | edge |
|---|---|---|
| generic argument | `List<Animal> animals = new ArrayList<>();` | `main` uses `Animal` |
| `extends` | `public class Dog extends Animal` | `Dog` uses `Animal` |
| `implements` | `public class Duck extends Animal implements Swimmable` | `Duck` uses `Swimmable` |
| return type | `public static Animal create(String type, String name)` | `create` uses `Animal` |
| parameter type | `public static String getAnimalCity(Animal animal)` | `getAnimalCity` uses `Animal` |
| local type | `Animal factoryAnimal = AnimalFactory.create(...)` | `main` uses `Animal` |
| `instanceof` | `if (!(o instanceof Container c)) return false;` | `equals` uses `Container` |

**Annotations are included on purpose**, and they are the entry most worth arguing about. `@Override` and `@Test` resolve to nothing and cost nothing, because the annotation type is not in the repository. An in-repo `@Component` or `@Mock` does resolve, and in a Spring or Mockito codebase that annotation *is* the wiring — the same argument as C#'s `[DependsOn(typeof(X))]`. If you would rather not have them, it is one entry in `_JAVA_ANNOTATION_NODE_TYPES`.

**Python** — 16 references:

| position | real example | edge |
|---|---|---|
| base class | `class Cat(Animal):` | `Cat` uses `Animal` |
| parameter annotation | `def _handle_cat(animal: Animal) -> str:` | `_handle_cat` uses `Animal` |
| return annotation | `def create_dog(name: str) -> Dog:` | `create_dog` uses `Dog` |
| generic argument | `def animal_names(animals: list[Animal]) -> list[str]:` | `animal_names` uses `Animal` |

Python's are the cleanest of the six languages — an annotation is *only* ever a type, so there is no incidental category, and its import table means a name is either bound in this file or is not ours. That is why it resolves 1,407 of 1,870 sites where C# resolves 18%.

The honest weakness is coverage rather than precision: an unannotated codebase yields nothing. django has 894 analysed files and only 1,870 type positions in them. Nothing is wrong — there is simply less to read.

## Measured, over the files a real run analyses

| language | repo | files | class nodes | type references | pass | resolved / sites |
|---|---|---:|---:|---:|---:|---|
| Java | mockito | 990 | 1,823 | **3,425** | 1.3 s | 8,041 / 25,933 |
| Python | django | 894 | 1,920 | **1,405** | 2.2 s | 1,407 / 1,870 |
| Python | CodeBoarding | 168 | 1,638 | **816** | 0.6 s | 1,058 / 6,813 |

Python resolves a far higher share of its sites than Java or C# because its import table is explicit — a name is either bound or it is not ours. The edges are the ones a reader wants: `XFrameOptionsMiddleware` → `MiddlewareMixin` is a django relationship no call carries.

## Tests

Reader tests per language (every type position Java writes; a created type left to its constructor; annotations, bases and nested generics in Python; package/wildcard/static imports; absolute, relative, aliased and module imports) and resolver tests that discriminate rather than confirm — a single-type import beating the file's own package, a wildcard import reaching another package, a static import *not* making its type nameable, a relative import beating a same-named module elsewhere, and a third-party import refusing to claim a same-named local class.

The two tests that asserted "other languages have no type sites" now use Rust, which stays unsupported.

Full suite 2,151 passed; the 3 `test_cli_parser` failures are pre-existing and fail identically on `main` at `e38c07e3`. mypy and black clean.

## Still missing

Go and PHP follow in the next PR of this stack. Rust is deliberately last: `use` paths over a module tree plus trait impls and macro-generated types are a different order of difficulty, and I would rather land the four cheap ones first.

## Summary by CodeRabbit

- **New Features**
  - Added Java and Python support for detecting type references in source code.
  - Improved handling of Java packages, imports, wildcard imports, and static imports.
  - Added Python import resolution for absolute, relative, aliased, and dotted imports.
  - Improved type resolution across local files, packages, modules, and imported names.
  - Preserved accurate behavior for unsupported languages and non-type expressions.

### DIFF
--- a/static_analyzer/engine/source_inspector.py
+++ b/static_analyzer/engine/source_inspector.py
@@ -205,6 +205,22 @@ _DECORATOR_SEARCH_DEPTH = 12
 _SCRIPT_IMPORT_NODE_TYPES = frozenset({"import_statement"})
 _TYPESCRIPT_SUFFIXES = frozenset({".ts", ".tsx", ".mts", ".cts"})
 _SCRIPT_SUFFIXES = _TYPESCRIPT_SUFFIXES | frozenset(LANGUAGE_EXTENSIONS[Language.JAVASCRIPT])
+_JAVA_SUFFIXES = frozenset(LANGUAGE_EXTENSIONS[Language.JAVA])
+_PYTHON_SUFFIXES = frozenset(LANGUAGE_EXTENSIONS[Language.PYTHON])
+# Java: every ``type_identifier`` is a type position — a declaration names itself with a plain
+# ``identifier`` — so the selector only has to exclude the segments of a dotted name and the
+# type of a creation, which is already a constructor call.
+_JAVA_SCOPED_TYPE_NODE_TYPES = frozenset({"scoped_type_identifier"})
+_JAVA_CREATION_NODE_TYPES = frozenset({"object_creation_expression"})
+_JAVA_ANNOTATION_NODE_TYPES = frozenset({"marker_annotation", "annotation"})
+_JAVA_IMPORT_NODE_TYPES = frozenset({"import_declaration"})
+_JAVA_PACKAGE_NODE_TYPES = frozenset({"package_declaration"})
+# Python: an annotation is always wrapped in a ``type`` node, and a base class is an argument
+# of the class's ``superclasses`` list.
+_PYTHON_TYPE_NODE_TYPES = frozenset({"type", "generic_type"})
+_PYTHON_SUPERCLASS_FIELD = "superclasses"
+_PYTHON_IMPORT_NODE_TYPES = frozenset({"import_statement", "import_from_statement"})
+_PYTHON_RELATIVE_IMPORT_NODE_TYPES = frozenset({"relative_import"})
 
 # Ceiling on retained tree-sitter nodes. Trees are by far the largest thing this
 # class touches — retaining one per file cost 2.2GB on a 5k-file C# repo — and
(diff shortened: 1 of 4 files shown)


ivanmilevtues:
<a href="https://app.codeboarding.org" target="_blank" rel="noopener noreferrer">Open CodeBoarding in a new tab</a>

PR #577, opened 2026-09-08
chore(analytics): tag events with the repository owner

Svilen-Stefanov:
## What

Every event already carries an origin: `source` (which invoker the run came
from) and `internal` (whether that invoker is a person using the product). This
adds a third field to the same block, `org` — the account the analysed
repository belongs to.

## Why

`source` says a run came from the CLI, or CI, or the extension. Nothing says
which deployment it came from, so two runs look alike however unrelated the
repositories behind them are. That makes anything measured per-deployment
impossible to measure: adoption inside one org, whether a reported failure is
widespread or isolated, whether a version regressed for everyone or for one
setup.

## How

Resolved in `_origin()`, the one place that already stamps the origin onto both
`capture` and `capture_exception`:

- `CODEBOARDING_ORG` when an embedding sets it (the VSCode extension resolves
  the owner itself and passes it down);
- otherwise the owner half of `GITHUB_REPOSITORY`, which Actions always sets,
  so CI runs need no change to the action at all.

`generate_analysis(repo_url=...)` sets it from the URL it was pointed at, since
`GITHUB_REPOSITORY` would otherwise name the workflow's own repository rather
than the one being analysed.

Scope, deliberately narrow:

- **Only the owner segment is read**, never the repository's own name.
- Nothing is inferred. A run with neither variable set sends no `org` at all —
  the key is absent rather than empty, so a query can tell "no owner" from
  "owner is blank".
- Lower-cased, because owners are case-insensitive on GitHub and a grouping that
  splits `Acme` from `acme` is a wrong one.
- Merged last, like the rest of the origin, so a payload cannot claim it.

## Documentation

An owner is an organisation name for an org-owned repository and a **username**
for a personal one, so this is a real change in what is collected and the docs
say so rather than working around it.

`TELEMETRY.md` now leads with the full field list instead of a broad promise. It
documents `org` plainly, states that the data is pseudonymous rather than
anonymous, and moves the opt-out near the top. `## What we never collect` keeps
only the claims that remain true: no source code, no file or repository names,
no prompts, no credentials. The sentence "There is no account, email, login, or
IP-based identity" is corrected, since it stops being true here.

## Tests

`tests/test_telemetry_service.py` covers the CI variable, the explicit
override and its precedence, owner-only extraction, absence on a local run,
case normalisation, precedence over a caller-supplied value, and the exception
path. Full suite: 2078 passed (the 3 `test_cli_parser` failures are pre-existing
on `main`).

Companion: CodeBoarding/CodeBoarding-vscode#230, which resolves the owner from
the git remote, passes it down, and adds the opt-out this field makes overdue.

## Summary by CodeRabbit

- **Documentation**
  - Updated telemetry documentation to explain pseudonymous data collection, including account and machine identifiers.
  - Added opt-out instructions for the VS Code telemetry settings.
  - Clarified collected fields, data usage, and information that is not collected.

- **Privacy**
  - Telemetry now reports the repository owner or username as an `org` value instead of a hashed identifier.

- **Bug Fixes**
  - Repository URLs with trailing slashes are now handled correctly, including proper `.git` suffixes and repository name extraction.


github_action.py
```diff
@@ -106,6 +107,10 @@ def generate_analysis(
 ) -> Path:
     """Generate analysis for a GitHub repository URL (GitHub Action entry point)."""
     os.environ.setdefault("CODEBOARDING_SOURCE", "github_action")
+    # Why: telemetry otherwise falls back to GITHUB_REPOSITORY, which names the
+    # workflow's own repository rather than the one this run was pointed at.
+    repo_path = urlparse(repo_url).path if "://" in repo_url else repo_url.partition(":")[2]
+    os.environ["CODEBOARDING_ORG"] = repo_path.strip("/").partition("/")[0]
     repo_root = Path(os.getenv("REPO_ROOT", "repos"))
     repo_name = clone_repository(repo_url, repo_root)
```
ivanmilevtues, line 110: I think that the repo_org should probably be part of the clone_repository function as it already gets the name so it should also get the org + i suspect that the github utils we use have an automatic org extractor.

ivanmilevtues (approved):
Looks good, the whole org extraction seems quite defensive but I couldn't come up in how to not have _org() which does a or b or c or .. and then again have an if _org()

My comment is that you might be able to reuse something in our repo_utils in the clone_repository to not have to parse things as it already extracts repo_name, it could also get the repo_org.

Svilen-Stefanov:
Thanks — both addressed.

**Reusing `repo_utils`.** You were right that the URL was being parsed twice.
There is now a `get_repo_org` next to `get_repo_name`, and `github_action` calls
it instead of doing its own `urlparse`.

I did not fold it into `clone_repository`'s return value, though. That would
change a signature three production callers and eight test mocks depend on, and
would hand an owner to callers that never asked for one — `health_main` already
pairs `clone_repository` with a separate `get_repo_name`, so a sibling function
matches how these are used today. Happy to change it if you'd rather have the
tuple.

On the "automatic org extractor" — nothing in the repo does it today. GitPython
gives us remote URLs, not parsed owners, so it would still come down to the same
`split("/")`. Keeping it as two lines next to `get_repo_name`, which does the
same thing for the name, seemed better than a dependency.

**One thing that reuse turned up.** `sanitize_repo_url` appends `.git` without
stripping a trailing slash, so `https://github.com/user/my-repo/` became
`https://github.com/user/my-repo/.git`. That does not clone, and
`get_repo_name` read `.git` back as the repository name:

```python
>>> get_repo_name("https://github.com/user/my-repo/")
'.git'
```

Pre-existing, unrelated to telemetry, and a one-line fix, so I took it here
rather than leaving it — with a test for `sanitize_repo_url`, `get_repo_name`
and `get_repo_org` each. Shout if you'd prefer it split out.

**The `if _org()`.** Fair — it is now `if org := _org():`, so the temporary and
the branch are one line. The `a or b` inside stays: the two sources genuinely
have a precedence (an explicit `CODEBOARDING_ORG` from an embedding beats CI's
`GITHUB_REPOSITORY`, which is what makes `generate_analysis(repo_url=...)`
attribute to the repository it was pointed at rather than the workflow's own).

Full suite: 2083 passed, same 3 pre-existing `test_cli_parser` failures as
`main`.

Svilen-Stefanov:
Bot review round — what I took and what I didn't.

**Fixed**

- **A payload could invent an `org`.** Merging the origin last stopped a caller
  overwriting `source`/`internal`, but `org` is *absent* when no owner resolves,
  so a caller's value survived into it. Reserved keys now come off the payload
  first, on both the capture and exception paths, with tests for each. This was
  the one real bug in the round.
- **`README.md` and `telemetry/__init__.py`** still said anonymous / no personal
  information. The README one sits next to the opt-out, so it mattered most.
- **`TELEMETRY.md` claimed to list every field** while omitting
  `lsp_analysis_result` — the highest-volume event we emit — and the two MCP
  events. Documented them rather than softening the claim.
- **`distinct_id` terminology** — the Identity section still called it anonymous
  and unmappable while the rest of the page said pseudonymous. One word now.
- **Comment narration** in `sanitize_repo_url` and a test comment, plus two test
  docstrings that restated their own names and two trimmed to the reason.
- **The no-code bullet** now carries the stack-trace caveat inline.

**Not doing, with reasons**

- *Scope the fallback to the repository being analysed.* Deliberate, and the
  reason changed since the first round: `org` answers "which deployment is using
  CodeBoarding", which is whoever runs the workflow. When one org analyses
  another's repository the customer is still the one running it, so overriding
  with the analysed repo's owner makes attribution worse. The override and its
  helper were removed for exactly this.
- *Move the private helpers below the public API.* `_telemetry_disabled` and
  `_origin` were already above `ProductTelemetry` before this PR, and the new
  helpers sit with the `INTERNAL_SOURCES` / `ORIGIN_KEYS` constants they read.
  Moving only the new ones would split that group and leave the module in two
  minds; happy to move the whole block in a separate pass if you'd rather.
- *Include the GitHub host in the owner identity.* Real in principle — a
  GitHub Enterprise `acme` and a github.com `acme` would collide. We have no
  Enterprise users, and adding a host would double the field's width for a
  collision that has never happened. Worth revisiting if that changes.

Full suite: 2080 passed, same 3 pre-existing `test_cli_parser` failures as
`main`. mypy clean.

PR #581, opened 2026-09-09
fix(csharp): let csharp-ls find restore assets in UseArtifactsOutput solutions

Svilen-Stefanov:
## Why

eShop's `AddSubscription<T, TH>(this IEventBusBuilder)` is called 18 times across 6 functions and had **zero** incoming call edges. Logged at the LSP boundary: csharp-ls answered every `textDocument/definition` with `[]`, no error. Not a generic/extension-method problem: the design-time build had no `project.assets.json`, so every package type and every *transitive* project reference was unresolved (`CS0012: IEventBusBuilder is defined in an assembly that is not referenced`), and the call could not bind.

Two independent causes, both in `CSharpAdapter`:

1. **The injected Directory.Build.props moved the artifacts root.** The SDK derives `ArtifactsPath` from `_DirectoryBuildPropsBasePath` (`Microsoft.NET.DefaultArtifactsPath.props`), which `Microsoft.Common.props` only computes when `DirectoryBuildPropsPath` was not supplied from outside. Ours is, so a `UseArtifactsOutput` repo evaluated `ProjectAssetsFile` as `src/WebApp/artifacts/obj/project.assets.json` while `dotnet restore` wrote `artifacts/obj/WebApp/`. The injected props now set the base path the way the SDK would have.
2. **One project killed the solution restore.** `dotnet restore eShop.slnx` fails on the MAUI test project (NETSDK1147, `maui-tizen` workload) and writes assets for nobody. On failure, each project is now restored on its own (17 of 19 succeed on eShop, ~15s).

## Measured on eShop (zero-LLM, static only)

| | before | after |
|---|---|---|
| called extension methods with no caller | 21 / 48 | 3 / 48 |
| call sites lost | 46 / 135 | 11 / 135 |
| call edges | 1376 | 1511 |
| INHERITS edges | 82 | 131 |
| error diagnostics | 2217 | 1077 (MAUI projects only) |

The residual 3 are `src/Shared/*.cs` files linked into several projects; csharp-ls serves no document symbols for those. Separate defect, next PR.

abp is not this class: its restore works and csharp-ls resolves cross-solution calls; its callerless methods are dropped later because each solution root is its own engine. Also next.

## Tests

- Unit: per-project fallback + no fallback after success + the base-path line in the props (`tests/static_analyzer/test_csharp_adapter.py`).
- Integration (paired PR https://github.com/CodeBoarding/CodeBoarding-tests/pull/38, same branch name): fixture `csharp_artifacts_output_project` (UseArtifactsOutput, App → Transport → Contracts, plus a project whose restore fails deterministically) and three tests that fail on main and pass here.

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering existing C# and F# projects listed in `.sln` and `.slnx` solution files, including projects nested in solution folders.

* **Bug Fixes**
  * Improved project restoration when a solution-level restore fails, allowing other projects to remain analyzable.
  * Preserved artifact and intermediate output locations during restoration for artifact-based builds.
  * Avoided unnecessary per-project restoration when the solution restore succeeds.
  * Direct project restore failures now stop without triggering additional retries.

## After review

The review comments pointed at two design gaps rather than local slips, so the fix changed shape:

- **Restore and csharp-ls ran under different MSBuild environments.** The base-path prop only papered over one consequence of that (and a user-exported `DirectoryBuildPropsPath` would have broken it again). Restore now runs under the exact environment csharp-ls gets (`get_lsp_env`), so the two cannot disagree on a path by construction; the base-path prop stays so both also match a plain `dotnet build`.
- **The fallback was unscoped.** It ran after any failed restore, including a plain `.csproj`, and re-restored every project under the root. It now runs only after a failed *solution* restore and restores that solution's own members, read from the `.sln`/`.slnx` (`static_analyzer/dotnet_solution.py`, also used by #582 for engine ownership). A failed project restore is not retried.
- `_restore` moved below the public API; the props docstring is a two-line Why.

Re-measured: eShop still resolves all 18 `AddSubscription` sites; assets land in `artifacts/obj/<Project>/` for `src/` and `obj/` for `tests/` (its own `Directory.Build.props`), matching the plain layout.

### DIFF
--- a/static_analyzer/dotnet_solution.py
+++ b/static_analyzer/dotnet_solution.py
@@ -1,0 +1,29 @@
+"""The projects a .NET solution file lists."""
+
+from __future__ import annotations
+
+import re
+import xml.etree.ElementTree as ElementTree
+from pathlib import Path
+
+PROJECT_SUFFIXES = (".csproj", ".fsproj")
+# ``Project("{type guid}") = "Name", "relative\path.csproj", "{project guid}"``
+_SLN_PROJECT_LINE = re.compile(r'^Project\("\{[^}]*\}"\)\s*=\s*"[^"]*",\s*"([^"]+)"', re.MULTILINE)
+
+
+def solution_projects(solution: Path) -> list[Path]:
+    """The project files a ``.sln`` or ``.slnx`` lists, absolute, in listing order.
+
+    Solution folders and files that do not exist are left out; a ``.sln`` names
+    its solution folders with the same ``Project(...)`` syntax as its projects.
+    """
+    if solution.suffix.lower() == ".slnx":
+        listed = [element.get("Path") or "" for element in ElementTree.parse(solution).iter("Project")]
+    else:
+        listed = _SLN_PROJECT_LINE.findall(solution.read_text(encoding="utf-8-sig", errors="replace"))
+    projects: list[Path] = []
+    for raw in listed:
+        path = (solution.parent / raw.replace("\\", "/")).resolve()
+        if path.suffix.lower() in PROJECT_SUFFIXES and path.is_file() and path not in projects:
+            projects.append(path)
+    return projects
--- a/static_analyzer/engine/adapters/csharp_adapter.py
+++ b/static_analyzer/engine/adapters/csharp_adapter.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from repo_utils.ignore import RepoIgnoreManager
 from static_analyzer.config import Language, NodeType
 from static_analyzer.dotnet_sdk import DotnetSdkError, resolve_dotnet_sdk, system_dotnet_env
+from static_analyzer.dotnet_solution import solution_projects
 from static_analyzer.engine.language_adapter import LanguageAdapter
 from static_analyzer.engine.lsp_client import LSPClient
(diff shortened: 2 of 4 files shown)


static_analyzer/dotnet_solution.py
```diff
@@ -0,0 +1,29 @@
...
+
+from __future__ import annotations
+
+import re
+import xml.etree.ElementTree as ElementTree
+from pathlib import Path
+
+PROJECT_SUFFIXES = (".csproj", ".fsproj")
+# ``Project("{type guid}") = "Name", "relative\path.csproj", "{project guid}"``
+_SLN_PROJECT_LINE = re.compile(r'^Project\("\{[^}]*\}"\)\s*=\s*"[^"]*",\s*"([^"]+)"', re.MULTILINE)
```
ivanmilevtues, line 11: That is a regex :D

ivanmilevtues (approved):
I think this is good to go. So effectively we are falling back to try to build the subprojects (build meaining doing the restore thing). I asked my agent and it said it is expensive to always do it bottom up. But I'd be curious to run an experiment if we alwasy do a bottom up if something will change.

You can merge it for now I think it is all g :)

PR #582, opened 2026-09-09
fix(csharp): name the members of a file compiled into several projects

Svilen-Stefanov:
Stacked on #581. Paired tests PR: https://github.com/CodeBoarding/CodeBoarding-tests/pull/39 (same branch name).

## Why

Second loss class from the same investigation as #581. csharp-ls answers **nothing** for a path that maps to more than one Roslyn document (`WorkspaceFolder.fs`: `| [ d ] -> Some d | _ -> None`, unchanged through 0.27 and master's unreleased fix, which only handles multi-TFM flavors and explicitly leaves linked files ambiguous). A source file linked into several projects is exactly that: Polly's `Guard.cs`, `ObjectPool.cs`, `ExceptionUtilities.cs`, `ResilienceValidationContext.cs` (Polly.Core's own files, also `<Compile Include>`d into Polly.Extensions and Polly.RateLimiting), eShop's `src/Shared/*.cs`.

Two effects, both logged at the LSP boundary:
- `documentSymbol` returns `[]` for the file, so every `textDocument/definition` that lands in it (the server still resolves *into* it fine) finds no symbol and the call is dropped. Polly: 196 call sites into four files. eShop: 11.
- Our `didOpen` of such a file makes csharp-ls add **one more** copy to the project whose directory holds it, after which every call into it is ambiguous. That is the source of Polly's 161 `CS0121: The call is ambiguous between 'Polly.Utils.Guard.NotNull<T>(T, string)' and 'Polly.Utils.Guard.NotNull<T>(T, string)'`.

## What changes

- `CallGraphBuilder._discover_symbols`: for a server that owns the workspace, symbols are queried before any file is opened, and only the files it answered for are opened afterwards.
- `LanguageAdapter.read_document_symbols` (default none); `CSharpAdapter` returns `SourceInspector.find_document_symbols`, a tree-sitter reading of the file shaped like csharp-ls's own `DocumentSymbol` tree (File → Namespace → types → members, names spelled the server's way: `NotNull<T>(T value, string argumentName = "")`, `this[int i]`, `operator +(Guard a, Guard b)`, no attributes, no `pb::` alias qualifiers). Checked against 213 symbols csharp-ls itself reported on eShop files: same position and spelling for all but nullable-reference `?` and generated protobuf properties.

Calls *from inside* such a file still get no answer from the server (no document to bind in); that half is an upstream csharp-ls change.

## Measured (zero-LLM, static only)

| | Polly before | after | eShop before | after |
|---|---|---|---|---|
| call edges | 5307 | 5544 | 1511 | 1523 |
| call sites resolved to an edge | 3477 | 3674 | 1336 | 1342 |
| files read from source | – | 15 | – | 2 |
| CS0121 errors | 161 | 0 | 0 | 0 |

## Second commit: a nested solution's files belong to its own engine

A solution root nested under another (Polly's `samples/`, eShop's `src/ClientApp`) already gets its own engine, but the outer engine walked the same files and asked its server about them. csharp-ls declines a file outside its solution, so those files silently had no symbols in the outer engine before; with the parse-tree fallback they would have had a second, differently named copy next to the nested engine's own (Polly: 73 duplicate symbols for the 20 sample files). `_create_engine_configs` now hands the outer engine its files minus every nested root.

## Tests

- Unit: `find_document_symbols` shape/spelling on a sample (`tests/static_analyzer/test_source_inspector.py`), discovery order and opened-file set for a workspace-owning server (`tests/static_analyzer/test_call_graph_builder.py`).
- Integration (paired PR): two sources of the hand-written C# edge-case solution are compiled into more than one project; its exact fixture moves from 94 to 102 edges and fails on the base engine with 8 missing edges. `tests/static_analyzer/test_engine_configs.py` covers the nested-root exclusion.

## Summary by CodeRabbit

* **Improvements**
  * Improved C# symbol discovery for namespaces, types, fields, events, methods, and nested declarations.
  * Added source-based fallback symbol detection when language services provide no results.
  * Reduced unnecessary file opening during analysis.
  * Improved source-file attribution for nested C# projects, reducing duplicate analysis.
  * Improved handling of files shared across multiple C# projects.

* **Tests**
  * Added coverage for C# symbol extraction, workspace document handling, and nested project source separation.

## After review

- **"Answered nothing" conflated two opposite states.** A file in two projects and a file in none both get `[]` from csharp-ls, but the first must stay closed (opening it duplicates the document) and the second must be opened (that is how the server serves it). The adapter now asks `workspace/symbol` for a type the file declares: a hit in that file means the loaded solution has it, so it is a linked file and is read from source; no hit means the server does not know it, so it is opened and asked again, exactly as before. A tree without any solution therefore keeps its symbols from the server, and a workspace that failed to load still trips `fail_on_empty_symbols`.
- **Engine ownership followed directory nesting; it now follows solution membership.** A project a nested solution lists, and the outer solution does not, belongs to the nested engine; a project the outer solution lists as well (eShop's `src/ClientApp/ClientApp.csproj`) stays with the outer engine; a root whose files all belong to nested solutions gets no engine, so the "empty means unresolved" sentinel never carries an authoritative empty list.
- Synthesized symbols: explicit interface members keep their interface (`IFoo.M()`), nested block namespaces are named in full (`A.Common`), whitespace inside a default-value literal is kept, `operator checked +` is distinct from `operator +`. `#if` regions are not an issue: the inspector blanks directive lines before parsing, so both branches are seen.
- `_walk` is annotated; cache tag v8 because the same source now yields more symbols and edges.

Re-measured on Polly: 5 files read from source (the genuinely linked ones; 10 others the server did not know are now opened and served), 0 CS0121, 5578 edges.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -49,6 +49,8 @@ class EngineConfig:
     adapter: LanguageAdapter
     project_path: Path
     source_files: list[Path] = field(default_factory=list)
+    # Retain discovery exclusions for incremental edits, including deleted files.
+    excluded_roots: list[Path] = field(default_factory=list)
 
 
 class StaticAnalysisFatalError(RuntimeError):
@@ -231,7 +233,11 @@ def _create_engine_configs(
                                 f"Every C# file under {csharp_config.root} belongs to a nested solution; skipping"
                             )
                             continue
-                        configs.append(EngineConfig(adapter, csharp_config.root, source_files=source_files))
+                        configs.append(
+                            EngineConfig(
+                                adapter, csharp_config.root, source_files=source_files, excluded_roots=elsewhere
+                            )
+                        )
                 else:
                     logger.info("No C# projects detected")
 
@@ -867,6 +873,11 @@ class StaticAnalyzer:
                 analysis = self._run_full_analysis(engine_config, engine_client)
                 self._absorb_into_results(results, language, analysis)
             else:
+                changed_files = {
+                    path
+                    for path in changed_files
+                    if not any(path.is_relative_to(root) for root in engine_config.excluded_roots)
+                }
                 logger.info(f"warmstart {adapter.language}: re-LSPing {len(changed_files)} changed file(s)")
                 cached_lang_dict = carried.get(language) or self._extract_language_dict(cached_results, language)
                 analysis = update_cfg_for_changed_files(
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -12,7 +12,8 @@ from static_analyzer.analysis_cache import StaticAnalysisCache
 from static_analyzer.analysis_result import StaticAnalysisResults
 from static_analyzer.cfg import CallGraph
 from static_analyzer.config import AdapterName, Language
-from static_analyzer.csharp_config_scanner import CSharpConfigScanner
+from static_analyzer.csharp_config_scanner import SOLUTION_GLOBS, CSharpConfigScanner, CSharpProjectConfig
+from static_analyzer.dotnet_solution import solution_projects
 from static_analyzer.engine.adapters import get_adapter
 from static_analyzer.engine.call_graph_builder import CallGraphBuilder
 from static_analyzer.engine.language_adapter import LanguageAdapter
@@ -39,8 +40,10 @@ class EngineConfig:
     """One adapter + project root the engine should run.
 
     ``source_files`` is non-empty only when a scanner has authoritatively
-    resolved file membership (currently TypeScript via ``tsc --showConfig``);
-    otherwise the adapter walks ``project_path`` itself in ``_run_full_analysis``.
+    resolved file membership (TypeScript via ``tsc --showConfig``, C# by what
+    each root's solution lists); otherwise the adapter walks ``project_path``
+    itself in ``_run_full_analysis``. A config whose membership resolved to no
+    file is not created at all.
     """
 
     adapter: LanguageAdapter
@@ -121,6 +124,15 @@ def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list
     return names
 
 
+def _csharp_solution_members(csharp_projects: list[CSharpProjectConfig]) -> dict[Path, list[Path]]:
+    """Per C# root, the project files its solution files list."""
+    members: dict[Path, list[Path]] = {}
+    for config in csharp_projects:
+        solutions = [path for pattern in SOLUTION_GLOBS for path in config.root.glob(pattern)]
+        members[config.root] = [project for solution in solutions for project in solution_projects(solution)]
+    return members
+
+
 def _create_engine_configs(
     programming_languages: list[ProgrammingLanguage],
     repository_path: Path,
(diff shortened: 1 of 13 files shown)


static_analyzer/__init__.py
```diff
@@ -121,6 +124,15 @@ def _adapter_names_for(programming_languages: list[ProgrammingLanguage]) -> list
     return names
 
 
+def _csharp_solution_members(csharp_projects: list[CSharpProjectConfig]) -> dict[Path, list[Path]]:
```
ivanmilevtues: This seems like it shouldn't be here, seems as it has ot be part of the adapter idk.

static_analyzer/__init__.py
```diff
@@ -198,12 +210,28 @@ def _create_engine_configs(
                 csharp_projects = csharp_scanner.scan()
 
                 if csharp_projects:
```
ivanmilevtues: ok i hadn't seen this before :D. Generally speaking probably this should be somehow part of the adapters. But this is larger refactor proly irreleacnt for this.

static_analyzer/__init__.py
```diff
@@ -198,12 +210,28 @@ def _create_engine_configs(
...
+                        # A project a nested root's solution lists, and this root's solutions do not,
+                        # is that engine's; naming it here as well would put it in the graph twice.
+                        elsewhere = [
+                            project.parent
+                            for root, projects in members.items()
+                            if root != csharp_config.root and root.is_relative_to(csharp_config.root)
+                            for project in projects
+                            if project not in members[csharp_config.root]
+                        ]
+                        source_files = adapter.discover_source_files(csharp_config.root, ignore_manager, elsewhere)
```
ivanmilevtues: hmm interesting, i don't care particularly for this comment. But is it possible that file exclusions mess up our analysis. For TS and other langs where we scan build artifacts it becomes exponentially slower, but just a thought.

I think for now nothing much to be done you can check the comment i think it is okay. THe leak is not that big of a deal (maybe)

static_analyzer/__init__.py
ivanmilevtues, line 129: This seems like it shouldn't be here, seems as it has ot be part of the adapter idk.

static_analyzer/__init__.py
ivanmilevtues, line 214: ok i hadn't seen this before :D. Generally speaking probably this should be somehow part of the adapters. But this is larger refactor proly irreleacnt for this.

static_analyzer/__init__.py
ivanmilevtues, line 230: hmm interesting, i don't care particularly for this comment. But is it possible that file exclusions mess up our analysis. For TS and other langs where we scan build artifacts it becomes exponentially slower, but just a thought.

I think for now nothing much to be done you can check the comment i think it is okay. THe leak is not that big of a deal (maybe)

static_analyzer/engine/source_inspector.py
```diff
@@ -647,6 +703,147 @@ def text(node: TreeSitterNode) -> str:
                 modifiers[(type_name, text(member_name_node))] = frozenset(found)
         return modifiers
 
+    def find_document_symbols(self, file_path: Path) -> list[dict]:
```
ivanmilevtues, line 706: interesting that your agent doesn't give a rats ass about agents.md :D

static_analyzer/engine/source_inspector.py
```diff
@@ -647,6 +703,147 @@ def text(node: TreeSitterNode) -> str:
...
+
+        Why: csharp-ls answers nothing for a file that belongs to more than one project
+        (a shared source file linked into several), so calls into it resolve to a
+        position no symbol covers. Names follow the server's display form --
+        ``Add<T>(this IList<T> items, int count = 0)`` -- so the two sources of a
+        symbol table agree on how a member is spelled.
+        """
+        if file_path.suffix.lower() != ".cs":
+            return []
+        parsed = self._parse(file_path)
```
ivanmilevtues, line 717: idk if parsing is super cheap, but we might want to have an eye on this. In order to parse a file only once, we can lazily have a collection of all files w/ inspector and parse them only once when we are calling for the different methods within the inspector. (proly for now not important as performance is not an issue)

static_analyzer/engine/source_inspector.py
ivanmilevtues, line 706: the other methods use a different api they use the nodes not directly the path maybe worth a second look. Not a blocker.

ivanmilevtues (approved):
Looks good no blockers, only concern if we parse more than needed, but that is mergable I think.

My understanding is we are using tree-sitter to find all the palces where the lsp should investigate for calls as we were missing some before. that is wahat muy agent said after some chat :D.

The code for this is quite entangled, i noticed that engine has a lot of specific code which seems more like it should be part of a strategy (problem for another day).

Merge if you want directly, up to you if you want to double check for the lazy parser.

static_analyzer/__init__.py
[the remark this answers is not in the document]
ivanmilevtues: Addressed for ivanmilevtues in https://github.com/CodeBoarding/CodeBoarding/commit/2108d937aab6260701fe1c25ad79f0cd2b7643ce.

C# discovery exclusions now remain on EngineConfig and filter each change set before incremental invalidation. This preserves the same ownership as full discovery while allowing added and deleted paths through their owning engine; filtering only against current source_files would miss deletions.

The paired private regression PR is https://github.com/CodeBoarding/CodeBoarding-tests/pull/42. The real csharp-ls regression failed before this fix and passes after it. Local unit suite: 2,128 passed, 6 skipped, 85.41% coverage. Mypy, formatting, and pre-commit checks also pass.

PR #583, opened 2026-09-09
fix(engine): finish call edges whose definition lies in another solution's project

Svilen-Stefanov:
Stacked on #582 (which is stacked on #581). Paired tests PR: https://github.com/CodeBoarding/CodeBoarding-tests/pull/40 (same branch name).

## Why

Third loss class from the same investigation. A repository with several solutions runs **one engine per solution root**, each with its own symbol table. abp has 30 solution roots; a module's `app.UseAuditing()` resolves at the LSP boundary to `framework/src/Volo.Abp.AspNetCore/.../AbpApplicationBuilderExtensions.cs` (csharp-ls follows the `ProjectReference` outside the module solution), but that file belongs to the framework engine, so `_resolve_definition_to_symbol` finds nothing and the call is dropped. On abp, 357 of the 594 callerless extension-method call sites (by a regex count) were cross-solution.

## What changes

- `_resolve_definitions` records an `ExternalCallSite` (caller, definition file/line/char, call site) whenever the definition lands in a file this engine holds **no** symbols for. Definitions in files it does hold symbols for but cannot match keep their current fate; framework and NuGet definitions carry no file location at all, so nothing is recorded for them.
- The sites travel on `LanguageAnalysisResult` → the converter dict → `_run_full_lsp_pass`, which after absorbing every engine calls `link_external_call_sites` (new module `static_analyzer/external_calls.py`) on the merged graph: exact selection-start match first, then same line, then ±2 lines, callables over classes, longest name; the same self/container and same-line guards as the engine; the class of a callable target gets its edge too, as the engine does. The incremental path already resolves definitions against the merged cached graph and needs nothing.

## Tests

- `tests/static_analyzer/test_external_calls.py`: edge + call site, overload picked by exact position, ±1 line tolerance, unknown file, own class, repeated site.
- `tests/static_analyzer/test_edge_builder.py`: the engine records a definition in a file it never named and builds no edge for it.
- Integration (paired PR): a second solution nested under the hand-written C# edge-case solution, whose project references a project outside it; the exact fixture moves from 102 to 107 edges and fails on the base engine with the 4 cross-solution edges missing.

## Summary by CodeRabbit

* **New Features**
  * Improved static analysis by linking calls across separately analyzed projects and languages.
  * Cross-project call targets are now included in full and incremental analysis results.
  * Unresolved external call sites are preserved with source locations and call-type details.
  * Improved resolution of constructors, overrides, method groups, collection initializers, and iterated calls.

* **Bug Fixes**
  * Prevented duplicate, self-referential, and invalid call edges.

* **Tests**
  * Added coverage for cross-project linking, incremental analysis, position matching, and unresolved targets.

## After review

The linker was a second, weaker implementation of something the warm-start updater already had. Reworked:

- **A recorded site keeps how it was found** (`ExternalCallSite.kind`: call, method-group probe, collection initializer, `foreach`), from both the definition loop and the iterated-type pass. A method-group probe that lands on a value (an enum member has no node; the enum around it must not stand in) adds nothing; a collection initializer reaches `Add`; a `foreach` reaches `GetEnumerator`; a base member reaches its overrides through the merged graph's INHERITS edges; a construction reaches the constructors where the adapter expands them.
- **One resolver for both paths.** The graph-level helpers the incremental orchestrator had (`_definition_nodes`, `_override_nodes`, `_members_named`, ...) move to `static_analyzer/graph_definitions.py` with a per-file index, and both the warm-start updater and the linker resolve through `targets_for`, so they cannot disagree.
- **Linking runs on every path**: the cold pass and both warm-start branches share `_absorb_and_link`; the updater hands back the definitions of changed files it could not resolve, so a caller changed in solution A and a destination added in solution B link after every engine merged, whichever ran first.
- A linked edge across packages is recorded in the package relations (health's cycle check reads them). Cache tag v9.

Not done, on purpose: a delegate-valued field in another engine that takes part in no local edge is not a node of the merged graph and cannot be linked; that is the converter's node filter, not the linker's.

### DIFF
--- a/static_analyzer/__init__.py
+++ b/static_analyzer/__init__.py
@@ -19,9 +19,11 @@ from static_analyzer.engine.call_graph_builder import CallGraphBuilder
 from static_analyzer.engine.language_adapter import LanguageAdapter
 from static_analyzer.engine.lsp_client import LSPClient
 from static_analyzer.engine.lsp_recycler import default_memory_budget, per_engine_memory_budget
+from static_analyzer.engine.models import ExternalCallSite
 from static_analyzer.engine.result_converter import convert_to_codeboarding_format
 from static_analyzer.engine.source_inspector import SourceInspector
 from static_analyzer.engine.utils import uri_to_path
+from static_analyzer.external_calls import link_external_call_sites
 from static_analyzer.incremental_orchestrator import update_cfg_for_changed_files
 from static_analyzer.java_config_scanner import JavaConfigScanner
 from static_analyzer.lsp_client.diagnostics import FileDiagnosticsMap
@@ -736,7 +738,7 @@ class StaticAnalyzer:
         # order. The merges replace on key collision, and overlapping configs
         # (nested solution roots) do collide, so completion order would let two
         # identical runs keep different nodes and produce different component IDs.
-        completed: dict[int, tuple[Language, dict]] = {}
+        completed: dict[int, tuple[LanguageAdapter, dict]] = {}
 
         def run_one(engine_config: EngineConfig, engine_client: LSPClient | None, order: int = 0) -> None:
             """Analyze one engine. Owns the client's lifetime when given none."""
@@ -760,7 +762,7 @@ class StaticAnalyzer:
                 duration_ms = round((time.monotonic() - t_lang_start) * 1000)
                 logger.info(f"Engine analysis for {adapter.language} completed in {duration_ms / 1000:.1f}s")
                 with absorb_lock:
-                    completed[order] = (language, analysis)
+                    completed[order] = (adapter, analysis)
                     self._collect_diagnostics_for(adapter, engine_client, analysis)
                     track_lsp_result(
                         language=adapter.language_enum.value,
@@ -817,9 +819,7 @@ class StaticAnalyzer:
                     f"(attempted: {', '.join(cfg.adapter.language for cfg in pending)}){details}"
                 )
 
-        for order in sorted(completed):
-            language, analysis = completed[order]
-            self._absorb_into_results(results, language, analysis)
+        self._absorb_and_link(results, [completed[order] for order in sorted(completed)])
 
         summaries = []
         for language in results.get_languages():
@@ -863,6 +863,8 @@ class StaticAnalyzer:
         # so a config that took its own copy of the cache would hand back the nodes another
         # config had just invalidated, and the merge would resurrect deleted files.
         carried: dict[Language, dict] = {}
+        carried_adapters: dict[Language, LanguageAdapter] = {}
+        rebuilt: list[tuple[LanguageAdapter, dict]] = []
         for engine_config, engine_client in self._live_clients("warm-start"):
             adapter, project_path = engine_config.adapter, engine_config.project_path
             language = adapter.results_language
@@ -871,7 +873,7 @@ class StaticAnalyzer:
 
             if changed_files is None:
                 analysis = self._run_full_analysis(engine_config, engine_client)
-                self._absorb_into_results(results, language, analysis)
+                rebuilt.append((adapter, analysis))
             else:
                 changed_files = {
                     path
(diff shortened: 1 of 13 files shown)


static_analyzer/analysis_cache.py
```diff
@@ -69,8 +69,9 @@
 # specification in analysis.json, replayed over the graph.
 # v8: a C# file compiled into several projects has symbols and callers, and a
 # nested solution's files are named by its own engine only.
+# v9: a call into a project of another solution root is an edge.
 # Older pickles are treated as cache misses and re-run.
-_TAG_VERSION = "v8"
+_TAG_VERSION = "v9"
```
ivanmilevtues: a classic at this point :D

ivanmilevtues, line 74: a classic at this point :D

static_analyzer/graph_definitions.py
```diff
@@ -0,0 +1,179 @@
+"""Resolve a definition position against a merged call graph, expanded the way the engine expands it.
```
ivanmilevtues, line 1: maybe not for now but i think this is additional index which again will make things harder to track later down the line.

I think the best solution would be to have a global grpah builder which dedups/resolve things as you use add_nodes and then add_edge links em correctly across solutions and so on.

So in Phase 0/1 you put in all the nodes and then in P3 you make them resolve correctly (as you have the full cross solution stuff in there).

Now we have multilpe lookups.

This is usually a blokerish for me but I am unsure how much work would this require, generally i think it will help as it might automagically resolve some custom work we've done to help things match instead of having a single unified index and builder on top of it.

I asked my agent to tell me whats going on, to be short: it wasn't as straight forward as i describe it here.

static_analyzer/graph_definitions.py
```diff
@@ -0,0 +1,179 @@
...
+from static_analyzer.engine.lsp_constants import EdgeStrategy
+from static_analyzer.engine.utils import definition_location
+from static_analyzer.internal_references import parent_qualified_name
+from static_analyzer.node import Node
+
+# How a call site was found in source; it decides what a resolved definition may become.
+CALL = "call"
+METHOD_GROUP = "method_group"
+COLLECTION_INITIALIZER = "collection"
+ITERATED = "iterated"
```
ivanmilevtues, line 23: nit: move to constants.

static_analyzer/__init__.py
```diff
@@ -754,7 +756,7 @@ def run_one(engine_config: EngineConfig, engine_client: LSPClient | None, order:
                 duration_ms = round((time.monotonic() - t_lang_start) * 1000)
                 logger.info(f"Engine analysis for {adapter.language} completed in {duration_ms / 1000:.1f}s")
                 with absorb_lock:
-                    completed[order] = (language, analysis)
+                    completed[order] = (adapter, analysis)
```
ivanmilevtues, line 765: blockerish: (this is really ugly ;d)
this change is very wierd, we used to have completions for langauges, now we have them for adapters?

It seems like a hack so that other things have access to the adapter. I don't think this is the right way to transfer it around idk.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -122,6 +122,7 @@ def build(self, source_files: list[Path], skip_hierarchy: bool = False) -> Langu
             cfg=cfg,
             package_dependencies=package_deps,
             source_files=abs_files,
+            external_call_sites=ctx.external_call_sites,
```
ivanmilevtues, line 125: I think if we have a unified call graph builder this hack would be avoided. I wrote a long comment on this in the new graph index.

ivanmilevtues (approved):
I will give this an approve, but I really don't like this new index i think it is adding again another layer of complexity and will make us search in which index things are not showing up.

also one nit for a very ugly way of sharing the adapters.

It is functional and working so we can merge it i suppose. But if you also agree i'd really like to see a global index with the graph builder, building across lsp runs/langs/solutions etc.

The static analysis index should be one not 3 different i asked and i see that there are multilpe lifecycles (phases) but i think it should rather be shared than created multiple times. I am unsure even if explained it well here ;d

PR #585, opened 2026-09-09
fix(lsp): end every bulk didOpen with a drain barrier

### LINKED ISSUE #529: fix(lsp): add a post-didOpen barrier for workspace-probing servers
## Context

This issue preserves the diagnosis, proposed fix, validation evidence, review feedback, and follow-up observations from [PR #462](https://github.com/CodeBoarding/CodeBoarding/pull/462), which is being closed without merging.

This follows the C# LSP work in #456/#457 and #459. After #457 shipped in 0.13.7, a 286-project / 7,605-file C# solution could load successfully but failed on the first Phase 1 `documentSymbol` request.

## Problem

For adapters with `probe_before_open = True`, symbol discovery currently does:

1. A synchronization probe using the scaled `probe_timeout`.
2. A bulk `didOpen` for every source file.
3. The first Phase 1 `documentSymbol` request using the adapter's fixed per-request timeout.

Bulk `didOpen` queues server work proportional to the file count. The next request therefore acts as an implicit drain barrier, but only receives C#'s 120-second default timeout.

Observed failure:

```
Analyzing 7605 CSharp files
Waiting for LSP server indexing (timeout=1800s)...
Sync probe: 56.4s (1 symbols)
Error during engine analysis for CSharp: Timeout waiting for LSP response to request 3
```

The timeout occurred exactly 120.001 seconds after opening all 7,605 files, matching `CSharpAdapter.get_lsp_default_timeout()`.

A direct csharp-ls 0.24.0 reproduction using the engine's handshake showed:

```
PROBE OK in 195.2s
did_open 8740 files in 27.2s
POST-OPEN REQUEST OK in 517.6s
```

The server remained healthy, published diagnostics, and used bounded memory (about 2.5 GB peak in the direct reproduction). This is a timeout-shaping problem, not a hang or unbounded-growth problem.

## Proposed change

In `CallGraphBuilder._discover_symbols`, add a second synchronization probe immediately after `_bulk_did_open` when:

- `probe_before_open` is true; and
- `interleave_open` is false.

Use the same scaled `probe_timeout`, label this wait as `"overlay processing"` in logs, and pass its result to the first file so the probe result is reused as before.

The intended sequence is:

```
pre-open documentSymbol probe
all didOpen notifications
post-open documentSymbol probe with scaled timeout
remaining Phase 1 documentSymbol requests
```

Do not change the other adapter paths:

- `interleave_open` adapters already apply per-file backpressure.
- Probe-after-open adapters already use their existing probe as the drain barrier.
- The 1,800-second `_PROBE_MAX_TIMEOUT` covered the measured 517.6-second wait and does not need adjustment based on current evidence.

## Acceptance criteria

- [ ] Workspace-probing, non-interleaved adapters perform a scaled-timeout probe both before and after bulk `didOpen`.
- [ ] The post-open probe is logged distinctly as overlay processing.
- [ ] The post-open probe result is reused for the first source file.
- [ ] Existing interleaved and probe-after-open behavior remains unchanged.
- [ ] A regression test checks the exact call order through `lsp.method_calls`: first probe, every `did_open`, then the second probe.
- [ ] The regression test also retains the scaled-timeout and `did_open` count assertions.
- [ ] The full unit suite, Black, and mypy pass.

## Validation evidence from PR #462

With the proposed barrier, the previously failing C# solution completed:

```
Sync probe: 169.1s (1 symbols)
did_open 7605 files: 17.6s
Waiting for LSP server overlay processing (timeout=1800s)...
Phase 1 (symbols): 100% (7605/7605 files) | symbols=54397
Phase 2 total (build edges): 428.9s, 3821 edges
Phase 3 (hierarchy): 810.6s
Pipeline complete: 7605 files, 54397 symbols, 3821 edges, 8368 hierarchy entries in 2170.0s
Reference edges for CSharp: 14855 ({'contains': 12063, 'inherits': 2792})
Static analysis complete: csharp: 7605 files, 42306 references, 20434 nodes, 3811 edges
```

Peak csharp-ls RSS was approximately 3.8 GB during Phase 2; the recycler armed but did not need to fire.

The PR reported:

- `uv run pytest`: 2,020 passed, 28 skipped
- `uv run mypy .`: clean on 313 files
- `black --check`: clean

CodeRabbit's only actionable review finding was to strengthen the regression test so it verifies that the second probe occurs after all `didOpen` calls, rather than checking only call counts and timeout values.

## Follow-up observation

Phase 3 hierarchy construction took 810.6 seconds of the 2,170-second run, longer than Phase 2. Treat hierarchy performance as a separate optimization if end-to-end action duration matters; it is not part of this synchronization fix.

RapidPoseidon:
Closes #529.

## WHAT

Bulk `didOpen` ends with a synchronization probe, and `_bulk_did_open` returns its result. The probe-after-open branch's separate probe becomes redundant and drops out.

The env var this PR originally carried is split out into #594; this is the fix on its own.

## WHY

`didOpen` queues server work proportional to the file count, and whichever request comes next drains it — on the adapter's fixed per-request timeout. For C# that is 120s, which a large solution exceeds with the server perfectly healthy.

We hit this from the outside, on the GitHub action against a 21-service .NET monorepo (8983 C# files):

```
Found 8983 CSharp files
LSP server: solution loaded (csharp-ls: Finished loading solution "Rapidata.slnx")
Sync probe: 95.9s (1 symbols)
did_open 8983 files: 26.8s
ERROR  Error during engine analysis for CSharp: Timeout waiting for LSP response to request 3
ERROR  LSP analysis result for csharp is unhealthy: zero nodes despite LOC, zero edges
       despite LOC, missing call graph, missing source files
```

Worth noting because the run still reports **success**: C# is dropped, only a generated TypeScript client survives, and the committed baseline ends up describing 3 of 10925 files as the whole architecture. Nothing in the action's exit status says the analysis is empty.

@Svilen-Stefanov reproduced both the original failure and the post-#582 shape on Umbraco-CMS (6011 files, csharp-ls throttled, real 120s timeout), and confirmed the fix absorbs a 401s drain on `main` 30451213.

## HOW

The barrier lives at the end of `_bulk_did_open` rather than at one call site. Every bulk open then ends with one wherever it sits — including the post-Phase-1 open on the `workspace_owns_documents` path, where an unserved first Phase 2 `textDocument/definition` round fails the entire run with `StaticAnalysisFatalError` rather than just dropping C#.

That placement is deliberate: the first version of this PR put the probe in the `probe_before_open` branch, and #582 moved C# onto the `workspace_owns_documents` path the day after, leaving the probe unreachable. Tying the barrier to the open keeps a later reorder from losing it again.

`_send_sync_probe` takes a `label` (default `"indexing"`); the barrier passes `"didOpen drain"`, which reads correctly both for C# and for the probe-after-open adapters that now take it as their only wait.

## Testing

- `uv run pytest` — 2243 passed, 6 skipped
- `uv run mypy .` — clean, 310 source files
- `uv run black --check .` — clean, 310 files

`test_bulk_open_ends_with_a_drain_barrier` sets `workspace_owns_documents = True`, the combination C# actually has, and asserts the barrier lands immediately after the post-Phase-1 `did_open`s on the scaled timeout:

```
("document_symbol", a.cs)   # pre-open probe
("document_symbol", b.cs)   # Phase 1
("did_open", a.cs)          # post-Phase-1 bulk open
("did_open", b.cs)
("document_symbol", a.cs)   # the barrier
```

It fails on plain `main`, verified by reverting only `call_graph_builder.py`. The two existing `workspace_owns_documents` tests each gained the trailing barrier call in their expected sequences.

`CodeBoarding-tests` skips on fork PRs per CONTRIBUTING §5e; @Svilen-Stefanov ran the edge-case (C#/Python/Java/TS) and incremental (C#/Python) suites locally with the same results as main.

## Not addressed

- `_restore_inbound_edges_via_definitions` in `incremental_orchestrator.py` opens caller files by hand and then sends definition batches on the fixed timeout, so "wherever the open happens" is not fully true yet. @Svilen-Stefanov is filing a ticket.
- The 1800s `_PROBE_MAX_TIMEOUT` is unchanged; #529's evidence says it does not need adjusting.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved symbol discovery reliability by ensuring document-opening work is fully processed before symbols are extracted.
  - Applied consistent synchronization across supported discovery paths, reducing the risk of incomplete or stale symbol results.
  - Improved handling of server processing delays by using an appropriate timeout while waiting for indexing work to finish.

- **Tests**
  - Added coverage verifying that document-opening operations are drained before symbol extraction.
  - Updated call-order expectations to reflect the more reliable processing sequence.


RapidPoseidon:
Pushed 18dda53 addressing the review.

**Fixed**

- *Reject invalid timeout overrides* — agreed, and the fallback was worse than it looked: it restored the exact ceiling the operator was overriding. An unparseable or non-positive value now raises. It raises `StaticAnalysisFatalError` rather than `ValueError` because both `_spawn_engine_client` call sites catch `Exception` and turn it into "skip this language and continue", so a `ValueError` would have degraded into the same silent partial analysis. The analysis loop already re-raises `StaticAnalysisFatalError`; I added the matching passthrough to the other call site so the two agree.
- *Assert reuse of the post-open probe result* — correct, and the test was weaker than I thought: every probe returned the same empty list, so it would have passed even if the first file reused the stale pre-open result. The two probes now return distinct payloads and the test asserts which one reaches `record_document_symbols`. Verified by keeping the post-open probe but not assigning its result — that now fails on the new assertion alone.
- *Trim the test module docstring* — done, issue reference dropped.

**Not doing**

- *Add `-> None` to the changed test methods* — skipping for consistency with the surrounding suite: `tests/` currently has 1740 unannotated `def test_*(self):` against 77 annotated, `test_max_concurrent_engines.py` (the file this one mirrors) is unannotated throughout, and `disallow_untyped_defs = false` in `pyproject.toml`. Happy to annotate if you'd rather the new files lead the way.

Gates on 18dda53: `uv run pytest` 2112 passed / 6 skipped, `uv run mypy .` clean on 304 files, `uv run black --check .` clean.

Note that `CodeBoarding-tests` skips here, as CONTRIBUTING §5e says it does for fork PRs — so the private integration suite has not run against this branch.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -170,6 +170,10 @@ def _discover_symbols(self, source_files: list[Path]) -> None:
             probe_result = self._send_sync_probe(source_files, probe_timeout)
             if not interleave_open:
                 self._bulk_did_open(source_files)
+                # Why: bulk didOpen queues server work proportional to the file
+                # count, and the next request drains it. Without a probe here that
+                # drain lands on the adapter's fixed per-request timeout.
+                probe_result = self._send_sync_probe(source_files, probe_timeout, label="overlay processing")
```
Svilen-Stefanov: blocker-ish: this fixes it on the branch base, but I don't think it survives a rebase. #582 landed the day after this was opened and C# now has `workspace_owns_documents = True`, so `_discover_symbols` goes through the `if owns_documents:` branch and never gets here. Go and PHP interleave, so on main no shipped adapter reaches this line. For C# the bulk didOpen now happens after Phase 1, so the drain lands on the first Phase 2 `textDocument/definition` round, and on main an unserved round raises `StaticAnalysisFatalError`, so the whole run dies instead of just dropping C#.

What I ran (Umbraco-CMS, 6011 files, throttled, default 120s):
- PR base: `Timeout waiting for LSP response to request 3` 120.1s after didOpen, C# unhealthy
- PR head: overlay probe blocked 405.8s, Phase 1 + Phase 2 fine
- main: first definition round blocked 120.1s, then `StaticAnalysisFatalError`
- main + this PR: same as main, the new probe never runs
- main + this PR + the barrier moved below: probe blocked 414.6s, Phase 2 served

```python
if owns_documents:
    already = set(read_from_source) | set(opened_early)
    self._bulk_did_open([file_path for file_path in source_files if file_path not in already])
    self._send_sync_probe(source_files, probe_timeout, label="overlay processing")
```

Could you rebase and move it there? I wonder if it'd be even nicer to put the barrier inside `_bulk_did_open` itself (and drop the separate probe in the `else` branch), so every bulk open ends with one and the next reorder can't lose it again.

tests/static_analyzer/test_call_graph_builder.py
```diff
@@ -188,6 +197,47 @@ def test_interleaves_did_open_with_symbol_queries(self):
         ]
         assert [item.kwargs.get("timeout") for item in lsp.document_symbol.call_args_list[1:]] == [64, 64]
 
+    def test_probes_again_after_bulk_did_open(self):
+        """Why: bulk didOpen queues server work that the next request drains, so the
+        drain must land on a probe with the scaled timeout, not the fixed default."""
+        lsp = _make_lsp()
+        adapter = _make_adapter()
+        adapter.probe_before_open = True
```
Svilen-Stefanov, line 195: Seems like this pins a combination no real adapter has on main anymore: `probe_before_open=True` with `workspace_owns_documents` left `False` from `_make_adapter`. I listed the flags of every adapter on main + this PR at runtime and none of them reaches the new probe, but this test still passes there, so it wouldn't catch that. Also heads up, this file conflicts with main (`CallGraphBuilder` takes `repository_path` now). Could we set `workspace_owns_documents = True` like the C# tests on main do and assert the probe comes right after the post-Phase-1 `did_open`s?

static_analyzer/__init__.py
```diff
@@ -96,6 +97,31 @@ def recommended_engine_concurrency(engine_count: int) -> int:
     return max(1, min(engine_count, cpu_bound, memory_bound))
 
 
+def lsp_request_timeout(adapter: LanguageAdapter) -> int:
```
Svilen-Stefanov: Do we really need this one once the barrier sits in the right place? It applies to every language at once (a 900 meant for csharp-ls also lets any hung server block 900s per request round), it isn't documented anywhere, and it needs a release to ship anyway. Not against an escape hatch per se, but since you offered, I'd lean towards splitting it into its own PR so the fix can land on its own. @ivanmilevtues what do you think?

static_analyzer/__init__.py
```diff
@@ -96,6 +97,31 @@ def recommended_engine_concurrency(engine_count: int) -> int:
...
+    """Seconds one LSP request may block, defaulting to the adapter's own ceiling.
+
+    Why overridable: a ceiling that fits a server on a mid-sized workspace can be
```
Svilen-Stefanov: nit: the docstring is longer than the function needs, AGENTS.md asks for a one-liner plus a short Why.

static_analyzer/__init__.py
```diff
@@ -306,6 +332,8 @@ def start_clients(self) -> None:
                 continue
             try:
                 started.append((engine_config, self._spawn_engine_client(engine_config)))
+            except StaticAnalysisFatalError:
+                raise
```
Svilen-Stefanov: just checking, this is a bit broader than the timeout config. `LSPClient.start()` can raise `StaticAnalysisFatalError` when the server process exits during `initialize`, so in a multi-language repo the eager path goes from "Proceeding with partial LSP coverage" to aborting the whole run. It matches the lazy path so maybe that's what we want, but if the goal is only to fail on a bad env value, could we validate it once before the spawn loop and leave this as is?

Svilen-Stefanov (changes requested):
Thanks a lot for digging into this and for the detailed write-up, the diagnosis is spot on. I could reproduce the exact failure on Umbraco-CMS (6011 C# files, csharp-ls throttled to the efficiency cores to mimic a small runner, real 120s timeout): Timeout waiting for LSP response to request 3 120.1s after didOpen, C# reported unhealthy, and the run still returned successfully. On the PR head the new probe absorbs a 405.8s drain and Phase 1/2 go through, so on its own base this works 👍

The catch is that main moved underneath it. After https://github.com/CodeBoarding/CodeBoarding/pull/582 C# takes the workspace_owns_documents path, so this probe is never reached, and the drain now lands on the first Phase 2 definition round, which on main fails the whole run with StaticAnalysisFatalError. Main + this PR fails exactly the same way. Moving the barrier after the post-Phase-1 bulk open fixed it in the same setup, details inline. So the blocker is a rebase + moving the barrier (and the test conflict), and I'd probably split the env var out.

RapidPoseidon:
Thanks — that's a much better catch than the bug itself, and the Umbraco numbers made it obvious. Rebased onto main; the PR is now one commit and the env var is split out into #594.

**The barrier moved, and then some.** I took the second suggestion rather than just relocating the call: the probe now lives at the end of `_bulk_did_open`, which returns its result. So every bulk open ends with a barrier wherever it sits — the post-Phase-1 open on the `workspace_owns_documents` path included — and a future reorder of the phases cannot lose it again, which is exactly how this PR broke. The separate probe in the probe-after-open `else` branch became redundant and drops out, so `_discover_symbols` reads a little shorter than before.

One consequence worth flagging: the probe-after-open branch's wait line now logs `overlay processing` instead of `indexing`, since it is the same barrier. The label still reads correctly there — it is a bulk open being drained — but it is a log-text change for Python/TS/Go, so say the word if you'd rather keep the old label on that path.

**The test.** You were right that it pinned a combination nothing has any more. It now sets `workspace_owns_documents = True` like the C# tests, and asserts the probe lands immediately after the post-Phase-1 `did_open`s, on the scaled timeout:

```
("document_symbol", a.cs)   # pre-open probe
("document_symbol", b.cs)   # Phase 1
("did_open", a.cs)          # post-Phase-1 bulk open
("did_open", b.cs)
("document_symbol", a.cs)   # the barrier
```

It fails on plain main, which the old one did not — I checked by reverting only `call_graph_builder.py`. The two existing `workspace_owns_documents` tests each gained the trailing barrier call in their expected sequences; that is the whole test diff besides the new case.

I have not reproduced your Umbraco run end to end — no throttled 6011-file C# checkout here — so the phase timings in this PR are still yours, not mine. The order guarantee is what the unit test pins.

Gates on 6e87ea3: `uv run pytest` 2243 passed / 6 skipped, `uv run mypy .` clean on 310 files, `uv run black --check .` clean. `CodeBoarding-tests` skips on fork PRs per CONTRIBUTING §5e, so the integration suite still needs a maintainer run.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -221,8 +220,12 @@ def _discover_symbols(self, source_files: list[Path]) -> None:
 
         logger.info("Discovered %d symbols across %d files", len(self._symbol_table.symbols), len(source_files))
 
-    def _bulk_did_open(self, source_files: list[Path]) -> None:
-        """Phase 0: Send didOpen for all files so the LSP server can index them."""
+    def _bulk_did_open(self, source_files: list[Path], probe_timeout: int) -> list[dict]:
+        """Phase 0: open every file, then block until the server has drained them.
+
+        Why the trailing probe: didOpen queues work proportional to the file count,
+        so the next request pays for it — on the scaled timeout, wherever it lands.
```
Svilen-Stefanov: These em dashes :D nit: could we swap this one for a normal hyphen or a comma?

static_analyzer/engine/call_graph_builder.py
```diff
@@ -234,11 +237,12 @@ def _bulk_did_open(self, source_files: list[Path]) -> None:
             time.sleep(0.1)
         pbar.finish()
         logger.info("did_open %d files: %.1fs", total, time.monotonic() - t_open_start)
+        return self._send_sync_probe(source_files, probe_timeout, label="overlay processing")
```
Svilen-Stefanov: nit: for the probe-after-open adapters (Python, TS, JS, Java, Rust) this is now their first and only wait, so a plain Flask run logs `Waiting for LSP server overlay processing`, which reads a bit odd there. Maybe pass the label into `_bulk_did_open` from the call sites, or pick something neutral like "didOpen drain"?

Svilen-Stefanov (approved):
Thanks for the quick turnaround, this looks great now and putting the barrier inside _bulk_did_open is exactly what I was hoping for.

I re-ran the Umbraco repro (6011 C# files, throttled, real 120s timeout) on the exact base, main 30451213: before, the first Phase 2 definition round blocks 120s and the whole run dies with StaticAnalysisFatalError. With this PR the overlay barrier absorbs a 401s drain and Phase 2 goes through 👍 Python on Flask behaves the same apart from the log label, the static_analyzer unit tests pass, and since CodeBoarding-tests skips on forks I ran the edge-case (C#/Python/Java/TS) and incremental (C#/Python) suites locally, same results as main.

Only nits left from my side. Could you also update the PR description and title? They still describe the two-commit version with the env var and the old tests.

One follow-up for later, not blocking: _restore_inbound_edges_via_definitions in incremental_orchestrator.py still opens the caller files by hand and then sends definition batches on the fixed timeout, so "wherever the open happens" isn't fully true yet. I'll make a ticket so we don't forget.

RapidPoseidon:
Both nits fixed in 831f3d6, and the title and description now describe the single-commit version.

- The em dash is gone.
- Label is now `didOpen drain` rather than a per-call-site parameter. You're right that `overlay processing` read oddly once the probe-after-open adapters take it as their only wait, and a neutral label keeps the barrier self-contained inside `_bulk_did_open` — threading it in from the call sites would give back some of the "the open carries its own barrier" property that made the fold worth doing. A plain Flask run now logs `Waiting for LSP server didOpen drain`.

Thanks for the Umbraco re-run on the exact base and for running the edge-case and incremental suites locally — that covers the gap I could not close from a fork.

Noted on `_restore_inbound_edges_via_definitions`; it does open caller files by hand and then batch definitions on the fixed timeout, so the invariant is not complete yet. Happy to pick that ticket up once it exists.

Gates on 831f3d6: `uv run pytest` 2243 passed / 6 skipped, `uv run mypy .` clean on 310 files, `uv run black --check .` clean.

PR #589, opened 2026-09-11
fix(engine): a definition names the declaration written at it, or nothing

Svilen-Stefanov:
Why: `_resolve_definition_to_symbol` fell back to any symbol on the definition's
line and then to +-1 and +-2 lines, and `nodes_at_location` took any node starting
on the line, preferring classes and callables whatever their name. Both bind a
definition that landed on a parameter, a local or a body statement to whatever
declaration is nearest, which is never the thing called: on the Go and TypeScript
edge-case projects every such match is a call site bound to the function around it
(`super(id);` -> the calling constructor, `for _, cb := range ...` -> the method).

What changes: one rule set, in the engine's resolver and in the merged graph's, so
a full run and a warm start cannot answer a position differently.
1. the declaration at the exact position;
2. else the innermost declaration that starts on that line and still contains the
   position (largest start column, then shortest span, then longest name);
3. else, when the identifier written at the definition names exactly one callable
   or class the file declares, that one -- the overload set whose signature line
   the server answers with, which the symbol table does not hold;
4. else nothing.
`_resolve_definition_to_symbol`/`_best_candidate` become `SymbolIndex`, which also
owns the by-position, by-line and by-name lookups `_build_definition_lookups` built;
`GraphIndex` gains the same rules and the source inspector they read names from.
Each pass logs one line: exact, containing, by name, rejected.

The client now advertises `definition.linkSupport`. Without it a server answers a
plain Location whose range starts at the modifiers, never at the name, so rule 1
cannot fire: on the PHP edge-case project 77 of 81 definitions resolved only through
the fallback this change removes. With it, intelephense answers LocationLink and all
81 are exact. No other server's answers change.

Measured, no LLM, one run each, edge-case projects (main -> this change):

| project    | nodes | edges      | dir pairs |
|------------|-------|------------|-----------|
| java       | 82    | 110 -> 110 | 4 -> 4    |
| python     | 67    | 60 -> 60   | 3 -> 3    |
| go         | 81    | 65 -> 65   | 8 -> 8    |
| php        | 85    | 75 -> 75   | 8 -> 8    |
| javascript | 77    | 61 -> 61   | 10 -> 10  |
| typescript | 97    | 78/82 both sides (references is not deterministic here) |

Definition results classified with the new rules, definitions forced on for every
language (exact / containing / by name / rejected, and matches the old fallback made
that are now dropped):

| project    | exact | containing | rejected | dropped fallback matches |
|------------|-------|-----------|----------|--------------------------|
| java       | 83    | 0         | 0        | 0                        |
| python     | 62    | 1         | 18       | 0                        |
| php        | 81    | 0         | 0        | 0                        |
| go         | 71    | 1         | 47       | 3, all wrong             |
| typescript | 125   | 2         | 66       | 2, all wrong             |

All five dropped matches were hand-checked: each is a statement inside a body bound
to the declaration one line above it.

Java real projects, full runs (main -> this change):

| project      | nodes | edges       | directory pairs |
|--------------|-------|-------------|-----------------|
| junit4       | 1890  | 3287 -> 3287 | 184 -> 184     |
| commons-lang | 4949 -> 4948 | 6830 -> 6656 | 76 -> 69 |

commons-lang is where the fallback was doing work, and all of it was wrong. Of 5770
definition results 5746 are exact and 24 contain the position; 59 matches are dropped,
every one a definition that landed on a lambda inside a method body, which the +-2 line
window then named after the declaration above it. Each produced an edge that reverses
the dependency it came from: `Functions.test` calls `getAsBoolean(() -> predicate.test(...))`
and the graph read `getAsBoolean -> test`; `TypeUtils.AMP_JOINER` builds an
`AppendableJoiner` with a lambda and the graph read `AppendableJoiner.joinArray ->
TypeUtils.AMP_JOINER`. All 7 lost directory pairs are those reversals (lang3 -> reflect,
tuple -> reflect, function -> time); 131 of the 174 lost edges leave the same caller still
reaching the same declaring type by the member it really calls.

Not measured: C#, and the eShop/Polly/abp warm-start population. This environment has no
.NET SDK and the egress proxy refuses builds.dotnet.microsoft.com, so csharp-ls cannot run.
mockito is Gradle and jdtls imports no dependencies here, so it reports zero symbols on both
sides. The Java and C# edge-case fixtures in CodeBoarding-tests already disagree with main
over #586's naming change, independently of this work.

Tests: `SymbolIndex.resolve` per rule, including the parameter, local and import
lines the old fallback mis-bound, a TypeScript overload set, a dual registration at
one position, and the counters; a parity test feeding nine positions to both
resolvers; the graph-side tests for the removed fallbacks now assert no match; the
cross-engine linker no longer links a site one line off a declaration.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_012jDDQdFvgzjEpNr3DCoq7v

## Summary by CodeRabbit

* **Improvements**
  * Static analysis now produces more consistent call graphs across full rebuilds and warm starts.
  * Improved resolution for method groups, function-valued calls, receiver-member calls, inheritance, implementations, and external symbols.
  * Expanded call detection for JSX, optional chaining, Rust macros, decorators, and keyword arguments.
  * Improved declaration matching near decorators, annotations, overloaded symbols, and imported names.
  * Retryable language-server request failures are retried automatically.
* **Bug Fixes**
  * Analysis failures that could produce misleading results are now surfaced instead of silently continuing.
  * Cached analysis results are automatically rebuilt when using older cache formats.


static_analyzer/engine/lsp_client.py
```diff
@@ -42,6 +48,18 @@ class MethodNotFoundError(Exception):
...
+def _is_protocol_failure(error: object) -> bool:
+    """Whether an error response means the request could not be served at all.
+
+    "Method not found" is the one reserved code that is still an answer: the server is
+    telling us it does not implement the request, and empty is the truth.
+    """
+    if not isinstance(error, dict):
+        return False
+    code = error.get("code")
+    return isinstance(code, int) and code <= LSP_RESERVED_ERROR_CODE_MAX and code != LSP_METHOD_NOT_FOUND
```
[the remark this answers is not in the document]
Svilen-Stefanov: The finding is right and is fixed, but not with `-32768 <= code <= -32000`: that interval excludes LSP's own codes, which sit *below* it. `LSPErrorCodes` reserves −32899…−32800 (`RequestCancelled` −32800, `ContentModified` −32801, `ServerCancelled` −32802, `RequestFailed` −32803), and treating `ContentModified` as an ordinary answer is exactly what an earlier commit in this PR fixed — rust-analyzer returns it for every query issued while it is still indexing, and reading those as empty silently drops edges.

So the predicate now tests the union of the two reserved bands and nothing else:

```python
LSP_RESERVED_ERROR_CODE_RANGES = ((-32899, -32800), (-32768, -32000))
```

`-40000` is outside both, so it is the server's own answer and no longer fails the run — which is the case you raised. Tests cover both band edges (−32899/−32800 and −32768/−32000), the gap between them (−32799), outside on either side (−32900, −31999), and an application code of −40000.

---
_Generated by [Claude Code](https://claude.ai/code)_

Svilen-Stefanov:
## Since the review: seven findings, each traced to its cause

Every one came back to a position or a code being read in the wrong system, and each drops edges silently rather than failing.

**Columns.** tree-sitter counts a column in bytes, LSP counts UTF-16 code units, and nothing converted between them. On any line holding non-ASCII text before a callee the definition query landed past the identifier, and the stored call site was wrong even where the server still answered. Every position now crosses the `SourceInspector` boundary through `ParsedSource`, which converts in both directions; an all-ASCII file — almost every file — keeps its columns untouched and pays one `bytes.isascii()` per parse. `s = 'héllo wörld'; target()` was asked at column 24 and is asked at 20.

**`this.#save()`.** `private_property_identifier` was in no name-node table, so `_select_query_node` fell through to `this`, which resolves to the class: every private-member call in JS/TS reached the class and never the method.

**Six more value bindings**, one per grammar spelling: Python's `default_parameter` and `typed_default_parameter`, JavaScript's `field_definition`, TypeScript's `public_field_definition`, Go's `var_spec`. Go's `const_spec` was considered and left out — a Go constant cannot hold a func, so querying its value only spends a request.

**Constructing is a property of the site, not of the kind.** The engine decides it per call site from the source; the graph-side resolvers asked `kind == CALL` first. `new Bag { 1 }` is one position with two shapes, and a full build gave it the class, the `Add` members *and* the constructors while a warm start gave it two of the three. `targets_for` also `continue`d past that expansion for a loop subject, so `foreach (var x in new Bag { 1 })` lost it on both paths.

**Reserved error codes are two bounded bands, not a half-line** — see the thread on `lsp_client.py`. The union of JSON-RPC's −32768…−32000 and LSP's own −32899…−32800; a single `-32768 <= code <= -32000` interval would let `ContentModified` (−32801) read as an empty answer.

**A dead server is fatal.** `_next_response` raised a bare `RuntimeError`, which `discover_file_dependencies`'s `except Exception` turned into "this file depends on nothing". It raises `StaticAnalysisFatalError` now — a `RuntimeError` subclass, so existing handlers still catch it.

**Cross-engine implementation expansion: not fixed, deliberately.** `link_external_call_sites` runs from `_absorb_and_link`, after `run_one`'s `finally` has already called `owned_client.shutdown()` on the engine owning the target file. There is no server left to ask. Fixing it means changing engine-client lifetime, which is its own change; documented at the call site.

Cache tag `v11` → `v12`: a stored call-site column moved.

### Measured

Real projects, Linux, one run each, no LLM. Only node and edge counts move — every `references_count`, `packages_count` and `source_files_count` came back exactly as the fixtures pin them, which is the check that the symbol table is untouched.

| project | nodes | edges |
|---|---|---|
| codeboarding_python | 642 (unchanged) | 705 → **827** |
| lodash_javascript | 880 → **950** | 1388 → **1477** |
| excalidraw_typescript | 5651 → **5855** | 6197 → **6798** |
| prometheus_go | 5489 (unchanged) | 13808 → **13117** |
| clap_rust | 1859 (unchanged) | 3407 → **5597** |
| wordpress_php | 19646 → **14502** | 90718 → **52324** |

excalidraw's 5855/6798 are exactly what a Windows CI runner measured, and clap's 5597 is within three edges of the prototype arm's 5594.

**WordPress is the one that falls, and the fall is the fan-out this change removes.** Both strategies over the same clone: the symbol table is identical (114342 symbols, 1644 files) and so is every callable node kind — 4762 functions, 805 classes, 6029 methods, 325 constructors, 10 interfaces. What leaves is 5686 `VARIABLE` nodes, in the graph only because a reference to a variable counted as an edge, and 22887 edges of which 82% have a destination whose simple name is shared by six or more callables. WordPress declares some forty REST controllers with identical member names, so `WP_REST_Comments_Controller::update_item` was linked to `get_item_schema` on every one of them; for 63% of the lost callable-to-callable edges the definitions run keeps a *different* same-named target from the same caller. The residue is the hook API — `add_filter('x', array($this, 'method'))` — where the callee is named by a string and no call is written to query at.

Edge-case projects gained the four shapes above, each as an exact expected edge, and **no project lost an edge**: python 75 → 78, javascript 69 → 74, typescript 105 → 112, go 65 → 68.

### Open: Rust and macro arguments

`rust_edge_cases` is 3 edges short and 2 edges richer, and both differences are correct behaviour rather than fixture drift.

The three missing — `main -> Cat.speak`, `main -> Dog.speak`, `main -> Speaker.speak` — are all `println!("{}", x.speak())`. tree-sitter parses a macro's arguments as an opaque `token_tree`, so no call site is found inside one. Across clap's 330 files that hides 6797 call-shaped names against 29942 found: 18.5% of all call sites. References found them, because rust-analyzer reports references inside macro bodies.

The two extra — `Speaker.shout -> Cat.speak/Dog.speak` — are a default trait method correctly expanded through its implementors, which references never found.

clap still gains 64% of its edges overall, so the aggregate does not fail. The question is whether to add macro-argument call sites (a name inside a `token_tree` applied to a parenthesised group) before flipping Rust, or to accept the loss and record it. That is a decision rather than a defect, so nothing has been changed and `rust_edge_cases.json` has not been regenerated.

Paired with CodeBoarding/CodeBoarding-tests#44.

---
_Generated by [Claude Code](https://claude.ai/code)_

Svilen-Stefanov:
## Correction: the Windows difference was ours, and it is fixed

My earlier comment attributed a Windows-only resolution difference to pyright. That was wrong, and the real cause is worth reading because it is not specific to this PR.

`LSPClient.did_open` read every source file as

```python
text = file_path.read_text(errors="replace")
```

with no encoding, so Python decoded it with the locale's — UTF-8 on Linux, **cp1252 on Windows**. Source is UTF-8 everywhere. On Windows a line holding `café — αβγ` therefore reached the server one character per extra byte longer than it is, while `SourceInspector` read the same file as raw bytes and produced the true columns. Every column sent after such a line named a different place in the server's buffer than in the file.

**So every non-ASCII source file has been analysed against a mismatched buffer on Windows, in every language, for as long as `did_open` has existed.** Nothing failed loudly; an edge attached to whatever name happened to sit at the shifted column, or to nothing.

### How it presented, and what finally gave it away

With two calls on the non-ASCII line, Windows attributed the edge to the *next* call — eight columns on, then thirteen when the line changed. With one call on the line, the edge disappeared entirely. The tell was `typescript` failing at the same site while `javascript` passed *only because its untyped receiver expects no edge there*: two independent servers behaving identically is almost never two server bugs.

Things I checked and ruled out along the way, each of which left something worth keeping:

* the byte → UTF-16 column conversion — correct, and `test_a_column_counts_utf16_units_not_bytes` proves it on every platform;
* `general.positionEncodings` not being declared — genuinely missing, now declared, with a fatal check if a server answers with a different encoding. It did not move this bug, because the server honestly reports UTF-16; the mismatch was in the text, not the units;
* CRLF and `.gitattributes` — `eol=lf` is enforced and byte-identical columns were verified locally.

### Fixed

`did_open` names `encoding="utf-8"`. Go's build-tag scan read files the same way and is named too; no other encoding-less text read remains in `static_analyzer/` or `repo_utils/`.

The test pins the invariant rather than the symptom — that the decode never defers to the locale — and fails if the fix is reverted. The three edge-case projects carry `café — αβγ` before a call again, which is the only end-to-end check that would catch this class: a unit test can pin the column the engine asks at, but only a real run shows the server answering there.

**CI on Windows, with that coverage in place:** `python` ✅ `javascript` ✅ `typescript` ✅ `go` ✅ `php` ✅ (and `php` on Linux ✅). Those first three had failed four runs running.

---
_Generated by [Claude Code](https://claude.ai/code)_

static_analyzer/incremental_orchestrator.py
```diff
@@ -346,60 +332,79 @@ def _add_outbound_edges_from_changed_files(
...
                     external.append(
                         ExternalCallSite(
                             src_node.fully_qualified_name, str(location[0]), location[1], location[2], site, kind
                         )
                     )
                     continue
-                added += _add_edges(call_graph, src_node, targets, site, changed_file_strs)
+                reached = True
+                added += _add_edges(call_graph, src_node, targets.nodes, site, changed_file_strs)
+                if str(location[0]) not in changed_file_strs:
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. Implementations are expanded for every resolved callable target, and the partial build only counts as owning a caller-to-implementation edge when it held both the declaration and the implementation. `test_a_changed_declaration_still_reaches_an_unchanged_implementation` covers the case.

static_analyzer/graph_definitions.py
```diff
@@ -23,36 +27,184 @@
...
 
 def position_inside_node(node: Node, zero_based_line: int, character: int) -> bool:
     line = zero_based_line + 1
     if line < node.line_start or line > node.line_end:
         return False
     if line == node.line_start and character < node.col_start:
         return False
+    # ``col_end`` is 0 on a node built before it was recorded, which reads as unbounded.
+    if line == node.line_end and node.col_end and character > node.col_end:
+        return False
```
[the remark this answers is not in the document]
Svilen-Stefanov: No longer applies after 6630844f. The signature rule that needed `col_end` is gone (on nine repositories it only ever matched parameters), `Node.col_end` went with it, and `position_inside_node` is back to main's behaviour.

static_analyzer/engine/adapters/go_adapter.py
```diff
@@ -104,6 +104,10 @@ def lsp_command(self) -> list[str]:
     def language_id(self) -> str:
         return "go"
 
+    @property
+    def resolves_method_groups(self) -> bool:
+        return True
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. `composite_literal`, `literal_value`, `keyed_element` and `literal_element` are value groups now, and a keyed element passes its value, never its key. `test_go_keyed_literal_passes_its_values_not_its_keys` covers it.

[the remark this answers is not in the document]
Svilen-Stefanov: Same as the thread above, fixed in 6630844f.

static_analyzer/engine/source_inspector.py
```diff
@@ -89,19 +134,68 @@
...
+    "short_var_declaration": "right",  # Go
+    "variable_declarator": "value",  # Java, JavaScript, TypeScript; C# names no field
+    "let_declaration": "value",  # Rust
+    "default_parameter": "value",  # Python
+    "typed_default_parameter": "value",  # Python
+    "field_definition": "value",  # JavaScript class field
+    "public_field_definition": "value",  # TypeScript class field
+    "var_spec": "value",  # Go
+}
 _VALUE_BODY_NODE_TYPES = frozenset({"return_statement", "arrow_expression_clause"})
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. `yield` (Python) and `yield_expression` (JS/TS) are value positions like `return`. `test_python_yield` covers it.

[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. `return_expression` is a value position, and Rust arrays and tuples are value groups. `test_rust_return` covers it.

static_analyzer/engine/call_graph_builder.py
```diff
@@ -67,10 +63,8 @@ def build(self, source_files: list[Path], skip_hierarchy: bool = False) -> Langu
         t_indices_done = time.monotonic()
         logger.info("Build indices: %.1fs", t_indices_done - t_symbols_done)
 
-        ctx = EdgeBuildContext(
-            self._lsp, self._symbol_table, self._source_inspector, recycler=self._build_recycler(source_files)
-        )
-        edge_set = self._build_edges(ctx, source_files)
+        ctx = EdgeBuildContext(self._lsp, self._symbol_table, self._source_inspector)
+        edge_set = build_edges_via_definitions(self._adapter, ctx, source_files)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Leaving this one. The same-start-line guard in `_is_valid_edge` predates this PR (Java and C# went through it on main) and keeps a declaration and the members registered on its line from linking to each other. Two functions declared on one line is minified output, which the analysis ignores anyway.

static_analyzer/engine/lsp_client.py
```diff
@@ -145,7 +189,10 @@ def start(self) -> dict | list | None:
                 "hierarchicalDocumentSymbolSupport": True,
             },
             "references": {},
-            "definition": {},
+            # Why linkSupport: a plain Location carries the whole declaration's range, which
+            # starts at the modifiers, so the position never matches the name the symbol table
+            # is keyed on. A LocationLink names the declared symbol itself.
+            "definition": {"linkSupport": True},
```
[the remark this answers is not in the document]
Svilen-Stefanov: Measured rather than normalised. pyright and JDTLS answer plain `Location`s whose range starts at the name: 13,190 of them matched a symbol exactly on mem0, 5,746 on commons-lang, 2,308 on junit4, and tsserver answers `LocationLink`s. Of the rejected answers that land on a line some symbol starts on, none sit on a modifier or keyword; they are parameters, `self`, locals and lambda parameters. No server we run starts a Location at `public` or `def`, so a check for it would only guard a server we don't ship.

static_analyzer/engine/adapters/go_adapter.py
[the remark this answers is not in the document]
Svilen-Stefanov: Checked against the bundled tree-sitter-go: `Map[User](value)` parses as a `type_conversion_expression` over a `generic_type`, not an `index_expression`, so no call site is found at all and nothing is sent to `User`. Telling a generic call from a conversion to a generic type needs type information tree-sitter doesn't have, so this stays a known Go gap rather than querying every conversion.

static_analyzer/incremental_orchestrator.py
```diff
@@ -141,85 +139,19 @@ def _rebuild_changed_file_edges(
...
     """
     source_inspector = SourceInspector()
-    index = GraphIndex(merged_analysis.call_graph)
-    _restore_cross_boundary_edges(index, invalidated_edges, changed_file_strs, adapter, engine_client, source_inspector)
-    return _add_outbound_edges_from_changed_files(
+    index = GraphIndex(merged_analysis.call_graph, source_inspector)
+    _restore_inbound_edges_via_definitions(
+        index, invalidated_edges, changed_file_strs, adapter, engine_client, source_inspector
+    )
+    external = _add_outbound_edges_from_changed_files(
```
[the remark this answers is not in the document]
Svilen-Stefanov: Leaving this as a known warm-start gap. Reaching the unchanged callers a new implementation now serves means re-querying every caller of every declaration the new type might implement, which is what a warm start avoids. main has the same gap in every language (Java and C# never expanded implementations on a warm start, and the references path never added them for a new file), so this PR doesn't widen it; a full run picks the edges up.

static_analyzer/engine/source_inspector.py
```diff
@@ -74,6 +75,50 @@
...
+_JSX_ELEMENT_NODE_TYPES = frozenset({"jsx_opening_element", "jsx_self_closing_element"})
+# Applying a decorator calls it. ``@cache(...)`` is already a call node, so only a bare
+# name is a site of its own.
+_DECORATOR_NODE_TYPES = frozenset({"decorator"})
+# Everything a grammar writes above a declaration to decorate it.
+_DECORATION_NODE_TYPES = frozenset({"decorator", "annotation", "marker_annotation", "attribute_list"})
+_DECORATOR_NAME_NODE_TYPES = frozenset({"identifier", "attribute", "member_expression"})
+# Declarations that hold a value, and the values that are functions: ``const f = () => ...``
+# is as callable a target as a declared function.
+_FUNCTION_VALUE_HOLDER_NODE_TYPES = frozenset(
```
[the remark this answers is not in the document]
Svilen-Stefanov: Not reachable. A PHP variable passed as an argument (`array_map($handler, [])`) is a `variable_name`, which is never a method-group site, so no definition query goes to `$handler` and `declares_function_value` is never asked. After 6630844f the holders derive from the binding table, which includes `assignment_expression`, but its name is a `variable_name` too. PHP callables passed by value are strings or `[$this, 'method']`, the hook-API residue described in the PR comment.

static_analyzer/engine/source_inspector.py
```diff
@@ -263,12 +424,21 @@ def get_file_lines(self, file_path: Path) -> list[str] | None:
...
+
+        Why an import binding is not one: the name written in ``import { target as t }`` is
+        declared in the file it comes from, and a server answers at that binding as readily
+        as at a declaration. Reading it as a name this file declares matches it to whatever
+        the file happens to declare under the same name.
+        """
+        source = self.get_source_line(file_path, line)
+        if source is None or not 0 <= character < len(source):
+            return ""
+        match = _IDENTIFIER.match(source, character)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. `declared_name_at` now reads the name off the parse tree at `byte_column(line, character)`, so the UTF-16 offset is converted before anything is indexed, and a declaration name is the `name` field of a node with parameters or a body. `test_a_column_past_an_astral_character_counts_utf16_units` pins it.

static_analyzer/incremental_orchestrator.py
```diff
@@ -447,13 +449,67 @@ def _add_iterated_type_edges(
...
+                if location is None:
+                    continue
+                receiver = index.declaration_at(str(location[0]), location[1], location[2]).declaration
+                if receiver is None:
+                    continue
+                for site, member in by_receiver[position]:
+                    target = index.call_graph.nodes.get(f"{receiver.fully_qualified_name}.{member}")
+                    src_node = containing_source_node(index, str(file_path), site.lsp_line, site.lsp_column)
+                    if target is not None and src_node is not None:
+                        added += _add_edges(index.call_graph, src_node, [target], site, changed_file_strs)
```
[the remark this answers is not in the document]
Svilen-Stefanov: Fixed in 6630844f. The warm receiver route expands its target through `targets_through`, the same expansion `targets_for` uses, so it reaches the owning class like `CallEdgeSink.add` does. `test_a_member_reached_through_its_receiver_also_reaches_its_class` covers it.

static_analyzer/analysis_cache.py
```diff
@@ -71,8 +71,10 @@
 # nested solution's files are named by its own engine only.
 # v9: a call into a project of another solution root is an edge.
 # v10: every name is spelled from the repository root; C# keeps ``src`` and Java its source root.
+# v11: every language builds its call edges from definitions at the call site.
+# v12: a call site's column counts UTF-16 code units, as LSP does, not tree-sitter's bytes.
 # Older pickles are treated as cache misses and re-run.
-_TAG_VERSION = "v10"
+_TAG_VERSION = "v12"
```
Svilen-Stefanov: This can be kept at v10 or v11 - the core hasn't been released yet, we don't have to bump it up yet.

static_analyzer/engine/adapters/php_adapter.py
```diff
@@ -59,6 +49,10 @@ def lsp_command(self) -> list[str]:
     def language_id(self) -> str:
         return "php"
 
+    @property
+    def resolves_method_groups(self) -> bool:
+        return True
```
Svilen-Stefanov: Does something set this to false? Is there a downside of having this set as True by default?

static_analyzer/analysis_cache.py
Svilen-Stefanov: Back to v11 in 22064ed5. One bump from main's v10 is still needed: the pickles main writes were built from references, so they must miss here.

static_analyzer/engine/adapters/php_adapter.py
Svilen-Stefanov: Only Java wants it off: Java hands a method on only as a method reference (`Dog::speak`), which is already a call site, so probing bare arguments and field reads there would only add JDTLS requests. Every other adapter overrode it to True, so in 22064ed5 the default is True, the six overrides are gone, and JavaAdapter sets False with that reason. No language changes behaviour.

static_analyzer/external_calls.py
```diff
@@ -1,86 +1,145 @@
...
-                _record_package_import(
-                    package_dependencies, adapter, caller.fully_qualified_name, destination.fully_qualified_name
-                )
+        reached.add(written)
+        edges.extend(_add_edges(call_graph, caller, targets.nodes, site.call_site))
+        for declaration in targets.implementations:
+            owed.setdefault(declaration, []).append((caller, site.call_site))
+
+    if client is not None and owed:
+        for declaration, implementations in implemented_by(index, client, owed).items():
```
ivanmilevtues: that seems like a suspicious way to use the client, i don't really get why does the implemented_by need a client passed by parameter, i was thinking and am still thinking are we doing extra client calls here idk. Feels like we shouldn't as this is a post analysis step or?

static_analyzer/external_calls.py
```diff
@@ -1,86 +1,145 @@
...
 def link_external_call_sites(
     call_graph: CallGraph,
     sites: list[ExternalCallSite],
     adapter: LanguageAdapter,
-    package_dependencies: dict[str, dict],
     inspector: SourceInspector,
-) -> int:
-    """Add the edges whose targets the merged graph now holds; return how many were added.
+    analysed_files: Collection[str],
+    client: LSPClient | None = None,
```
ivanmilevtues: Questionable nullable type.

static_analyzer/external_calls.py
ivanmilevtues: Follow-up on my own comment, I asked an agent and I tend to agree with what it spit out.

**My review take:** the dependency has a concrete purpose, but your layering concern is reasonable: this module now mixes graph linking with live semantic queries. A cleaner separation could return pending implementation queries to the orchestrator, although that would move the work rather than remove it.

One limitation worth questioning in the PR: **a declaration first resolved at the final cross-engine merge gets no fresh implementation query.** Previously recorded implementation answers and graph-known overrides can still link, but that is not equivalent to querying the server. I would ask whether that asymmetry is intentional before assuming full/incremental parity.

In essence what I do not like about this is that it both does linking and resolution, I was expecting resolution to be done from the client when were going through our phases and this to be a part after that or just being independent of the client (or part of the client itself). Now it is hard to really trace when this happens as it has double purpose:

Yes—**`link_external_call_sites()` specifically takes the optional `client` because it does more than attach saved call sites to graph nodes.** It also finishes implementation lookup for declarations it resolves.

ivanmilevtues: that seems like a suspicious way to use the client, i don't really get why does the implemented_by need a client passed by parameter, i was thinking and am still thinking are we doing extra client calls here idk. Feels like we shouldn't as this is a post analysis step or?

static_analyzer/external_calls.py
ivanmilevtues: Questionable nullable type.

static_analyzer/external_calls.py
ivanmilevtues: Follow-up on my own comment, I asked an agent and I tend to agree with what it spit out.

**My review take:** the dependency has a concrete purpose, but your layering concern is reasonable: this module now mixes graph linking with live semantic queries. A cleaner separation could return pending implementation queries to the orchestrator, although that would move the work rather than remove it.

One limitation worth questioning in the PR: **a declaration first resolved at the final cross-engine merge gets no fresh implementation query.** Previously recorded implementation answers and graph-known overrides can still link, but that is not equivalent to querying the server. I would ask whether that asymmetry is intentional before assuming full/incremental parity.

In essence what I do not like about this is that it both does linking and resolution, I was expecting resolution to be done from the client when were going through our phases and this to be a part after that or just being independent of the client (or part of the client itself). Now it is hard to really trace when this happens as it has double purpose:

Yes—**`link_external_call_sites()` specifically takes the optional `client` because it does more than attach saved call sites to graph nodes.** It also finishes implementation lookup for declarations it resolves.

static_analyzer/graph_definitions.py
```diff
@@ -1,58 +1,227 @@
...
 ITERATED = "iterated"
+# Answers rather than call sites: what a call's target is implemented by, and a receiver whose member a call names.
+IMPLEMENTATION = "implementation"
+RECEIVER = "receiver"
+# A dispatched declaration a call reached inside its engine, whose overrides may lie in files that engine never read.
+OVERRIDE = "override"
+
+
+@dataclass(frozen=True)
+class CallShapes:
```
ivanmilevtues: Maybe rename this to PotentialCalls?

I am unsure of CallShapes, i didn't really udnerstand that. From asking agents and looking around i got that this gives potential positions for calls which we then check if are real. using role-specific rules?

static_analyzer/graph_definitions.py
```diff
@@ -1,58 +1,227 @@
...
+    @classmethod
+    def none(cls) -> "CallTargets":
+        return cls([], [])
 
 
 class GraphIndex:
     """A graph's nodes by file, so a position lookup does not scan every node."""
 
-    def __init__(self, call_graph: CallGraph) -> None:
+    def __init__(self, call_graph: CallGraph, inspector: SourceInspector) -> None:
```
ivanmilevtues: it is weird for the index ot have the inspector as a dependency, as I think of this as the index of the call graph.

static_analyzer/__init__.py
```diff
@@ -654,6 +651,9 @@ def discover_file_dependencies(self, file_path: Path) -> list[str]:
 
             logger.debug(f"Discovered {len(unique_paths)} dependencies for {file_path}")
             return list(unique_paths)
+        except StaticAnalysisFatalError:
+            # An unanswered batch is not an answer of "nothing"; the caller must hear it.
+            raise
```
ivanmilevtues: raise with the from thing that you showed me so we have the full stack trace. raise e from

static_analyzer/engine/lsp_client.py
```diff
@@ -571,15 +546,41 @@ def _send_batch(
...
+                first = chunk[unserved[0]]
+                raise StaticAnalysisFatalError(
+                    f"{len(unserved)} of {len(chunk)} {method} requests went unserved, first at "
+                    f"{first[0]}:{first[1] + 1}:{first[2] + 1}. The server timed out or failed the request "
+                    "rather than declining it; reading those as empty would silently drop call-graph edges."
+                )
+            results.extend(answer.results for answer in by_query)
+        return results
+
+    def _ask(
```
ivanmilevtues: ask sounds more like a send :D. timeout probably better to have a default or non nullable as a parameter so we always have a timeout.

static_analyzer/engine/lsp_client.py
```diff
@@ -14,15 +14,26 @@
...
 from static_analyzer.engine.utils import uri_to_path
+from static_analyzer.exceptions import StaticAnalysisFatalError
 from static_analyzer.lsp_client.diagnostics import FileDiagnosticsMap, LSPDiagnostic
 
 logger = logging.getLogger(__name__)
 
 LSP_METHOD_NOT_FOUND = -32601
+# Codes LSP (-32899..-32800) and JSON-RPC (-32768..-32000) reserve for a request that went unserved;
+# any other code is the server declining for itself, as gopls does for a free function.
+LSP_RESERVED_ERROR_CODE_RANGES = ((-32899, -32800), (-32768, -32000))
```
ivanmilevtues, line 31: Cool finding. It became again interesting to me if there isn't some LSP library/client which can handle these for us. However we are a very advanced users so i suspect the answer will be no, that used to be the case last summer.

static_analyzer/engine/source_inspector.py
```diff
@@ -203,17 +284,85 @@ def _error_node_count(tree: Tree) -> int:
...
+
+    def lsp_position(self, point: Point) -> tuple[int, int]:
+        """The line and character an LSP request must carry to name *point*."""
+        if not self.line_starts or point.row >= len(self.line_starts):
+            return point.row, point.column
+        start = self.line_starts[point.row]
+        prefix = self.content[start : start + point.column]
+        if prefix.isascii():
+            return point.row, point.column
+        return point.row, sum(2 if ord(char) > 0xFFFF else 1 for char in prefix.decode("utf8", "replace"))
```
ivanmilevtues: here there are some hex numbers maybe we should put them into a constant or add a small comment on what the heck are we returning as a second sum param here :D

ivanmilevtues (approved):
No blockers from my side. The one thing I didn't like (but it is not a blocker) is the way we are plugging in the external_calls. I think it is really unclear at the moment what it does and hard to track down where it fits into the flow.

If you consider that also the case you can redact and merge, if not you can merge.

Apart from that mostly nits about naming stuff which didn't make sense to me, I suggested names where I had an idea or found what the agent suggested senisble.

static_analyzer/external_calls.py
Svilen-Stefanov: Agreed, and addressed in 5f66ff99. The linker no longer talks to the server at all: `link_external_call_sites` only links, and returns the edges, the sites nothing in the graph could finish, and the declarations still owed an implementation query. The warm start then runs that query as its own step (`implemented_by`) and links the answers (`link_implementations`), so the one place the server is asked is visible in `update_cfg_for_changed_files`. The full build's final merge only links, as it did on main: the owning servers may already be down there. The module docstring of `external_calls` now says where each path runs.

static_analyzer/external_calls.py
Svilen-Stefanov: Gone in 5f66ff99: there is no `client` parameter any more, so nothing is nullable.

static_analyzer/external_calls.py
Svilen-Stefanov: Your reading is right, and the asymmetry the agent points at is real: a declaration first resolved at the final cross-solution merge gets no implementation query. That predates this PR (main's linker has the same limit, its comment says the owning server is already down), so I documented it rather than widening this PR. For the split between resolution and linking, see my reply above: resolution now happens in the engine phases and in one explicit warm-start step, and linking is a pure graph operation. I'm also prototyping a follow-up that lets the warm start's engine run resolve against the cached declarations directly, which removes the warm-start linking step altogether.

static_analyzer/graph_definitions.py
Svilen-Stefanov: Renamed to `PotentialCalls` / `potential_calls` in 5f66ff99. And yes, that is exactly it: tree-sitter finds the positions that may be calls and how each is written (a call, a method passed as a value, a member read or write, a collection initializer, a loop), and the definition answer at the position decides whether it is a call by the rule for that shape. The docstring now says so.

static_analyzer/graph_definitions.py
Svilen-Stefanov: Agreed. In 5f66ff99 `GraphIndex(call_graph)` only indexes the graph; the inspector is passed explicitly to `declaration_at`, `targets_for`, `targets_through` and `implemented_by`, the places that read source.

static_analyzer/__init__.py
Svilen-Stefanov: That line is a bare `raise`, which already re-raises with the full traceback; `from` only matters when raising a new exception. But your question made it clear the pass-through only existed because of the catch-all right below it, so 5f66ff99 removes both: `discover_file_dependencies` now lets failures propagate. Its only caller, the wrapper's live status sync, already catches and logs per cycle, so a failing query skips that overlay update loudly instead of silently falling back.

static_analyzer/engine/lsp_client.py
Svilen-Stefanov: Both done in 5f66ff99: `_ask` is now `_send_round`, and the private request helpers (`_send_batch`, `_send_round`, `_send_request`, `_collect_batch_responses`) take a non-nullable `timeout: int`. The public methods keep an optional override that resolves to the client's configured default once.

static_analyzer/engine/lsp_client.py
Svilen-Stefanov: Worth a timeboxed look at some point; I would not expect one to cover the unserved-vs-declined error codes and the retry rules we need, but I have not checked what is current. Not changing it in this PR.

static_analyzer/engine/source_inspector.py
Svilen-Stefanov: Named `_MAX_SINGLE_UTF16_UNIT_CODE_POINT` with a one-line comment in 5f66ff99: a code point above it takes a surrogate pair, i.e. two UTF-16 units, which is what LSP columns count.

PR #594, opened 2026-09-15
feat(lsp): allow overriding the per-request LSP timeout

RapidPoseidon:
Split out of #585 at @Svilen-Stefanov's request, so the barrier fix can land on its own. Close this if you decide the barrier is enough — #585 does not depend on it.

## WHAT

`CODEBOARDING_LSP_REQUEST_TIMEOUT` overrides the per-request LSP timeout that each adapter fixes today (120s for C#, 60s elsewhere).

## WHY

The ceiling is calibrated on a mid-sized workspace. When a larger one exceeds it with the server still healthy, there is currently no way to get past it without a release — nothing in the CLI, the action, or the environment reaches the value. That is the position we were in with #529.

## HOW

Resolved once in `StaticAnalyzer.__init__`, not inside `_spawn_engine_client`. That is the change from the version in #585: both spawn call sites catch `Exception` and turn it into "skip this language and continue", so validating at the spawn would have let a typo degrade into a partial analysis, and making it not degrade meant broadening a handler well beyond this feature. Resolving in `__init__` runs before any engine starts, so a bad value fails the run on its own and the handlers stay as they are.

An unparseable or non-positive value raises rather than falling back, because the fallback is the exact ceiling the operator set the variable to escape — the run would then time out looking like the variable was ignored.

## On the scope concern

You're right that it applies to every language at once: a `900` meant for csharp-ls also lets a hung server block 900s per request round. I have kept it a single global rather than adding per-language keys, because a per-language surface is a bigger thing to commit to than this is worth. It stays unset by default, so nothing changes unless someone opts in.

Also documented, since you noted it was not: a short table under **Configuration** in the README covering this variable and the existing `CODEBOARDING_MAX_CONCURRENT_ENGINES`, which had no documentation either. Happy to drop that second row if you'd rather keep the doc to new surface.

## Testing

- `uv run pytest` — 2249 passed, 6 skipped
- `uv run mypy .` — clean, 311 source files
- `uv run black --check .` — clean, 311 files

`tests/static_analyzer/test_lsp_request_timeout.py` covers unset / whitespace / explicit / zero / negative / non-integer.

## Summary by CodeRabbit

- **New Features**
  - Added an environment variable to configure language-server request timeouts, with validation for positive whole-number values.
  - Added an environment variable to limit the number of language servers resident at once; `0` allows unlimited servers.

- **Documentation**
  - Documented both configuration options, including their defaults, scope, and validation behavior.


README.md
```diff
@@ -147,6 +147,13 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
 
 Shell environment variables such as `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GOOGLE_API_KEY`, and `OLLAMA_BASE_URL` take precedence over the config file. For private repositories, set `GITHUB_TOKEN` in your environment.
 
+Two environment variables tune the static analysis itself:
+
+| Variable | Effect |
+|---|---|
+| `CODEBOARDING_LSP_REQUEST_TIMEOUT` | Seconds a single language-server request may block, overriding the per-language default (120s for C#, 60s elsewhere). Applies to every language in the run. Unset by default; an unusable value fails the run rather than falling back. |
```
Svilen-Stefanov: I think this is related to the above P1 from codex. This reads like it caps every request, but the startup and overlay probes pass their own scaled timeout (60s + 2s per file, up to 1800s), so they aren't touched by this. On Flask with `=900` the probe still logs `timeout=110s` while the normal requests get 900s. Could we say it applies to requests that use the per-language default and that the probes keep their own budget? Otherwise someone setting it to 30 expecting a hard cap will be surprised.

README.md
```diff
@@ -147,6 +147,13 @@ On first run, CodeBoarding creates `~/.codeboarding/config.toml`. Set one provid
...
+| `CODEBOARDING_MAX_CONCURRENT_ENGINES` | How many language servers may be resident at once. `0` (the default) leaves the bound off. |
```
Svilen-Stefanov, line 155: Interesting that this one warns and keeps the bound off on a bad value, while the new one raises. Not for this PR, but good refactor for later to make them behave the same?

static_analyzer/__init__.py
```diff
@@ -436,7 +460,7 @@ def _spawn_engine_client(self, engine_config: EngineConfig) -> LSPClient:
             command=command,
             project_root=project_path,
             init_options=init_options,
-            default_timeout=adapter.get_lsp_default_timeout(),
+            default_timeout=self._request_timeout_override or adapter.get_lsp_default_timeout(),
```
Svilen-Stefanov, line 463: Could we add a small test that the spawned client actually gets the override as `default_timeout`? The new tests only cover the parsing, so if this line regressed nothing would catch it.

Svilen-Stefanov (changes requested):
Thanks, I think the main blocker is the P1 from codex. I left some other nits but after they are addressed, this should be good to go as well.

RapidPoseidon:
Pushed 0fc62a0.

**The P1 — fixed as a doc change, deliberately.** You and codex are both right that the README overpromised: the probes pass their own scaled budget explicitly, so it wins over the client default, and on Flask with `=900` the probe really does still log `timeout=110s`.

I went with your framing rather than widening the override to cover probes, because widening it would be actively harmful: the didOpen-drain barrier in #585 works *because* the probe budget scales with file count, so a `=30` set to bound normal requests would shorten the drain and hand back exactly the `Timeout waiting for LSP response to request 3` that #585 exists to prevent. Making one knob control both means any value low enough to be useful as a request bound is low enough to re-break large C# solutions.

The row now reads:

> Seconds a language-server request may block *when it uses the per-language default* (120s for C#, 60s elsewhere). Not a hard cap on every request: the indexing and didOpen-drain probes pass their own scaled budget (60s plus 2s per file, capped at 1800s) and are unaffected. Applies to every language in the run. Unset or empty leaves the defaults alone; any other unusable value fails the run rather than falling back.

If you'd rather the probes were tunable too, I think that wants its own variable with its own name, and I'd do it as a separate PR.

**Propagation test — added.** You and CodeRabbit both flagged it and you were right that nothing covered that line. `TestOverrideReachesTheSpawnedClient` starts an engine through `start_clients` with the variable set and asserts the patched `LSPClient` receives `default_timeout=900`, plus the unset case asserting the adapter's 120 still wins. It fails with `assert 120 == 900` if that expression regresses — checked by reverting it.

**Not doing: raise on an empty value.** CodeRabbit wants `CODEBOARDING_LSP_REQUEST_TIMEOUT=""` to fail rather than fall back. I've left empty as "unset", because `FOO=` is the ordinary way to neutralise a variable in a shell or a compose file, and hard-failing every run on it would be unfriendly — and because `CODEBOARDING_MAX_CONCURRENT_ENGINES` right next to it treats blank the same way. The README now says "unset or empty leaves the defaults alone" so the documented contract matches. Happy to flip it if you disagree.

**On the warn-vs-raise inconsistency** you spotted between this and `CODEBOARDING_MAX_CONCURRENT_ENGINES`: agreed, and noted as out of scope here. The two want the same parsing helper; I'm happy to take that refactor if you open a ticket.

Gates on 0fc62a0: `uv run pytest` 2251 passed / 6 skipped, `uv run mypy .` clean on 311 files, `uv run black --check .` clean.

Svilen-Stefanov (approved):
Aweosme, lgtm.

PR #604, opened 2026-09-18
fix(csharp): build analyzer project references so Roslyn can resolve them

### LINKED ISSUE #603: textDocument/implementation fails solution-wide when analyzer ProjectReferences are unresolved (restore does not build them)
## Summary

On a repository that references source generators via `ProjectReference ... OutputItemType="Analyzer"`, `sync` aborts in Phase 2b with:

```
StaticAnalysisFatalError: 16 of 50 textDocument/implementation requests went unserved,
first at .../IRequestExecutor.cs:73:7. The server timed out or failed the request rather
than declining it; reading those as empty would silently drop call-graph edges.
```

The cause is not the queried symbol, the repository size, or request concurrency. `prepare_project()` runs `dotnet restore` but never builds the projects referenced as analyzers, so Roslyn holds an `UnresolvedAnalyzerReference` and every `textDocument/implementation` request in the solution fails.

Minimal reproduction (two projects, ten lines of C#): https://github.com/RapidataAI/csharp-ls-unresolved-analyzer-repro

## Root cause

csharp-ls stderr, captured by launching it through `CSharpAdapter.get_lsp_command` / `get_lsp_env`:

```
System.AggregateException: One or more errors occurred.
  (Unexpected value 'Microsoft.CodeAnalysis.Diagnostics.UnresolvedAnalyzerReference'
   of type 'Microsoft.CodeAnalysis.Diagnostics.UnresolvedAnalyzerReference')
---> System.InvalidOperationException: Unexpected value 'UnresolvedAnalyzerReference' ...
  at Microsoft.CodeAnalysis.Serialization.SerializerService.CreateChecksum(AnalyzerReference reference)
  at Microsoft.CodeAnalysis.Serialization.ChecksumCache.GetOrCreateChecksumCollection(...)
```

`SerializerService.CreateChecksum` has no case for `UnresolvedAnalyzerReference` and throws. The chain:

1. An analyzer `ProjectReference` resolves to the referenced project's **output assembly**.
2. `dotnet restore` never produces one — only a build does.
3. Roslyn substitutes an `UnresolvedAnalyzerReference` placeholder.
4. `textDocument/implementation` reaches `CreateChecksum` through the out-of-process symbol finder, which checksums the whole solution.
5. It throws, csharp-ls returns `-32603`, and `_send_batch` raises `StaticAnalysisFatalError`.

`textDocument/definition` does not take that path, which is why Phase 2a completes and Phase 2b dies seconds later. Because the failure is in solution-wide checksum computation, one unbuilt analyzer project breaks queries in **every** project.

## Evidence

Reproduced locally with `codeboarding==0.14.3` and csharp-ls 0.24.0 against a 9,041-file / 301-project solution. Same three positions, before and after `dotnet build`:

| Position | restore only | after build |
|---|---|---|
| `IAuditingCreator.cs:8:10` | `-32603` | OK, 1 result |
| `IAccessAuthorizer.cs:20:10` | `-32603` | OK, 1 result |
| `IRequestExecutor.cs:73:7` | `-32603` | OK, 1 result |

All 16 unserved requests returned `code=-32603 Internal error: AggregateException`.

Things ruled out experimentally, in case they save someone else the time:

- **Symbol shape.** `IAuditingCreator` has one implementation in one project and still fails. A synthetic solution with 18 generic overloads served 18/18.
- **Scale.** A generated 80-project / 5,281-file solution served 1,440/1,440 implementation requests in 2.7s.
- **Concurrency.** Re-asking the 16 failing positions one request in flight at a time recovered 0 of 16, each failing in ~10ms.
- **`.slnx`.** csharp-ls has supported it since 0.18.0.
- **Multi-targeting / linked sources.** Neither is present in the affected repo.

## Two suggested fixes

**1. Build analyzer projects during `prepare_project` (PR attached).**

Discover `ProjectReference` entries carrying `OutputItemType="Analyzer"` and build them after restore. Surgical — it builds only those projects, not the solution.

**2. Log every unserved position, not only the first.**

`_send_batch` names `chunk[unserved[0]]` and discards the other 15, and `_collect_batch_responses` logs the error *message* but not its *code*. Diagnosing this required patching both locally. Two small changes would have made the cause visible from CI logs alone.

Related but separate: a single `-32603` aborts the entire analysis. Worth reconsidering, though in this case the edges genuinely are unavailable — serial retry recovered none of them — so the current fail-fast behaviour is defensible once the underlying resolution is fixed.

## Environment

- codeboarding 0.14.3 (via CodeBoarding-action v1.15.1)
- csharp-ls 0.24.0
- .NET SDK 10.0.400
- Ubuntu, GitHub-hosted runner and local

RapidPoseidon:
Fixes #603

## Problem

`prepare_project()` runs `dotnet restore` but never builds the projects referenced with `OutputItemType="Analyzer"`. An analyzer `ProjectReference` resolves to the referenced project's **output assembly**, which restore does not produce, so Roslyn substitutes an `UnresolvedAnalyzerReference`.

`SerializerService.CreateChecksum` has no case for that type and throws:

```
System.InvalidOperationException: Unexpected value 'UnresolvedAnalyzerReference'
  at Microsoft.CodeAnalysis.Serialization.SerializerService.CreateChecksum(AnalyzerReference reference)
  at Microsoft.CodeAnalysis.Serialization.ChecksumCache.GetOrCreateChecksumCollection(...)
```

`textDocument/implementation` reaches it through the out-of-process symbol finder, which checksums the whole solution, so csharp-ls answers `-32603` and `_send_batch` raises `StaticAnalysisFatalError`. Because the failure is in solution-wide checksum computation, one unbuilt analyzer project breaks implementation queries in **every** project. `textDocument/definition` does not take that path, which is why Phase 2a completes and Phase 2b dies seconds later.

## Change

- `dotnet_solution.analyzer_project_references()` — the projects a set of project files reference as analyzers, deduplicated, skipping references whose target is missing and projects that will not parse.
- `CSharpAdapter._build_analyzers()` / `_build()` — build exactly those projects after restore, on both the solution-restore and per-project-restore paths.

Only the analyzer projects are built, never the solution. A failed build is logged and does not abort, matching how `_restore` already behaves.

## Verification

Against a 9,041-file, 301-project solution, the same three positions before and after the change:

| Position | before | after |
|---|---|---|
| `IAuditingCreator.cs:8:10` | `-32603` | OK, 1 result |
| `IAccessAuthorizer.cs:20:10` | `-32603` | OK, 1 result |
| `IRequestExecutor.cs:73:7` | `-32603` | OK, 1 result |

Running the patched `prepare_project()` against the two-project reproduction at https://github.com/RapidataAI/csharp-ls-unresolved-analyzer-repro:

```
INFO    dotnet restore completed for Repro.sln
INFO    dotnet build completed for MyGenerator.csproj
INFO    dotnet build analyzer projects: 1 of 1 built
```

and the query that previously returned `-32603` returns `OK 1 implementation(s)` with a clean server stderr.

## Checks

- `uv run pytest tests/static_analyzer/test_dotnet_solution.py tests/static_analyzer/test_csharp_adapter.py` — 67 passed
- `uv run black --check static_analyzer/ tests/static_analyzer/test_dotnet_solution.py` — clean
- `uv run mypy static_analyzer/dotnet_solution.py static_analyzer/engine/adapters/csharp_adapter.py` — clean

## Note

Four new tests cover discovery. The build step itself is not covered by a unit test — it shells out to `dotnet`, and the existing suite has no fixture for that. Happy to add one if you have a preferred pattern.

Separately, and not addressed here: `_send_batch` names only `chunk[unserved[0]]` when it raises, and `_collect_batch_responses` logs the error message but not its code. Diagnosing this needed both patched locally; surfacing them would make the next occurrence readable from CI logs alone.

## Summary by CodeRabbit

* **New Features**
  * Analyzer projects referenced by .NET solutions and individual project targets are automatically discovered and built.
  * Analyzer references support legacy namespaced project files, case-insensitive matching, and relative path resolution.
  * Duplicate, missing, or unsupported references are excluded.

* **Bug Fixes**
  * Unreadable or malformed project files no longer block valid analyzer references.
  * Analyzer builds are attempted even when project restoration fails.

* **Tests**
  * Added coverage for analyzer discovery and build behavior.


RapidPoseidon:
Both findings were valid — fixed in dccfc91.

**Namespaced `ProjectReference`.** Confirmed before changing anything: `iter("ProjectReference")` matches 0 elements on a legacy project, because every tag carries `{http://schemas.microsoft.com/developer/msbuild/2003}`. Those analyzers stayed unresolved and the bug persisted for that repo. Now matched on the local name, with a namespaced-project test.

**Bare project target.** Also real. `prepare_project` restored a `.csproj` target and returned without building its analyzers, so a single-project repository still failed. Now builds with `[target]` when restore succeeds, and skips the build when restore fails.

That second one also filled a gap I had flagged in the PR description: there is now a `prepare_project` test class covering both the success and the restore-failure branch, patching `_restore` and `_build_analyzers` rather than shelling out. The `_build` subprocess call itself is still uncovered.

Checks after the change:

- `uv run pytest tests/static_analyzer/test_csharp_adapter.py tests/static_analyzer/test_dotnet_solution.py` — 70 passed
- `uv run black --check static_analyzer/ tests/static_analyzer/` — clean, 125 files
- `uv run mypy static_analyzer/dotnet_solution.py static_analyzer/engine/adapters/csharp_adapter.py` — clean

Re-verified end to end against the two-project reproduction: `dotnet build analyzer projects: 1 of 1 built`, and the query that returned `-32603` returns `OK 1 implementation(s)` with clean server stderr.

static_analyzer/dotnet_solution.py
```diff
@@ -27,3 +27,30 @@ def solution_projects(solution: Path) -> list[Path]:
...
+def analyzer_project_references(projects: list[Path]) -> list[Path]:
+    """The projects referenced with ``OutputItemType="Analyzer"``, absolute, in listing order.
+
+    Why: an analyzer reference resolves to the referenced project's output assembly, so
+    it stays unresolved until that project is built.
+    """
+    analyzers: list[Path] = []
+    for project in projects:
+        try:
+            root = ElementTree.parse(project).getroot()
```
Svilen-Stefanov, line 41: This parses the raw csproj, so only references written literally in it are found. I tried a few variants of the repro on this branch and these still die in Phase 2b with the same `-32603`:
- the reference coming from a `Directory.Build.props`. jellyfin does this for its `Jellyfin.CodeAnalysis` analyzer, the scan finds 0 analyzer projects across its 41 projects, and a full run dies on the first Phase 2b batch (`ILocalizationManager.cs:73:17`) on both main and this branch
- `Include="$(MSBuildThisFileDirectory)../MyGenerator/MyGenerator.csproj"`, which is how Avalonia's `build/SourceGenerators.props` does it
- metadata as child elements (`<OutputItemType>Analyzer</OutputItemType>`). I think that's actually the usual form in the legacy namespaced projects the second commit adds, since attribute metadata only came with MSBuild 15
- a bare `App.csproj` target whose analyzer is one `ProjectReference` away (App -> Lib -> Gen), since only `[target]` gets scanned

`dotnet msbuild <proj> -getItem:ProjectReference` finds all four, but that's ~0.3-0.7s per project even on these tiny ones, so probably too slow for 300 projects. Not sure we need to chase all of these here if we add a safety net though (see my comment on `_build_analyzers`), wdyt?

static_analyzer/dotnet_solution.py
```diff
@@ -27,3 +27,30 @@ def solution_projects(solution: Path) -> list[Path]:
...
+        for reference in root.iter():
+            # Legacy-format projects namespace every tag, so match on the local name.
+            if reference.tag.rpartition("}")[2] != "ProjectReference":
+                continue
+            if (reference.get("OutputItemType") or "").strip().lower() != "analyzer":
+                continue
+            raw = reference.get("Include") or ""
+            if not raw:
+                continue
+            path = (project.parent / raw.replace("\\", "/")).resolve()
```
Svilen-Stefanov: nit: these lines are the same as in `solution_projects` above (resolve, suffix check, `is_file`, dedupe). Could we pull them into a small helper, e.g. `_project_file(base, raw) -> Path | None`, and use it in both?

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -411,6 +417,53 @@ def _top_level_types(self, symbols: list[dict]) -> list[str]:
...
+        """Build the projects referenced as analyzers so Roslyn can resolve them.
+
+        Why: restore does not produce an output assembly, so Roslyn holds an
+        ``UnresolvedAnalyzerReference`` whose checksum throws, failing every
+        ``textDocument/implementation`` request across the whole solution.
+        """
+        analyzers = analyzer_project_references(projects)
+        if not analyzers:
+            return
+        built = sum(self._build(dotnet_path, analyzer, project_root, env) for analyzer in analyzers)
```
Svilen-Stefanov: Also wondering about the failure mode here. If the build fails (I tried a generator with a compile error), we log a warning, csharp-ls still gets the `UnresolvedAnalyzerReference`, and Phase 2b dies anyway with an error that points at some random interface and not at this warning. Same for anything the scan misses.

What do you think about a safety net in `_SINGLE_TARGET_FRAMEWORK_TARGETS`, which we already inject into every project csharp-ls loads?

```xml
  <Target Name="CodeBoardingDropMissingAnalyzers" BeforeTargets="CoreCompile">
    <ItemGroup>
      <Analyzer Remove="@(Analyzer)" Condition="!Exists('%(Analyzer.FullPath)')" />
    </ItemGroup>
  </Target>
```

On top of main, without any build step, this made all six cases pass (the repro, the broken generator and the four from my other comment), and the repro comes out with the same three edges as this branch. jellyfin then goes through the whole of Phase 2b too (3879 targets, 20626 call edges). On top of this branch the built generator is kept and the broken one is no longer fatal. An analyzer whose dll doesn't exist wouldn't run anyway, so dropping it shouldn't lose anything. I'd still keep the build though, since then the generators actually run and code that goes through generated members can resolve.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -411,6 +417,53 @@ def _top_level_types(self, symbols: list[dict]) -> list[str]:
...
+        logger.info("dotnet build analyzer projects: %d of %d built", built, len(analyzers))
+
+    def _build(self, dotnet_path: str, target: Path, project_root: Path, env: dict[str, str]) -> bool:
```
Svilen-Stefanov: This is pretty much a copy of `_restore` below with `restore` swapped for `build`. Could we have one `_dotnet(verb, dotnet_path, target, project_root, env) -> bool` and use it for both? Then the `600` (in the timeout and the log message, in both) can become a named constant too, e.g. `_DOTNET_TIMEOUT_SECONDS`.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -357,7 +357,12 @@ def prepare_project(self, project_root: Path) -> None:
 
         env = os.environ.copy()
         env.update(self.get_lsp_env(project_root))
-        if self._restore(resolution.dotnet_path, target, project_root, env) or target.suffix not in (".sln", ".slnx"):
+        if target.suffix not in (".sln", ".slnx"):
+            if self._restore(resolution.dotnet_path, target, project_root, env):
```
Svilen-Stefanov: just checking, why skip the build when restore fails here? `dotnet build` restores the analyzer project on its own, and the per-project fallback below builds even when nothing restored, so the two paths behave differently now. If we skip, a repo with an analyzer reference then dies in Phase 2b anyway (unless we add the safety net), so maybe just always build?

tests/static_analyzer/test_csharp_adapter.py
```diff
@@ -696,3 +696,43 @@ def test_a_file_without_types_asks_nothing(self, tmp_path: Path):
 
         assert CSharpAdapter().read_document_symbols(usings, SourceInspector(), client) == []
         client.workspace_symbol.assert_not_called()
+
+
+class TestPrepareProjectBuildsAnalyzers:
```
Svilen-Stefanov: Re the "preferred pattern" question in the description: `TestPrepareProject` above already fakes `subprocess.run` and records every command, so these could live there and cover `_build` for real too. I tried a quick sketch with a slnx whose app references `src/Gen/Gen.csproj` as an analyzer, asserting `[cmd[1:3] for cmd in self.commands] == [["restore", "App.slnx"], ["build", "src/Gen/Gen.csproj"]]`, plus the same with `self.failing = {"App.slnx"}` for the per-project fallback. Both pass on this branch. Those two solution paths are the ones your 301-project repo actually hits and they don't have a test yet, so it would be nice to have them.

Svilen-Stefanov (changes requested):
Thanks for the really detailed write-up and the repro, made this super easy to check!

I reproduced it on main through the actual engine (added a small Consumer that calls through IGreeter, since Phase 2b only asks for implementations of interface calls and the repro has none) and got the same StaticAnalysisFatalError in Phase 2b. On this branch MyGenerator gets built, Phase 2b is served and the Consumer.Run -> Greeter.Greet edge shows up, so it does fix the repro.

The main thing I'm not sure about is coverage. The scan only sees references written literally in the listed csproj files, and a few pretty common variants still die the same way on this branch (jellyfin is a real one, details inline). I left a suggestion for a small safety net in the targets file we already inject, which fixed all of them in my runs, jellyfin included. Would be great to get that in before merging, otherwise any repo we miss still fails the whole analysis. The rest are reuse/test nits.

Didn't land on a changed line:

nit: the prepare_project docstring still only talks about restore ("only writes under obj/"), might be worth a line about the analyzer build now that it also compiles and writes bin/.
The logging bits from the description (all unserved positions + the error code) sound good to me, happy to take those as a separate small PR.

RapidPoseidon:
All six addressed in 91e4d92. The safety net was the right call — thanks for testing it that thoroughly, the coverage gap was real and I had not seen it.

**Safety net.** Added `CodeBoardingDropMissingAnalyzers` to `_SINGLE_TARGET_FRAMEWORK_TARGETS` as you wrote it. I verified it independently rather than taking it on trust, on the `Directory.Build.props` variant specifically, because that is the one my scan cannot see at all:

```
scan finds analyzers: 0
analyzer built?       0

with the target:     OK     1 implementation(s)   exceptions: 0
without the target:  ERROR  -32603 AggregateException   exceptions: 2
```

Same solution, same unbuilt analyzer, only the target differs. I also confirmed the plain repro passes on the safety net alone with nothing built. I have not independently reproduced your jellyfin or Avalonia runs — taking those from your testing.

Kept the build, for the reason you gave: the generators then actually run, so code going through generated members resolves. The safety net only covers what is missing.

**Always build.** You were right, and this reverses what I changed for CodeRabbit's earlier comment. `dotnet build` restores the analyzer itself, the per-project fallback already built unconditionally, and skipping was worst exactly when the reference is most likely unresolved. Now unconditional, with a test naming the reason.

**Reuse.** `_restore`/`_build` collapsed into `_dotnet(verb, ...)`, `600` is now `_DOTNET_TIMEOUT_SECONDS`, and `_project_file(base, raw)` is shared with `solution_projects`.

**Tests.** I had missed `TestPrepareProject` entirely — my standalone class is gone and the tests live there now, so `_dotnet` is covered for real through the fake. Both solution paths are covered as you sketched:

```python
[["restore", "App.slnx"], ["build", "src/Gen/Gen.csproj"]]
[["restore", "App.slnx"], ["restore", "App.csproj"], ["build", "src/Gen/Gen.csproj"]]
```

plus the bare-target, restore-failed and no-analyzer cases.

**Docstring.** Updated — it now mentions the analyzer build and that it writes `bin/`.

Checks: `uv run pytest tests/static_analyzer/` 1223 passed · black clean, 203 files · mypy clean, 63 files.

On the remaining scan gaps — MSBuild-property `Include`, child-element metadata, transitive references: with the safety net in place they degrade to a missing generator rather than a failed analysis, so I have left them rather than adding `dotnet msbuild -getItem` at 0.3-0.7s per project. Happy to revisit if you would rather close them properly.

Will send the logging changes (all unserved positions plus the error code) as a separate PR.

RapidPoseidon:
Partly taken, in 2eed97a — but not the main suggestion, and I want to be explicit about why.

**Taken: the drop should not be silent.** AGENTS.md §5 permits graceful degradation on "genuinely non-critical, *clearly-logged* paths", and the logging half was missing. A partial analyzer build now logs at warning level rather than info, with a test:

```
dotnet build analyzer projects: only 0 of 1 built
```

That is the case where something real is lost — the generator does not run, so its symbols are absent.

**Not taken: raising a partial-analysis error instead of removing the item.** That reinstates the bug this PR exists to fix. Today a single unresolved analyzer aborts the entire analysis via `StaticAnalysisFatalError`; raising a different exception in the same situation changes the error text, not the outcome. @Svilen-Stefanov measured the alternative on real repositories: with the target, jellyfin completes Phase 2b (3879 targets, 20626 call edges); without it, jellyfin dies on the first batch. Failing those repositories loudly is still failing them.

The trade is also not symmetric. An analyzer whose assembly does not exist cannot run either way — Roslyn will not load a file that is not there. Dropping the item costs the generator's symbols; keeping it costs every `textDocument/implementation` result in the solution, including for projects that have no analyzers at all.

**Not taken: resolving from evaluated MSBuild items.** Correct in principle, and it is what would close the `Directory.Build.props` and MSBuild-property cases properly. `dotnet msbuild -getItem:ProjectReference` was measured at 0.3-0.7s per project, which is several minutes on the 301-project solution that prompted this. Deferred deliberately, and noted in my previous comment.

One correction to the finding: the analyzer `ProjectReference` items contributed by the *injected* props are not the issue. Those files add the framework and target-framework plumbing, not analyzer references. The gap is analyzer references in the *repository's own* `Directory.Build.props`, which is what I verified against, and it is a scan gap rather than something the injected imports introduce.

Checks: `uv run pytest tests/static_analyzer/` 1224 passed · black clean, 203 files · mypy clean, 63 files.

Deferring to @Svilen-Stefanov on the design here, since he specified and tested this approach.

RapidPoseidon:
Leaving this one as is.

The annotations follow each file's local style rather than a single rule. In `test_dotnet_solution.py` every test is annotated, so the four tests I added there are too:

```python
def test_analyzer_project_references_finds_projects_referenced_as_analyzers(tmp_path: Path) -> None:
```

In `TestPrepareProject` none of the eight pre-existing tests are, so the six I added are not either:

```python
def test_runs_dotnet_restore_when_csproj_present(self, tmp_path):        # already there
def test_warns_when_an_analyzer_project_fails_to_build(self, tmp_path, caplog):   # mine
```

Annotating only my additions would leave that one class six annotated and nine not, which is worse than either consistent choice.

`disallow_untyped_defs = false` in `pyproject.toml`, and `uv run mypy tests/static_analyzer/test_csharp_adapter.py` passes as written, so nothing is being suppressed here.

Happy to annotate the whole class if @Svilen-Stefanov wants it, but that is a cleanup of existing tests rather than part of this fix, so I would rather not fold it in.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -47,6 +48,14 @@
     <TargetFrameworks>$([System.String]::Copy('$(CodeBoardingTargetFrameworks)').Trim(';').Split(';')[0])</TargetFrameworks>
     <TargetFrameworks Condition="'$(CodeBoardingPreferredTargetFramework)' != '' and $(CodeBoardingTargetFrameworks.Contains(';$(CodeBoardingPreferredTargetFramework);'))">$(CodeBoardingPreferredTargetFramework)</TargetFrameworks>
   </PropertyGroup>
+  <!-- An analyzer whose assembly is absent cannot run, but Roslyn keeps it as an
+       UnresolvedAnalyzerReference whose checksum throws, failing every
+       textDocument/implementation request in the solution. -->
+  <Target Name="CodeBoardingDropMissingAnalyzers" BeforeTargets="CoreCompile">
```
Svilen-Stefanov: One thing I noticed while testing with a real incremental generator (it emits a factory the app then calls through): if the generator comes in via `Directory.Build.props`, the scan misses it, nothing gets built and this target drops it without any trace. The run then "succeeds" with 0 call edges instead of 3, since nothing downstream of the generated code resolves, and there's no log line at all because the new warning only fires for analyzers the scan found.

Not a blocker imo since main doesn't get those edges either, but per AGENTS.md §5 it would be nice if this was at least visible. Maybe the target could also write what it removes to a file we pass in via env, and we log one warning after the load? Happy to track it as a follow-up ticket if you'd rather keep this PR small.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -357,16 +367,24 @@ def prepare_project(self, project_root: Path) -> None:
...
         # rest without ``project.assets.json`` -- and csharp-ls then resolves no package
         # or transitive project reference anywhere. Each member on its own only fails
         # for itself.
         projects = solution_projects(target)
-        restored = sum(self._restore(resolution.dotnet_path, project, project_root, env) for project in projects)
+        restored = sum(
+            self._dotnet("restore", resolution.dotnet_path, project, project_root, env) for project in projects
+        )
         logger.info("dotnet restore per project: %d of %d restored", restored, len(projects))
+        self._build_analyzers(resolution.dotnet_path, projects, project_root, env)
```
Svilen-Stefanov: Something I only noticed this round: restore and build leave their MSBuild nodes running (node reuse), and `dotnet build` also starts the compiler server (VBCSCompiler). All of them outlive `prepare_project` and sit there while csharp-ls loads. On Avalonia that's 11 MSBuild nodes + VBCSCompiler, ~2.6 GB, right when csharp-ls needs the memory. The nodes mostly come from restore so that part is on main already, the compiler server is new here.

I tried two ways to fix it:
- `-nodeReuse:false -p:UseSharedCompilation=false` on every `_dotnet` call: nothing left behind, but Avalonia's per-project restore fallback goes 106s -> 174s
- one `dotnet build-server shutdown` at the end of `prepare_project`: nothing left behind and the time stays the same (104s), and the generator fixture still gets its edges

So the second one. Maybe only when we're on our private SDK, since on a system SDK it would also stop the user's own build servers. Would you be up for adding it here? Otherwise I'll make a ticket.

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -411,29 +429,48 @@ def _top_level_types(self, symbols: list[dict]) -> list[str]:
                 found.append(sym.get("name", ""))
         return found
 
-    def _restore(self, dotnet_path: str, target: Path, project_root: Path, env: dict[str, str]) -> bool:
+    def _build_analyzers(self, dotnet_path: str, projects: list[Path], project_root: Path, env: dict[str, str]) -> None:
+        """Build the projects referenced as analyzers so Roslyn can resolve them.
+
+        Why: restore does not produce an output assembly, so Roslyn holds an
```
Svilen-Stefanov: nit: now that the target is what keeps Phase 2b alive, this Why reads a bit off. The build is really about making the generators run, so code that goes through generated members resolves. Maybe something like "Why: a built generator runs, so calls through generated code resolve; whatever stays missing is dropped by the injected target"?

static_analyzer/engine/adapters/csharp_adapter.py
```diff
@@ -29,6 +29,7 @@
 logger = logging.getLogger(__name__)
 
 _MULTI_TARGET_PROJECT_THRESHOLD = 25
+_DOTNET_TIMEOUT_SECONDS = 600
 _SINGLE_TARGET_FRAMEWORK_TARGETS = r"""<Project>
```
Svilen-Stefanov: nit (my bad for suggesting it here): this now does more than the single-target fold, but the name (and `csharp-ls-single-target.targets`, `_single_target_framework_env`) doesn't hint that it also carries the analyzer drop. Maybe rename it to something like `_INJECTED_DIRECTORY_BUILD_TARGETS`, or at least add a line about the drop to the `_single_target_framework_env` docstring so people can find it?

Svilen-Stefanov (approved):
Ran it and seems to work. I'll merge it.

